PR #1, opened 2015-06-17
FireCloud List Methods API

rushtong:
Initial implementation of listing all available Methods from the Agora service.


.gitignore
```diff
@@ -1,9 +1,68 @@
...
+
+### JetBrains template
+# Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm
+
 *.iml
-.classpath_nb
+
+## Directory-based project format:
+.idea/
+# if you remove the above rule, at least ignore the following:
```
dmohs: This is a surprising comment. Why would someone do this?

rushtong: This is a standard gitignore configuration for Intellij. It's generally best to leave intellij config files out of a repo, but some folks (https://github.com/samtools/htsjdk) like to have them in.

dmohs:
Maybe add a TODO to update the README with a link to the Swagger UI when that becomes available.

dmohs:
This looks good to me, but since I don't know Scala, maybe @MattPutnam or @davidangb should take a quick look before merge.

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
```diff
@@ -0,0 +1,69 @@
...
+  // connects the services environment to the enclosing actor or test
+  override def actorRefFactory = context
+
+  trait ActorRefFactoryContext {
+    def actorRefFactory = context
+  }
+
+  val methodsService = new MethodsService with ActorRefFactoryContext
+
+  private implicit val ec = context.dispatcher
```
davidangb: as an implicit, you might not be able to rename ... but I like "executionContext" better than "ec" for clarity.

src/main/scala/org/broadinstitute/dsde/firecloud/MethodsClient.scala
```diff
@@ -0,0 +1,67 @@
...
+      Get(FireCloudConfig.Methods.baseUrl)
+    }
+
+    responseFuture onComplete {
+      case Success(methods) =>
+        log.debug("methods found")
+        requestContext.complete(methods)
+        context.stop(self)
+
+      case Failure(error) =>
```
davidangb: this case clause catches an http error from Agora, then rethrows an http 500. Instead of catching and rethrowing, can we just bubble up the Agora error? Or at minimum, call failWith the same error code that Agora gave us? (this was an issue in vault-api too)

src/main/scala/org/broadinstitute/dsde/firecloud/model/MethodEntity.scala
```diff
@@ -0,0 +1,33 @@
...
+import com.wordnik.swagger.annotations.{ApiModel, ApiModelProperty}
+import spray.json.DefaultJsonProtocol
+
+import scala.annotation.meta.field
+
+object MethodEntityJsonProtocol extends DefaultJsonProtocol {
+  implicit val impMethodEntity = jsonFormat10(MethodEntity)
+}
+
+@ApiModel(value = "Agora Method")
```
davidangb: Before we get too far in our own model classes ... can we share models with Agora via a common lib?

src/main/scala/org/broadinstitute/dsde/firecloud/service/MethodsService.scala
```diff
@@ -0,0 +1,46 @@
...
+  @ApiOperation(
+    value = "list methods",
+    nickname = "listMethods",
+    httpMethod = "GET",
+    response = classOf[MethodEntity],
+    responseContainer = "List",
+    notes = "response is a list of methods from the methods repository")
+  @ApiResponses(Array(
+    new ApiResponse(code = 200, message = "Successful"),
+    new ApiResponse(code = 400, message = "Bad Request"),
```
davidangb: Under what conditions would the 400 and 404 be thrown? See my previous comment - I think we catch those errors and rethrow as a 500.

src/main/scala/org/broadinstitute/dsde/firecloud/MethodsClient.scala
```diff
@@ -0,0 +1,67 @@
...
+
+}
+
+class MethodsClient(requestContext: RequestContext) extends Actor {
+
+  import DefaultJsonProtocol._
+  import system.dispatcher
+
+  implicit val system = context.system
+  val log = Logging(system, getClass)
```
davidangb: the level for this logger is controlled by the akka.logLevel in application.conf, not the appropriate logger in logback.xml :(

src/main/scala/org/broadinstitute/dsde/firecloud/service/MethodsService.scala
```diff
@@ -0,0 +1,46 @@
...
+import org.broadinstitute.dsde.firecloud.model.MethodEntity
+import org.slf4j.LoggerFactory
+import spray.routing.{HttpService, Route}
+
+class MethodsServiceActor extends Actor with MethodsService {
+  def actorRefFactory = context
+  def receive = runRoute(listRoute)
+}
+
+@Api(value = "/methods", description = "Methods Service", produces = "application/json")
```
davidangb: This may be a red herring, but I couldn't get the swagger-ui to work until I added consumes="application/json" as well. Before adding that, I would get a 406 error that mentioned it was expecting text/plain.

HOWEVER, I then removed consumes="application/json" to verify that was the fix ... and it still worked. So I dunno what happened. If it works for you, feel free to ignore.

davidangb:
review complete with a few comments ... this is close!

src/main/scala/org/broadinstitute/dsde/firecloud/MethodsClient.scala
rushtong: Ah - I was catching things from different processes - I can refine this a bit. The bigger problem I had was that spray was swallowing everything during the unmarshalling. If it isn't json, it chokes, so we don't know what Agora was telling us.

src/main/scala/org/broadinstitute/dsde/firecloud/model/MethodEntity.scala
rushtong: Will follow up with the green team and try to make that happen somewhere.

rushtong:
Thank you - I'll look into all of these and re-submit.

src/main/scala/org/broadinstitute/dsde/firecloud/model/MethodEntity.scala
rushtong: Green team is not keen on extracting their model classes to a commons lib. Will discuss further, but is not likely. Will investigate swagger codegen to see if it can help.

davidangb: don't hold up this story for codegen - we can do that as follow-on.

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
rushtong: Turns out that line isn't needed at all. Removing.

src/main/scala/org/broadinstitute/dsde/firecloud/MethodsClient.scala
davidangb: I still can't figure out which conf file controls the level for this logger. I'm fine pushing that out to another story, but don't want to lose this issue.

rushtong: I tested this - by removing the akka lines in the config, that let logback take effect.

src/main/scala/org/broadinstitute/dsde/firecloud/MethodsClient.scala
```diff
@@ -0,0 +1,69 @@
...
+
+    responseFuture onComplete {
+      case Success(response) =>
+        response.status match {
+          case OK =>
+            log.debug("OK response")
+            requestContext.complete(unmarshal[List[MethodEntity]].apply(response))
+            context.stop(self)
+          case _ =>
+            // Bubble up other successful status codes
```
davidangb: nitpick on the comment - this bubbles up any status code where the Future completed, including e.g. 404s from Agora. The term "successful status codes" makes it sound like only HTTP 20x codes.

src/main/scala/org/broadinstitute/dsde/firecloud/MethodsClient.scala
```diff
@@ -0,0 +1,69 @@
...
+            log.warning("Unanticipated response: " + response.status.defaultMessage)
+            requestContext.complete(response)
+            context.stop(self)
+        }
+      case Failure(error) =>
+        // Bubble up all other failures
```
davidangb: Another comment nitpick - this case is for anything where the pipeline itself failed, not for specific http error codes, so it's not a bubble.

src/main/scala/org/broadinstitute/dsde/firecloud/MethodsClient.scala
rushtong: Will update to "non-failure" status codes.

davidangb:
Re-review complete:
* 2 comments could/should be re-worded
* Need to spawn a new story around logging configuration

Other than that, ready to go!

src/main/scala/org/broadinstitute/dsde/firecloud/MethodsClient.scala
rushtong: I'm probably mis-understanding, but I thought a responseFuture that received a 500 error will fall into the Failure() bucket. If that's wrong, then all 500's fall into the default response.status match case.

davidangb: I just tested this by changing the methods.baseUrl to /configurations/a/b/c, which causes agora to throw an error. The 500 from agora is still being caught in the Success block, under "case _".

Think of it this way - the Success/Failure of the Future is completely about whether or not the code specified in the Future ran without error. When spray-client calls out to agora for a response, it did successfully get a response, so we enter the Success case. That's where we actually look at what is IN the response ... and oh look! The response was an http error!

PR #2, opened 2015-06-25
Implement create workspace api

rushtong:
First pass at implementing a create workspace api call. 
User Story: https://broadinstitute.atlassian.net/browse/DSDEEPB-429
Implementation Story: https://broadinstitute.atlassian.net/browse/DSDEEPB-652

Some notes on what I did and why:

* The workspace entity currently requires an attribute map, creation date and created by fields. That is subject to change at the rawls layer in that creation and creator may be eliminated. The attribute map functionality is not needed for this story - fleshing that out should be in it's own task. See the rawls workspace case class for the complexity there.
* I moved the implicit model jsons to their own object class. I was having a lot of trouble with DefaultJsonProtocol imported in different places - I'm pretty sure there were some duplicate implicit conflicts in that case. 
* Currently pulls the create user from the authentication token. This will need to change if the rawls API is updated.
* I split out the mock servers into separate start-up/shutdowns for efficiency reasons. I'm thinking that the test classes will generally only need one or the other server to do any real work. That might not be an accurate assumption though, but I think I prefer the more granular approach so that if tests do need more than one server, they can start up whichever ones they want. 
* While I was in there, I removed the akka block from logback.xml and added it to the config. As per a different story, if you want to see logging in an actor class, you have to adjust both the akka block in the config file and the logger in logback.xml. Bit of a pain, but actor classes are funneled through both of those configurations.
* Added a callback handler for the mock test framework that returns different responses depending on the validity of the incoming workspace ingest object. Should consider this an example of how to handle custom return values on the same path but with variable inputs.


src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -0,0 +1,64 @@
...
+    httpMethod = "POST",
+    produces = "application/json",
+    consumes = "application/json",
+    response = classOf[WorkspaceEntity],
+    notes = "response is a the workspace created by the workspace service")
+  @ApiImplicitParams(Array(
+    new ApiImplicitParam(name = "body", required = true, dataType = "org.broadinstitute.dsde.firecloud.model.WorkspaceIngest", paramType = "body", value = "Workspace to create")
+  ))
+  @ApiResponses(Array(
+    new ApiResponse(code = 201, message = "Successful"),
```
davidangb: Created, not Successful

src/main/resources/application.conf
```diff
@@ -23,7 +27,11 @@ swagger {
 }
 
 methods {
-  baseUrl="https://agora-ci.broadinstitute.org/methods"
+  baseUrl="https://agora-prod.broadinstitute.org/methods"
+}
+
+workspace {
+  workspacesUrl="https://rawls-dev.broadinstitute.org/workspaces"
```
davidangb: we might want to separate these out into base server and path before we get too far ... (a la https://github.com/broadinstitute/vault-api/blob/develop/application.conf)

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -0,0 +1,64 @@
...
+    new ApiResponse(code = 401, message = "Unauthorized"),
+    new ApiResponse(code = 500, message = "Internal Error")))
+  def createWorkspaceRoute: Route =
+    path(ApiPrefix) {
+      post {
+        entity(as[WorkspaceIngest]) { ingest =>
+          commonNameFromOptionalCookie() { username =>
```
davidangb: add a //TODO here to remove this directive as soon as workspace service supports it. We don't want the overhead of the extra calls.

src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceServiceSpec.scala
```diff
@@ -0,0 +1,80 @@
...
+
+  "WorkspaceService" - {
+
+    val openAMSession = OpenAMSession(()).futureValue(timeout(Span(5, Seconds)), interval(scaled(Span(0.5, Seconds))))
+    val token = openAMSession.cookies.head.content
+    val workspaceIngest = WorkspaceIngest(
+      name = Some(MockServers.randomAlpha()),
+      namespace = Some(MockServers.randomAlpha()))
+
+    "when calling POST on the workspaces path with a valid WorkspaceIngest" - {
```
davidangb: I'd love to see an additional test that POSTs an invalid/incomplete WorkspaceIngest.

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -0,0 +1,64 @@
...
+  def receive = runRoute(createRoute)
+}
+
+@Api(value = "/workspaces", description = "Workspace Services", produces = "application/json")
+trait WorkspaceService extends HttpService with FireCloudDirectives {
+
+  private final val ApiPrefix = "workspaces"
+  private implicit val executionContext = actorRefFactory.dispatcher
+
+  val createRoute = createWorkspaceRoute
```
davidangb: for each service class, let's create a routes variable, and run all tests against all routes. This avoids test problems due to incorrect ordering of routes. See https://github.com/broadinstitute/vault-datamanagement/commit/07bbc6acd04e172766e6e441cc64ac11a5c3f9d7.

davidangb:
Review complete, with pretty small comments. However, I haven't gotten this to successfully run against the non-mock workspace service yet, and I want to see that working before merge. Will discuss in person tomorrow.

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
rushtong: :+1:

src/main/resources/application.conf
rushtong: :+1:

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
rushtong: :+1:

src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceServiceSpec.scala
rushtong: :+1:

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
rushtong: :+1:

src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceServiceSpec.scala
rushtong: Rawls gives us a 400 response (which is undocumented) when passed an entity with no name or namespace, but only errors out on one of the two. I'll update the service class to check for both values.

```
HttpResponse(400 Bad Request,HttpEntity(text/plain; charset=UTF-8,The request content was malformed:
Object is missing required member 'namespace'),List(Date: Fri, 26 Jun 2015 01:40:59 GMT, Server: spray-can/1.3.2, Content-Type: text/plain; charset=UTF-8, Content-Length: 80, Connection: close),HTTP/1.1)
The request content was malformed:
Object is missing required member 'namespace'
```

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -0,0 +1,75 @@
...
+    new ApiResponse(code = 500, message = "Internal Error")))
+  def createWorkspaceRoute: Route =
+    path(ApiPrefix) {
+      post {
+        entity(as[WorkspaceIngest]) { ingest =>
+          // TODO: Revisit getting the username from the auth token if the rawls service removes it from their API.
+          commonNameFromOptionalCookie() { username =>
+            respondWithJSON { requestContext =>
+              ingest match {
+                case x if x.name.isEmpty || x.namespace.isEmpty =>
```
davidangb: I love that you went the extra mile and added this validation. BUT ... I'd rather keep this kind of logic in the workspace service; they own that domain, and we should just pass through to them. If they are not validating these yet, let's disable our test (with a TODO to re-enable when necessary).

This means you'd also have to add a 400 response to the mock server.

davidangb:
Looking good, one comment, so close!

PR #3, opened 2015-06-26
Adds CORS headers to orchestration layer

dmohs:
Task: https://broadinstitute.atlassian.net/browse/DSDEEPB-660

This is a provisional implementation to document the headers and help unblock local development. There is a follow-up task to refactor this into a correct implementation:
https://broadinstitute.atlassian.net/browse/DSDEEPB-667

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/MethodsClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/MethodsClient.scala
@@ -5,8 +5,9 @@ import akka.event.Logging
 import org.broadinstitute.dsde.firecloud.MethodsClient.MethodsListRequest
 import org.broadinstitute.dsde.firecloud.model.MethodEntity
 import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
+import scala.reflect.classTag
 import spray.client.pipelining._
-import spray.http.HttpHeaders.Cookie
+import spray.http.HttpHeaders.{Cookie, Origin, RawHeader}
 import spray.http.StatusCodes._
 import spray.http.{HttpRequest, HttpResponse, RequestProcessingException, StatusCodes}
 import spray.httpx.SprayJsonSupport._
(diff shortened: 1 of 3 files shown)


rushtong:
FYI, this is based off of my incomplete branch that still needs to be squashed. I would suggest either waiting for that one to be sqashed/merged and then rebasing, or perhaps rebase this off of current master instead.

davidangb:
Agreed - could you please rebase this off current master to reduce (redundant) commits?

dmohs:
Done. Thank you.

rushtong:
My apologies @dmohs - I did a force push which probably means you have to rebase again.

dmohs:
Not a problem, Greg. Thanks for letting us know.

src/main/scala/org/broadinstitute/dsde/firecloud/MethodsClient.scala
```diff
@@ -43,24 +44,30 @@ class MethodsClient(requestContext: RequestContext) extends Actor {
       addHeader(Cookie(requestContext.request.cookies)) ~> sendReceive
 
     val responseFuture: Future[HttpResponse] = pipeline { Get(FireCloudConfig.Methods.methodsListUrl) }
+    val corsHeaders = List(
+      RawHeader("Access-Control-Allow-Credentials", "true"),
+      RawHeader("Access-Control-Allow-Origin", "http://local.broadinstitute.org:8000")
```
davidangb: how about allowing * instead of local.b.o? That would allow https://firecloud-ci.broadinstitute.org/ to work as well. Since we require OpenAM authentication, we're protected.

Additionally, looks like we need to add a Vary: Origin header - see https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Access-Control-Allow-Origin

davidangb:
Review done - one comment inline. I agree that this should be refactored, but I'm okay waiting on DSDEEPB-667 for that.

dmohs:
Thanks for the review, @davidangb. Comments addressed.

dmohs:
Here is another version based on our HipChat discussion.

src/main/scala/org/broadinstitute/dsde/firecloud/MethodsClient.scala
```diff
@@ -44,23 +45,41 @@ class MethodsClient(requestContext: RequestContext) extends Actor {
 
     val responseFuture: Future[HttpResponse] = pipeline { Get(FireCloudConfig.Methods.methodsListUrl) }
 
+    def isWhitelisted(origin: Origin): Boolean = {
+      // TODO(dmohs): Check this against a whitelist.
+      true
+    }
+
+    def addCorsHeaders(requestContext: RequestContext): RequestContext = {
```
davidangb: no need to change for this PR. As a comment for the longer term (see DSDEEPB-667), I think we'll want to create a custom spray directive for this, likely built on top of spray's optionalHeaderValueByName directive. The custom directive would look for an Origin header, compare it to a whitelist, and add response headers as appropriate.

See https://github.com/broadinstitute/vault-api/commit/32b3f7f3f11a00ae80c0686152485cf14f2e2690 for an example of wrapping all (most) routes in a custom directive - in that case it was logOpenAMRequest.

PR #5, opened 2015-06-30
Add list workspaces method

dmohs:
Also factors out some of the CORS stuff.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/WorkspaceClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/WorkspaceClient.scala
@@ -3,16 +3,18 @@ package org.broadinstitute.dsde.firecloud
 import java.text.SimpleDateFormat
 import java.util.Date
 
-import akka.actor.{Actor, Props}
+import akka.actor.{Actor, ActorRef, Props}
 import akka.event.Logging
-import org.broadinstitute.dsde.firecloud.WorkspaceClient.WorkspaceCreate
+import org.broadinstitute.dsde.firecloud.WorkspaceClient.{WorkspaceCreate, WorkspacesListRequest}
 import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
 import org.broadinstitute.dsde.firecloud.model.{WorkspaceEntity, WorkspaceIngest}
+import org.broadinstitute.dsde.firecloud.service.ServiceUtils
 import spray.client.pipelining._
 import spray.http.HttpHeaders.Cookie
 import spray.http.StatusCodes._
 import spray.http.{HttpRequest, HttpResponse, RequestProcessingException, StatusCodes}
 import spray.httpx.SprayJsonSupport._
+import spray.json.DefaultJsonProtocol._
 import spray.routing.RequestContext
 
 import scala.concurrent.Future
@@ -22,6 +24,7 @@ import scala.util.{Failure, Success}
 object WorkspaceClient {
 
   case class WorkspaceCreate(workspaceIngest: WorkspaceIngest, username: Option[String])
+  case class WorkspacesListRequest()
 
   def props(requestContext: RequestContext): Props = Props(new WorkspaceClient(requestContext))
 
@@ -39,6 +42,7 @@ class WorkspaceClient (requestContext: RequestContext) extends Actor {
 
     case WorkspaceCreate(workspaceIngest: WorkspaceIngest, username: Option[String]) =>
       createWorkspace(workspaceIngest, username)
+    case WorkspacesListRequest => listWorkspaces(sender())
 
   }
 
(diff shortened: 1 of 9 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/MethodsService.scala
```diff
@@ -17,10 +17,19 @@ trait MethodsService extends HttpService with FireCloudDirectives {
 
   private final val ApiPrefix = "methods"
 
-  val routes = listMethodsRoute
+  val routes = { optionsRoute ~ listMethodsRoute }
```
davidangb: The braces shouldn't be necessary here. Stylistically, I'd leave 'em out - willing to hear otherwise.

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -22,10 +22,19 @@ trait WorkspaceService extends HttpService with FireCloudDirectives {
   private final val ApiPrefix = "workspaces"
   private implicit val executionContext = actorRefFactory.dispatcher
 
-  val routes = createWorkspaceRoute
+  val routes = {optionsRoute ~ createWorkspaceRoute ~ listWorkspacesRoute}
 
   lazy val log = LoggerFactory.getLogger(getClass)
 
+  def optionsRoute: Route =
```
davidangb: Any reason to not swaggerize this (and the equivalent in MethodsService)? We may only use this for cors preflighting, but might as well be transparent that it exists.

src/test/scala/org/broadinstitute/dsde/firecloud/MockServers.scala
```diff
@@ -45,6 +45,18 @@ object MockServers {
       attributes = Some(Map.empty)
     )
   }
+  val mockWorkspaceEntities: List[WorkspaceEntity] = {
+    List.tabulate(randomPositiveInt())(
+    n =>
+      WorkspaceEntity(
+      namespace = Some(randomAlpha()),
+      name = Some(randomAlpha()),
+      createdDate = Some(randomAlpha()),
```
davidangb: make this a date, not a string - e.g. SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZZ").format(new Date()). Randomize if you feel like it - I don't think the randomization is important right now.

src/main/scala/org/broadinstitute/dsde/firecloud/WorkspaceClient.scala
```diff
@@ -3,12 +3,14 @@ package org.broadinstitute.dsde.firecloud
...
-import akka.actor.{Actor, Props}
+import akka.actor.{Actor, ActorRef, Props}
 import akka.event.Logging
-import org.broadinstitute.dsde.firecloud.WorkspaceClient.WorkspaceCreate
+import org.broadinstitute.dsde.firecloud.WorkspaceClient.{WorkspaceCreate, WorkspacesListRequest}
 import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
 import org.broadinstitute.dsde.firecloud.model.{WorkspaceEntity, WorkspaceIngest}
+import org.broadinstitute.dsde.firecloud.service.ServiceUtils
 import spray.client.pipelining._
+import spray.json.DefaultJsonProtocol._
```
davidangb: give this a command-option-O!

src/main/scala/org/broadinstitute/dsde/firecloud/service/ServiceUtils.scala
```diff
@@ -0,0 +1,67 @@
...
+import spray.http.{HttpRequest, HttpResponse, RequestProcessingException, StatusCodes}
+import spray.routing.RequestContext
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.reflect._
+import scala.util.{Failure, Success}
+
+object ServiceUtils {
+
+  def listEntities(
```
davidangb: I like factoring this out into a common def. Though ...
1) This could be used for just about any get request; it doesn't have to be just for listing entities. Might want to rename it.
2) I could imagine passing in a return type for unmarshalling, and providing a default implementation of completeSuccessfully. Callers would only implement completeSuccessfully if they needed custom logic.
3) I could even imagine passing in the request method (i.e. get vs post) and an optional post payload.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ServiceUtils.scala
```diff
@@ -0,0 +1,67 @@
...
+      log: LoggingAdapter,
+      context: ActorContext,
+      url: String,
```
davidangb: instead of passing all these arguments into the method, either create a new case class with those args as members and pass the case class, OR make those args members of ServiceUtils, then instantiate a new ServiceUtils when you want to run it.

dmohs:
Comments addressed. Please squash before merging.

PR #8, opened 2015-07-10
Lighter request handling

dmohs:
Passes requests to the remote services to handle them and returns the responses unmodified. This was the only way I could figure out how to accept the "text/plain" content type, which is nice for easy curling.
Also removes CORS headers since we no longer need them.


src/main/scala/org/broadinstitute/dsde/firecloud/service/MethodsService.scala
```diff
@@ -1,60 +1,36 @@
...
+    notes = "The response is forwarded unmodified from the methods repository.")
   def listMethodsRoute: Route =
     path(ApiPrefix) {
-      get {
-        respondWithJSON { requestContext =>
-          val methodsClient = actorRefFactory.actorOf(Props(new MethodsClient(requestContext)))
-          methodsClient ! MethodsClient.MethodsListRequest
-        }
+      get { requestContext =>
+        val workspaceClient = actorRefFactory.actorOf(Props(new HttpClient(requestContext)))
```
rushtong: The client here is retrieving methods, no? Might be better to keep it named as such.

dmohs:
Greg, Thank you for catching that. I refactored a bit to make that type of copy-paste error less likely.

src/main/scala/org/broadinstitute/dsde/firecloud/service/MethodsService.scala
```diff
@@ -1,60 +1,35 @@
...
 
+import org.broadinstitute.dsde.firecloud.{FireCloudConfig, HttpClient}
+
 class MethodsServiceActor extends Actor with MethodsService {
   def actorRefFactory = context
   def receive = runRoute(routes)
 }
 
-@Api(value = "/methods", description = "Methods Service", produces = "application/json")
+@Api(value = "/methods", description = "Methods Service", produces = "application/json, text/plain")
```
rushtong: I don't think it is true that text/plain can be returned from agora for the ```/methods``` call. It can be in the get ```/methods/{namespace}/{name}/{snapshotId}``` call, but I think it would be better to annotate it on that service call.

bradtaylor: Correct. We return application/json everywhere except:
/methods/{namespace}/{name}/{snapshotId} where onlyPayload=true.

This is so we can return bare WDL as text/plain to Execution Service

rushtong:
Summary of my initial thoughts:

* I think that by extracting all client requests to a single class introduces an architectural restriction that doesn't feel right to me. By having each remote service accessed by a dedicated client, we gain the ability to react to idiosyncrasies of the remote server. We can channel different responses, un-marshal responses, validate if need be, etc. By moving all of that into a single class, we lose most of that, if not all. It also channels future service client development into using that class which will mean its complexity will only grow over time in an effort to handle *all* remote servers. I think the vault team had a cleaner approach with dedicated clients and a service handler that sat between the route and the client if there needed to be any non-trivial work done between the two. 

* I think we have lost the respond-with-json in the refactoring. 

* We've removed a lot of annotation from the routes which cuts back on the amount of information in the swagger UI. That seems like the wrong approach to me. By just returning the raw response from the remote servers, we're pushing a lot of work to the firecloud-UI layer to handle everything.

rushtong:
I'm finished reviewing, but I'm not exactly sure where to go from here.

src/main/scala/org/broadinstitute/dsde/firecloud/service/MethodsService.scala
dmohs: Even in error conditions? Are there any servers in front of Agora that could potentially return text in an error condition? Or maybe I'm looking at this the wrong way in that it doesn't have to be correct all the time just most of the time.

dmohs:
@rushtong I generally agree with your comments. Let me address each one:
* "It also channels future service client development into using that class..." I don't necessarily agree with this. I would argue that it makes sense to create a separate client when service-specific functionality is needed. However, I don't see a need for that at the moment.
* The intent (for now) is to respond with service's response, including the content type.
* I still think this is better, even though I agree that it provides less information in the docs. In my testing, the services themselves return better error information than we do, so until we match that information in quality (which we may want to do, but that is a longer discussion), we should simply document that we return the raw response. Previously, the docs were actually providing incorrect information, which is worse than providing less information in my opinion.

PR #9, opened 2015-07-14
added entity service endpoints

MatthewBemis:


build.sbt
```diff
@@ -63,6 +63,8 @@ libraryDependencies ++= {
 
 Revolver.settings.settings
 
+Revolver.enableDebugging(port = 5051, suspend = false)
```
dmohs: This is awesome. I was wondering how to do this. Two things: (1) Will this cause debugging to be enabled when this is deployed to production, and (2) can you call this out in the README so developers are aware it exists?

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -0,0 +1,66 @@
...
+class EntityClient (requestContext: RequestContext) extends Actor {
+
+  import system.dispatcher
+
+  implicit val system = context.system
+  val log = Logging(system, getClass)
+  val format = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZZ")
+
+  override def receive: Receive = {
+    case EntityListRequest(workspaceNamespace: String, workspaceName: String, entityType: String) => listEntities(workspaceNamespace, workspaceName, entityType)
```
dmohs: Nit: Our style conventions specify a 100-column max line length. Please fix all occurrences.

dmohs:
This looks good, but there are no tests. I'm fine with merging this without tests provided we track that task somewhere.

src/main/resources/application.conf
```diff
@@ -36,6 +36,11 @@ workspace {
   workspacesPath="/workspaces"
 }
 
+entities {
+  baseUrl="https://rawls-dev.broadinstitute.org"
+  entitiesPath="/entities"
```
rushtong: This should be in the workspace configuration block since it is pointing to rawls.

rushtong:
I would have added the route in EntityClient as another route in WorkspaceClient since that class points to everything rawls-like. I think this is fine as is though - we are having some internal discussion about architecture that will likely render this point moot.

src/main/scala/org/broadinstitute/dsde/firecloud/service/EntityService.scala
```diff
@@ -0,0 +1,78 @@
...
+        }
+      }
+    }
+
+  @Path("/{entityType}")
+  @ApiOperation(value = "list all entities of given type in a workspace",
+    nickname = "listEntities",
+    httpMethod = "GET",
+    produces = "application/json",
+    responseContainer = "List",
```
rushtong: It is nice to have a response class defined as well so that the swagger UI can populate an example data structure. That would require a case class that defines Entity.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -0,0 +1,66 @@
...
+      context,
+      formUrl(workspaceNamespace, workspaceName) + "/" + entityType,
+      requestContext,
+      completeSuccessfully
+    ))
+  }
+
+  def formUrl(workspaceNamespace: String, workspaceName: String): String = {
+    s"${FireCloudConfig.Workspace.baseUrl}/workspaces/${workspaceNamespace}/${workspaceName}/entities"
+  }
```
rushtong: This should be broken out in a method in FireCloudConfig as something like this.

In the application config, the entity path should be something like
```
entitiesPath="/workspace/%s/%s/entities"
```
and in FireCloudConfig
```
def entityPathFromWorkspace(namespace: String, name: String) = baseUrl + entitiesPath.getString("entitiesPath).format(namespace, name)
```
That way, your service client can call the url with

```
FireCloudConfig.Workspace.entityPathFromWorkspace(namespace, name)
```

We don't have any examples of this in firecloud, but there are lots of examples in the configuration class of https://github.com/broadinstitute/vault-api

rushtong:
Given that this PR describes entities in a workspace, we should probably refactor our internal WorkspaceEntity class as the naming is getting confusing. Entity should likely still be Entity, but WorkspaceEntity should become Workspace instead. Subject for a follow-on story, not for this though.

Setting up a mock server for this will require a bit of refactoring of the current workspace service handlers due to entities being a sub-path under workspace. We will definitely need to handle that, but perhaps also for a different PR.

build.sbt
rushtong: On production, the system should not be started in this way - so this should not be a problem. The only case where this comes up is if you run re-start/re-stop in the sbt console.

src/main/scala/org/broadinstitute/dsde/firecloud/service/EntityService.scala
```diff
@@ -0,0 +1,78 @@
...
+
+  @ApiOperation(
+    value = "entities options",
+    nickname = "entitiesOptions",
+    httpMethod = "OPTIONS",
+    notes = "response is an OPTIONS response")
+  @ApiResponses(Array(
+    new ApiResponse(code = 200, message = "OK")))
+  def optionsRoute: Route =
+    path(ApiPrefix) {
```
rushtong: I'm not following this. How is the application handling "/entities" but the swagger @Api is using "/workspaces/{workspaceNamespace}/{workspaceName}/entities"? Doesn't this route need to handle "/entities" as a path suffix of some sort, along with a path prefix of "/workspace"?

rushtong: Never mind ... I see what this is doing - it's only for the options request. In that case - it is confusing to me to have the top-level /entities path handled at all. Since an entity is only nested inside a workspace, it makes more sense to me to have this use the same path as handled below (path("workspaces" / Segment / Segment / "entities")) which would work for any provided namespace/name segment provided, or even shorten that to "workspaces/entities".

rushtong:
Added follow-on story for adding tests:
https://broadinstitute.atlassian.net/browse/DSDEEPB-824

PR #13, opened 2015-07-17
Send uploaded entities to Rawls 

mattsooknah:
Don't merge this yet - just looking for comments.

This code finishes the basic entity upload functionality. The JSON form data is parsed into a ```Seq[Entity]```. Each entity is sent in a separate ```POST``` to Rawls, and we get back a response for each one. The response messages, and whether each was a success or failure, are wrapped up in an ```Seq[EntityCreateResult]``` which could then be displayed in the UI somehow.


src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -62,7 +73,30 @@ class EntityClient (requestContext: RequestContext) extends Actor {
         requestContext.failWith(
           new RequestProcessingException(StatusCodes.InternalServerError, error.getMessage))
     }
+  }
 
+  def createEntities(wsNamespace: String, wsName: String, entities: Seq[Entity]) = {
+    val results = entities.map(entity => createEntityOrReportError(wsNamespace, wsName, entity))
+    // have to manually construct an HttpResponse here because we're aggregating the results of multiple calls to Rawls
+    val response = HttpResponse(OK, HttpClient.createJsonHttpEntity(results.toJson.compactPrint))
```
mattsooknah: Not sure what the right thing to do here is - should we return ```OK``` for the batch operation even if there are some individual failures, or only if everything succeeds? And does ```OK``` or ```Created``` make more sense in that case?

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -62,7 +73,30 @@ class EntityClient (requestContext: RequestContext) extends Actor {
...
+    requestContext.complete(response)
+  }
+
+  def createEntityOrReportError(workspaceNamespace: String, workspaceName: String, entity: Entity): EntityCreateResult = {
+    val url = s"${FireCloudConfig.Workspace.entityPathFromWorkspace(workspaceNamespace, workspaceName)}"
+    val externalRequest = Post(url, HttpClient.createJsonHttpEntity(entity.toJson.compactPrint))
+    val pipeline = addHeader(Cookie(requestContext.request.cookies)) ~> sendReceive
+    // TODO add a timeout?
+    // TODO is there a way to return the EntityCreateResult from within an onComplete, instead of doing this?
+    Try(Await.result(pipeline(externalRequest), Duration.Inf)) match {
```
mattsooknah: Since this function returns an explicit object (```EntityCreateResult```) based on the contents of the response from Rawls, I used ```Await.result``` instead of  the usual ```onComplete```. If there's a better / more idiomatic solution, I'm happy to change this.

src/main/scala/org/broadinstitute/dsde/firecloud/model/Entity.scala
```diff
@@ -11,12 +11,12 @@ import scala.annotation.meta.field
...
                               @(ApiModelProperty@field)(required = true, value = "The type of the entity")
-                              entityType: Option[String] = None,
+                              entityType: String,
                               @(ApiModelProperty@field)(required = true, value = "The name of the entity")
-                              entityName: Option[String] = None,
+                              entityName: String,
                               @(ApiModelProperty@field)(required = true, value = "The attributes of the entity")
-                              attributes: Option[Map[String, String]] = None)
\ No newline at end of file
+                              attributes: Map[String, String])
```
mattsooknah: I de-Optionized these because it doesn't really make sense, at least from the Rawls perspective, to have an entity without e.g. an entityType or entityName. I've noticed that all of the model classes seem to be fully Optioned, though, so maybe there's a deeper reason for it?

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
rushtong: I think you need an intermediary service handler to track entity handling at the individual level. If all succeed, then definitely `Created` should be the return, potentially with a list of all of the created entities or ids, whatever makes the most sense.

For the negative case, track any failures and possibly return an error with the reasons for the failures. I think a 422, "Unprocessable Entity" might be the most logical response here. The service handler should likely call delete on any of the successfully created entities so as to clean up after itself.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -62,7 +73,30 @@ class EntityClient (requestContext: RequestContext) extends Actor {
...
 
+  def createEntities(wsNamespace: String, wsName: String, entities: Seq[Entity]) = {
+    val results = entities.map(entity => createEntityOrReportError(wsNamespace, wsName, entity))
+    // have to manually construct an HttpResponse here because we're aggregating the results of multiple calls to Rawls
+    val response = HttpResponse(OK, HttpClient.createJsonHttpEntity(results.toJson.compactPrint))
+    requestContext.complete(response)
+  }
+
+  def createEntityOrReportError(workspaceNamespace: String, workspaceName: String, entity: Entity): EntityCreateResult = {
+    val url = s"${FireCloudConfig.Workspace.entityPathFromWorkspace(workspaceNamespace, workspaceName)}"
```
rushtong: I don't think you need to wrap the config call with `s""`.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -62,7 +73,30 @@ class EntityClient (requestContext: RequestContext) extends Actor {
...
+    val externalRequest = Post(url, HttpClient.createJsonHttpEntity(entity.toJson.compactPrint))
+    val pipeline = addHeader(Cookie(requestContext.request.cookies)) ~> sendReceive
+    // TODO add a timeout?
```
rushtong: Time-outs should be handled at the app layer - I'd be hesitant to add them here unless *really* needed.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -62,7 +73,30 @@ class EntityClient (requestContext: RequestContext) extends Actor {
...
+    // TODO is there a way to return the EntityCreateResult from within an onComplete, instead of doing this?
+    Try(Await.result(pipeline(externalRequest), Duration.Inf)) match {
+      case Success(response) => response.status match {
+        case StatusCodes.Created => EntityCreateResult(entity.entityType, entity.entityName, true, "Entity created successfully")
+        case _ => EntityCreateResult(entity.entityType, entity.entityName, false, s"Bad response from workspace service: ${response.message}")
+      }
+      // TODO how should we handle the case where an exception is thrown?
```
rushtong: There is an example in this class for handling exceptions. Update the logging message if you would like to see more detail
```
      case Failure(error) =>
        // Failure accessing service
        log.error(error, "Service API call failed")
        requestContext.failWith(
          new RequestProcessingException(StatusCodes.InternalServerError, error.getMessage))
```

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
rushtong: Since we're dealing with actors, I am hesitant to block with Await. I think using onComplete is more idiomatic in this case.

src/main/scala/org/broadinstitute/dsde/firecloud/model/Entity.scala
rushtong: Using option-ed values is generally a lot easier to deal with when working with potential null values, matching blocks, etc.

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -92,22 +95,22 @@ trait WorkspaceService extends HttpService with FireCloudDirectives {
     value = "import entities (JSON)",
     nickname = "importEntitiesJSON",
     httpMethod = "POST",
-    response = classOf[EntityCreateResult],
+    response = classOf[Seq[EntityCreateResult]],
```
rushtong: You could keep the  response = classOf[EntityCreateResult] and add another line for responseContainer = "Seq" instead of merging them. 

On a somewhat related note, we have not been adding these to our orchestration services due to the high churn on the model classes. Also, given that there is a lot of code duplication since we have to re-write the case class in orchestration to match what's in other service endpoints - we have been avoiding that and waiting for some kind of dsde-commons code to come about so we could reference the model classes there. Feel free to approach this as you see fit - we don't have much insight into the level of model churn in this case.

src/test/resources/test.conf
```diff
@@ -34,6 +34,7 @@ methods {
 workspace {
   baseUrl="http://localhost:8990"
   workspacesPath="/workspaces"
+  entitiesPath="/workspaces/%s/%s/entities"
```
rushtong: Good catch - that wasn't in the last entity PR so it was on my list to add it in test.conf.

src/test/scala/org/broadinstitute/dsde/firecloud/MockServers.scala
```diff
@@ -136,6 +184,47 @@ object MockServers {
...
+        request()
+          .withMethod("POST")
+          .withPath(s"/workspaces/${mockSampleConflict.wsNamespace}/${mockSampleConflict.wsName}/entities")
+          .withBody(mockSampleConflict.toJson.compactPrint)
+          .withCookies(cookie)
+      ).respond(
+        response()
+          .withHeaders(header)
+          .withStatusCode(Conflict.intValue)
+      )
```
rushtong: Nice additions to the mock cases. We should add some more to handle the case where no cookie is provided and return an auth error response.

rushtong: We might want to have another mock case for handling invalid entity models too. That is done with a callback class - see `ValidWorkspaceCallback`.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -62,7 +73,30 @@ class EntityClient (requestContext: RequestContext) extends Actor {
...
+    val response = HttpResponse(OK, HttpClient.createJsonHttpEntity(results.toJson.compactPrint))
+    requestContext.complete(response)
+  }
+
+  def createEntityOrReportError(workspaceNamespace: String, workspaceName: String, entity: Entity): EntityCreateResult = {
+    val url = s"${FireCloudConfig.Workspace.entityPathFromWorkspace(workspaceNamespace, workspaceName)}"
+    val externalRequest = Post(url, HttpClient.createJsonHttpEntity(entity.toJson.compactPrint))
+    val pipeline = addHeader(Cookie(requestContext.request.cookies)) ~> sendReceive
+    // TODO add a timeout?
+    // TODO is there a way to return the EntityCreateResult from within an onComplete, instead of doing this?
```
rushtong: Here is an example from some code that has been refactored away. Look for the MethodsClient class in this commit: https://github.com/broadinstitute/firecloud-orchestration/commit/5557d8b7a743ba16ce606b544b48c1b160b83347#diff-e9a809a0f437e83f07e2870935cd480d

```
responseFuture onComplete {
      case Success(response) =>
        response.status match {
          case OK =>
            log.debug("OK response")
            requestContext.complete(unmarshal[List[MethodEntity]].apply(response))
            context.stop(self)
          case _ =>
            // Bubble up all other unmarshallable responses
            log.warning("Unanticipated response: " + response.status.defaultMessage)
            requestContext.complete(response)
            context.stop(self)
        }
      case Failure(error) =>
        // Failure accessing service
        log.error(error, "Could not access the methods repository")
        requestContext.failWith(new RequestProcessingException(StatusCodes.InternalServerError, error.getMessage))
        context.stop(self)
```

rushtong:
@mattsooknah - I think that's as much feedback as I can muster at the moment. Feel free to hipchat/google-chat me if you want to.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
abaumann: We will be implementing the full featured upload next sprint (DSDEEPB-4), this is just a first pass effort we could finish in this sprint.  For now, we don't need to worry about clean up, or whether or not all succeeded.  I think if we just do OK upon successful call and return success/failure for each entity.  In the next sprint we will be adding a batch create/update method that will work on a single transaction and rollback any changes if there are failures.

src/main/scala/org/broadinstitute/dsde/firecloud/model/Entity.scala
abaumann: I know Matt had removed the options so he didn't have to have a check for every required param to make sure its not a None - do you foresee some sort of automated way that Options would be verified against required parameters?

src/test/scala/org/broadinstitute/dsde/firecloud/MockServers.scala
abaumann: In Rawls we had the issue of having to implement mock functionality for every service call we provide, which was a lot of duplicated effort - this was instead changed to using the existing functionality with an in-memory database.  I'd like to see if it would be possible to also use Rawls as a mock in-memory service with in-memory database so that we could avoid having to mock all Rawls functionality here if that makes sense.

src/main/scala/org/broadinstitute/dsde/firecloud/model/Entity.scala
rushtong: To clarify here - the original code is Matt Bemis', but I agree in general with using optioned values. Orchestration sits between the UI and the remote services so being able to enforce some manner of business logic is one of the goals for this layer. Having optioned values makes that job a bit easier, IMHO.

src/test/scala/org/broadinstitute/dsde/firecloud/MockServers.scala
rushtong: We did not want to be hitting the live servers for all of our testing, which is why we have the mock-server approach. Hitting them live means we are effectively integration testing rather than unit testing.

abaumann: I don't mean any sort of live server, I'm wondering if we can mock by instead running a local service using an in-memory database during the life of the unit test - it may turn out this isn't very feasible to do, but I just worry about mocking out the functionality the same as Rawls provides and running unit tests that pass despite changes to Rawls that would break integration tests

abaumann: and what I mean is the orchestration layer unit tests would spin up this "mock" server

rushtong: If I understand correctly, you're suggesting the orchestration test harness would spin up the latest rawls master branch and run that locally? I do think that is feasible, if a bit of a coordination challenge wrt keeping orchestration's remote server configurations in sync with rawls'. I'll bring that up with the team at today's standup.

abaumann: Yes - I think it's just something to discuss with the team as you suggest and if we all think this makes sense then coordinate some work on doing so - I think if all we need are server configs to keep in sync that is significantly less to deal with that all the mock functionality

rushtong: The other approach that I think is worthwhile is moving mock server functionality to a common dsde library. Everyone could use the same mocked responses in that way. If we did have that, we could put common models there too. Entity, EntityCreateResult, Method, etc. should all be there so we might all be able to reduce duplicated efforts.

abaumann: Also a great idea - you could rely on that for mock functionality, and whoever provides the mock (for us presumably we would provide Rawls mock services) we would implement that however we see fit - which would ideally be using the existing code.  I like the idea of all of these mocks living in a common place along with the models

mattsooknah:
I've updated the PR:
1) ```Entity``` has been re-Optionized, for now. I added an extra check to ```createEntityOrReportError``` to make sure that ```entityType``` and ```entityName``` are defined before trying to post to Rawls.
2) The ```Await.result``` code has been left in for now (with a TODO), since as per our discussion today, there's not a good way to aggregate the result of multiple calls to Rawls (some subset of which might fail) using just ```requestContext.complete```.

mattsooknah:
Not sure what you guys want to do with this PR, since the batch functionality is going to get implemented directly in Rawls in the upcoming sprint. If you do plan to merge it and there are any other changes you'd like me to make, let me know,

PR #14, opened 2015-07-20
Es dsdeepb 816

eddiebroad:
this branch is dedicated to implementing subtask DSDEEPB-816 (Create "list method configurations" endpoint in orchestration layer) of story DSDEEPB-10 (View method configurations in my workspace).
Greg R's "tour" helped give me background!  Thank you Greg!

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -35,6 +35,7 @@ workspace {
   baseUrl="https://rawls-dev.broadinstitute.org"
   workspacesPath="/workspaces"
   entitiesPath="/workspaces/%s/%s/entities"
+  methodConfigsListPath="/workspaces/%s/%s/methodconfigs"
 }
 
 openam {
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
@@ -37,8 +37,10 @@ object FireCloudConfig {
     private val workspace = config.getConfig("workspace")
     lazy val baseUrl= workspace.getString("baseUrl")
     lazy val workspacesPath = workspace.getString("workspacesPath")
+    lazy val methodConfigsListPath=workspace.getString("methodConfigsListPath")
     lazy val workspaceCreateUrl = baseUrl + workspacesPath
     lazy val workspacesListUrl = baseUrl + workspacesPath
+    lazy val listMethodConfigurationsUrl=baseUrl+methodConfigsListPath
     def entityPathFromWorkspace(namespace: String, name: String) = baseUrl + workspace.getString("entitiesPath").format(namespace, name)
   }
 
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
@@ -40,7 +40,7 @@ object FireCloudConfig {
     lazy val methodConfigsListPath=workspace.getString("methodConfigsListPath")
     lazy val workspaceCreateUrl = baseUrl + workspacesPath
     lazy val workspacesListUrl = baseUrl + workspacesPath
-    lazy val listMethodConfigurationsUrl = baseUrl+methodConfigsListPath
+    lazy val listMethodConfigurationsUrl = baseUrl + methodConfigsListPath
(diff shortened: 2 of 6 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/model/Workspace.scala
```diff
@@ -30,3 +30,36 @@ case class EntityCreateResult(
   namespace: String,
   name: String,
   theJSON: String)
+
+
+//swagger annotation
+@ApiModel(value="method configuration entity")
+//scala definitions
+case class MethodConfigurationEntity(
```
rushtong: Please rename this to MethodConfiguration. We have added 'Entity' to other case class names in a confusing manner and I don't want to continue with that.

rushtong: On a second read, this class is not used. This can be removed.

src/main/resources/application.conf
```diff
@@ -35,6 +35,7 @@ workspace {
   baseUrl="https://rawls-dev.broadinstitute.org"
   workspacesPath="/workspaces"
   entitiesPath="/workspaces/%s/%s/entities"
+  methodConfigsListPath="/workspaces/%s/%s/methodconfigs"
```
rushtong: Also need this in test.conf as well.

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
```diff
@@ -37,8 +37,10 @@ object FireCloudConfig {
     private val workspace = config.getConfig("workspace")
     lazy val baseUrl= workspace.getString("baseUrl")
     lazy val workspacesPath = workspace.getString("workspacesPath")
+    lazy val methodConfigsListPath=workspace.getString("methodConfigsListPath")
```
rushtong: Nitpick - please use consistent spacing in assignments, i.e:
```
lazy val methodConfigsListPath = workspace.getString("methodConfigsListPath")
```

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -88,6 +88,34 @@ trait WorkspaceService extends HttpService with FireCloudDirectives {
...
+          actorRefFactory.actorOf(Props(new HttpClient(requestContext))) !
+          HttpClient.PerformExternalRequest(Get(FireCloudConfig.Workspace.listMethodConfigurationsUrl))
+        }
+      }
+
+
+
+
+
+
```
rushtong: Nitpick - lots of extra lines here - trim down single breaks.

src/main/scala/org/broadinstitute/dsde/firecloud/model/Workspace.scala
eddiebroad: I guess you mean Greg rename "MethodConfigurationEntity" to just "MethodConfiguration" ?

rushtong:
Could use a test case for this method. Re-assigning back to Eddie for minor re-work.

src/main/scala/org/broadinstitute/dsde/firecloud/model/Workspace.scala
rushtong: At that time, yes, I meant that, but it can be removed since it isn't used.

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
eddiebroad: I see you mean, for example, around the "=" here.

src/main/resources/application.conf
eddiebroad: okay

src/main/scala/org/broadinstitute/dsde/firecloud/model/Workspace.scala
eddiebroad: okay!!! even better to remove something unused!

rushtong:
@eddiebroad - I would suggest closing this PR, finishing the work that needs to be done to make the tests work, squashing all of your commits, and then submitting a new PR.

eddiebroad:
okay.  I wasn't sure if it's best or if its the customary to have one pull-request live then end in either A) 1) continued review and 2 ) merging (or closure) (depending on review outcome)   or B) end by quick resolution (quick edits then merger and close or just possibly edit and close, without merge).

PR #16, opened 2015-07-21
Batch upload entities to Rawls

mattsooknah:
Temporary upload entities solution. This code will be refactored in the next sprint.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
@@ -2,23 +2,32 @@ package org.broadinstitute.dsde.firecloud
 
 import java.text.SimpleDateFormat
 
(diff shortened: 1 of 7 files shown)


rushtong:
I get this warning when running tests"
```
[warn] /Users/grushton/develop/firecloud-orchestration/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala:94: match may not be exhaustive.
[warn] It would fail on the following input: Failure((x: Throwable forSome x not in Exception))
[warn]         Try(Await.result(pipeline(externalRequest), Duration.Inf)) match {
[warn]            ^
[warn] one warning found
```
Is that expected?

rushtong:
:+1: Tests pass aside from the warning. Is blue team OK with that? If so, merge away and we can deal with the refactoring later.

mattsooknah:
Blue team is OK with it. I will merge.

PR #17, opened 2015-07-21
Change swagger baseUrl to be at "" rather than "/api" which was causing

abaumann:
swagger to build the wrong calls


rushtong:
For consistency sake, can we add this to test.conf too? Will need to be manually set by Matt N. when it gets merged.

PR #22, opened 2015-07-23
Added TSV parser

helgridly:


src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVParser.scala
```diff
@@ -0,0 +1,31 @@
...
+object TSVParser {
+  private def parseLine( tsvLine: String, tsvLineNo: Int, headers: Seq[String] ) : Map[String, String] = {
+    val fields = tsvLine.split('\t') //note we don't trim the line here in case the last element is deliberately empty
+    if( fields.length != headers.length ) {
+      throw new RuntimeException("TSV parsing error in line " + tsvLine.toString + ": wrong number of fields")
+    }
+    headers.zip(fields).toMap
+  }
+
+  def parse( tsvString : String ) : TSVLoadFile = {
```
dvoet: are we really to be loading the whole tsv file into a string in memory?

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVParser.scala
```diff
@@ -0,0 +1,31 @@
+package org.broadinstitute.dsde.firecloud.utils
+
+/**
+ * Created with IntelliJ IDEA.
+ * User: hussein
+ * Date: 07/23/2015
+ * Time: 14:01
+ */
+
+case class TSVLoadFile( firstColumnHeader:String, tsvData:Seq[Map[String, String]] )
```
dvoet: a little more doc so people know what is in tsvData.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVParser.scala
```diff
@@ -0,0 +1,31 @@
...
+
+object TSVParser {
+  private def parseLine( tsvLine: String, tsvLineNo: Int, headers: Seq[String] ) : Map[String, String] = {
+    val fields = tsvLine.split('\t') //note we don't trim the line here in case the last element is deliberately empty
+    if( fields.length != headers.length ) {
+      throw new RuntimeException("TSV parsing error in line " + tsvLine.toString + ": wrong number of fields")
```
dvoet: did you intend to put tsvLineNo here too? Otherwise why do you have tsvLineNo?

dvoet:
This parser will break on blank lines. Also, does the spec mention anything about comments?

helgridly:
The spec mentions nothing about comments. Technically, blank lines are disallowed by the spec:

> Each line in a file must contain the same number of fields.

...but I'll trim off any newlines occurring at the end of the file.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVParser.scala
helgridly: yes

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVParser.scala
helgridly: good catch

PR #23, opened 2015-07-24
Configurations API

rushtong:
This PR has a configuration update so a merge to master will cause a jenkins failure until that is updated.

Required for DSDEEPB-506 and DSDEEPB-881
Add api call for agora configurations
Refactor mock frameworks as it is getting out of control
Refactor the agora model classes into a common object


eddiebroad:
the tests passed and the code compiled

PR #24, opened 2015-07-27
Created model schema and interrogated it to determine how to parse TSVs

helgridly:


model.json
```diff
@@ -0,0 +1,35 @@
...
+            "case_sample_id" : "sample",
+            "control_sample_id" : "sample",
+            "participant_id" : "participant"
+        },
+        "plural" : "pairs"
+    },
+    "participant_set" : {
+        "requiredAttributes" : {},
+        "plural" : "participant_sets",
+        "memberType" : "participant"
```
helgridly: Note: The existence of the `memberType` field is the only way to determine whether an entity type is a collection type or not. I initially had an `isCollectionType` boolean field but then figured that was asking for one to be defined and not the other, so the code decides whether the type is a collection by checking for the field.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -111,5 +119,106 @@ class EntityClient (requestContext: RequestContext) extends Actor {
...
+          requestContext.complete( HttpResponse(BadRequest, "TSV is missing required attributes: " + (requiredAttributes.keySet -- headers).mkString(", ")) )
+        } else {
+          op(requiredAttributes)
+        }
+      }
+    }
+  }
+
+  def addToCollectionMembersAttribute( entityType: String, entityName: String, newMembers: Seq[(String, String)], membersAttributeName: String ) = {
+    //STUB: return a partial batch Rawls call for this:
```
helgridly: Ted is looking at these.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -111,5 +119,106 @@ class EntityClient (requestContext: RequestContext) extends Actor {
...
+  def upsertEntitiesFromTSV(workspaceNamespace: String, workspaceName: String, tsv: TSVLoadFile): Unit = {
+    //the first column header of the tsv defines the entity type we're trying to import or update.
+    //This is magically verified in the withFoo functions below.
+    val entityType = tsv.firstColumnHeader.stripSuffix("_id")
+
+    withMemberCollectionType(entityType) { memberTypeOpt =>
+      //This magical list of conditions determines whether the TSV is populating the "members" attribute of a collection type entity.
+      if( memberTypeOpt.isDefined
+          && tsv.headers.length == 2
+          && tsv.headers == Seq( tsv.firstColumnHeader, memberTypeOpt.get + "_id" ) ) {
```
helgridly: I still find this branchy and hard to read, and I'm not really a fan of the `withFoo` thing, but it puts the Bad Request responses out of sight and just shows the golden path.

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelSchema.scala
```diff
@@ -0,0 +1,52 @@
...
+    }
+  }
+
+  def isCollectionType(entityType: String): Try[Boolean] = {
+    getTypeSchema(entityType).map(_.memberType.isDefined)
+  }
+
+  def getCollectionMemberType(collectionEntityType: String): Try[Option[String]] = {
+    getTypeSchema(collectionEntityType).map(_.memberType)
+  }
```
helgridly: It seems strange that I'm repeating myself here, but I guess that's what I get for writing a facade.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -45,6 +51,8 @@ class EntityClient (requestContext: RequestContext) extends Actor {
       listEntities(workspaceNamespace, workspaceName, entityType)
     case CreateEntities(workspaceNamespace: String, workspaceName: String, entities: JsValue) =>
       createEntities(workspaceNamespace, workspaceName, entities)
+    case UpsertEntitiesFromTSV(workspaceNamespace: String, workspaceName: String, tsvFile: JsValue) =>
```
abaumann: tsvFile is misleading if its the content, tsvFileContent?

helgridly: actually that's a bug, it should be a TSVLoadFile

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -111,5 +119,106 @@ class EntityClient (requestContext: RequestContext) extends Actor {
...
+  //TSVs can be of two forms: they can either populate the "members" list of a collection entity,
+  //or they can populate the attributes on an(y kind of) entity.
+  def upsertEntitiesFromTSV(workspaceNamespace: String, workspaceName: String, tsv: TSVLoadFile): Unit = {
+    //the first column header of the tsv defines the entity type we're trying to import or update.
+    //This is magically verified in the withFoo functions below.
+    val entityType = tsv.firstColumnHeader.stripSuffix("_id")
+
+    withMemberCollectionType(entityType) { memberTypeOpt =>
+      //This magical list of conditions determines whether the TSV is populating the "members" attribute of a collection type entity.
+      if( memberTypeOpt.isDefined
```
abaumann: I find this big if hard to understand, what is the exact meaning of the cases in here?  can we just make that a function that returns boolean with a readable name instead?

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -111,5 +119,106 @@ class EntityClient (requestContext: RequestContext) extends Actor {
...
+          && tsv.headers.length == 2
+          && tsv.headers == Seq( tsv.firstColumnHeader, memberTypeOpt.get + "_id" ) ) {
+
+        withPlural(memberTypeOpt.get) { memberPlural =>
+          val rawlsCalls = addToCollectionTypeFromTSV(workspaceNamespace, workspaceName, entityType, memberTypeOpt.get, memberPlural, tsv)
+          batchCallToRawls(rawlsCalls)
+        }
+      } else {
```
abaumann: should this be an else, or are there cases where the above might not be true due to user error and we will try to proceed (should it instead be an else if?)

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
helgridly: Sure.

helgridly: Although, y'know, that's what the comment is for.

helgridly:
Alex says yay.

PR #25, opened 2015-07-29
TSV parsing calling out to model schema to determine type information…

tedsharpe:
…, filled in stubs for parsed TSV --> rawls

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
@@ -18,7 +18,7 @@ import spray.json.DefaultJsonProtocol._
 
 import org.broadinstitute.dsde.firecloud.EntityClient._
 import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
-import org.broadinstitute.dsde.firecloud.model.{EntityCreateResult, ModelSchema}
+import org.broadinstitute.dsde.firecloud.model._
 import org.broadinstitute.dsde.firecloud.utils.TSVLoadFile
 
 object EntityClient {
@@ -142,9 +142,9 @@ class EntityClient (requestContext: RequestContext) extends Actor {
   //Bail with a 400 Bad Request if the first column of the tsv has duplicate values.
   //Otherwise, carry on.
   private def checkFirstColumnsDistinct( tsv: TSVLoadFile )(op: Unit): Unit = {
-    val entitiesToUpdate = tsv.tsvData.map(_(tsv.firstColumnHeader))
+    val entitiesToUpdate = tsv.tsvData.map(_(0))
     val distinctEntities = entitiesToUpdate.distinct
-    if( entitiesToUpdate != distinctEntities ) {
+    if ( entitiesToUpdate.size != distinctEntities.size ) {
       requestContext.complete( HttpResponse(BadRequest, "TSV has duplicated entities: " + entitiesToUpdate.diff(distinctEntities).distinct.mkString(", ")) )
     } else {
       op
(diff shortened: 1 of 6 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -142,9 +142,9 @@ class EntityClient (requestContext: RequestContext) extends Actor {
   //Bail with a 400 Bad Request if the first column of the tsv has duplicate values.
   //Otherwise, carry on.
   private def checkFirstColumnsDistinct( tsv: TSVLoadFile )(op: Unit): Unit = {
-    val entitiesToUpdate = tsv.tsvData.map(_(tsv.firstColumnHeader))
+    val entitiesToUpdate = tsv.tsvData.map(_(0))
```
helgridly: Why?

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -6,7 +6,7 @@ import javax.ws.rs.Path
 
 import akka.actor.{Actor, Props}
 import com.wordnik.swagger.annotations._
-
+import org.broadinstitute.dsde.firecloud.utils.TSVParser
```
helgridly: Nit: This should move down per the [import ordering conventions](https://broadinstitute.atlassian.net/wiki/display/DSDE/Code+Style+Conventions)

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVParser.scala
```diff
@@ -10,16 +10,16 @@ package org.broadinstitute.dsde.firecloud.utils
 case class TSVLoadFile(
   firstColumnHeader:String, //The first header column, used to determine the type of entities being imported
   headers:Seq[String], //All the headers
-  tsvData:Seq[Map[String, String]] //List of rows of the TSV. Each row is a map from header to value.
+  tsvData:Seq[Seq[String]] //List of rows of the TSV, broken out into fields
 )
```
helgridly: I don't care enough to kick up a *huge* stink, but: why?

tedsharpe: 1. I needed to know the reference type, too, in addition to the column name when that's defined, so the info was incomplete.  (Yes, I could do a lookup while processing every field, but that's silly, because it can be done once for all time for each column and saved ahead of time.)
2. It's crazy extravagant with memory to zip column info with every field up front rather than to do it row by row on demand.

helgridly: I understand 2; for 1 you're saying "and it didn't even give me all the information I needed anyway so no great loss"?

tedsharpe: yes, exactly.  also added a 3.

helgridly: (thumbsup)

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
tedsharpe: Not only was it out of order, it was repeated later (in correct position).

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -177,30 +177,62 @@ class EntityClient (requestContext: RequestContext) extends Actor {
...
   }
 
-  def batchCallToRawls( calls: Any ) = {
-    //STUB: stitch the batch calls together and fire off the Rawls request.
-    //requestContext.complete(...)
+  val upsertAttrOperation = "op" -> AttributeString("AddUpdateAttribute")
+  val removeAttrOperation = "op" -> AttributeString("RemoveAttribute")
+  val addListMemberOperator = "op" -> AttributeString("AddListMember")
+
+  def setAttributesOnEntity(entityType: String, row: Seq[String], colInfo: Seq[(String,Option[String])]) = {
```
abaumann: comments on this function would be helpful to follow what this is doing

helgridly: This PR has already been merged.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -177,30 +177,62 @@ class EntityClient (requestContext: RequestContext) extends Actor {
...
-    //group together updates to distinct entities, construct the list of entities to add references to, and poke it into whatever the "members" field is called.
-    val tsvFieldName = memberType + "_id"
-    tsv.tsvData.groupBy( _(tsv.firstColumnHeader) ).map({
-      case (entityName, rows) => addToCollectionMembersAttribute( entityType, entityName, rows.map(row => (memberType, row(tsvFieldName))), membersAttributeName )
-    })
+  def addToCollectionTypeFromTSV( entityType: String, memberType: String, membersAttributeName: String, tsv: TSVLoadFile ) = {
+    // group together updates to distinct entities
+    tsv.tsvData groupBy(_(0)) map { case (entityName, rows) =>
+      val ops = rows map { row =>
+        val attrRef = AttributeReference(memberType,row(1))
```
abaumann: here and other places knowing what 0 and 1 are would be helpful, 0 in most places i get, first column, but val some_useful_variable_name = row(1) would be nice

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
helgridly: (but we can probably do some tidying before it goes from FEATURE_ into master)

src/test/scala/org/broadinstitute/dsde/firecloud/utils/TSVParserSpec.scala
```diff
@@ -67,7 +67,7 @@ class TSVParserSpec extends FlatSpec {
   }
 
   it should "be fine with a bunch of newlines at the end of file" in {
-    val parseResult = TSVLoadFile("foo", Seq("foo", "bar", "baz"), Seq(Map("foo" -> "woop", "bar" -> "de", "baz" -> "doo")))
+    val parseResult = TSVLoadFile("foo", Seq("foo", "bar", "baz"), Seq(Array("woop", "de", "doo")))
```
abaumann: failure case tests?

helgridly: Are at the top of the file.

abaumann: son of a crap

PR #26, opened 2015-07-30
Copy method configuration into workspace

rushtong:
This is an initial attempt at pulling Method Configuration API calls into a single, top-level service endpoint. Future work will require that other method config services are relocated here and will be addressed in additional technical debt stories.

Notes:
* The Rawls endpoint is currently undergoing work so integration does not function.
* Running integration tests will fail as they require existing, valid configurations and workspaces that will certainly not exist for the test user. 
* This requires a change to the configuration so please check that this has happened with devops before merging.


eddiebroad:
Greg, did you get this error in running a test?

[info]   when calling PUT on the /workspaces*/*/method_configs/copyFromMethodRepo path
[info]   - MethodNotAllowed error is returned
[ERROR] [08:20:30.322] [pool-5-thread-2-ScalaTest-running-WorkspaceServiceSpec] o.b.d.f.service.WorkspaceServiceSpec - No authenticated username provided.
[info] WorkspaceServiceSpec:
[info] WorkspaceService

rushtong:
@eddiebroad - Yes, that is expected and is not a test failure and is a successful test result. The test cases cover the negative condition of trying to access a service without an authentication token.

src/main/scala/org/broadinstitute/dsde/firecloud/model/WorkspaceName.scala
```diff
@@ -0,0 +1,70 @@
...
+                              @(ApiModelProperty@field)(required = true, value = "The name of the workspace")
+                              name: Option[String] = None,
+                              @(ApiModelProperty@field)(required = true, value = "The date the workspace was created in yyyy-MM-dd'T'HH:mm:ssZZ format")
+                              createdDate: Option[String] = None,
+                              @(ApiModelProperty@field)(required = true, value = "The user who created the workspace")
+                              createdBy: Option[String] = None,
+                              @(ApiModelProperty@field)(required = true, value = "The attributes of the workspace")
+                              attributes: Option[Map[String, String]] = None)
+
+case class EntityCreateResult(entityType: String, entityName: String, succeeded: Boolean, message: String)// TODO: This is a stub case class until we know what we're returning from the batch entity create endpoint.
```
dmohs: It's tough to review this with the long lines. Would you please wrap them at 100? To make it less painful, I suggest you use a two-space indent for the `@(ApiModelProperty@field)` values.

src/main/scala/org/broadinstitute/dsde/firecloud/service/MethodConfigurationService.scala
```diff
@@ -0,0 +1,72 @@
...
+  lazy val log = LoggerFactory.getLogger(getClass)
+
+  @Path(value = "/copyFromMethodRepo")
+  @ApiOperation(
+    value = "copy a method repository configuration to a workspace",
+    nickname = "copyMethodRepositoryConfigurationToWorkspace",
+    httpMethod = "POST",
+    notes = "Copy a Method Repository Configuration into a workspace")
+  @ApiImplicitParams(Array(
+    new ApiImplicitParam(name = "workspaceNamespace", required = true, dataType = "string", paramType = "path", value = "Workspace Namespace"),
```
dmohs: Who is the customer for this API? Is there a reason they cannot use JSON?

rushtong: Sorry ... just saw this. The customer is the UI and the CLI. This field is not provided as a json object - it is a path parameter.

src/main/scala/org/broadinstitute/dsde/firecloud/model/WorkspaceName.scala
rushtong: This is idiomatic scala, unfortunately. We can update our code, but we'd be updating everyone else's with every reformat.

src/main/scala/org/broadinstitute/dsde/firecloud/service/MethodConfigurationService.scala
```diff
@@ -0,0 +1,72 @@
...
+    new ApiImplicitParam(name = "workspaceName", required = true, dataType = "string", paramType = "path", value = "Workspace Name"),
+    new ApiImplicitParam(name = "configurationNamespace", required = true, dataType = "string", paramType = "form", value = "Configuration Namespace"),
```
dmohs: Sorry. Previous question was for this line.

dmohs:
I just tried this:
```
curl -v --cookie "$OPENAM_COOKIE" -X POST --header "Content-Type: application/x-www-form-urlencoded" --header "Accept: application/json" -d "configurationNamespace=adsf&configurationName=adsf&configurationSnapshot=adsf&destinationNamespace=adsf&destinationName=adsf" "http://local.broadinstitute.org:8080/workspaces/asdf/fdas/method_configs/copyFromMethodRepo"; echo
* Hostname was NOT found in DNS cache
*   Trying 127.0.0.1...
* Connected to local.broadinstitute.org (127.0.0.1) port 8080 (#0)
> POST /workspaces/asdf/fdas/method_configs/copyFromMethodRepo HTTP/1.1
> User-Agent: curl/7.37.1
> Host: local.broadinstitute.org:8080
> Cookie: iPlanetDirectoryPro=********
> Content-Type: application/x-www-form-urlencoded
> Accept: application/json
> Content-Length: 124
> 
* upload completely sent off: 124 out of 124 bytes
< HTTP/1.1 404 Not Found
* Server spray-can/1.3.3 is not blacklisted
< Server: spray-can/1.3.3
< Date: Tue, 04 Aug 2015 16:47:03 GMT
< Content-Type: application/json; charset=UTF-8
< Content-Length: 83
< 
* Connection #0 to host local.broadinstitute.org left intact
The requested resource could not be found but may be available again in the future.
```
I believe Rawls returns better error messages than that. Until we are ready to provide our own good error messages, we should at least be returning theirs. The "create workspace" call does this, even though the Swagger UI for that call seems to be broken, so you can use that as a reference.

src/main/scala/org/broadinstitute/dsde/firecloud/service/MethodConfigurationService.scala
rushtong: During standup last week, we discussed the two approaches of either sending a full json object to orchestration or sending the data values as form fields. Everyone, at the time, agreed with the simpler approach of using form fields in conjunction with path parameters to then construct the object that gets passed to rawls.

rushtong:
In this endpoint, we are not modifying or intercepting the Rawls response
in any way. What is sent back from rawls should be what is returned to the
client.

rushtong:
Merging per discussion with @dmohs

PR #30, opened 2015-08-04
Enforce java 8 in build.sbt

helgridly:
Note that this change will throw an error in Scala versions < 2.11.2.


dmohs:
:+1: Works for me. Thanks for this.

PR #35, opened 2015-08-05
Add tests for update method configuration endpoint.

rushtong:


src/test/scala/org/broadinstitute/dsde/firecloud/mock/MockWorkspaceServer.scala
```diff
@@ -36,6 +36,10 @@ object MockWorkspaceServer {
     Some("name")
   )
 
+  val mockConfiguration = MethodConfiguration(
```
dmohs: Is this right?

dmohs:
One question. Otherwise :+1:

src/test/scala/org/broadinstitute/dsde/firecloud/mock/MockWorkspaceServer.scala
rushtong: No, that should be gone ... thanks for the catch. Will remove and then merge.

PR #36, opened 2015-08-05
Support multiple configurations.

rushtong:
This PR breaks out secrets into the `test.conf` and `application.conf` files and adds another configuration file where we can hold all of the custom endpoint paths. Support for new endpoint paths should be placed in `configurations.conf`

### DIFF
--- a/src/test/resources/test.conf
+++ b/src/test/resources/test.conf
@@ -1,45 +1,10 @@
-akka {
-  loglevel = "DEBUG"
-  loggers = ["akka.event.slf4j.Slf4jLogger"]
-}
-
-spray.can.server {
-  request-timeout = infinite
-}
-
-http {
-  interface="localhost"
-  port=8080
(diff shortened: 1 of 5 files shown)


src/test/resources/test.conf
```diff
@@ -1,54 +1,19 @@
...
 openam {
   deploymentUri = "https://openam-ci.vault.broadinstitute.org/opensso"
   realm = "/apollo"
   // The next three values should be provided by your environment (e.g. in an .sbtopts file)
-  username = ""
-  password = ""
-  commonName = ""
+  username = "apollo_test"
+  password = "TH3yDL5JSMcH0zr"
+  commonName = "TestAccount"
```
MattPutnam: Are these supposed to be included?

rushtong: NO!!!! That is a bad mistake.

PR #38, opened 2015-08-05
Updated model to match rawls change, added proper update path

MattPutnam:
I was getting some form of success with this.  Please carefully review though, this is the first I've really mucked with this codebase.


rushtong:
@MattPutnam - I would prefer that the configuration PR happen first - then your change to the path would go into a different file and not require any devops help.

rushtong:
Otherwise, :+1:

PR #40, opened 2015-08-06
Remove swagger from the methods service

rushtong:
It was already broken and not visible to the API. Follows decision from retro and is related to DSDEEPB-988


dmohs:
This is nice. It makes the code lighter. I encourage you to give other parts of the code a similar treatment :) :+1:

rushtong:
I have a ticket for more of this: DSDEEPB-988.

PR #42, opened 2015-08-07
Initial support for Submissions API. See DSDEEPB-978 for user story.

rushtong:


dmohs:
:+1: Tested GET request, which returns what Rawls returns.

PR #45, opened 2015-08-10
Created service for get entity types

MattPutnam:


helgridly:
seems legit

src/main/scala/org/broadinstitute/dsde/firecloud/service/EntityService.scala
```diff
@@ -33,10 +33,45 @@ trait EntityService extends HttpService with FireCloudDirectives {
   private final val ApiPrefix = "entities"
   private implicit val executionContext = actorRefFactory.dispatcher
 
-  val routes = listEntitiesPerTypeRoute
+  val routes = listEntityTypesRoute ~ listEntitiesPerTypeRoute
 
   lazy val log = LoggerFactory.getLogger(getClass)
 
+  @Path("")
```
rushtong: You can lose this line, the path is inherited from the class level `Api` annotation value property.

src/main/scala/org/broadinstitute/dsde/firecloud/service/EntityService.scala
```diff
@@ -33,10 +33,45 @@ trait EntityService extends HttpService with FireCloudDirectives {
...
+  @ApiOperation(value = "list all entity types in a workspace",
+    nickname = "listEntityTypes",
+    httpMethod = "GET",
+    produces = "application/json",
+    response = classOf[String],
+    responseContainer = "List")
```
rushtong: I recommend condensing these lines to something more scala-like

src/main/scala/org/broadinstitute/dsde/firecloud/service/EntityService.scala
```diff
@@ -33,10 +33,45 @@ trait EntityService extends HttpService with FireCloudDirectives {
   private final val ApiPrefix = "entities"
```
rushtong: Not your code, but this is unused, and incorrect if it were.

rushtong:
As noted in hipchat, there are no tests for this service yet. See https://broadinstitute.atlassian.net/browse/DSDEEPB-824

:+1: - looks good. Mostly minor comments. Feel free to address or merge as is.

PR #48, opened 2015-08-11
Passthrough to rawls list submissions endpoint

helgridly:
As per discussion in Silver Team hipchat room (2015-08-11 17:54) I'm not testing the response of `GET .../submissions` or `GET .../submissions/id` (and have removed testing of the latter).

Feel free to debate the merits of such a change here.


src/test/scala/org/broadinstitute/dsde/firecloud/service/SubmissionServiceSpec.scala
```diff
@@ -76,8 +85,6 @@ class SubmissionServiceSpec extends FreeSpec with ScalaFutures with ScalatestRou
       "OK response is returned" in {
         Get(localSubmissionIdPath) ~> Cookie(HttpCookie("iPlanetDirectoryPro", token)) ~> sealRoute(routes) ~> check {
           status should equal(OK)
-          val submission = responseAs[SubmissionIngest]
-          submission shouldNot be (None)
         }
```
rushtong: Good changes, I should have defaulted to a simple OK response without proper response classes. Best to avoid them completely until they are stable and/or in a dsde commons library.

rushtong:
:+1:

PR #49, opened 2015-08-12
change Exception to Throwable to eliminate compile warning

davidangb:
compile would previously show:

```
[info] Compiling 19 Scala sources to /Users/davidan/work/src/firecloud-orchestration/target/scala-2.11/classes...
[warn] /Users/davidan/work/src/firecloud-orchestration/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala:113: match may not be exhaustive.
[warn] It would fail on the following input: Failure((x: Throwable forSome x not in Exception))
[warn]     Try(Await.result(pipeline(externalRequest), Duration.Inf)) match {
[warn]        ^
[warn] one warning found
[success] Total time: 11 s, completed Aug 12, 2015 10:47:31 AM
```
but now it doesn't.


rushtong:
:+1:

dmohs:
Thank you for this. It was definitely annoying.

PR #50, opened 2015-08-12
Added endpoint to get method config

MattPutnam:


rushtong:
Looking at this now:
```
[info]   - MethodNotAllowed error is returned *** FAILED ***
[info]     404 Not Found did not equal 405 Method Not Allowed (MethodConfigurationServiceSpec.scala:69)
```

PR #53, opened 2015-08-12
Add "get workspace details" endpoint...

dmohs:
- also removes namespace so it can be provided by the client per a change in Rawls
- also refactors WorkspaceService to route in a more Spray-like way


rushtong:
:+1:

PR #55, opened 2015-08-13
added method config rename path

MattPutnam:


dmohs:
My only suggestion would be to remove the Swagger stuff since we don't have a good way to use it without significantly complicating the routes. Up to you. :+1:

PR #57, opened 2015-08-15
Implement combined call for entities with type information

rushtong:
Notes:
* This has been tested against Rawls-dev.
* There is a lot of entity information that comes back from rawls. This only captures the name and entity type. The model class can be expanded to capture more information if the UI requires it.
* Format of the response is a list of entity name and entity type for easy tabular display. This can be re-modeled into a map of type -> list of names if that is easier for the UI to work with.

My first approach was to implement the ask pattern for the various requests, but I ran into weirdness where the sender would not get the response back. When changing this to the tell pattern, the actor can handle the response itself and that works.

This (https://github.com/NET-A-PORTER/spray-actor-per-request/blob/master/src/main/scala/com/netaporter/core/DummyAppCore.scala) is another pattern that we could use. It is part of the Per Request pattern which we are not currently using, but might at some point. 

Example Output:
```
curl -X GET -H 'Content-type: application/json' --cookie 'iPlanetDirectoryPro=AQI...' http://local.broadinstitute.org:8080/workspaces/broad-dsde-dev/alexb_test_submission/entities_with_type
[{
  "name": "participant_01",
  "entityType": "participant"
}, {
  "name": "participant_02",
  "entityType": "participant"
}, {
  "name": "participant_03",
  "entityType": "participant"
}, {
  "name": "sample_01",
  "entityType": "sample"
}, {
  "name": "sample_02",
  "entityType": "sample"
}, {
  "name": "sample_03",
  "entityType": "sample"
}]
```

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/core/GetEntitiesWithType.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/core/GetEntitiesWithType.scala
@@ -1,0 +1,83 @@
+package org.broadinstitute.dsde.firecloud.core
+
+import akka.actor.{Actor, Props}
+import akka.event.Logging
+import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
+import org.broadinstitute.dsde.firecloud.core.GetEntitiesWithType.{AddEntities, EntityWithType, ProcessUrl}
+import spray.client.pipelining._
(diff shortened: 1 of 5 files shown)


rushtong:
Testing reveals that the results are not consistent. The return values are sometimes accurate, but often are only partial. It seems that request completion sometimes happens before all of the results have been assembled.

PR #60, opened 2015-08-19
template config for actor logging

davidangb:
with @dmohs ' changes in #47, I think we're in a good place:
* if you want to affect logging for our own log messages in our classes, change in logback.xml
* if you want to affect akka's internal logging, change in application.conf

while this means there are two files that would need editing under different conditions, I think this is a clear and understandable separation.

This PR should change nothing functionally - everything will log as it did before. The changes just provide a template for what you _would_ change if you wanted to affect akka logging.

The changes to application.conf will need to get deployed to jenkins, etc. to truly complete this story.

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -19,3 +19,25 @@ openam {
     value = "DataStore"
   }
 }
+
+akka {
+  loglevel = "INFO"
+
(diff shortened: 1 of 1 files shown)


dmohs:
Works as advertised (nothing broke in my testing). :+1:

PR #61, opened 2015-08-21
Orchestration support for removing a workspace method config

rushtong:
Support for DSDEEPB-421


eddiebroad:
it is my understanding that tests are running and passing!  great!

PR #62, opened 2015-08-24
added test tsv files and removed importJSON related activities

MatthewBemis:
just a small PR to add some tsv files that can be used to test against the current spec. ultimately, the goal will be to have these test files used in some unit tests- that way we'll know immediately when the tsv spec is updated. also removed some old code related to importing entities with json, since we now use tsv


dmohs:
:+1: Nice to have these available for testing.

PR #63, opened 2015-08-24
First pass at Swagger via YAML

dmohs:
The new version of Swagger prefers the definition to be all in one file. This is not ideal, but because the YAML syntax is compact, I think it's okay for now.

There are some TODOs for fixing stuff, but I think this is a good place for us to review/discuss.

I think this will work in prod with the transparent proxy, since I tested it locally, but because the environments are slightly different, it might not. I will monitor the deploy.


rushtong:
:+1: This works! How far do you want to take this PR? Are we OK with starting with workspaces and documenting other services in other stories/subtasks?

dmohs:
@rushtong I have already created subtasks for the other services, so we can keep this one small.

davidangb:
:+1: works for me locally, code changes look good.

PR #64, opened 2015-08-26
DSDEEPB-979 PerRequest implementation

davidangb:
I feel like I'm missing something. This implements the PerRequest pattern, as spearheaded by rawls and agora ... but it doesn't change a whole lot in the orchestration code? It does clean up the routes a little bit but not a lot.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/PerRequest.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/PerRequest.scala
@@ -1,0 +1,117 @@
+package org.broadinstitute.dsde.firecloud.service
+
+import akka.actor.SupervisorStrategy.Stop
+import akka.actor._
+import org.broadinstitute.dsde.firecloud.{HttpClient, FireCloudExceptionWithStatusCode}
+import org.broadinstitute.dsde.firecloud.service.PerRequest._
+import spray.client.pipelining._
+import spray.http.{RequestProcessingException, HttpRequest, HttpHeader}
+import spray.http.StatusCodes._
+import spray.httpx.marshalling.ToResponseMarshaller
+import spray.routing.RequestContext
+
+import scala.concurrent.duration._
+
+/**
+ * This actor controls the lifecycle of a request. It is responsible for forwarding the initial message
+ * to a target handling actor. This actor waits for the target actor to signal completion (via a message),
+ * timeout, or handle an exception. It is this actors responsibility to respond to the request and
+ * shutdown itself and child actors.
+ *
+ * Request completion can be signaled in 2 ways:
+ * 1) with just a response object
+ * 2) with a RequestComplete message which can specify http status code as well as the response
+ */
+trait PerRequest extends Actor {
+  import context._
+
+  // JSON Serialization Support
+  import spray.httpx.SprayJsonSupport._
+  import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
+
+  def r: RequestContext
+  def target: ActorRef
+  def message: AnyRef
+  def timeout: Duration
+
+  setReceiveTimeout(timeout)
+  target ! message
+
+  def receive = {
+    case RequestComplete_(response, marshaller) => complete(response)(marshaller)
+    case RequestCompleteWithHeaders_(response, headers, marshaller) => complete(response, headers: _*)(marshaller)
+    case ReceiveTimeout => complete(GatewayTimeout)
+    case x =>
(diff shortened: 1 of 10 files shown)


dmohs:
The PerRequest object is more Scala-foo than I can handle, but the usages show that it simplifies the code, which is definitely an improvement. I would prefer a 100-column limit to make it easier to review and fit on my screen, but otherwise I'd be happy to see this merged. :+1:

src/main/scala/org/broadinstitute/dsde/firecloud/service/PerRequest.scala
```diff
@@ -0,0 +1,117 @@
...
+ */
+trait PerRequestCreator {
+  implicit def actorRefFactory: ActorRefFactory
+
+  def perRequest(r: RequestContext, props: Props, message: AnyRef, name: String = java.lang.Thread.currentThread.getStackTrace()(1).getMethodName, timeout: Duration = 1.minutes) =
+    actorRefFactory.actorOf(Props(new WithProps(r, props, message, name + System.nanoTime(), timeout)), name + System.nanoTime())
+
+  /** convenience for HttpClient */
+  def externalHttpPerRequest(r: RequestContext, request: HttpRequest) =
+    perRequest(r, Props(new HttpClient(r)), HttpClient.PerformExternalRequest(request))
```
bradtaylor: I worry this is that antipattern where one should not be declaring a new actor in a Props() call from within another actor. @geoffjentry comment?

rushtong: Isn't that exactly what perRequest does when you call it WithProps? The props in that code-path is actor-creation.

rushtong:
:+1:

bradtaylor:
What I don't see here is a change to HttpClient to respond to PerRequest level for request completion.

so in createResponseFutureFromExternalRequest, instead of:
requestContext.complete(response.withHeaders(response.headers.filterNot(isAutomaticHeader)))

You would do:
context.parent ! RequestCompleteWithHeaders(response.status, response.entity, response.headers.filterNot(isAutomaticHeader))

That syntax is probably wrong, but the general idea as I understand it is this:
* Your do-something actors simply produce something and hand it back to whatever parent actor called them. In your case, this is a method or a configuration. In your case, it's the response body

* Per request is the only one who cares about requestContexts and is responsible for serializing things for return to the user, etc. Anything involved in the HTTP cycle. Its primary function is to be the single place where you worry about serialization and request completion. It also supervises lower level actors such that a request can be completed appropriately even if they throw an exception.

davidangb:
Ahhh ... @bradtaylor thank you. I was pretty sure I was missing something. That makes much more sense. Back to me for more implementation ...

src/main/scala/org/broadinstitute/dsde/firecloud/service/PerRequest.scala
bradtaylor: Yes there is a call to new in the perRequest function. good call. For the props of the "target" actor, we usually define a separate function somewhere (usually in the companion object to that actor), ala:
def queryHandlerProps = Props(classOf[QueryHandler])

Then we call perRequest like
perRequest(<requestContext>, quertyHandlerProps, <serviceMessage>)

Again, I /think/ we're safe with 'new' because we are in a trait, but the worry here is documented:
http://doc.akka.io/docs/akka/2.3.12/scala/actors.html#actors-scala
(see "Props" section)

rushtong:
Following Brad's suggestion, GetEntitiesWithType.scala will need to be revised with the same context.parent tell pattern. That class concerns me under this paradigm, but I may be over-thinking it.

davidangb:
okay, ready for re-review. I tried to find some elegant way to reduce logic in HttpClient, but coldn't really find it. *sigh*

dmohs:
:+1: Looks good. I think there are some optimize import calls that could be done. For example, I think you've removed usages of HttpClient in a few files. My previous comment about line-length still applies, but your call.

davidangb:
I've run IntelliJ's "Organize Imports" on all files I touched, and I adjusted line lengths on every file *except* PerRequest. I left PerRequest alone because it is primarily a copy from agora, and I wanted to keep the code with as few diffs from agora as possible.

davidangb:
oh crap, and now it looks like I need to manually merge :(

davidangb:
one more time for re-review. I haven't squashed commits to make it easier to review, but can squash before this goes in.

dmohs:
Good call on keeping the diff from agora minimal. Looks good to me.

davidangb:
(verbal OK from @dmohs to merge)

PR #65, opened 2015-08-26
added some YAML-based documentation for submisssions

eddiebroad:
also did some rearrangement of  order within submissions.
The reordering isn't reflected in a single squashed commit however.

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -26,3 +26,66 @@ paths:
       responses:
         200:
           description: List of workspaces.
+  /workspaces/{workspaceNamespace}/{workspaceName}/submissions:
+    get:
+      tags:
+      - submissions
+      operationId: listSubmissions
+      summary: |
+        Lists submissions.
+      produces:
+      - application/json
+      - text/plain
+      responses:
+        200:
+          description: List of submissions
+    post:
+      tags:
+      - submissions
+      operationId: createSubmission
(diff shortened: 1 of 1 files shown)


eddiebroad:
This is an incremental addition to the .YAML file.  it will need additional work to more fully represent the API

rushtong:
Each of the path specifications need parameters so that they can be entered into the swagger UI. As it is now, these do not function. Look at this example for some inspiration, specifically the {petId} path parameter:
https://github.com/swagger-api/swagger-codegen/blob/master/samples/yaml/pet.yml

rushtong:
Closing this PR for now. When the functionality is complete, we can re-open a new PR.

rushtong:
My apologies - the link I provided was for an older spec. The newer spec is here:
https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md

eddiebroad:
I'm not sure what you mean by "do not function".

I modeled the commit after PR 63 https://github.com/broadinstitute/firecloud-orchestration/pull/63/ and since the API isn't fully defined as far as I know (I think David An mentioned something about this during stand-up this morning) I wasn't sure how extensive to fill stuff in.  


-eddie

rushtong:
@eddiebroad - The swagger UI does not work on this branch. What I mean by that is  one cannot enter a workspace name/namespace and see submissions. See my latest PR for an example of providing path parameters in the yml so that the swagger UI will function.

PR #67, opened 2015-08-28
Update the actor so that it doesn't return before the results are col…

rushtong:
…lected.

This also removes the source of a nagging suspicion I had around an expected timeout when using the Aggregator pattern (http://doc.akka.io/docs/akka/snapshot/contrib/aggregator.html). Removing the expect, the timeout, and moving the entity collection call seems to make all the difference.


davidangb:
code looks good to me; I have a bunch of local changes right now so I can't test right away, but soon ...

davidangb:
tested locally, works for me, looks good. :+1:

PR #68, opened 2015-08-28
Add methods and configurations swagger yml

rushtong:
For DSDEEPB-1088
Notes:
* I put these two services into the same "tag" so they show up together in the UI. 
* We should standardize on naming convention and capitalization. Agora capitalizes the endpoint descriptions, Rawls does not. As this is user-facing, I like the Agora approach better.


davidangb:
:+1:

PR #69, opened 2015-08-28
Add swagger yml for submission services

rushtong:
For DSDEEPB-1089
Notes:
* This was a bit more complicated as it required defining the expected SubmissionIngest json as an object in the yaml.


davidangb:
added some property descriptions since Greg is out; otherwise :+1: .

PR #70, opened 2015-08-31
added delete pass-through to  SubmissionService.scala

eddiebroad:
added some DELETE tests

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -393,6 +393,37 @@ paths:
         500:
           description: Internal Error
   /workspaces/{workspaceNamespace}/{workspaceName}/submissions/{submissionId}:
+    delete:
+      tags:
+      - Submissions
+      operationId: deleteSubmission
+      summary: |
+        abort a submission
+      produces:
+      - application/json
+      parameters:
+        - name: workspaceNamespace
+          description: Workspace Namespace
+          required: true
+          type: string
+          in: path
+        - name: workspaceName
+          description: Workspace Name
+          required: true
+          type: string
+          in: path
+        - name: submissionId
+          description: Submission ID
+          required: true
+          type: string
+          in: path
+      responses:
+        204:
+          description: successful deletion
+        404:
+          description: Submission not found
+        500:
+          description: Internal Error
     get:
       tags:
       - Submissions
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/SubmissionService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/SubmissionService.scala
@@ -46,6 +46,14 @@ trait SubmissionService extends HttpService with PerRequestCreator with FireClou
         externalHttpPerRequest(requestContext, extReq)
       }
     } ~
+    path("workspaces" / Segment / Segment / "submissions" / Segment) {
+      (workspaceNamespace, workspaceName, submissionId) =>
+        delete { requestContext =>
+          val extReq: HttpRequest = Delete(FireCloudConfig.Rawls.
+            submissionByIdUrl(workspaceNamespace, workspaceName, submissionId))
+          externalHttpPerRequest(requestContext, extReq)
+        }
+    } ~
     path("workspaces" / Segment / Segment / "submissions" / Segment / "workflows" / Segment / "outputs") {
       (workspaceNamespace, workspaceName, submissionId, workflowId) =>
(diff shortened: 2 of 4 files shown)


eddiebroad:
I've put in a pass-through "Route" (I believe is the term) and two tests.  Are these sufficient for the back-end?

davidangb:
now that master has swagger documentation for the submission service, could you please update the swagger doc to reflect this change?

eddiebroad:
"added a deletion entry for the submissions path" should be written as "added a deletion entry for the submissions path section in src/main/resources/swagger/api-docs.yaml"

src/main/scala/org/broadinstitute/dsde/firecloud/service/SubmissionService.scala
```diff
@@ -46,6 +46,15 @@ trait SubmissionService extends HttpService with PerRequestCreator with FireClou
         externalHttpPerRequest(requestContext, extReq)
       }
     } ~
+    path("workspaces" / Segment / Segment / "submissions" / Segment) {
+      (workspaceNamespace, workspaceName, submissionId) =>
+        delete { requestContext =>
+          val deleteSubmission: HttpRequest = Delete(FireCloudConfig.Rawls.
+            submissionByIdUrl(workspaceNamespace, workspaceName, submissionId))
+          actorRefFactory.actorOf(Props(new HttpClient(requestContext))) !
+            HttpClient.PerformExternalRequest(deleteSubmission)
```
davidangb: with the PerRequest implementation merged in, this block should change to something like:
```
delete { requestContext =>
          val extReq: HttpRequest = Delete(FireCloudConfig.Rawls.
            submissionByIdUrl(workspaceNamespace, workspaceName, submissionId))
        }
```
instead of creating the actor explicitly.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -202,6 +202,37 @@ paths:
...
+          required: true
+          type: string
+          in: path
+        - name: submissionId
+          description: Submission ID
+          required: true
+          type: string
+          in: path
+      responses:
+        200:
```
davidangb: rawls API says this returns 204, not 200. Are you sure?

davidangb:
the two new tests fail for me on a clean build of this branch.

eddiebroad:
@davidangb re "the two new tests fail for me on a clean build of this branch."
hmm.... I rand the tests on a clean build too and everything passed! ?
Is "clean" the sbt command to clean before a build?
image

src/main/resources/swagger/api-docs.yaml
eddiebroad: re "rawls API says this returns 204, not 200. Are you sure?"
I am not sure.  However, I am sure that Greg told me that we are not using RAWLS as a model however.

eddiebroad: re "rawls API says this returns 204, not 200. Are you sure?"
Do the values we put here ultimately depend on the definition of the orchestration API?  Are there any decisions on this?

rushtong: We don't use Rawls as a model for what we design as endpoint urls in orchestration. This is a different scenario as we are returning whatever Rawls does. If Rawls says the response is a 204, then our api docs should also say that.

eddiebroad: @rushtong I would have expected you to be enjoying some vacation....!

davidangb:
hmmm ... may have been a problem on my end. I re-cleaned, re-checked out, re-tested, etc. and all is well now. Sorry about that!

src/main/resources/swagger/api-docs.yaml
```diff
@@ -393,6 +393,37 @@ paths:
         500:
           description: Internal Error
   /workspaces/{workspaceNamespace}/{workspaceName}/submissions/{submissionId}:
+    delete:
+      tags:
+      - Submissions
+      operationId: deleteSubmission
+      summary: |
+        delete a submission
```
davidangb: nitpick: this should be "Abort a submission" not delete.

davidangb:
one minor change to the yaml description, otherwise good to go. Feel free to make the change and merge yourself, or make the change and send back to me if you'd rather.

src/main/resources/swagger/api-docs.yaml
eddiebroad: I've reworded "delete" as "abort" locally and am about to force push and merge

PR #71, opened 2015-08-31
first pass

davidangb:
Creating a PR on @MattPutnam 's behalf for easier commenting

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -283,10 +283,12 @@ paths:
           schema:
             $ref: MethodConfigRename
       responses:
-        200:
+        204:
           description: Successful request
         404:
           description: Workspace or Method Configuration not found
+        409:
+          description: Method Configuration with that name already exists
         500:
           description: Internal Server Error
   /workspaces/{workspaceNamespace}/{workspaceName}/method_configs/copyFromMethodRepo:
@@ -318,7 +320,7 @@ paths:
       responses:
         201:
           description: Successful request
-        403:
+        404:
           description: Source method configuration does not exist
         409:
           description: Destination method configuration by that name already exists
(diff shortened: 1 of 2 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -137,6 +137,195 @@ paths:
...
+          in: path
+        - name: rename
+          description: Method Config Rename
+          required: true
+          type: string
+          in: body
+          schema:
+            $ref: MethodConfigRename
+      responses:
+        200:
```
davidangb: rawls API documents a 204 here, not 200. Are you sure?

src/main/resources/swagger/api-docs.yaml
```diff
@@ -137,6 +137,195 @@ paths:
...
+          description: Successful request
+        404:
+          description: Workspace or Method Configuration not found
+        500:
```
davidangb: rawls API documents it can return a 409	Method config by that name already exists in workspace

src/main/resources/swagger/api-docs.yaml
```diff
@@ -137,6 +137,195 @@ paths:
...
+          description: Method Configuration to Copy
+          required: true
+          type: string
+          in: body
+          schema:
+            $ref: CopyConfigurationIngest
+      responses:
+        201:
+          description: Successful request
+        403:
```
davidangb: 404 not 403

src/main/resources/swagger/api-docs.yaml
```diff
@@ -295,3 +484,80 @@ definitions:
...
+        type: string
+        description: Method configuration namespace
+      name:
+        type: string
+        description: Method configuration name
+      rootEntityType:
+        type: string
+        description: Root entity type
+      workspaceName:
+        type: string
```
davidangb: these all need to be Map[String, String], not string. Looks like in rawls, the methodRepoMethod and methodRepoConfig and workspaceName are strongly typed, even if we don't type them in orchestration.

MattPutnam: Does Swagger support that?  I didn't see it in the documentation.

davidangb: Swagger supports it at least sort of - see https://rawls-dev.broadinstitute.org/api-docs/listings/methodconfigs (or the rawls swagger UI).

I just tried
```
type: Map[string,string]
```
locally. That makes it appear correctly in the "Model" view, but makes it disappear from the "Model Schema" view. Hmmmmm.

davidangb:
scala refactor looks good. YAML looks mostly good with a few comments inline.

PR #72, opened 2015-09-01
Use X-Forwarded-Host header when present for correct redirect URLs

dmohs:
Apache's ProxyPass includes this header explicitly for this purpose. Tested locally using modified script (watch for PR in firecloud-ui).


davidangb:
:+1:

PR #73, opened 2015-09-02
DSDEEPB-958: pass on new Google/OAuth headers

davidangb:
DO NOT MERGE without the new Google/OAuth proxy in place, and without reading through this description fully.

This is a *theoretical* and *untested* change. I haven't had a chance to spin up the new OAuth proxy locally, and rawls/agora don't support the new OAuth proxy yet, so I have no idea if this works. It'll be up to you to try it and see.

The heart of this change is the new FireCloudRequestBuilder trait and its authHeaders function. This looks for auth headers in the incoming request and copies those headers to the outgoing request.
**TODO: verify this is looking for the correct headers.**

_Other teams, like rawls, may want a copy of this trait!_

Then, I've modified all our endpoints to remove handling of the OpenAM cookie and replace with a call to the authHeaders function.

Finally, I did the same for all our tests, which use a dummy auth header instead of a real one.
**TODO: this means tests will only work against mock data; they'll fail against a real version of rawls/agora because the dummy auth header will be rejected as invalid.**
Alternately, we could choose to remove auth headers from our tests entirely, cleaning them up a bit.

The FireCloudRequestBuilding trait could probably be optimized from the pure Scala perspective.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
@@ -2,7 +2,7 @@ package org.broadinstitute.dsde.firecloud
 
 import java.text.SimpleDateFormat
 
-import org.broadinstitute.dsde.firecloud.service.FireCloudTransformers
+import org.broadinstitute.dsde.firecloud.service.FireCloudRequestBuilding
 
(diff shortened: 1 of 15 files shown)


dmohs:
:+1: More cleanup to do, but this works.

PR #75, opened 2015-09-10
DSDEEPB-1235: Move filtering of automatic headers from HttpClient to PerRequest

davidangb:
This should eliminate all the "Explicitly set response header 'Date: Wed, 09 Sep 2015 22:11:12 GMT' is ignored, the spray-can HTTP layer sets this header automatically!" messages in the orchestration layer log.


MattPutnam:
:+1:

PR #76, opened 2015-09-10
Add endpoints for ACL get/patch

MattPutnam:


davidangb:
Code looks good. I'd love to see some tests for this, which means some mock responses.

PR #77, opened 2015-09-11
add patch endpoint for editing workspace attributes

MatthewBemis:
This is part of the UI work that Blue Team has taken on for [DSDEEPB-1137](https://broadinstitute.atlassian.net/browse/DSDEEPB-1137).

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -16,7 +16,7 @@ paths:
   /workspaces:
     get:
       tags:
-      - workspaces
+      - Workspaces
       operationId: listWorkspaces
       summary: |
         Lists workspaces.
@@ -53,7 +53,7 @@ paths:
   /workspaces/{workspaceNamespace}/{workspaceName}:
     get:
       tags:
-        - workspaces
+        - Workspaces
       operationId: getWorkspace
       summary: Get workspace
       produces:
@@ -78,7 +78,7 @@ paths:
           description: Internal Server Error
     delete:
       tags:
-        - workspaces
+        - Workspaces
       operationId: deleteWorkspace
       summary: Delete workspace
       parameters:
@@ -102,7 +102,7 @@ paths:
   /workspaces/{workspaceNamespace}/{workspaceName}/acl:
     get:
       tags:
-        - workspaces
+        - Workspaces
       operationId: getWorkspaceAcl
       summary: Get workspace ACL
       produces:
@@ -129,7 +129,7 @@ paths:
           description: Internal Server Error
     patch:
       tags:
-        - workspaces
+        - Workspaces
       operationId: updateWorkspaceACL
       summary: Update workspace ACL
       parameters:
@@ -599,6 +599,43 @@ paths:
           description: Submission or Workflow not found
         500:
           description: Internal Error
+          
+  /workspaces/{workspaceNamespace}/{workspaceName}/updateAttributes:
+    patch:
+      tags:
+      - Workspaces
+      operationId: updateAttributes
+      summary: |
+        Modify attributes on a workspace.
+      produces:
+      - application/json
+      parameters:
+        - name: workspaceNamespace
+          description: Workspace Namespace
+          required: true
+          type: string
+          in: path
+        - name: workspaceName
+          description: Workspace Name
+          required: true
+          type: string
+          in: path
+        - name: workspaceUpdateJson
+          description: Attribute operations
+          required: true
+          type: array
+          items:
+            $ref: AttributeUpdateOperation
+          in: body
+      responses:
+        200:
+          description: Success
+        400:
+          description: Invalid operation
+        404:
+          description: Workspace not found
+        500:
+          description: Internal Error
 
 definitions:
   SubmissionIngest:
@@ -701,3 +738,17 @@ definitions:
       - configurationSnapshotId
       - destinationNamespace
       - destinationName
+  AttributeUpdateOperation:
+    properties:
+      op:
+        type: string
+        description: Operation to perform on the attribute
+      attributeName:
+        type: string
+        description: The name of the attribute
+      addUpdateAttribute:
+        type: string
+        description: The value of the attribute (if adding/updating)
+    required:
+      - op
+      - attributeName
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
@@ -9,6 +9,7 @@ import spray.client.pipelining.{Get, Post, Patch, Delete}
 import spray.http.StatusCodes._
 import spray.json.DefaultJsonProtocol._
 import spray.json._
+import spray.httpx.SprayJsonSupport._
 import spray.routing._
 
 import org.broadinstitute.dsde.firecloud.{EntityClient, FireCloudConfig, HttpClient}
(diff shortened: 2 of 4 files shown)


eddiebroad:
it is my understanding that all orchestration PRs that include the implementation of a new service or pass-through require (by mutual agreement of group members) the addition of tests under /src/test/scala/org/broadinstitute/dsde/firecloud/.  I don't see such commits here

MatthewBemis:
I'll add those!

eddiebroad:
@MatthewBemis If you ask me, I would suggest adding a minimum of two tests.

eddiebroad:
@MatthewBemis Also, tests are required to pass before any merge is allowed

MatthewBemis:
@eddiebroad That was a mistake, but I've corrected that now. I replaced the two tests with just one because it's just a pass-through to rawls and (I think) 1 test should be sufficient to test that it works. I'm a bit unfamiliar with how your mocks work, but it looks like I can't test that an invalid attribute update operation returns Bad Request, because that's a response that rawls would provide, and rawls has it's own unit tests to check for this type of thing. Am I right in assuming this?

Also, I'll rebase again since there are new conflicts.

rushtong:
I'm looking at this now Matt, thanks!

src/main/resources/configurations.conf
```diff
@@ -25,4 +25,5 @@ workspace {
   submissionsPath="/workspaces/%s/%s/submissions"
   submissionByIdPath="/workspaces/%s/%s/submissions/%s"
   workflowOutputsByIdPath="/workspaces/%s/%s/submissions/%s/workflows/%s/outputs"
+  updateAttributesOnWorkspacePath="/workspaces/%s/%s/updateAttributes"
```
rushtong: Which endpoint is using this path? I don't see any usage of it in the rest of the code.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -599,6 +599,41 @@ paths:
...
+          in: path
+        - name: workspaceName
+          description: Workspace Name
+          required: true
+          type: string
+          in: path
+        - name: workspaceUpdateJson
+          description: Attribute operations
+          required: true
+          type: Array[AttributeUpdateOperation]
```
rushtong: I'm pretty sure this isn't the correct syntax for the type. I think it is something like this, where `AttributeUpdateOperation` is defined in the definitions:
```
  type: array
    items:
      $ref: "#/definitions/AttributeUpdateOperation"
```

src/main/resources/swagger/api-docs.yaml
```diff
@@ -599,6 +599,41 @@ paths:
           description: Submission or Workflow not found
         500:
           description: Internal Error
+          
+  /workspaces/{workspaceNamespace}/{workspaceName}/updateAttributes:
+    get:
+      tags:
+      - Workspaces
```
rushtong: This introduces a new header section in the swagger UI. Could you also update the other "workspaces" tags (lowercased) with the upper case version as part of this? It would be purely for the sake of consistency.

src/main/resources/swagger/api-docs.yaml
rushtong: See spec here: http://swagger.io/specification/
Lots of examples of `type: array`

src/test/scala/org/broadinstitute/dsde/firecloud/mock/MockWorkspaceServer.scala
```diff
@@ -364,6 +368,20 @@ object MockWorkspaceServer {
...
+          .withPath(s"/workspaces/%s/%s"
+          .format(mockValidWorkspace.namespace.get, mockValidWorkspace.name.get))
+          .withBody(mockUpdateAttributeOperation.toJson.prettyPrint)
+          .withHeader(authHeader))
+      .respond(
+        response()
+          .withHeaders(header)
+          .withStatusCode(OK.intValue)
+      )
+
```
rushtong: It isn't too bad creating a negative test case. If you define a when case that listens for the values `mockInvalidWorkspace`, and then responds with what Rawls would respond with, you could then call that invalid case in your test class. 

If you did do that, you'd need to modify this when handler to only respond to a valid case, such as the values in `mockValidWorkspace` and call it appropriately in the test class.

src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceServiceSpec.scala
```diff
@@ -41,6 +41,10 @@ class WorkspaceServiceSpec extends FreeSpec with ScalaFutures with ScalatestRout
       MockWorkspaceServer.mockValidWorkspace.namespace.get,
       MockWorkspaceServer.mockValidWorkspace.name.get)
 
+    val attributeUpdatePath = ApiPrefix + "/%s/%s/updateAttributes".format(
+      MockWorkspaceServer.mockValidWorkspace.namespace.get,
+      MockWorkspaceServer.mockValidWorkspace.name.get)
```
rushtong: Here's where you'd add an invalid attributeUpdatePath using mockInvalidWorkspace.

src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceServiceSpec.scala
```diff
@@ -185,6 +189,16 @@ class WorkspaceServiceSpec extends FreeSpec with ScalaFutures with ScalatestRout
...
 
+    "when calling PATCH on the workspaces/*/*/updateAttributes to add/update an attribute" - {
+      "OK response is returned" in {
+        (Patch(attributeUpdatePath, MockWorkspaceServer.mockUpdateAttributeOperation)
+        ~> dummyAuthHeaders
+        ~> sealRoute(routes)) ~> check {
+          status should equal(OK)
+        }
+      }
+    }
```
rushtong: Here you would add the negative case that would return what Rawls would return as an error case.

rushtong:
@MatthewBemis - I'm done reviewing. I think there's a small bit of work left to do, specifically around the yaml and the test. Let me know when you'd like me to re-review.

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -86,6 +90,14 @@ trait WorkspaceService extends HttpService with PerRequestCreator with FireCloud
             }
           }
         } ~
+        path("updateAttributes") {
+          patch {
+            entity(as[Seq[AttributeUpdateOperation]]) { attributeOp =>
```
rushtong: I am reconsidering my initial approval of this. I think we should just pass whatever entity is sent through as the entity without coercing it to `[Seq[AttributeUpdateOperation]]`. This will save you a bunch of code and simplify the passthrough. You can be more clear about what `AttributeUpdateOperation` is in the yaml, but I don't think we need this in the code.

rushtong:
@MatthewBemis - Ignore my comments about adding a negative test case. There's no real value in the effort.

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
MatthewBemis: I'm almost done addressing the comments, but just a tiny bit confused on this part since I couldn't find any examples of it in the code. If we're just going to pass it through without caring about what it is so rawls can deal with it, what would be the entity type here? Would it be String? i.e.:

```
        path("updateAttributes") {
          patch {
            entity(as[String]) { attributeOp =>
              requestContext =>
                externalHttpPerRequest(requestContext, Patch(workspacePath, attributeOp))
            }
          }
        }
```
Or would it be some sort of wildcard type?

rushtong: How about spray.json.JsObject ?

MatthewBemis:
@rushtong Ready for re-review!

rushtong:
:+1: Code looks good. I can't actually run the code, but I'm pretty sure that's a dev environment thing.

rushtong:
@MatthewBemis - Please squash and merge when complete.

PR #78, opened 2015-09-11
Update for new deployment configuration

dmohs:
- in production, this lives at /service/
- /service/ is Swagger, /service/api/ is the authenticated endpoint

Should be deployed with https://github.com/broadinstitute/firecloud-ui/pull/174


dmohs:
Note: Should be deployed with https://github.com/broadinstitute/firecloud-ui/pull/174

Dockerfile
```diff
@@ -1,25 +1,38 @@
-FROM centos:7
+FROM phusion/baseimage
 
-RUN curl https://bintray.com/sbt/rpm/rpm > /etc/yum.repos.d/bintray-sbt-rpm.repo
-RUN yum -y install git sbt && yum clean all
+RUN add-apt-repository ppa:openjdk-r/ppa
+RUN echo "deb https://dl.bintray.com/sbt/debian /" | tee -a /etc/apt/sources.list.d/sbt.list
+RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 642AC823
```
rushtong: Is this for handling auth tokens?

Dockerfile
```diff
@@ -1,25 +1,38 @@
-FROM centos:7
+FROM phusion/baseimage
 
-RUN curl https://bintray.com/sbt/rpm/rpm > /etc/yum.repos.d/bintray-sbt-rpm.repo
-RUN yum -y install git sbt && yum clean all
+RUN add-apt-repository ppa:openjdk-r/ppa
```
rushtong: I have security concerns about this line (an untrusted ppa), but will defer to devops for their opinion

dmohs:
@rushtong Added comments to Dockerfile to address your concerns.

.gitignore
```diff
@@ -10,8 +11,8 @@ dist/*
 target/
 lib_managed/
 src_managed/
-project/boot/
-project/plugins/project/
+/project/boot/
```
davidangb: are the absolute paths here necessary?

Dockerfile
```diff
@@ -1,25 +1,44 @@
-FROM centos:7
+FROM phusion/baseimage
```
davidangb: broad has one (or more) baseimage(s) too - can we use that instead?

Dockerfile
```diff
@@ -1,25 +1,44 @@
...
+COPY project project
+RUN sbt compile
 
-RUN alternatives --set java /usr/lib/jvm/java-1.8.0-openjdk-1.8.0.51-1.b16.el7_1.x86_64/jre/bin/java
+# Compile first to cache it.
+COPY src src
+RUN sbt compile
 
-RUN sbt assembly -Dconfig.file=/usr/firecloud-orchestration/test.conf
+# TODO(dmohs): Re-enable when tests pass.
```
davidangb: are these failing?

Dockerfile
rushtong: https://github.com/broadinstitute/baseimage/blob/master/Dockerfile inherits from phusion/baseimage:0.9.17

Dockerfile
```diff
@@ -1,25 +1,44 @@
...
+# RUN sbt test -Dconfig.file=src/test/resources/test.conf
 
-#No longer needed, remove it. 
-RUN rm -rf /usr/firecloud-orchestration/test.conf
+RUN sbt assembly -Dconfig.file=src/main/resources/application.conf
```
rushtong: Isn't this supposed to be the managed file from Jenkins? If we are pulling it from src, we'll always be getting whatever is hard-coded in master for agora and rawls urls.

rushtong: And I now see that they are gone from application.conf. Never mind.

README.md
```diff
@@ -71,4 +71,19 @@ start sbt with `src/main/resources/application.conf`. For mock testing, use `src
...
+Build the docker container:
+```
+docker build -t firecloud-orch .
+```
+
+Run the container:
+```
+docker run --name orch \
+  -e RAWLS_URL_ROOT='https://rawls.dsde-dev.broadinstitute.org' \
+  -e AGORA_URL_ROOT='https://agora.dsde-dev.broadinstitute.org' \
```
rushtong: Thank you for adding these envs to the readme!!!

.gitignore
dmohs: I will revert. Was trying to get better info from `git status`.

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
```diff
@@ -51,62 +55,50 @@ class FireCloudServiceActor extends HttpServiceActor {
...
-    val uri = forwardedHost match {
-      case Some(x) => baseUri.withAuthority(hostAndPortToAuthority(x))
-      case None => baseUri
+  private def serveIndex(accessToken: String): Route = {
+    val authLine = "$(function() { window.swaggerUi.api.clientAuthorizations.add(" +
+      "'key', new SwaggerClient.ApiKeyAuthorization('Authorization', 'Bearer " + accessToken +
+      "', 'header')); });"
+    val indexHtml = getResourceFileContents(swaggerUiPath + "/index.html")
+    complete {
+      HttpEntity(ContentType(MediaTypes.`text/html`),
```
davidangb: sooooo ... ... this is what we have to do to modify the default swagger-ui without copying their code, huh? Ouch.

src/main/resources/application.conf
```diff
@@ -1,25 +1,3 @@
...
-  // The next three values should be provided by your environment (e.g. in an .sbtopts file)
-  username = ""
-  password = ""
-  commonName = ""
-  authIndex {
-    type = "module"
-    value = "DataStore"
-  }
-}
-
```
davidangb: at this point there's really nothing left in application.conf. The akka section is just boilerplate defaults in case we ever want to change them. We could remove the file altogether, or move all the info from configurations.conf into here.

Dockerfile
davidangb: I thought we even had a scala baseimage too - might come with a jdk and sbt. I don't know where/if that exists, so this could be a red herring.

dmohs: Let's punt on this and revisit if we find one later.

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
dmohs: Yeah, I know.

src/main/resources/application.conf
dmohs: Moved the akka stuff into configurations.conf and removed application.conf.

PR #79, opened 2015-09-14
Delete workspace

MattPutnam:
* Added service call to delete workspace
* Fixed indentation and removed some code duplication in workspace service
* Added swagger docs for get/delete workspace and get/patch acl

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -50,6 +50,113 @@ paths:
       responses:
         200:
           description: List of Method Repository configurations.
+  /workspaces/{workspaceNamespace}/{workspaceName}:
+    get:
+      tags:
+        - workspaces
+      operationId: getWorkspace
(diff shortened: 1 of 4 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -50,6 +50,113 @@ paths:
...
+          in: path
+        - name: workspaceName
+          description: Workspace Name
+          required: true
+          type: string
+          in: path
+        - name: aclUpdates
+          description: Series of ACL updates for workspace
+          required: true
+          type: List[Map[string, string]]
```
davidangb: This is more generic than it should be, though I realize orchestration doesn't enforce anything. No need to change anything right now; just commenting.

davidangb:
Existing code looks good (and thanks for adding GET /workspaces/{workspaceNamespace}/{workspaceName}), but this really should have some tests.

PR #80, opened 2015-09-18
fixed updateAttributes endpoint

MatthewBemis:
Quick fix for something that didn't quite work with rawls in my previous PR


davidangb:
:+1:

PR #81, opened 2015-09-18
DSDEEPB-1164: change method config snapshot from String to Int

davidangb:
Agora and Rawls are both now using an int for config snapshot id. Orchestration was still using a String. This fixes that incompatibility.

*importing configs still has problems* but this unblocks development of those other problems. After this change, when I try to import a config, I get an error
```
{
message: "Error parsing Method Repo response: Object is missing required member 'methodRepoMethod'"
}
```
Which I'm guessing is going to require some UI changes.


eddiebroad:
I think that the "MethodRepoMethod" vs "MethodStoreMethod" is a different problem compared to the string/id issue.  I think it is in the payload

davidangb:
yes, I agree the payload is missing methodRepoMethod, and I didn't try to address that. I'm trying to land this change to unblock work on that incomplete payload.

eddiebroad:
Re: "yes, I agree the payload is missing ..." I only mentioned the methodRepoMethod vs. methodStoreMethod is because you mentioned it in the main PR

eddiebroad:
I'm going to go ahead and merge

PR #83, opened 2015-09-21
fix TSV attribute names on pairs to be case and control

tedsharpe:
rather than case_sample_id and control_sample_id

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
@@ -38,6 +38,28 @@ object EntityClient {
 
   def props(requestContext: RequestContext): Props = Props(new EntityClient(requestContext))
 
+  // Fix up certain attribute names for reference types:
+  // If the attribute name is of the form "something_{entityType}_id", we'll fix it up to be "something" --
+  //   for example, "case_sample_id" becomes "case" if we're loading "sample"s.  This will make "pair" entities prettier.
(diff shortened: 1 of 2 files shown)


davidangb:
I've looked briefly at the code. But ... what's the motivation for this change? Is this driven by any user-facing requirement or product owner request? If not, I'd prefer to *not* include this - I'd rather that we transparently use whatever the user uploaded, instead of transforming names.

tedsharpe:
David, this is my attempt to resolve DSDEEPB-1111.

I think the intent of the author of the TSV spec was to specify, in a 
single column name such as case_sample_id, the name of the attribute 
(case), the type of entity on which the datum is a foreign key (sample), 
and the fact that it is a reference type (id).

I'm just decoding what I believe to be that intent.

Happy to hear suggestions for other resolutions.  For example, we could 
just change the TSV spec to give the column names as "case" and 
"control" in the case of the "pair" object.  This would work because we 
have secret metadata (the json file) describing the schema, but makes 
the overall scheme less flexible.

Regards,
Ted

dvoet:
this code reflects the intent and should be a better and expected user experience.

tedsharpe:
FWIW I think this is weird too. I believe in Firehose the attributes are
called case_sample and control_sample, which would imply just stripping off
the trailing _id.

But I don't really like the idea of encoding meaning in column headers. IMO
said meaning should be stored explicitly in an external schema, which is a
feature we've punted on for MVP.

/$.02

PR #87, opened 2015-09-22
allow for arbitrary renaming of TSV attributes as dictated by model.json

tedsharpe:
ugly, but maybe less so


davidangb:
code looks good and works good. I'd love to see at least one test for this. See https://github.com/broadinstitute/firecloud-orchestration/pull/88 for a suggestion on a test. Ready to merge once test(s) exist(s).

tedsharpe:
Nice.  Thanks for organizing and providing the test.

PR #91, opened 2015-09-23
Add directive to pass through endpoints

MattPutnam:


MattPutnam:
David Mohs tested and approved :+1:

PR #92, opened 2015-09-23
Added endpoints for method/config import/export

MattPutnam:
* Method->config template
* Post config to workspace
* Get individual config details

Tested with live data

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -50,6 +50,61 @@ paths:
       responses:
         200:
           description: List of Method Repository configurations.
+  /template:
+    post:
+      tags:
+        - Method Repository
+        - Workspaces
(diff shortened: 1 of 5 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -50,6 +50,61 @@ paths:
...
+          description: Internal Server Error
+  /configurations/{namespace}/{name}/{snapshotId}:
+    get:
+      tags:
+        - Method Repository
+      operationId: getMethodRepositoryConfiguration
+      summary: Get a Method Repository configuration
+      produces:
+        - application/json
+      parameters:
```
davidangb: could this be $ref: MethodID instead?

davidangb: ah, nm; ignore - I see the difference now.

PR #93, opened 2015-09-24
implemented pass through to publish config to method repo

dvoet:


src/main/resources/application.conf
```diff
@@ -27,7 +27,7 @@ spray.can.server {
 http {
   interface="0.0.0.0"
   port=8080
-  timeoutSeconds = 1
+  timeoutSeconds = 10
```
davidangb: I don't object to this, but I'm curious .. were you running into any problems? Or just changing this proactively?

src/main/resources/swagger/api-docs.yaml
```diff
@@ -435,6 +435,43 @@ paths:
...
+          in: path
+        - name: configToCopy
+          description: Method Configuration to Copy
+          required: true
+          type: string
+          in: body
+          schema:
+            $ref: PublishConfigurationIngest
+      responses:
+        201:
```
davidangb: these response codes don't match what's documented for rawls at https://swagger.dsde-dev.broadinstitute.org/#!/methodconfigs/copy_configuration_to_repo. I assume this is a problem with the rawls swagger?

src/main/resources/application.conf
dvoet: I ran into a problem running locally on docker. It could have just been a blip but...

PR #97, opened 2015-09-24
Passthrough API for copying entities from one workspace to another.

rushtong:
Orchestration pass-through to copy an entity from one workspace to another.

*Note*: Rawls is currently throwing exceptions for this endpoint and it is not usable with real data.

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -266,6 +266,43 @@ paths:
           description: Workspace does not exist
         500:
           description: Internal Server Error
+  /workspaces/{workspaceNamespace}/{workspaceName}/entities/copy:
+    post:
+      tags:
+      - Entities
+      operationId: copyEntities
+      summary: |
+        Copy entities from one workspace to another
+      produces:
+        - application/json
+      parameters:
+        - name: workspaceNamespace
+          description: Workspace Namespace
+          required: true
+          type: string
+          in: path
+        - name: workspaceName
+          description: Workspace Name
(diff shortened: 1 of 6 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -211,6 +211,43 @@ paths:
...
+          description: Workspace Namespace
+          required: true
+          type: string
+          in: path
+        - name: workspaceName
+          description: Workspace Name
+          required: true
+          type: string
+          in: path
+        - name: body
```
davidangb: if the body of this request contains both the source and destination workspaces, why duplicate the namespace/name in the url as well? Ideally, we'd remove it from one of those locations. At minimum, specify in the swagger doc if the url params are source or destination workspace.

rushtong: I was hesitant about going that route. The endpoint of /workspaces/entities/copy is the same as /workspaces/{workspacenamespace}/{workspacename} and there is potential for url collision. I chose this path more for consistency (and to avoid potential collisions) rather than efficiency.

davidangb: how about removing it from the body, then? Make the url parameters the destination workspace, and include the source workspace in the body? You'd have to have extra manipulation code inside the endpoint to translate into the rawls format, but it makes for a better end-user API.

rushtong: I don't disagree with that approach - but it won't be feasible (for me) in this sprint.

PR #100, opened 2015-09-28
Workflow tasks passthrough

dmohs:


src/main/scala/org/broadinstitute/dsde/firecloud/service/SubmissionService.scala
```diff
@@ -22,20 +18,5 @@ trait SubmissionService extends HttpService with PerRequestCreator with FireClou
...
-          passthrough(FireCloudConfig.Rawls.
-            submissionByIdUrl(workspaceNamespace, workspaceName, submissionId), HttpMethods.GET, HttpMethods.DELETE)
-        } ~
-        path("workflows" / Segment / "outputs") { workflowId =>
-          passthrough(FireCloudConfig.Rawls.
-            workflowOutputsByIdUrl(workspaceNamespace, workspaceName, submissionId, workflowId), HttpMethods.GET)
-        }
-      }
-    }
+    passthroughAllPaths("workspaces", FireCloudConfig.Rawls.baseUrl + "/workspaces")
```
davidangb: I like the simplification, but this conflicts with routes in WorkspaceService. Admittedly, FireCloudServiceActor checks WorkspaceService first, so it shouldn't conflict in *practice*, but it feels wrong.

davidangb:
I can see a few paths forward from here, in order of preference:
1) Get passthroughAllPaths working for "workspaces" / Segment / Segment / "submissions" only
2) Remove SubmissionService entirely and move passthroughAllPaths into WorkspaceService, keeping the two special cases there intact. This would require updates to SubmissionServiceSpec, too
3) Write explicit passthroughs for the new submission endpoints

davidangb:
:+1:

PR #102, opened 2015-10-01
the orch handler only needs to handle GET and POST requests from

eddiebroad:


eddiebroad:
Dave,

The orch layer, I envisioned as only handling GET and POST requests from the user.
GET requests should essentially passed through I think.
POST requests however are more complicated!
A POST request pseudo code I think should go like this:

parsedPostData=Parse(request.GetPostData()
uiACLPermsArray=ParseAsPermsArray(parsedPostData)
for(uiACLPerm in uiACLPermsArray){
     agoraACL=translateToAgora(uiACLPerm)
     if(isEmptyList(agoraACL)) {
          sendRequest(method=DELETE,user=agoraACL.getUser(),AgoraACLURL)
          }
     else {
          sendRequest(method=POST,user=agoraACL.getUser(),AgoraACLURL)
          }
     }

What do you think?????

-eddie

eddiebroad:
After we talk about this I can do more PRs this way?  What do you think?

PR #103, opened 2015-10-01
Passthrough for workspace clone

MattPutnam:
As of now (Thursday 8/1 at 3:05PM) Rawls has a pair of bugs, one that bombs no matter what and one that bombs when the source workspace has data in it.  Fixes for both are in review and should be merged soon.  However, this code should be correct.


dmohs:
:+1:

PR #104, opened 2015-10-04
Update entity service and tests

rushtong:
Addresses:
* Entity Service Failure: https://broadinstitute.atlassian.net/browse/DSDEEPB-1458
* Entity Service Tests: https://broadinstitute.atlassian.net/browse/DSDEEPB-824

This PR takes a slightly different approach to mock server tests. I've removed the entity-with-type tests (which will be addressed in an upcoming PR) and instead added mock responses in the same test class that are called by the service.


MattPutnam:
:+1: Looks good to me

PR #106, opened 2015-10-04
Implement orchestration call for export entities by type

rushtong:
This PR addresses exporting entities as a TSV file from the Orchestration layer: https://broadinstitute.atlassian.net/browse/DSDEEPB-1148

Notes:
* This does not fulfill all requirements of DSDEEPB-1148. If you take this PR, do not move the story along as there still needs to be a UI component to invoke this endpoint.
* Please look closely at `EntityMatrix`. As a class, it feels a bit more like java than scala and feedback on construction/implementation would be very much appreciated. 
* I split out tests for this separately from EntityService due to the specialized nature of what it's doing. 
* There are three cases of data conversion that happen on import and need to be reversed on export. `participant_id`, `case_sample`, and `control_sample` columns require special processing.
* I have not been able to test this from the UI (currently blocked), but this can be tested locally with curl. Add `-OJ` to the curl which will tell curl to respect the attachment filename. The rest of the curl command can be gotten from the swagger UI.

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -46,8 +46,5 @@ workspace {
   copyToMethodRepoConfig="/methodconfigs/copyToMethodRepo"
   template="/methodconfigs/template"
   importEntitiesPath="/workspaces/%s/%s/importEntities"
-  submissionsPath="/workspaces/%s/%s/submissions"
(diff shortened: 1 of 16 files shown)


rushtong:
There are previous merges to master in here somehow ... will re-generate a new branch and new PR.

PR #109, opened 2015-10-06
Implement orchestration call for export entities by type

rushtong:
This PR addresses exporting entities as a TSV file from the Orchestration layer: https://broadinstitute.atlassian.net/browse/DSDEEPB-1148

Notes:
* This does not fulfill all requirements of DSDEEPB-1148. If you take this PR, do not move the story along as there still needs to be a UI component to invoke this endpoint.
* Please look closely at EntityMatrix. As a class, it feels a bit more like java than scala and feedback on construction/implementation would be very much appreciated.
* I split out tests for this separately from EntityService due to the specialized nature of what it's doing.
* There are three cases of data conversion that happen on import and need to be reversed on export. participant_id, case_sample, and control_sample columns require special processing.
* I have not been able to test this from the UI (currently blocked), but this can be tested locally with curl. Add -OJ to the curl which will tell curl to respect the attachment filename. The rest of the curl command can be gotten from the swagger UI.

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -427,6 +427,39 @@ paths:
           description: Workspace or entity type does not exist
         500:
           description: Internal Server Error
+  /workspaces/{workspaceNamespace}/{workspaceName}/entities/{entityType}/tsv:
+    get:
+      type: file
+      tags:
+      - Entities
+      operationId: downloadEntitiesTSV
+      summary: |
+        TSV file containing workspace entities of the specified type
+      produces:
+        - text/plain
+      parameters:
+        - name: workspaceNamespace
+          description: Workspace Namespace
+          required: true
+          type: string
+          in: path
+        - name: workspaceName
+          description: Workspace Name
+          required: true
+          type: string
+          in: path
+        - name: entityType
+          description: Entity Type
+          required: true
+          type: string
+          in: path
+      responses:
+        200:
+          description: Workspace entities of specified type in TSV format
+        404:
+          description: Workspace or entity type does not exist
+        500:
+          description: Internal Server Error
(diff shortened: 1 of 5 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/utils/EntityMatrix.scala
```diff
@@ -0,0 +1,114 @@
+package org.broadinstitute.dsde.firecloud.utils
+
+import org.broadinstitute.dsde.firecloud.core.GetEntitiesWithType.EntityWithType
+import spray.json.JsValue
+
+import scala.collection.mutable
+
+class EntityMatrix {
```
dmohs: What about this:
```scala
object EntityMatrix {
  def makeTsvString(entities: List[EntityWithType], entityType: String) = {
    ...
  }
}
```
I think it would be more Scala-y to not store mutable values like headers and rows.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -427,6 +427,39 @@ paths:
           description: Workspace or entity type does not exist
         500:
           description: Internal Server Error
+  /workspaces/{workspaceNamespace}/{workspaceName}/entities/{entityType}/export:
```
dmohs: I would prefer the endpoint `/tsv` or `/as-tsv` rather than `/export`. To me, "export" sounds like you're kicking off an asynchronous task to export the data somewhere else.

rushtong:
@dmohs - Thank you, excellent ideas. Implemented and ready for re-review. I will submit a UI PR to correspond with the endpoint change.

dmohs:
:+1: I like that EntityMatrix nicely contains the functionality.

dmohs:
Good to merge in my opinion.

PR #111, opened 2015-10-09
for methods and configs ACL translation and passthrough to Agora

eddiebroad:
wip - agora permissions endpoints and translations to FireCloud roles

hadle put, delete requests by responding with BadRequest

prepare for fanning

pass both URL and permissions object to handler

cleanup

fix errors for mapping and post passthru

fix declarion to match implementaiton

rearrange and comment

rearrange close brace

it compiles

compiles

cleanup

more cleanup

methods ACL override route

remove unused delete from handler

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -130,6 +130,63 @@ paths:
           description: Method Repository configuration detail
         500:
           description: Internal Server Error
+  /methods/{namespace}/{name}/{snapshotId}/permissions
+    get:
+      tags:
+        - Method Repository
+      operationId: get ACL permissions on a Method Repository method
+      summary: get ACL permissions on a Method Repository method
+      produces:
+        - application/json
+      parameters:
+        - name: namespace
+          description: Method namespace
+          required: true
+          type: string
+          in: path
+        - name: name
+          description: Method name
+          required: true
+          type: string
+          in: path
+        - name: snapshotId
+          description: Method snapshot ID
+          required: true
+          type: string
+          in: path
+      responses:
+        200:
+          description: the indicated method ACL
+        500:
+          description: Internal Server Error
+    post:
+      tags:
+        - Method Repository
+      operationId: get ACL permissions on a Method Repository method
+      summary: get ACL permissions on a Method Repository method
+      produces:
+        - application/json
+      parameters:
+        - name: namespace
+          description: Method namespace
+          required: true
+          type: string
+          in: path
+        - name: name
+          description: Method name
+          required: true
+          type: string
+          in: path
+        - name: snapshotId
+          description: Method snapshot ID
+          required: true
+          type: string
+          in: path
+      responses:
+        200:
+          description: the indicated method ACL once the posted changes have been applied
+        500:
+          description: Internal Server Error
   /configurations/{namespace}/{name}/{snapshotId}/permissions
     get:
       tags:
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -154,6 +154,8 @@ paths:
           required: true
           type: string
           in: path
+        schema:
+          $ref: '#/definitions/MethodConfigACL'
       responses:
         200:
           description: the indicated method ACL
@@ -182,9 +184,13 @@ paths:
           required: true
           type: string
           in: path
+        schema:
+          $ref: '#/definitions/MethodConfigACL'
       responses:
         200:
           description: the indicated method ACL once the posted changes have been applied
+          schema:
+            $ref: '#/definitions/MethodConfigACL'
         500:
           description: Internal Server Error
   /configurations/{namespace}/{name}/{snapshotId}/permissions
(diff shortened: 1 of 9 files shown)


davidangb:
I'd like to see all the TODOs in this PR get addressed - there are TODOs for tests, checks for valid data, and exception-throwing/logging.

src/main/scala/org/broadinstitute/dsde/firecloud/core/AgoraPermissionHandler.scala
```diff
@@ -0,0 +1,144 @@
...
+import spray.routing.RequestContext
+
+import scala.concurrent.Future
+import scala.util.{Failure, Success}
+
+object AgoraPermissionHandler {
+  case class Delete(url: String, user: String)
+  case class Get(url: String)
+  case class Post(url: String, agoraPermissions: List[AgoraPermission])
+  case class Put(url: String, agoraPermission: AgoraPermission)
```
davidangb: If Put and Delete case classes are never used in this code, they should be deleted.

src/main/scala/org/broadinstitute/dsde/firecloud/service/MethodsService.scala
```diff
@@ -1,21 +1,108 @@
...
+
+  val methodsACLOverrideRoute: Route =
+    path("methods" / Segment / Segment / IntNumber / "permissions") { (namespace, name, snapshotId) =>
+
+      // generate the Agora url we'll call
+      val url = FireCloudConfig.Agora.methodsPermissionsFromMethod(namespace, name, snapshotId)
+
+      delete {
+        //this functionality is achieve via "NO ACCESS" in the Post
+        requestContext => requestContext.complete(BadRequest)
```
davidangb: These should be MethodNotAllowed, not BadRequest. However, have you verified with Agora that these ednpoints should be blocked? It's possible they're in use from the CLI.

src/main/scala/org/broadinstitute/dsde/firecloud/service/MethodsService.scala
```diff
@@ -1,21 +1,108 @@
...
+          requestContext => requestContext.complete(BadRequest)
+        }}
+
+
+
+  val configurationsACLOverrideRoute: Route =
+    path("configurations" / Segment / Segment / IntNumber / "permissions") { (namespace, name, snapshotId) =>
+
+      // generate the Agora url we'll call
+      val url = FireCloudConfig.Agora.methodsPermissionsFromMethod(namespace, name, snapshotId)
```
davidangb: should be configurationsPermissionsFromConfig

src/main/scala/org/broadinstitute/dsde/firecloud/service/MethodsService.scala
```diff
@@ -1,21 +1,108 @@
...
 
   lazy val log = LoggerFactory.getLogger(getClass)
 
-  val routes: Route =
-    passthroughAllPaths("methods", FireCloudConfig.Agora.methodsBaseUrl) ~
-    passthroughAllPaths("configurations", FireCloudConfig.Agora.configurationsBaseUrl)
+  // Agora routes that require special handling
+
+  val methodsACLOverrideRoute: Route =
+    path("methods" / Segment / Segment / IntNumber / "permissions") { (namespace, name, snapshotId) =>
```
davidangb: Since the methods and configs endpoints (here and line 64) are so similar, can they be combined? Perhaps use a regex or a Map in place of the hardcoded "methods" and "configurations" in the path. See http://spray.io/documentation/1.2.2/spray-routing/path-directives/pathmatcher-dsl/#basic-pathmatchers for info. Is the only other difference the creation of the Agora url? That could be handled dynamically too.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -104,6 +104,120 @@ paths:
           description: Method Repository configuration detail
         500:
           description: Internal Server Error
+  /methods/{namespace}/{name}/{snapshotId}/permissions
```
davidangb: The swagger definitions should have model objects for post bodies and responses.

davidangb:
How does this code handle the new "public" permission in Agora? I don't know the details of how that permission is supposed to work, so I don't know if this code handles it well.

src/main/scala/org/broadinstitute/dsde/firecloud/service/MethodsService.scala
eddiebroad: Re "These should be MethodNotAllowed, not BadRequest. ....." last I heard the CLI hadn't gotten started on this, but I can check in with Brad about it and see if he can offer any status update

eddiebroad:
Re "How does this code handle the new "public" permission in Agora?...."  It doesn't!  The UI doesn't either!

eddiebroad:
Re: "How does this code handle the new "public" permission in Agora? "  One possibility might be to scan for the user->role assignments for the public user and if the role is determined to be OWNER then to respond with "BAD request"....but is this okay?

davidangb:
Please recreate this PR against the develop branch! This will need to happen before merging, but can be done at your convenience in order to preserve comment history, etc.

PR #112, opened 2015-10-13
encode uris to allow for spaces and other special characters in path

MatthewBemis:
This addresses a bug that would cause an exception to be thrown if there were special characters in a resource name (i.e. a space in a workspace name). I'm not a regex expert so if somebody who is could double check my pattern, that would be great!

Also fixed a little thing in swagger that I noticed was inconsistent with the actual orchestration code.

Jira ticket: https://broadinstitute.atlassian.net/browse/DSDEEPB-1479

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/FireCloudDirectives.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/FireCloudDirectives.scala
@@ -6,8 +6,10 @@ import spray.http.MediaTypes._
 trait FireCloudDirectives extends spray.routing.Directives with PerRequestCreator with spray.httpx.RequestBuilding {
   def respondWithJSON = respondWithMediaType(`application/json`)
 
-  def passthrough(path: String, methods: HttpMethod*) = methods map { inMethod =>
+  def passthrough(unencodedPath: String, methods: HttpMethod*) = methods map { inMethod =>
     val outMethod = new RequestBuilder(inMethod)
+
+    val path = encodeUri(unencodedPath)
 
     // POST, PUT, PATCH
     if (inMethod.isEntityAccepted) {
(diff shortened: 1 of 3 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/FireCloudDirectives.scala
```diff
@@ -35,4 +37,14 @@ trait FireCloudDirectives extends spray.routing.Directives with PerRequestCreato
       }
     }
   }
+
+  def encodeUri(path: String): String = {
+    val pattern = """(https|http)://([^/\r\n]+)(/[^\r\n]*)?""".r
```
rushtong: I'm not seeing how this replaces white space in an url. I see returns and newlines, but no \s or \t, etc. Do you have an example poorly named workspace for this?

MatthewBemis: That regex is just to split the string into the scheme, host, and path so that a uri can be formed from it. The forming of the uri is where the encoding takes place, which is done so automatically. I can share a relevant workspace with you so that you can check the before/after and make sure that everything works (it's called "mbemis space").

MatthewBemis:
@rushtong I've added the tests that you suggested, ready for re-review

src/test/scala/org/broadinstitute/dsde/firecloud/service/FireCloudDirectivesSpec.scala
```diff
@@ -0,0 +1,34 @@
...
+class FireCloudDirectivesSpec extends FreeSpec with ScalatestRouteTest with FireCloudDirectives {
+
+  def actorRefFactory = system
+
+  "FireCloudDirectives" - {
+    "Unencoded passthrough URLs" - {
+      "should be encoded" in {
+        val unencoded = "http://abc.com/path with spaces/"
+        val encoded = encodeUri(unencoded)
+        println(encoded)
```
rushtong: Can we add an assertion here instead of just `println`s? The example I sent you in hipchat was meant as a starter - sorry if I wasn't more clear on that.

MatthewBemis: My bad! I've updated them now. Will squash when given the go-ahead.

rushtong:
:+1: Looks good. Feel free to merge after squashing.

PR #113, opened 2015-10-13
Allow port in URI to enable unit tests

jmthibault79:
After #112 unit tests did not work, because they require local mock server URLs with ports in them, like `http://localhost:8989`


rushtong:
:+1:

PR #114, opened 2015-10-14
DSDEEPB-1339: ErrorReport to enable chained exceptions

jmthibault79:
See https://github.com/broadinstitute/rawls/pull/162 for the Rawls counterpart to this PR.  We intend for this to be a common error reporting structure among the services, which should eventually move to common code.


src/main/scala/org/broadinstitute/dsde/firecloud/service/MethodConfigurationService.scala
```diff
@@ -42,7 +42,8 @@ trait MethodConfigurationService extends HttpService with PerRequestCreator with
               val extReq = Post(FireCloudConfig.Rawls.copyFromMethodRepoConfigUrl, copyMethodConfig)
               externalHttpPerRequest(requestContext, extReq)
             }
-          }
+          } ~
+          complete(StatusCodes.MethodNotAllowed, ErrorReport(StatusCodes.MethodNotAllowed, "HTTP method not allowed"))
```
rushtong: Doesn't spray respond appropriately with a 405 if the path is handled, but method is not? Or are we capturing it here to filter it through ErrorReport? If it's the latter, I think it would be a lot less boilerplate code if we instead customized the default RejectionHandler.

jmthibault79: Right, the only purpose is to wrap it in an ErrorReport.

I really like your default RejectionHandler idea.  I'll look into that.  Do you know of any convenient examples?

rushtong: I don't think we have anything - I only know about it from the documentation: http://spray.io/documentation/1.2.2/spray-routing/key-concepts/rejections/#rejectionhandler

jmthibault79: OK, I gave it a try.

src/test/scala/org/broadinstitute/dsde/firecloud/service/MethodConfigurationServiceSpec.scala
```diff
@@ -162,7 +174,7 @@ class MethodConfigurationServiceSpec extends FreeSpec with ScalaFutures with Sca
       "MethodNotAllowed error is returned" in {
         Put(validCopyFromRepoUrl) ~> sealRoute(routes) ~> check {
           status should equal(MethodNotAllowed)
-          responseAs[String] === "HTTP method not allowed, supported methods: GET"
+          errorReportCheck("FireCloud", MethodNotAllowed)
```
jmthibault79: Orchestration only passes `POST` to Rawls `copyFromMethodRepo` so this check indicates that it's Orchestration returning this error, not Rawls.  See below for the alternate scenario.

src/test/scala/org/broadinstitute/dsde/firecloud/service/MethodsServiceSpec.scala
```diff
@@ -52,7 +50,7 @@ class MethodsServiceSpec extends FreeSpec with ScalaFutures with ScalatestRouteT
       "MethodNotAllowed error is returned" in {
         Put("/methods") ~> sealRoute(routes) ~> check {
           status should equal(MethodNotAllowed)
-          responseAs[String] === "HTTP method not allowed, supported methods: GET"
+          errorReportCheck("Agora", MethodNotAllowed)
```
jmthibault79: Orchestration passes all `/methods` calls to Agora regardless of full path or method.  This check indicates that the passthrough to Agora succeeded, and it's Agora returning this error, not Orchestration.  See above for the alternate scenario.

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
```diff
@@ -30,6 +30,9 @@ class FireCloudServiceActor extends HttpServiceActor {
     route(requestContext)
   }
 
+  // wraps route rejections in an ErrorReport
+  import org.broadinstitute.dsde.firecloud.model.ErrorReport.errorReportRejectionHandler
```
jmthibault79: This import covers all services in *production* code, but it must be added to each *test* individually.  I was unable to make that work as part of the new `ServiceSpec`.

src/main/scala/org/broadinstitute/dsde/firecloud/core/GetEntitiesWithType.scala
```diff
@@ -39,14 +40,14 @@ class GetEntitiesWithTypeActor(requestContext: RequestContext) extends Actor wit
           val entityTypes: List[String] = unmarshal[List[String]].apply(response)
           new EntityAggregator(requestContext, url, entityTypes)
         case Success(response) =>
-          context.parent ! RequestComplete(response.status)
+          context.parent ! RequestComplete(response)
```
jmthibault79: Not sure why it was removing the response body previously.

src/test/scala/org/broadinstitute/dsde/firecloud/service/MethodConfigurationServiceSpec.scala
rushtong: This is correct - the 405 error checks in tests should never hit the downstream endpoints - they should be captured at the orchestration layer.

src/test/scala/org/broadinstitute/dsde/firecloud/service/MethodsServiceSpec.scala
rushtong: Most of these tests were written when we didn't have a concept of passthrough. In my opinion, the vast majority of these tests should go away. Anything that is a passthrough - we really should only be testing that orchestration can perform the passthrough. We should not care about what is passed or what is returned - that should be something we do in when integration testing.

jmthibault79: My thought was to write these as unit/integration agnostic, so they could be migrated to integration tests as appropriate.

rushtong:
I maintain that any test that looks for a 405 is in fact a bad test, but that is beyond the scope of this PR. There are at least two tests for 405s (MethodsService "configurations" and "methods") that are in fact flat-out wrong now that we have a passthrough on all methods for those paths. See https://broadinstitute.atlassian.net/browse/DSDEEPB-1612 for my suggestion. 

Code looks good, tests pass. I can't test with live data - rawls is persnickety. Will re-test when that becomes available and I can get a free block of time.

rushtong:
@jmthibault79 - Can you rebase this please? I'm trying to test against rawls and I think endpoints have been changed in both rawls and the latest orchestration.

rushtong:
:+1:

PR #115, opened 2015-10-14
DSDEEPB-1592: Move all rawls endpoints under /api

davidangb:
All rawls endpoints move under /api for protection. This involved some cleanup and consolidation of hardcoded endpoints into application.conf.

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -37,15 +37,20 @@ methods {
 
 workspace {
   model="model.json"
+  authPrefix="/api"
+  workspacesPath="/workspaces"
   entitiesPath="/workspaces/%s/%s/entities"
   methodConfigsListPath="/workspaces/%s/%s/methodconfigs"
   getMethodConfigPath="/workspaces/%s/%s/methodconfigs/%s/%s"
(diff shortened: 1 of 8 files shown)


dmohs:
:+1: Code looks good to me. Since you're cleaning this up, I hope you can answer this question: What is the advantage of putting our URLs in application.conf?

I see two disadvantages:
1) It requires additional "lazy val..." boilerplate in FireCloudConfig.scala
2) It de-localizes the URL from the code using it: When I'm implementing "create workspace" and thinking about that, I have to go somewhere else when I want to create/verify the URL.

My second question would be, is that advantage worth these sacrifices? That's really the answer I want to figure out. No need to hold up this PR, just hoping you could provide insight.

davidangb:
Well, the GOAL of putting those urls in application.conf is to define them once in an easy-to-change place, in case they ever change. For instance, if rawls ever decides to change their endpoints, like putting them under /api, we could just update application.conf and everything else magically works! Anywhere we need to reference an endpoint - tests/mocks, service implementations, route definitions, etc etc would all be updated.

But as you can see that's not the case, because the calling code hasn't consistently honored that approach - sometimes because we wrote lazy code, and sometimes because there were complications in using that approach. *sigh*

PR #116, opened 2015-10-15
Update workspace service

rushtong:
* Add swagger api doc for missing create call.
* Remove the custom handling of create workspace in favor or using a passthrough request. It was doing unnecessary things that Rawls is now handling correctly.


davidangb:
:+1:

PR #117, opened 2015-10-16
DSDEEPB-1605: first pass at user profile service endpoints

davidangb:
@dmohs , I want to get you a PR/branch early, even though this PR might (probably) needs additional work.

This adds endpoints into Thurloe, in support of user profile/registration. Some important notes about this change:

* user profile endpoints **require** an OIDC_access_token and other OIDC headers. The mod_auth_openidc proxy provides these transparently, provided the requestor sends an Authentication: Bearer xxx token. However, if you run orchestration outside the proxy (e.g. from a simple sbt re-start), you'll need to provide these in your request.
* the UserInfoDirectives/StandardUserInfoDirectives I'm using to get the currently-logged-in user's unique id are stolen from rawls. However, we're not using them quite right; see the TODO at the top of UserService.
* I have not written any tests yet.
* I want to provide a batch POST endpoint to accept an entire user profile; currently, the only way to update a user is a single kvp at a time. I have not done so yet. It's worth asking Thurloe if they will provide a batch endpoint.
* because this adds a requirement for THURLOE_URL_ROOT in the environment, we'll need to update build systems, etc. before this works.

Given that this PR is a blocker for user registration work, I defer to the reviewer on how much needs to be fixed before merging, and how much can be follow-on work.


src/main/resources/swagger/api-docs.yaml
```diff
@@ -1044,6 +1044,138 @@ paths:
...
+          description: Internal server error
+          produces: text/plain
+  /profile/{key}:
+    post:
+      tags:
+        - Profile
+      operationId: setKey
+      summary: Adds or resets a single key/value pair in the user profile service for the currently logged-in user.
+      consumes:
+      - text/plain
```
davidangb: This may be the only place we POST a text/plain value instead of a json object. Should I change this to json?

dmohs: I think text/plain was right for the first iteration.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1044,6 +1044,138 @@ paths:
...
+        200:
+          description: OK
+        400:
+          description: Bad request
+        500:
+          description: Internal server error
+    get:
+      tags:
+        - Profile
+      operationId: getKvp
```
dmohs: I really like these endpoint names.

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
```diff
@@ -62,4 +62,13 @@ object FireCloudConfig {
     def importEntitiesPathFromWorkspace(namespace: String, name: String) = authUrl + importEntitiesPath.format(namespace, name)
   }
 
+  object Thurloe {
+    private val profile = config.getConfig("userprofile")
+    lazy val baseUrl = sys.env.get("THURLOE_URL_ROOT").get
+    lazy val setKey = profile.getString("setKey")
```
dmohs: This is the only part of this PR I don't like, but I would not change it (so your code remains consistent with the rest of the codebase). I recommend we do away with this pattern via https://github.com/broadinstitute/firecloud-orchestration/pull/119

dmohs:
:+1: I've tested this via Swagger. Everything works nicely. Good call on grabbing whichever header is available for the user ID.

PR #119, opened 2015-10-19
Eliminate URL boilerplate

dmohs:
- unrelated Dockerfile tweak
- removes requirement to duplicate URLs in FireCloudConfig.scala
- URLs are defined closer to use
- a side-benefit to this change is that it is more obvious that the "submission ID" URL is only used in tests

As far as I can tell, the only drawback to this change is that URLs can no longer be changed without recompiling code, but I'm not sure that is a use case we want to support.

If we agree that this is better, I will update the other URLs either as part of this PR or in a follow-up.

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/firecloud/service/SubmissionServiceSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/firecloud/service/SubmissionServiceSpec.scala
@@ -1,6 +1,5 @@
 package org.broadinstitute.dsde.firecloud.service
 
-import org.broadinstitute.dsde.firecloud.FireCloudConfig
 import org.broadinstitute.dsde.firecloud.mock.{MockUtils, MockWorkspaceServer}
 import org.broadinstitute.dsde.firecloud.model.SubmissionIngest
 import spray.http.StatusCodes._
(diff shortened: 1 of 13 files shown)


davidangb:
I'm okay with this in general. A few comments/TODOs:

* I want to make sure there's a convention for where we define agora/rawls/thurloe/etc endpoints. "Top of the service definition" is a fine convention - just want something consistent
* I want to make sure URLs are defined once as a constant and referenced by anything that needs it - they should not be defined inline. In other words, there should be one or a small number of code blocks where someone would go to change a URL. Again, "top of the service definition" is a fine convention.

Sadly, I _don't_ like that the submissionID url is only defined in the tests. I also don't like having test-only cruft inside runtime code, so I don't have a great solution for this - but it means going to multiple places if those endpoints ever change.

dmohs:
I agree with consistency. I recommend we don't merge this PR until all URLs are changed so the pattern is well-defined at merge time.

Avoiding code duplication is still important. This change should only reduce the duplication.

Here is why I like making the test-only URL more explicit: Since it is now obvious that URL is only used in test code, we should probably remove it. We should not be writing tests (that are not integration tests) that don't exercise our code, so I think this change makes it more obvious that this particular test is unnecessary.

I like this change because I may not have otherwise noticed this problem.

src/test/scala/org/broadinstitute/dsde/firecloud/service/SubmissionServiceSpec.scala
```diff
@@ -20,18 +19,13 @@ class SubmissionServiceSpec extends ServiceSpec with SubmissionService {
     MockWorkspaceServer.stopWorkspaceServer()
   }
 
-  val localSubmissionsPath = FireCloudConfig.Rawls.submissionsPath.format(
+  val localSubmissionsPath = submissionsPath.format(
```
rushtong: Yes! It is critical that we remember we should not be using a Rawls endpoint as an orchestration endpoint.

davidangb:
I could live with a convention of "define urls at the top of service classes or the top of tests". Keep it somehow consistent between those. (and yes I realize that's what you did, I'm just coming around)

davidangb:
Please recreate this PR against the develop branch!

PR #120, opened 2015-10-19
Mirror for GCS get-metadata call to help with downloads

davidangb:
Create an endpoint within orchestration that contacts GCS to get info about a file, in support of downloading that file (or checking its existence, etc).

See https://cloud.google.com/storage/docs/json_api/v1/objects/get for details of the API.

UI can safely call the orchestration mirror via ajax without any CORS issues.

Remember you can browse GCS via their console, here: https://console.developers.google.com/project/broad-dsde-dev/storage/browser.

A few safe files you can use for manual testing (please don't delete these from GCS):
https://console.developers.google.com/project/broad-dsde-dev/storage/browser/bossdev/vault/test/

Note that it looks like this call returns a **safer** download url than the one we're generating in the UI. Because Google generates the download url instead of us reverse-engineering their downloads, we might want to switch over to this.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/PerRequest.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/PerRequest.scala
@@ -63,6 +63,7 @@ trait PerRequest extends Actor {
     case _:HttpHeaders.Server => true
     case _:HttpHeaders.`Content-Type` => true
     case _:HttpHeaders.`Content-Length` => true
+    case _:HttpHeaders.`Transfer-Encoding` => true
     case _ => false
   }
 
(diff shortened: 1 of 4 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/PerRequest.scala
```diff
@@ -63,6 +63,7 @@ trait PerRequest extends Actor {
     case _:HttpHeaders.Server => true
     case _:HttpHeaders.`Content-Type` => true
     case _:HttpHeaders.`Content-Length` => true
+    case _:HttpHeaders.`Transfer-Encoding` => true
```
davidangb: I found through testing that we were missed this definition in isAutomaticHeader.

davidangb:
Please use https://github.com/broadinstitute/firecloud-orchestration/pull/122 instead.

PR #121, opened 2015-10-20
remove _id from all column keys in imported TSV

MatthewBemis:
For https://broadinstitute.atlassian.net/browse/DSDEEPB-1457.

Remove _id from the end of all column keys, which are specified in the TSV spec: https://docs.google.com/document/d/1iBhW0LwrJYBihQw0qa0c-6hc3kd_mnOqlUlvghlxVMs


rushtong:
:+1:

PR #123, opened 2015-10-20
 for methods and configs ACL translation and passthrough to Agora

eddiebroad:


src/main/scala/org/broadinstitute/dsde/firecloud/core/AgoraPermissionHandler.scala
```diff
@@ -0,0 +1,307 @@
...
+    for( a <- 1 to 256){
+      val randomRoles=generateRandomAgoraRoles()
+      val randomAgoraPerm=AgoraPermission(
+        user=Some("Random_user_"+a),
+        roles=Some(randomRoles))
+      val randomTranslation=randomAgoraPerm.toFireCloudPermission
+      println("For Agora *RANDOM* "+randomAgoraPerm+" the FC translation is "+randomTranslation)
+
+    }
+    return false
```
eddiebroad: this is marked as false here to draw attention to it, but it should return true
visual inspection of output suggests that everything is functioning.  However,  it is suggested to look at it because for some cases it might be worth it to verify that the operation is what is actually desired

src/main/scala/org/broadinstitute/dsde/firecloud/core/AgoraPermissionHandler.scala
```diff
@@ -0,0 +1,307 @@
...
+
+import scala.concurrent.Future
+import scala.util.{Failure, Success}
+
+object AgoraPermissionHandler {
+  case class Get(url: String)
+  case class Post(url: String, agoraPermissions: List[AgoraPermission])
+  def props(requestContext: RequestContext): Props = Props(new GetEntitiesWithTypeActor(requestContext))
+
+  def testFromFCToAgora():Boolean = {
```
rushtong: Why is this test not in a *Spec class? We should try to keep our tests there for consistency sake.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -130,6 +130,128 @@ paths:
           description: Method Repository configuration detail
         500:
           description: Internal Server Error
+  /methods/{namespace}/{name}/{snapshotId}/permissions
```
davidangb: need to end this line with a colon:
```
  /methods/{namespace}/{name}/{snapshotId}/permissions:
```

src/main/resources/swagger/api-docs.yaml
```diff
@@ -130,6 +130,128 @@ paths:
...
+          description: Method name
+          required: true
+          type: string
+          in: path
+        - name: snapshotId
+          description: Method snapshot ID
+          required: true
+          type: string
+          in: path
+        schema:
```
davidangb: indentation matters in yaml ... this line and the next line ($ref) need to be un-indented by one level.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -130,6 +130,128 @@ paths:
...
+          $ref: '#/definitions/MethodConfigACL'
+      responses:
+        200:
+          description: the indicated method ACL once the posted changes have been applied
+          schema:
+            $ref: '#/definitions/MethodConfigACL'
+        500:
+          description: Internal Server Error
+  /configurations/{namespace}/{name}/{snapshotId}/permissions
```
davidangb: needs to end with a colon

src/main/resources/swagger/api-docs.yaml
```diff
@@ -130,6 +130,128 @@ paths:
...
+          description: Method name
+          required: true
+          type: string
+          in: path
+        - name: snapshotId
+          description: Method snapshot ID
+          required: true
+          type: string
+          in: path
+        schema:
```
davidangb: unindent by one level

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1220,6 +1342,17 @@ definitions:
...
+    properties:
+      userId:
+        type: string
+        description: a user's ID or "public"
+      accessLevel:
+        type: string
+        description: one of "NO ACCESS", "READER", or "OWNER"
+    required:
+      - userId
+      - accessLevel
```
davidangb: this file did not work for me when I loaded swagger locally; you can see the syntax problems I found above. Please test locally!

eddiebroad: Re: "this file did not work for me when I loaded swagger locally..." okay, will do!

src/main/scala/org/broadinstitute/dsde/firecloud/core/AgoraPermissionHandler.scala
davidangb: test code should live only within the test package; there should be no test-only code in runtime classes.

rushtong:
My first suggestion is to fix the yaml and then squash. It's hard to test the functionality without the swagger UI.

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
```diff
@@ -25,6 +25,12 @@ object FireCloudConfig {
     lazy val methodsBaseUrl = authUrl + methodsPath
     lazy val configurationsPath = methods.getString("configurationsPath")
     lazy val configurationsBaseUrl = authUrl + configurationsPath
+
+    def methodsPermissionsFromMethod(namespace:String, name:String, snapshotId:Int) = authUrl +
```
rushtong: Suggestion: shorten name. `methodsPermissions` is explanatory enough without being redundant.

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
```diff
@@ -25,6 +25,12 @@ object FireCloudConfig {
...
+      methods.getString("methodsPermissionsPath").format(namespace, name, snapshotId)
+
+    def configurationsPermissionsFromConfig(namespace:String, name:String, snapshotId:Int) = authUrl +
```
rushtong: Same suggestion, `configurationsPermissions` is enough to convey the intent.

src/main/scala/org/broadinstitute/dsde/firecloud/core/AgoraPermissionHandler.scala
```diff
@@ -0,0 +1,307 @@
...
+      val randomTranslation=randomAgoraPerm.toFireCloudPermission
+      println("For Agora *RANDOM* "+randomAgoraPerm+" the FC translation is "+randomTranslation)
+
+    }
+    return false
+  }
+
+
+
+  def testFromAgoraToFC():Boolean = {
```
rushtong: This comment applies to all of the smaller tests in this method. Once the tests are pulled out and into a *Spec class, these should each be individual tests - if we really need them. I'm not really clear what it is you are trying to test. It's important to remember that we are unit testing Orchestration logic and not Agora.

src/main/scala/org/broadinstitute/dsde/firecloud/core/AgoraPermissionHandler.scala
```diff
@@ -0,0 +1,307 @@
...
+      }
+      case Some(u) => FireCloudPermission(Some(u), Some(toFireCloudRole(agoraPermission.roles)))
+    }
+  }
+
+
+  def hasValidUserAndRole(fireCloudPermission: FireCloudPermission):Boolean = {
+    val vRole=this.hasValidRole(fireCloudPermission)
+    val vUser=this.hasValidUser(fireCloudPermission)
+    return (vRole && vUser)
```
rushtong: There is no need for returns in scala. The last line that is executed is automatically the return value. This should just be `vRole && vUser`

The whole thing could be a one-liner... 
``` 
hasValidRole(fireCloudPermission) && hasValidUser(fireCloudPermission)
```

src/main/scala/org/broadinstitute/dsde/firecloud/core/AgoraPermissionHandler.scala
```diff
@@ -0,0 +1,307 @@
...
+  }
+
+
+  def hasValidUser(fireCloudPermission: FireCloudPermission):Boolean = {
+    fireCloudPermission.user match {
+      case Some("") => false
+      case None => false
+      //any non-empty user is okay!
+      case _ => true
+    }
```
rushtong: By convention, most uses of case checks would look for the positive first and handle negative secondarily. I would write this something along the lines of:
```
case x if x.nonEmpty => true
case _ => false
```

src/main/scala/org/broadinstitute/dsde/firecloud/core/AgoraPermissionHandler.scala
```diff
@@ -0,0 +1,307 @@
...
+  }
+
+
+  def hasValidRole(fireCloudPermission: FireCloudPermission):Boolean = {
+    fireCloudPermission.role match {
+      case Some("NO ACCESS") => true
+      case Some("OWNER") => true
```
rushtong: I see these strings a lot ... "OWNER", etc. They should be constants somewhere so that we are reducing the chance for copy/paste errors and also so that the definition is consolidated in a single location. If the word "OWNER" is changed in Rawls to "CREATOR", for instance, it would be a lot nicer if we changed that in one place. I recommend constants in a companion object class defined along side AgoraPermissionsHandler.

src/main/scala/org/broadinstitute/dsde/firecloud/core/AgoraPermissionHandler.scala
```diff
@@ -0,0 +1,307 @@
...
+
+
+  // translation between a FireCloud role and a list of Agora roles
+  def toAgoraRoles(fireCloudRole:Option[String]) = {
+    fireCloudRole match {
+      case None => List("Nothing")
+      case Some("OWNER") => List("Read","Write","Create","Redact","Manage") // Could use "All" instead but this is more precise
+      case Some("READER") => List("Read")
+      case Some("NO ACCESS") => List("Nothing")
+      case _ => List("Nothing") // TODO: throw an exception instead? Log something?
```
rushtong: No need to introduce an exception - `Nothing` should be good.

src/main/scala/org/broadinstitute/dsde/firecloud/core/AgoraPermissionHandler.scala
```diff
@@ -0,0 +1,307 @@
...
+
+  val log = Logging(system, getClass)
+
+  // TODO: tests
+
+  def receive = {
+    // GET requests
+    case AgoraPermissionHandler.Get(url: String) =>
+      val pipeline = authHeaders(requestContext) ~> sendReceive
+      val permissionListFuture: Future[HttpResponse] = pipeline { Get(url) }
```
rushtong: See FireCloudDirectiveUtils.encodeUri and make sure that whatever url you are using has gone through that code path.

src/main/scala/org/broadinstitute/dsde/firecloud/core/AgoraPermissionHandler.scala
```diff
@@ -0,0 +1,307 @@
...
+      permissionListFuture onComplete {
+        case Success(response) =>
+          response.status match {
+            case StatusCodes.OK =>
+              val agoraPermissions = unmarshal[List[AgoraPermission]].apply(response)
+              val fireCloudPermissions = agoraPermissions.map(x => x.toFireCloudPermission)
+              context.parent ! RequestComplete(OK, fireCloudPermissions)
```
rushtong: See `RequestCompleteWithErrorReport` and make sure all `RequestComplete` calls go through that code path.

src/main/scala/org/broadinstitute/dsde/firecloud/core/AgoraPermissionHandler.scala
```diff
@@ -0,0 +1,307 @@
...
+          }
+        case Failure(e) =>
+          context.parent ! RequestComplete(StatusCodes.InternalServerError, e.getMessage)
+          context stop self
+      }
+
+    case AgoraPermissionHandler.Post(url: String, agoraPermissions: List[AgoraPermission]) =>
+      val pipeline = authHeaders(requestContext) ~> sendReceive
+      val permissionListFuture: Future[HttpResponse] = pipeline { Post(url.toString,
+        HttpClient.createJsonHttpEntity(agoraPermissions.toString()))
```
rushtong: I don't like all of this casting/conversion. You should be able to just write this:
```
Post(url, agoraPermissions)
```
and it should just *work*. If it doesn't, then url and agoraPermissions might need extra attention.

src/main/scala/org/broadinstitute/dsde/firecloud/model/MethodRepository.scala
```diff
@@ -1,49 +1,48 @@
...
+    namespace: Option[String] = None,
+    name: Option[String] = None,
+    snapshotId: Option[Int] = None,
+    synopsis: Option[String] = None,
+    documentation: Option[String] = None,
+    owner: Option[String] = None,
+    payload: Option[String] = None,
+    excludedField: Option[String] = None,
+    includedField: Option[String] = None
+  )
```
rushtong: Nice that you removed all of the extraneous annotations.

eddiebroad: I think @davidangb  did that.....somehow his mark disappeared on the commits

src/main/scala/org/broadinstitute/dsde/firecloud/model/MethodRepository.scala
```diff
@@ -1,49 +1,48 @@
...
+
+  // represents a method/config permission as exposed to the user from the orchestration layer
+  case class FireCloudPermission(
+    user: Option[String] = None,
+    role: Option[String] = None
+  ) {
+    //verifies non-empty user and that role is one of READER, OWNER or NO ACCESS
+    def verifyValidUserAndRole = AgoraPermissionHandler.hasValidUserAndRole(this)
+    def toAgoraPermission = AgoraPermissionHandler.toAgoraPermission(this)
```
rushtong: I'm concerned about having these two methods here and referencing `(this)`. I'm pretty sure there's a better approach - but I'll make those comments closer to where it's appropriate.

src/main/scala/org/broadinstitute/dsde/firecloud/model/MethodRepository.scala
```diff
@@ -1,49 +1,48 @@
...
+  }
+
+  // represents a method/config permission as exposed by Agora
+  case class AgoraPermission(
+    user: Option[String] = None,
+    roles: Option[List[String]] = None
+  ) {
+    def toFireCloudPermission = AgoraPermissionHandler.toFireCloudPermission(this)
```
rushtong: Same concern as previous comment.

dmohs:
Note that the style conventions in this PR are not consistent with the rest of the project or Scala style guidelines. These should be fixed before merging.

src/main/scala/org/broadinstitute/dsde/firecloud/service/MethodsService.scala
```diff
@@ -1,21 +1,86 @@
...
+        post {
+          // take the body of the HTTP POST and construct a FireCloudPermission from it
+          entity(as[List[FireCloudPermission]]) {
+            fireCloudPermissions => requestContext =>
+              //scan user and roles of each input to make sure they're valid
+              //this validityFlag represents a list of booleans where each item
+              //is true if an individual input is valid and okay to send to Agora
+              val validityFlags=fireCloudPermissions.map(x => x.verifyValidUserAndRole)
+              //see http://stackoverflow.com/questions/22518773/applying-logical-and-to-list-of-boolean-values
+              val andAllByUsersAndRoles=validityFlags.foldLeft(true)(_ && _)
```
rushtong: This can be simplified a lot. You could get rid of validityFlags completely:
```
val andAllByUsersAndRoles = { fireCloudPermissions.map(_.verifyValidUserAndRole).forall _ }
```

src/main/scala/org/broadinstitute/dsde/firecloud/model/MethodRepository.scala
davidangb: I think because you squashed :)

eddiebroad: Re : "I think because you squashed".....I think so, but it squashing supposed to do that?  I guess if only a single name can go onto a given single commit it makes sense

src/main/scala/org/broadinstitute/dsde/firecloud/service/MethodsService.scala
```diff
@@ -1,21 +1,86 @@
...
+                }
+
+              // translate the FireCloudPermissions into  AgoraPermissions
+              val agoraPermissions = fireCloudPermissions.map(x => x.toAgoraPermission)
+              // pass to AgoraPermissionHandler
+              perRequest(
+                requestContext,
+                Props(new AgoraPermissionActor(requestContext)),
+                AgoraPermissionHandler.Post(url, agoraPermissions))
+          }
```
rushtong: So ... if you find yourself in scala-land creating lots of vals and using if conditions, those are classic signs of java-thinking. This whole method could be simplified to something like this:
```
fireCloudPermissions match {
	case x if x.map(_.verifyValidUserAndRole).forall _ =>
	  perRequest(
		requestContext,
		Props(new AgoraPermissionActor(requestContext)),
		AgoraPermissionHandler.Post(url, fireCloudPermissions.map(_.toAgoraPermission)))
	case _ =>
	  RequestCompleteWithErrorReport(BadRequest,
		"Invalid user or access setting detected! Each permission posted"+
		  " must have a non-empty user and a permission of READER, OWNER, or NO ACCESS!")
}
```
I'm not a fan of the `verifyValidUserAndRole` and `toAgoraPermission` methods ... let me think on that one for a bit.

src/main/scala/org/broadinstitute/dsde/firecloud/service/MethodsService.scala
```diff
@@ -1,21 +1,86 @@
...
   lazy val log = LoggerFactory.getLogger(getClass)
 
-  val routes: Route =
-    passthroughAllPaths("methods", FireCloudConfig.Agora.methodsBaseUrl) ~
-    passthroughAllPaths("configurations", FireCloudConfig.Agora.configurationsBaseUrl)
+  // Agora permissions that require special handling
+
+  val methodsAndConfigsACLOverrideRoute: Route =
+    path( "configurations|methods".r / Segment / Segment / IntNumber / "permissions") { ( agora_ent, namespace, name, snapshotId) =>
+      val url = getUrlFromBasePath(agora_ent, namespace, name, snapshotId)
```
rushtong: Good place for encoding the url with FireCloudDirectiveUtils.encodeUri

src/test/scala/org/broadinstitute/dsde/firecloud/mock/MockMethodsServer.scala
```diff
@@ -16,13 +16,165 @@ import DefaultJsonProtocol._
...
+    else if(randomNum==1)
+      {
+        return "READER"
+      }
+    else
+      {
+        return "OWNER"
+      }
+
+  }
```
rushtong: This method should be a one-liner:
```
Random.shuffle(List("NO ACCESS", "READER", "OWNER")).head
```

src/test/scala/org/broadinstitute/dsde/firecloud/mock/MockMethodsServer.scala
```diff
@@ -16,13 +16,165 @@ import DefaultJsonProtocol._
...
+    {
+      aList=(aList :+ "Redact")
+    }
+    val randomNum5=scala.util.Random.nextInt(2)
+    if(randomNum5 ==0)
+    {
+      aList=(aList :+ "Manage")
+    }
+    return aList
+  }
```
rushtong: This needs to be re-thought. Do we really need this level of complexity for a random list of strings?

src/test/scala/org/broadinstitute/dsde/firecloud/mock/MockMethodsServer.scala
```diff
@@ -16,13 +16,165 @@ import DefaultJsonProtocol._
...
+    aPermList
+  }
+
+  val mockAgoraOwnerAllFivePermission:List[AgoraPermission] ={
+    var aList=List[String]()
+    aList=(aList:+"Read")
+    aList=(aList:+"Write")
+    aList=(aList:+"Create")
+    aList=(aList:+"Redact")
+    aList=(aList:+"Manage")
```
rushtong: The above is a very strong anti-pattern. Just create a list:
```
val agoraRoles = List("Read", "Write", "Create", "Redact", "Manage")
```
And then you can use that in the roles assignment below.

rushtong:
We need to review the test cases - I think there are far too many. There are two basic conditions we're checking for in the main route, one for when the user has *all* user/role permissions and one where they do not. Testing those two cases should be the primary goal - remove any that do not.

rushtong:
@eddiebroad - I've haven't completely reviewed this - but I think it would be good if you spent some time refactoring this before we continue further. Tests all need to pass and the swagger UI needs to work as well so I can do functional testing too.

src/main/scala/org/broadinstitute/dsde/firecloud/core/AgoraPermissionHandler.scala
eddiebroad: I don't understand "No need to introduce an exception - Nothing should be good."

rushtong: Your line of code is this
```
case _ => List("Nothing") // TODO: throw an exception instead? Log something?
```
I responded to the question you asked about what to do in this case.

davidangb:
Okay, I'm submitting this for re-review. A lot of code has changed; here's the highlights:
* I've moved validation of the POST payloads from the UI into the spray json unmarshalling layer, by making the case class fields Strings instead of Option[String]s, and by adding require() validation inside the case class. This significantly reduces code elsewhere.
* I've added a RejectionHandler to give good error messages when/if the above validation fails.
* I did NOT do the same validation for the data retrieved from Agora. We are more lenient here. If we somehow end up with bad data in Agora, that lenience will give us a little more flexibility in debugging.
* I've totally refactored the tests; there's a bunch of unit testing and targeted endpoint testing. I separated MethodsServiceSpec and MethodsServiceACLSpec
* I have NOT fixed the "AgoraPermissionHandler.toAgoraPermission(this)" non-idiomatic Scala. Let's save that for an optimization.

Note that this does not currently function with the UI/develop branch. I believe there are bugs in UI.

Once review is complete, this definitely needs a rebase/squash.

src/main/scala/org/broadinstitute/dsde/firecloud/core/AgoraPermissionHandler.scala
```diff
@@ -0,0 +1,121 @@
...
+      case Success(response) =>
+        response.status match {
+          case StatusCodes.OK =>
+            try {
+              val agoraPermissions = unmarshal[List[AgoraPermission]].apply(response)
+              val fireCloudPermissions = agoraPermissions.map(x => x.toFireCloudPermission)
+              context.parent ! RequestComplete(OK, fireCloudPermissions)
+              context stop self
+            } catch {
+              // TODO: more specific and graceful error-handling
```
rushtong: I'm thinking about this too. Not in scope for this PR, but definitely a tech-debt thing that we should try to address,

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala
```diff
@@ -84,4 +91,11 @@ object ModelJsonProtocol {
 
   implicit val impFireCloudKeyValue = jsonFormat2(FireCloudKeyValue)
   implicit val impThurloeKeyValue = jsonFormat2(ThurloeKeyValue)
+
+  // don't make this implicit! It would be pulled in by anything including ModelJsonProtocol._
+  val entityExtractionRejectionHandler = RejectionHandler {
+    case MalformedRequestContentRejection(errorMsg, _) :: _ =>
+      complete(BadRequest, errorMsg)
+  }
```
rushtong: My only comment would be that this could be located with ErrorResponse.

src/main/scala/org/broadinstitute/dsde/firecloud/service/MethodsService.scala
```diff
@@ -15,17 +23,58 @@ object MethodsService {
...
+                Props(new AgoraPermissionActor(requestContext)),
+                AgoraPermissionHandler.Post(url, fireCloudPermissions.map(_.toAgoraPermission)))
+          }
+        }
+      } ~
+      delete {
+        complete(StatusCodes.MethodNotAllowed)
+      } ~
+      put {
+        complete(StatusCodes.MethodNotAllowed)
```
rushtong: Do we need to explicitly handle the put and delete cases? Won't spray do that for us if it can't find the path + method combination?

src/test/scala/org/broadinstitute/dsde/firecloud/mock/MockAgoraACLServer.scala
```diff
@@ -0,0 +1,117 @@
...
+import org.mockserver.model.HttpResponse._
+import spray.http.StatusCodes._
+import spray.json._
+
+import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol.impAgoraPermission
+import DefaultJsonProtocol._
+
+object MockAgoraACLServer {
+
+  val methodsServerPort = 8989
```
rushtong: This val is also in MockUtils

src/test/scala/org/broadinstitute/dsde/firecloud/service/MethodsServiceACLSpec.scala
```diff
@@ -0,0 +1,148 @@
...
+  val sourceBadUser = """[{"user":"foo@broadinstitute.org","role":"OWNER"},{"user":"","role":"READER"}]"""
+  val jsonBadUser = sourceBadUser.parseJson.asInstanceOf[JsArray]
+
+    /* ACL endpoints.
+       We unit test the individual translations, so we only need to augment the unit tests here. We need to test:
+        * handling of lists (unit tests work with single objects)
+        * configs vs. methods paths are not mixed up (we use a single code path for both)
+        * rejections when invalid data POSTed
+        * DELETE/PUT methods are rejected (we supersede those with POST, and should block passthroughs)
+     */
```
rushtong: In light of this approach (which I really like, i.e. splitting out orch logic testing from passthroughs) - we might want to start packaging our tests a little more granularly. Having everything in org.broadinstitute.dsde.firecloud.mock will become difficult to sort through over time.

rushtong:
:+1: No real concerns here - I think this is great especially in terms of documentation clarity.

davidangb:
See #146.

PR #124, opened 2015-10-20
DSDEEPB-1708: move Agora urls under /api/v1

davidangb:
I *think* this fixes everything. However, it's hard to test by clicking through the UI, because sometimes our communication with agora goes through rawls, and rawls isn't updated yet.

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -31,6 +31,7 @@ http {
 }
 
 methods {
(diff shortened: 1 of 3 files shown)


rushtong:
Local testing is problematic at the moment - I get a 404 on pretty much everything.

davidangb:
got the hipchat okay from Greg; merging.

PR #125, opened 2015-10-21
encode uris in GET entities_with_type

MatthewBemis:
Now you can load entities in workspaces that have fancy characters in their names


rushtong:
:+1: - Matt's looking into the possibility of an entity type having invalid characters which would mean further encoding is necessary in the aggregator class.

rushtong:
:+1:

PR #126, opened 2015-10-21
Move encode to object and use in GetEntitiesWithType url construction

rushtong:
@MatthewBemis - Let me know what you think about this approach. I move the encode method so it was easier to call from classes that don't work closely with PerRequest and used the encode method on url construction just to be safe.


MatthewBemis:
:+1: I like this approach! It's a lot cleaner than what I was working on. There still seems to be a problem in rawls so when testing this we shouldn't expect to actually get any entities that have spaces in their types, but as far as orchestration is concerned, this solves everything.

rushtong:
If that works for you, feel free to merge this into your branch and then I'll revisit your PR.

PR #130, opened 2015-10-23
Rawls passthrough for deleting entities

MattPutnam:
Passthrough all 3 endpoints at that path because why not, and document GET and DELETE.  Not swaggering PATCH for now because we're not using it (we're not using GET either but whatever).


dmohs:
:+1: Code looks good. I'm just going to assume the Swagger is valid.

PR #131, opened 2015-10-24
Add travis and coveralls support

rushtong:
This is ready for review. 
* I have found that there are sporadic timeout issues when travis runs tests, especially when there are simultaneous runs triggered from a PR and a push. We need to look into ways to try and minimize that noise. 
* Notifications is turned off by default. I think having the notification checks clear in the PR is enough for a reviewer to determine status.

### DIFF
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,0 +1,1 @@
+service_name: travis-ci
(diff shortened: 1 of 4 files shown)


dmohs:
I don't know Travis or Coveralls, but this looks clean and straightforward to me. :+1:

PR #132, opened 2015-10-26
Make sure url is encoded for download entity feature

rushtong:
Does not complete the story of DSDEEPB-1148, but implements the new `encodeUri` functionality.


dmohs:
:+1: Nice catch.

PR #135, opened 2015-10-26
DSDEEPB-1730 (rev 2): increase spray's content-length limit to 50m

davidangb:
change spray.can.server.parsing, not spray.can.parsing when setting max Content-Length; see spray/spray#346.


dmohs:
:+1:

PR #136, opened 2015-10-27
Update tests so that they more accurately reflect what the endpoints …

rushtong:
…handle.


dmohs:
:+1: Huge improvement over previous approach. Love it. Only two tests fail for me (EntitiesWithTypeServiceSpec) which I think always do, so this is good to go I think.

PR #137, opened 2015-10-27
Refactor remote service urls. Significant update to tests. Clean up s…

rushtong:
…ome extraneous mock handlers.


rushtong:
As an aside, I had to re-start the travis build 3 times to get it to pass. It was failing on another, unrelated TSV Parser test.

dmohs:
:+1: Would prefer a 100-column limit to make this easier to review, but love this PR overall.

rushtong:
There is a test failure after rebasing - looking at that now and will update once that is fixed.

PR #139, opened 2015-10-28
Batch delete actor implemented

MattPutnam:
Also found a cosmetic error in the EntitiesWithType actor as I was referencing it.

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -420,6 +420,31 @@ paths:
           description: One or more entities of that name/type already exist
         500:
           description: Internal Error
+  /workspaces/{workspaceNamespace}/{workspaceName}/entities/delete:
+    post:
+      tags:
+      - Entities
+      operationId: deleteEntities
+      summary: Bulk delete entities from a workspace
+      produces:
+        - application/json
+      parameters:
+        - name: workspaceNamespace
+          description: Destination Workspace Namespace
+          required: true
+          type: string
+          in: path
+        - name: workspaceName
+          description: Destination Workspace Name
+          required: true
+          type: string
+          in: path
+        - name: body
+          description: Entities to delete
+          required: true
+          in: body
+          schema:
+            $ref: DeleteEntities
   /workspaces/{workspaceNamespace}/{workspaceName}/entities/{entityType}:
     get:
       type: Entity
@@ -1475,6 +1500,27 @@ definitions:
       name:
         type: string
         description: Entity name
+  DeleteEntities:
+    properties:
+      recursive:
+        type: boolean
+        description: Whether or not to recursively delete (currently unimplemented)
+      entities:
+        type: array
+        items:
+          type: EntityID
+        description: List of entities to be deleted
+    required:
+      - recursive
+      - entities
+  EntityID:
+    properties:
+      entityType:
+        type: string
+        description: Entity Type
+      entityName:
+        type: string
+        description: Entity Name
   MethodID:
     properties:
       methodNamespace:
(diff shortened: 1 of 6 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -1475,6 +1500,27 @@ definitions:
       name:
         type: string
         description: Entity name
+  DeleteEntities:
+    properties:
+      recursive:
+        type: boolean
```
davidangb: until ui/orch/rawls supports recursion, I see you're setting this to false in the UI. Add "default: false" to this property definition to keep it aligned. False is probably the right default anyway.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1475,6 +1500,27 @@ definitions:
...
+        description: Whether or not to recursively delete (currently unimplemented)
+      entities:
+        type: array
+        items:
+          type: EntityID
```
davidangb: use "$ref: EntityID" here to get full schema support!

src/main/scala/org/broadinstitute/dsde/firecloud/core/GetEntitiesWithType.scala
```diff
@@ -79,7 +79,7 @@ class GetEntitiesWithTypeActor(requestContext: RequestContext) extends Actor wit
           val missingReports = errors collect { case (originalError, Failure(_)) => originalError }
 
           val errorMessage = {
-            val baseMessage = "%d failures out of %d attempts retrieving entityUrls.  Errors: %s".format(entityTypes.size, errors.size, errors mkString ",")
+            val baseMessage = "%d failures out of %d attempts retrieving entityUrls.  Errors: %s".format(errors.size, entityTypes.size, errors mkString ",")
```
davidangb: thank you :)

src/main/resources/swagger/api-docs.yaml
```diff
@@ -420,6 +420,31 @@ paths:
           description: One or more entities of that name/type already exist
         500:
           description: Internal Error
+  /workspaces/{workspaceNamespace}/{workspaceName}/entities/delete:
```
davidangb: add expected response codes: 200 OK, 400 Malformed Response, 500 Internal Server Error

src/main/scala/org/broadinstitute/dsde/firecloud/core/DeleteEntities.scala
```diff
@@ -0,0 +1,61 @@
...
+        case Success(responses) if responses.forall(_.status.isSuccess) =>
+          context.parent ! RequestComplete(OK)
+          context stop self
+        case Success(responses) =>
+          val errors = responses.filterNot(_.status.isSuccess) map { e => (e, ErrorReport.tryUnmarshal(e)) }
+          val errorReports = errors collect { case (_, Success(report)) => report }
+          val missingReports = errors collect { case (originalError, Failure(_)) => originalError }
+
+          val errorMessage = {
+            val baseMessage = "%d failures out of %d attempts deleting entities.  Errors: %s".format(errors.size, entities.size, errors mkString ",")
```
davidangb: Cosmetic fix - since this will likely end up being shown to the end user, I'd prefer something like:
%d entities deleted, with %d failures
followed by the error messages.

davidangb:
Review complete - a few minor comments inline. Big issue is that this functionality is missing tests. See #141 .

src/main/resources/swagger/api-docs.yaml
MattPutnam: I'm having local swagger gen issues.  Do you mean that this would become:

    items:
      $ref: EntityID

Or something else?

davidangb: yup, just replace "type: EntityID" with "$ref: EntityID". Worked for me.

PR #140, opened 2015-10-29
Change fields to match Rawls

MattPutnam:
I find it a tad suspicious that this isn't causing unmarshalling errors, but it seems to work.


davidangb:
:+1:

PR #143, opened 2015-11-03
Profile registration service.

rushtong:
Orchestration side of profile registration service. 

Parent story (https://broadinstitute.atlassian.net/browse/DSDEEPB-1815) is not complete without the UI side which is not implemented yet.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileActor.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileActor.scala
@@ -1,0 +1,87 @@
+package org.broadinstitute.dsde.firecloud.core
+
+import akka.actor.{Actor, Props}
+import akka.event.Logging
+import org.broadinstitute.dsde.firecloud.core.ProfileActor.UpdateProfile
+import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
+import org.broadinstitute.dsde.firecloud.model._
+import org.broadinstitute.dsde.firecloud.service.{UserService, FireCloudRequestBuilding}
+import org.broadinstitute.dsde.firecloud.service.PerRequest.RequestComplete
+import spray.client.pipelining._
+import spray.http.HttpResponse
+import spray.http.StatusCodes._
+import spray.httpx.SprayJsonSupport._
+import spray.routing.RequestContext
+
+import scala.collection.immutable.Iterable
+import scala.concurrent.Future
+import scala.util.{Failure, Success}
+
+object ProfileActor {
+  case class UpdateProfile(userInfo: UserInfo, profile: Profile)
+  def props(requestContext: RequestContext): Props = Props(new ProfileActor(requestContext))
+}
+
+class ProfileActor(requestContext: RequestContext) extends Actor with FireCloudRequestBuilding {
+
+  implicit val system = context.system
+  import system.dispatcher
+  val log = Logging(system, getClass)
+
+  override def receive: Receive = {
+
+    case UpdateProfile(userInfo: UserInfo, profile: Profile) =>
+      val pipeline = authHeaders(requestContext) ~> sendReceive
+      val profilePropertyMap = profile.propertyValueMap
+      val posts: Iterable[Future[HttpResponse]] = profilePropertyMap map {
+        case (key, value) =>
+          val kv = FireCloudKeyValue(Some(key), Some(value))
+          pipeline {
+            Post(UserService.remoteSetKeyURL, ThurloeKeyValue(Some(userInfo.getUniqueId), Some(kv)))
+          }
+      }
+      val propertyUpdates: Future[List[HttpResponse]] = Future sequence posts.toList
+
+      val kv2 = FireCloudKeyValue(Some("isRegistrationComplete"), Some("true"))
+      val completionUpdate = pipeline {
+        Post(UserService.remoteSetKeyURL, ThurloeKeyValue(Some(userInfo.getUniqueId), Some(kv2)))
+      }
+
+      propertyUpdates onComplete {
+        case Success(responses) if responses.forall { _.status == OK } =>
+          log.debug("Profile is saved, update isRegistrationComplete status.")
+          completionUpdate onComplete {
+            case Success(response) if response.status == OK =>
+              context.parent ! RequestComplete(OK, response.entity)
+              context stop self
+            case Success(response) =>
+              context.parent ! RequestCompleteWithErrorReport(response.status, response.toString)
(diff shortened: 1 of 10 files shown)


davidangb:
Take a look at the validation approach I used in #123; I think it might be simpler. Specifically, look at the FireCloudPermission case class. I put the require() statements right in the case class, and made its vals Strings instead of Option[String]s. This means we validate at the time of creation of the case class, instead of allowing it to be created and validating after the fact.

Additionally, note that I added handleRejections(entityExtractionRejectionHandler) to the POST route in MethodsService; this gives a nice error message to the user if the POST payload can't be unmarshalled into the case class because of the validation.

(this is not a full-scale code review)

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileActor.scala
```diff
@@ -0,0 +1,86 @@
...
+          log.debug("Profile is saved, update isRegistrationComplete status.")
+          completionUpdate onComplete {
+            case Success(response) if response.status == OK =>
+              context.parent ! RequestComplete(OK, response.toString)
+              context stop self
+            case Success(response) =>
+              context.parent ! RequestCompleteWithErrorReport(response.status, response.toString)
+              context stop self
+            case _ =>
+              context.parent ! RequestCompleteWithErrorReport(InternalServerError, "Failure posting profile: [%s]".format(profileProperties))
```
davidangb: nitpick: all the profile keys were saved; it's just the isRegistrationComplete that failed. Update the log message so it's not confused with the one at line 80.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileActor.scala
```diff
@@ -0,0 +1,86 @@
...
+              context stop self
+          }
+
+        case Success(responses) =>
```
davidangb: This block is also used in the entity-deletion code. It would be nice to extract this into a reusable utility or some Spray command (a route? a directive? I'm not even sure what the right object is). Don't hold up the PR for this; it's an optimization.

src/main/scala/org/broadinstitute/dsde/firecloud/service/FireCloudRequestBuilding.scala
```diff
@@ -31,4 +31,13 @@ trait FireCloudRequestBuilding extends spray.httpx.RequestBuilding {
     addCredentials(OAuth2BearerToken("mF_9.B5f-4.1JqM"))
   }
 
+  def dummyUserIdHeaders(userId: String) = {
+    addCredentials(OAuth2BearerToken("mF_9.B5f-4.1JqM")) ~>
+      addHeader(RawHeader("OIDC_CLAIM_sub", userId)) ~>
```
davidangb: remove the _sub. I believe we've determined that the openid-connect proxy creates _sub, while the oauth proxy creates _user_id. Let's use just _user_id to mimic the current proxy.

src/test/scala/org/broadinstitute/dsde/firecloud/service/UserServiceSpec.scala
```diff
@@ -0,0 +1,153 @@
...
+  "UserService" - {
+
+    "when GET-ting or DELETE-ing a single profile key" - {
+      "MethodNotAllowed response is not returned" in {
+        List(HttpMethods.GET, HttpMethods.DELETE) map {
+          method =>
+            new RequestBuilder(method)(s"/profile/$exampleKey") ~>
+              dummyUserIdHeaders(uniqueId1) ~> sealRoute(routes) ~> check {
+              log.debug(s"$method /profile/$exampleKey: " + status)
+              status shouldNot equal(MethodNotAllowed)
```
davidangb: here (and elsewhere), I'd rather see:
```
status should equal(OK)
```
... in other words, test for the specific positive case, not a single negative case.

rushtong: I'm going to disagree here. If we test for OK, then what we're testing Thurloe logic, not orchestration logic. Where the call is a passthrough, the only thing orchestration is doing is to pass the call onto some downstream service. Orchestration doesn't know what the response will be nor does it care what the response is. I could see the point if orchestration did something with the response, but in these passthrough cases, it does not.

rushtong:
FireCloudPermission is much cleaner - thanks for the pointer!

src/test/scala/org/broadinstitute/dsde/firecloud/service/UserServiceSpec.scala
davidangb: Hmmmm, good point, I agree with you after hearing your stance. All set.

rushtong:
@davidangb - I was able to successfully test this through swagger by running both UI and Orch locally and it is now ready for re-review. 
Summary:
* PR comments addressed
* Moved profile validation to require statements in Profile
  * This allowed for cleaning up `UserService` quite a bit
  * Maybe we should move common validation methods (nonEmpty, etc.) to a common object? This seems like a thing we will need more of.
* Updated tests somewhat. In one non-passthrough case, I was mistakenly not checking for OK status - that is fixed.
* `fullName` is now just `name` - UI is sending over name, not fullName. 
* Added `MalformedRequestContentRejection` to `ErrorReport`. This seems to work correctly in all tests.

src/main/scala/org/broadinstitute/dsde/firecloud/model/Profile.scala
```diff
@@ -10,3 +8,28 @@ case class ThurloeKeyValue(
...
+      f =>
+        f.setAccessible(true)
+        f.getName -> f.get(this).asInstanceOf[String]
+    } toMap
+  }
+
+}
+
+object ProfileValidator {
+  def nonEmpty(field: String): Boolean = {
```
davidangb: couldn't this just be:
```
def nonEmpty(field: String) = !field.trim.isEmpty
``
?

I know the match/case is all Scala-like, but still ...

davidangb:
One minor code comment inline; otherwise good to go, works for me locally.

Feel free to fix the comment, squash if desired, and merge ... or push up another rev if that's easier, and I'll click the merge button.

src/main/scala/org/broadinstitute/dsde/firecloud/model/Profile.scala
rushtong: Why yes, yes it can - I like the one liners where appropriate.

rushtong:
Got the OK to merge from David in hipchat.

PR #144, opened 2015-11-03
DSDEEPB-1723: OAuth implementation in orchestration

davidangb:
ready-for-review implementation of OAuth in orchestration layer.

You will need to set up new env vars for this:
* FIRECLOUD_URL_ROOT: what's the base url for the UI? (https://firecloud.dsde-dev.broadinstitute.org or https://local.broadinstitute.org). We need this to cleanly redirect the browser.
* GOOGLE_SECRET_JSON: the json, as downloaded from Google Developers Console, representing the client id/secret in use.

These are set up for the dev environment, but will need to be configured in staging/alpha/etc before this code goes out there.

You'll also need to make sure your credentials as set up in Google have the FIRECLOUD_URL_ROOT as an allowed origin, and have FIRECLOUD_URL_ROOT/service/oauth2callback as an authorized redirect URI.

To test, go to https://local.broadinstitute.org/service/login?path=foo in your browser to start the OAuth process. Your browser will be redirected a few times and you should end up at https://local.broadinstitute.org/?token={yourtokenhere}#foo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,8 +89,8 @@ Run the container:
 ```
 docker run --rm --name orch \
   -p 8080:8080 \
-  -e GOOGLE_CLIENT_ID=GOOGLE_CLIENT_ID \
-  -e GOOGLE_SECRET_JSON=GOOGLE_SECRET_JSON \
+  -e GOOGLE_CLIENT_ID=$GOOGLE_CLIENT_ID \
+  -e GOOGLE_SECRET_JSON=$GOOGLE_SECRET_JSON \
   -e RAWLS_URL_ROOT='https://rawls.dsde-dev.broadinstitute.org' \
   -e AGORA_URL_ROOT='https://agora.dsde-dev.broadinstitute.org' \
   -e THURLOE_URL_ROOT='https://thurloe.dsde-dev.broadinstitute.org' \
--- a/README.md
+++ b/README.md
@@ -39,8 +39,8 @@ sbt
 > assembly
 ```
 
-Ensure that `GOOGLE_CLIENT_ID` (with origins set to `http://localhost:8080` and `https://localhost:8080` - see below), 
-`AGORA_URL_ROOT`, `RAWLS_URL_ROOT`, and `THURLOE_URL_ROOT` are defined in your environment, and execute the jar with the path to the jar and path of the desired config file:
+Ensure that `GOOGLE_CLIENT_ID` and `GOOGLE_SECRET_JSON` (with origin set to `https://local.broadinstitute.org` - see below), 
+`AGORA_URL_ROOT`, `RAWLS_URL_ROOT`, `THURLOE_URL_ROOT`, and `FIRECLOUD_URL_ROOT` are defined in your environment, and execute the jar with the path to the jar and path of the desired config file:
 
 ```
 java -Dconfig.file=src/main/resources/application.conf \
@@ -56,7 +56,8 @@ sbt
 ## Testing
 
 ```
-AGORA_URL_ROOT='http://localhost:8989' RAWLS_URL_ROOT='http://localhost:8990' THURLOE_URL_ROOT='http://localhost:8991' sbt test
+AGORA_URL_ROOT='http://localhost:8989' RAWLS_URL_ROOT='http://localhost:8990' \
+    THURLOE_URL_ROOT='http://localhost:8991' FIRECLOUD_URL_ROOT='https://local.broadinstitute.org' sbt test
 ```
 
 ## Debugging
@@ -80,16 +81,20 @@ Add your docker host as an authorized JavaScript origin, one entry per line.  Na
 Set your client ID in your environment:
 ```
 export GOOGLE_CLIENT_ID='...'
+export GOOGLE_SECRET_JSON='...'
+
 ```
 
 Run the container:
 ```
 docker run --rm --name orch \
   -p 8080:8080 \
-  -e GOOGLE_CLIENT_ID=$GOOGLE_CLIENT_ID \
+  -e GOOGLE_CLIENT_ID=GOOGLE_CLIENT_ID \
+  -e GOOGLE_SECRET_JSON=GOOGLE_SECRET_JSON \
   -e RAWLS_URL_ROOT='https://rawls.dsde-dev.broadinstitute.org' \
   -e AGORA_URL_ROOT='https://agora.dsde-dev.broadinstitute.org' \
   -e THURLOE_URL_ROOT='https://thurloe.dsde-dev.broadinstitute.org' \
+  -e FIRECLOUD_URL_ROOT='https://local.broadinstitute.org' \
   broadinstitute/firecloud-orchestration
 ```
 
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -3,5 +3,6 @@ addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.12.0")
 addSbtPlugin("io.spray" % "sbt-revolver" % "0.7.2")
 
 addSbtPlugin("org.scoverage" % "sbt-coveralls" % "1.0.0")
+
 // latest coveralls (1.0.0) is *only* compatible with scoverage 1.0.4 or less: https://github.com/scoverage/sbt-coveralls/issues/49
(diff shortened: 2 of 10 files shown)


project/plugins.sbt
```diff
@@ -3,5 +3,6 @@ addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.12.0")
 addSbtPlugin("io.spray" % "sbt-revolver" % "0.7.2")
 
 addSbtPlugin("org.scoverage" % "sbt-coveralls" % "1.0.0")
+
```
davidangb: This blank line allowed my local IntelliJ to parse this file; it got errors before.

src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
```diff
@@ -0,0 +1,83 @@
...
+            redirect(gcsAuthUrl, StatusCodes.TemporaryRedirect)
+          } catch {
+            /* we don't expect any exceptions here; if we get any, it's likely
+                a misconfiguration of our client id/secrets/callback. But since this is about
+                authentication, we are extra careful.
+             */
+            case e: Exception => {
+              log.error("problem during OAuth redirect", e)
+              // TODO: here and elsewhere, redirect to a UI error page instead of issuing a 401?
+              complete(StatusCodes.Unauthorized, e.getMessage)
```
davidangb: See the TODO: should this redirect to a UI error page instead?

README.md
```diff
@@ -79,16 +80,20 @@ Add your docker host as an authorized JavaScript origin, one entry per line.  Na
...
+
 ```
 
 Run the container:
 ```
 docker run --rm --name orch \
   -p 8080:8080 \
-  -e GOOGLE_CLIENT_ID=$GOOGLE_CLIENT_ID \
+  -e GOOGLE_CLIENT_ID=GOOGLE_CLIENT_ID \
+  -e GOOGLE_SECRET_JSON=GOOGLE_SECRET_JSON \
```
dmohs: I don't think these lines work.

src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
dmohs: I would suggest removing the TODO and adding a story for this if/when we actually run into this issue.

src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
```diff
@@ -0,0 +1,83 @@
...
+          try {
+            // is it worth breaking this out into an async/perrequest actor, so we don't block a thread?
+            // not sure what the google library does under the covers - is it blocking?
+            val gcsTokenResponse = HttpGoogleServicesDAO.getTokens(actualState, expectedState, code)
+            val accessToken = gcsTokenResponse.access_token
+
+            val uiRedirect = Uri(FireCloudConfig.FireCloud.baseUrl)
+              .withFragment(actualState)
+              .withQuery(("token", accessToken))
+            redirect(uiRedirect, StatusCodes.TemporaryRedirect)
```
dmohs: This doesn't look up-to-date with my last email on the subject.

dmohs:
:+1: Code looks good. Just needs to be updated with a better UI redirect URL, otherwise good to go.

src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
davidangb: Could you be more specific? I just re-read your email and noticed that I'm using "token" instead of "access_token" for the query param. I'll fix that. But is there more? Help me out ...

src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
davidangb: Logged as DSDEEPB-1914

README.md
davidangb: you're right; they're missing the dollar signs. Was there more?

davidangb:
Questions back to you, @dmohs ... once you respond, I'll update the PR

README.md
dmohs: That was it.

src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
dmohs: Sorry, my bad. You're getting the "path" parameter at the login endpoint. Ideally, you will persist that somehow (it looks like you're using some sort of `state` thingy). Now, all you have to do is:
```scala
.withFragment(path_from_state_thingy+"?access_token="accessToken)
```
and you're done! All fragment, no query. If you include a query string, then removing it causes the browser to reload the page which is fine but unnecessary.

davidangb: Have you tried it? Other than token vs. access_token, this is working as you expect in my local testing.

    https://local.broadinstitute.org/service/login?path=methods

handles authentication, then redirects to

    https://local.broadinstitute.org/?token={my_access_token}#methods

this *is* using the `state` thingy, which keeps us from having to persist anything; the `state` param is passed back and forth between the client and Google during auth, totally round-tripped. See DSDEEPB-1907 for a follow-on story about persistence - I didn't want to go there until we had to, it's nice to keep it stateless.

davidangb:
Updated based on PR feedback. Inline comment threads are now outdated; please see discussion there!

dmohs:
This result URL is not ideal:
`https://local.broadinstitute.org/?token={my_access_token}#methods`
whereas this result URL would be ideal:
`https://local.broadinstitute.org/#methods?token={my_access_token}`

Can you turn the former into the latter?

davidangb:
I *can* formulate the url as you ask ... but that's against the url spec. Standard says fragment comes after query. Are you sure?
https://en.wikipedia.org/wiki/Fragment_identifier

src/main/config/docker-compose.yaml.ctmpl
```diff
@@ -8,10 +8,12 @@ app:
...
-    GOOGLE_CLIENT_ID: {{.Data.env_client_id}}
-    AGORA_URL_ROOT: {{.Data.env_agora_url}}
-    RAWLS_URL_ROOT: {{.Data.env_rawls_url}}
-    THURLOE_URL_ROOT: {{.Data.env_thurloe_url}}
+    - GOOGLE_CLIENT_ID={{.Data.env_client_id}}
+    - AGORA_URL_ROOT={{.Data.env_agora_url}}
+    - RAWLS_URL_ROOT={{.Data.env_rawls_url}}
+    - THURLOE_URL_ROOT={{.Data.env_thurloe_url}}
+    - FIRECLOUD_URL_ROOT={{.Data.env_firecloud_url}}
+    - GOOGLE_SECRET_JSON={{.Data.env_google_json}}
```
davidangb: I needed to switch to this yaml syntax to support storing json in the value for GOOGLE_SECRET_JSON. (I think. I couldn't get it to work in the other format. Whatevs.)

src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
```diff
@@ -0,0 +1,86 @@
...
+             */
+            case e: Exception => {
+              log.error("problem during OAuth redirect", e)
+              complete(StatusCodes.Unauthorized, e.getMessage)
+            }
+          }
+        }
+      }
+    } ~
+    path("callback") {
```
davidangb: I needed to move the callback url, because "oauth2callback" conflicted with the standalone OAuth proxy.

davidangb:
Updated based on PR feedback and merge with develop branch:
* changed the docker-compose yaml to add the firecloud url root and the google json
* moved the callback url from "oauth2callback" to "callback" because the former conflicted with the standalone OAuth proxy
* changed the final redirect to the UI to embed the access token in the fragment, instead of including the access token in a querystring.

I've updated Vault contents for local. I don't know right now where to update them for the dev env, will need to talk to devops for that.

davidangb:
:+1: from @dmohs via hipchat, merging!

PR #145, opened 2015-11-04
DSDEEPB-1911: allow GCS object names containing slashes

davidangb:
quick change to support GCS objects with slashes - i.e. "subdirectories" of buckets.

Note: this fixes a problem in orch. In UI testing, I saw this working with my own test cases, but NOT with the test case mentioned in DSDEEPB-1911. That may be a UI bug.


rushtong:
:+1:

PR #147, opened 2015-11-04
increase route test timeouts from 1 to 5 seconds to pass Travis.

davidangb:
We have a test in Travis that fails most of the time but not all the time, and I have a gut feeling Travis is just being slow.


rushtong:
:+1:

PR #149, opened 2015-11-06
DSDEEPB-1757: store OAuth refresh tokens via rawls

davidangb:
with this PR:
* we *always* request a refresh token during login. This means that you'll *always* see the Google screen asking your permission. We have DSDEEPB-1759 as a follow-on story to make this more graceful.
* once we have a refresh token, we PUT it to rawls.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
@@ -2,13 +2,28 @@ package org.broadinstitute.dsde.firecloud.service
 
 import org.broadinstitute.dsde.firecloud.FireCloudConfig
 import org.broadinstitute.dsde.firecloud.dataaccess.HttpGoogleServicesDAO
-import org.broadinstitute.dsde.firecloud.model.OAuthException
+import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
+import org.broadinstitute.dsde.firecloud.model.{RawlsToken, OAuthException}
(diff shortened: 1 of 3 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
```diff
@@ -59,6 +74,30 @@ trait OAuthService extends HttpService with PerRequestCreator with FireCloudDire
...
+              // we can't use the standard externalHttpPerRequest here, because the requestContext doesn't have the
+              // access token yet; we have to add the token manually
+              val pipeline = addCredentials(OAuth2BearerToken(accessToken)) ~> sendReceive
+              val tokenStoreFuture: Future[HttpResponse] = pipeline { tokenReq }
+
+              // we intentionally don't gate the login process on storage of the refresh token. Token storage
+              // will happen async. If token storage fails, we rely on underlying services to notice the
+              // missing token and re-initiate the oauth grants.
+              tokenStoreFuture onComplete {
+                case Success(response) => {
```
rushtong: Nitpick: we don't need the wrapping braces inside the case block

rushtong:
I would assume that `GoogleAuthorizationCodeFlow` is a blocking process just to be on the safe side. It probably does make sense to delegate this to an actor process.

rushtong:
When running orchestration via the docker command, the swagger interface no longer works. This might be an acceptable tech-debt problem.

rushtong:
:+1: Works!

PR #150, opened 2015-11-09
Add firecloud-oauth2.json template

coreone:


davidangb:
@coreone As is, this needs secrets populated in the local env (client id 806222273987-o97cd3100qndg8vi39ns4sviohrg7474.apps.googleusercontent.com). 

However, the approach I've taken up to now is to include this information inside the secret/dsde/%s/firecloud-orchestration/docker-compose.yaml path (see dev or local for an example). Are you advocating for your approach instead?

davidangb:
:+1: after in-person discussion

PR #151, opened 2015-11-11
allow login process to cross environments with whitelist check

davidangb:
Changes in this PR:
* swagger doc (via hack) for login endpoint
* "path" param, representing the user's current nav state, is now optional
* add new "prompt" param, controlling whether or not to force-request a refresh token. This satisfies DSDEEPB-1760; typical usage should omit this param and allow orchestration to use its default value
* add new "callback" param, controlling where the OAuth process should redirect to after successful auth. This allows an orchestration instance running in dev to redirect back to local.broadinstitute.org after authn - in turn, this allows developers to run a local UI instance talking to a dev orchestration instance.
* during the login dance, check with rawls for a cached refresh token. If one does not exist, or if the cached token is too old, restart auth and force-request a token. The restart should be invisible to the user (unless the user watches the multiple browser redirects)

Note: the "callback" param checks against the whitelist of allowed JS origins defined in the Google credential used by the orchestration instance. To work, this will require:
* any host, such as http://local.broadinstitute.org, that we want to redirect to MUST be listed in the allowed JS origins for the Google credential
* the Google credential json MUST be redeployed to the relevant orchestration instance's secrets

I realize these conditions are a PITA. But, this redirect is not part of the OAuth flow itself, and is therefore not enforced by Google; we must have some enforcement on it.

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1469,6 +1469,41 @@ paths:
           description: Key not found
         500:
           description: Internal server error
+  /../login:
+    get:
+      tags:
+        - OAuth
+      operationId: login
+      summary: Starts the authentication flow for a user
+      description: This is an unauthenticated endpoint at /login (not /api/login) that initiates the OAuth flow.
+        Authentication cannot be initiated from within swagger, because auth requires the entire browser
+        to redirect, and swagger uses ajax. To test, fill in your desired query params, then copy and paste the
+        request url into your browser address bar.
+      parameters:
+      - name: path
+        in: query
+        description: The url fragment to use after finishing authentication
+        required: false
+        type: string
+      - name: prompt
+        in: query
+        description: use "force" to always request a new refresh token
+        required: false
+        type: string
+        enum: ["auto","force"]
+        default: "force"
+      - name: callback
+        in: query
+        description: The host to use after finishing authentication. Include the scheme, but no trailing slash, e.g. https://firecloud.broadinstitute.org
+        required: false
+        type: string
+      responses:
+        307:
+          description: Successful redirect after authentication
+        401:
+          description: Unauthorized
+        500:
+          description: Internal server error
 
 definitions:
   WorkspaceIngest:
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
@@ -3,7 +3,8 @@ package org.broadinstitute.dsde.firecloud.service
 import org.broadinstitute.dsde.firecloud.FireCloudConfig
 import org.broadinstitute.dsde.firecloud.dataaccess.HttpGoogleServicesDAO
 import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
-import org.broadinstitute.dsde.firecloud.model.{RawlsToken, OAuthException}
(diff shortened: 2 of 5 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -1469,6 +1469,41 @@ paths:
           description: Key not found
         500:
           description: Internal server error
+  /../login:
```
davidangb: yes, this is a total hack. I'm still looking for a way to define this endpoint without changing the basePath for EVERY other endpoint.

src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
```diff
@@ -122,4 +150,12 @@ trait OAuthService extends HttpService with PerRequestCreator with FireCloudDire
       }
     }
 
+
+  def initiateAuth(state: String, approvalPrompt: String) =  {
+    val gcsAuthUrl = HttpGoogleServicesDAO.getGoogleRedirectURI(state, approvalPrompt=approvalPrompt)
+    redirect(gcsAuthUrl, StatusCodes.TemporaryRedirect)
+  }
```
davidangb: this was factored out into its own method to support future functionality; it's overkill for this PR, but I'll be using this in other places soon.

src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
```diff
@@ -33,27 +34,32 @@ trait OAuthService extends HttpService with PerRequestCreator with FireCloudDire
...
     path("login") {
       get {
-        parameter("path") { userpath =>
-          // TODO: future story: generate and persist unique token along with the user path
-          val state = userpath
-
-          // approval prompt - "force" to force a refresh token
-          // TODO: future story: make dynamic based on whether or not the user has a refresh token stored already
-          val approvalPrompt = "force"
+        parameters("path".?, "prompt".?, "callback".?) { (userpath, prompt, callback) =>
```
dmohs: Option: the `callback` parameter is a URL-encoded:
```
https://www.example.com/whatever-i-want/somepath#more-stuff?my-access-token-param={access_token}
```
You store this entire thing in `state` below. Then...

src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
```diff
@@ -122,4 +153,48 @@ trait OAuthService extends HttpService with PerRequestCreator with FireCloudDire
...
+    log.info(s"initiateAuth: $approvalPrompt")
+    val gcsAuthUrl = HttpGoogleServicesDAO.getGoogleRedirectURI(state, approvalPrompt=approvalPrompt)
+    redirect(gcsAuthUrl, StatusCodes.TemporaryRedirect)
+  }
+
+  private def completeAuthWithRedirect(actualState: String, accessToken: String) = {
+    // as created in the /login endpoint, the actual state contains the desired callback url
+    // and the desired callback fragment, separated by a hash. But, either value may be the empty string.
+
+    val redirectParts = actualState.split("#")
```
dmohs: ... Replace `"{access_token}"` with `accessToken` in `actualState`, then parse it as a URL.
Check the parts you want against the whitelist, redirect.

dmohs:
:+1: Looks good to me. I have a couple of comments on how I think the redirection could be simplified and more robust, but I have no reason to think this won't work correctly as-is, so that could be a follow-up story. If you merge without addressing these, please open a bug.

davidangb:
Filed DSDEEPB-2003 and merging.

PR #152, opened 2015-11-13
Post-swagger cleanup

rushtong:
* Minor organization name refactor in build.sbt
* Remove unused swagger library in build.sbt
* Pulled in joda libraries that were needed after removing swagger
* Removed all swagger annotations from case classes


dmohs:
:+1: Yayyyyyyyyyy!!!!!!!!!!

PR #153, opened 2015-11-13
Move profile endpoints under register.

rushtong:
This PR goes hand in hand with a UI component: https://github.com/broadinstitute/firecloud-ui/pull/284


davidangb:
looks good, works good - thanks especially for cleaning up the swagger stuff.

PR #154, opened 2015-11-13
Vault Update for .ctmpl

mmonnar:
Updated .ctmpl. WAIT for D Mohs to finish putting in the server.crt and server.key for dev and local into vault the proper way, with the value as "value".


dmohs:
:+1:

PR #155, opened 2015-11-16
DSDEEPB-1714: store the access token in a cookie during login

davidangb:
This is important for download proxy functionality; our rewrite rules look for a cookie.

I value any feedback on this, especially security-related feedback. If we think it's a bad idea to store the token in a cookie, we can come up with another approach - perhaps sending it in the url for download requests (and only for download requests)

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
@@ -7,7 +7,7 @@ import org.broadinstitute.dsde.firecloud.model.{RawlsTokenDate, RawlsToken, OAut
 import org.joda.time.{Days, DateTime}
 import org.slf4j.LoggerFactory
 import spray.http.Uri._
-import spray.http.{HttpResponse, OAuth2BearerToken, StatusCodes, Uri}
+import spray.http._
 import spray.routing._
 import spray.client.pipelining._
 import spray.httpx.SprayJsonSupport._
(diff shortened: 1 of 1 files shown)


dmohs:
My preference would be to create this cookie only when a user clicks a download link in the UI, and with a very short (5 minute) expiration.

rushtong:
I agree with David on this when the cookie should get triggered.

src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
```diff
@@ -192,7 +192,13 @@ trait OAuthService extends HttpService with PerRequestCreator with FireCloudDire
       .withPath(Uri.Path./)
       .withFragment(s"$redirectFragment?access_token=$accessToken")
 
-    redirect(Uri(uiRedirect.toString), StatusCodes.TemporaryRedirect)
+    // we require a spray DateTime, not a joda DateTime
+    // expire in one day
+    val expiration = spray.http.DateTime.now + (1000 * 60 * 60 * 24)
```
rushtong: If we don't trigger the cookie on the actual download endpoint, then I think this duration should be shortened. I think the current login expires after an hour - but I can't pin down where that is configured.

davidangb: Shortening to an hour is fine. I think you're looking for this code: https://github.com/broadinstitute/firecloud-orchestration/blob/develop/src/main/resources/swagger/auth-page.html

If we only place the cookie upon download, we can make it even shorter. I'll work on that.

davidangb:
Okay, please see https://github.com/broadinstitute/firecloud-ui/pull/293. If that PR goes through, this entire PR can be deleted; all the changes are in the UI.

davidangb:
this is now obsolete, given  broadinstitute/firecloud-ui#293. closing.

PR #158, opened 2015-11-17
changed pointer for firecloud-oauth2.json.ctmpl to go to secret/commo…

mmonnar:
…n/firecloud-oauth2.json in vault


davidangb:
Does this PR mean that the secret is not environment-specific? That seems wrong ... I believe we want different credentials in dev vs. prod.

davidangb:
:+1:

PR #160, opened 2015-11-19
Tweaks to run orch locally for development

dmohs:
Minor Dockerfile cleanups.

run.sh now attempts to find the .jar file. If it does not find one, it runs sbt '~ reStart' so incremental building works on every file save.

./script/create-configs.sh creates the config files from vault and, without arguments, modifies the resulting configs so that local.broadinstitute.org points to the orchestration containers.

Tested both sbt reStart and .jar modes locally.

### DIFF
--- a/script/create-configs.sh
+++ b/script/create-configs.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+set -euox pipefail
+IFS=$'\n\t'
+
+
+run_with_ui=${1:-'false'}
+
+docker run --rm -it -v "$PWD":/working broadinstitute/dsde-toolbox \
+  render-templates.sh local "$(<~/.vault-token)"
+
+if [ "$run_with_ui" = 'false' ]; then
+  sed -i '' 's/"20080:80"/"80:80"/1' target/config/proxy-compose.yaml
+  sed -i '' 's/"20443:443"/"443:443"/1' target/config/proxy-compose.yaml
+  sed -i '' 's/basePath: \/service/basePath: \//1' src/main/resources/swagger/api-docs.yaml
+fi
--- a/src/docker/run.sh
+++ b/src/docker/run.sh
@@ -1,3 +1,11 @@
-#!/bin/sh
+#!/bin/bash
+set -euox pipefail
+IFS=$'\n\t'
 
-exec java -jar /app/target/scala-2.11/FireCloud-Orchestration-assembly-*.jar
+
+if [ -e /app/target/scala-2.11/FireCloud-Orchestration-assembly-*.jar ]; then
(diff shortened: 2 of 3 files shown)


src/docker/run.sh
```diff
@@ -1,3 +1,11 @@
...
+IFS=$'\n\t'
 
-exec java -jar /app/target/scala-2.11/FireCloud-Orchestration-assembly-*.jar
+
+if [ -e /app/target/scala-2.11/FireCloud-Orchestration-assembly-*.jar ]; then
+  exec java -jar /app/target/scala-2.11/FireCloud-Orchestration-assembly-*.jar
+fi
+
+cd /app
+exec sbt '~ reStart'
```
davidangb: make this an if-else-fi ? Otherwise ... will we run sbt reStart when java exits? (not sure under what condition that would happen, but still)

davidangb:
This all looks good to me, with the one comment inline.

You and I disagree philosophically on what the default for create-configs.sh should be (standalone orch vs. composed orch/ui). Might be worth polling the team for tiebreaker opinions. I'm fine either way.

script/create-configs.sh
```diff
@@ -0,0 +1,15 @@
...
+
+run_with_ui=${1:-'false'}
+
+docker run --rm -it -v "$PWD":/working broadinstitute/dsde-toolbox \
+  render-templates.sh local "$(<~/.vault-token)"
+
+if [ "$run_with_ui" = 'false' ]; then
+  sed -i '' 's/"20080:80"/"80:80"/1' target/config/proxy-compose.yaml
+  sed -i '' 's/"20443:443"/"443:443"/1' target/config/proxy-compose.yaml
+  sed -i '' 's/basePath: \/service/basePath: \//1' src/main/resources/swagger/api-docs.yaml
```
davidangb: ah - here's an annoying thing. Running this script (with run_with_ui=false) writes to a git-controlled file. Suddenly your git env is not clean :(

src/docker/run.sh
dmohs: Good call. Definitely clearer.

script/create-configs.sh
dmohs: Yep. This is the penalty I have to pay for running a non-standard configuration (orch by itself).

dmohs:
create-config.sh is a helper script for running orch under my own preferred config (by itself without UI). The script is only necessary because I want to be able to do that without stepping on anyone who does not share my preference. My assumption is that the "normal" config is already well-scripted, so this wouldn't be used in that case.

davidangb:
:+1:

PR #161, opened 2015-11-23
DSDEEPB-1355: centralized swagger instance

davidangb:
I *think* this is what's needed to use swagger via DSDE's centralized swagger instance. Without deploying it to there, I can't know for sure (unless I set up a new JS harness just for testing, but I'd rather not do that). It's possible this will need revision later.

Sadly, swagger-ui has bugs (https://github.com/swagger-api/swagger-ui/issues/1517, https://github.com/swagger-api/swagger-ui/issues/1330) that prevent us from defining security once at the top level - we have to define it once for each endpoint. *sigh*

I have NOT tackled any changes to basePath (DSDEEPB-2069). This is only the work for centralized swagger.

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -26,6 +26,12 @@ paths:
       responses:
         200:
           description: List of workspaces.
+      security:
+        - googleoauth:
+            - openid
+            - email
+            - profile
+            - devstorage.full_control
     post:
       tags:
         - Workspaces
@@ -52,6 +58,12 @@ paths:
           description: Workspace by that name already exists
         500:
           description: Internal Server Error
+      security:
+        - googleoauth:
+            - openid
+            - email
+            - profile
+            - devstorage.full_control
   /api/methods:
     get:
       tags:
@@ -64,6 +76,12 @@ paths:
       responses:
         200:
           description: List of Method Repository methods.
+      security:
+        - googleoauth:
+            - openid
+            - email
+            - profile
+            - devstorage.full_control
   /api/configurations:
     get:
       tags:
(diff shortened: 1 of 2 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -901,43 +1075,55 @@ paths:
...
+          description: Method Configuration Name
+          name: configName
+          required: true
+          type: string
+      security:
+        - googleoauth:
+            - openid
+            - email
+            - profile
+            - devstorage.full_control
```
davidangb: fixing some indentation in this block.

rushtong:
Code looks good. OAuth doesn't work for me locally, but I think that needs to be a devop task. 

screen shot 2015-11-24 at 9 56 27 am

rushtong:
This is definitely broken on firecloud-dev. I will work with devops on sorting out what needs to happen.

screen shot 2015-11-24 at 10 56 35 am

davidangb:
This is not expected to work on dev; only on the centralized swagger. The OAuth functionality within the swagger ui requires some customizations to the swagger index page that I haven't implemented.

However, swagger on https://firecloud-orchestration.dsde-dev.broadinstitute.org/ should work exactly as it did before, without requiring you to toggle the oauth buttons. I see that working fine.

Getting swagger to work on our choice of either firecloud-orchestration/ or firecloud/service/ is a separate jira story (DSDEEPB-2069).

davidangb:
oh and the oauth stuff isn't expected to work locally, either (local should still work the same way it always has, without using oauth). The oauth functionality requires JS customization within swagger-ui - try viewing source on swagger.dsde-dev - and I'd rather not make those changes since we're pulling in swagger via webjars.

rushtong:
Sounds good. I'll move the story back to in progress as there is more to do.

PR #162, opened 2015-11-24
Fix "close over sender" anti-pattern

rushtong:
* Fix the close over sender problem (see http://helenaedelson.com/?p=879 for more details)
* Allow for PerRequest stop the child actors instead of explicitly stopping them.
* Implemented pipeTo pattern as per PR Feedback

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/core/AgoraPermissionHandler.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/core/AgoraPermissionHandler.scala
@@ -2,22 +2,23 @@ package org.broadinstitute.dsde.firecloud.core
 
 import akka.actor.{Actor, Props}
 import akka.event.Logging
+import akka.pattern.pipe
+
 import org.broadinstitute.dsde.firecloud.model.MethodRepository.{AgoraPermission, FireCloudPermission}
 import org.broadinstitute.dsde.firecloud.model.MethodRepository.ACLNames._
 import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
 import org.broadinstitute.dsde.firecloud.model.RequestCompleteWithErrorReport
-import org.broadinstitute.dsde.firecloud.service.{FireCloudDirectiveUtils, FireCloudRequestBuilding}
-import org.broadinstitute.dsde.firecloud.service.PerRequest.RequestComplete
-import org.broadinstitute.dsde.firecloud.HttpClient
+import org.broadinstitute.dsde.firecloud.service.FireCloudRequestBuilding
+import org.broadinstitute.dsde.firecloud.service.PerRequest.{PerRequestMessage, RequestComplete}
+
 import spray.client.pipelining._
 import spray.http.StatusCodes._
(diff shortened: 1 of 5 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/core/AgoraPermissionHandler.scala
```diff
@@ -96,26 +93,19 @@ class AgoraPermissionActor (requestContext: RequestContext) extends Actor with F
...
             }
           case x =>
-            context.parent ! RequestCompleteWithErrorReport(x, response.entity.asString)
-            context stop self
+            parent ! RequestCompleteWithErrorReport(x, response.entity.asString)
         }
       case Failure(e) =>
-        context.parent ! RequestCompleteWithErrorReport(StatusCodes.InternalServerError, e.getMessage)
-        context stop self
+        parent ! RequestCompleteWithErrorReport(StatusCodes.InternalServerError, e.getMessage)
```
dvoet: My recommendation would be to use the pipeTo pattern:
```
import akka.pattern.pipe
future pipeTo actor
```

instead of passing in the actor to pass the message to, handleAgoraResponse should return Future[PerRequestMessage] then your code looks like
```
handleAgoraResponse(context.parent, permissionListFuture) pipeTo context.parent
```
this avoid the whole "close over..." business entirely AND forces you to return something useful from handleAgoraResponse

dvoet:
my general comment is that any place you have `Future { actorRef ! message }` you should instead have `Future { message } pipeTo actorRef

dvoet:
this will also help in your exception handling

src/main/scala/org/broadinstitute/dsde/firecloud/core/DeleteEntities.scala
```diff
@@ -21,41 +23,44 @@ object DeleteEntities {
...
+        val allSucceeded = responses.forall(_.status.isSuccess)
+        allSucceeded match {
+          case true =>
+            RequestComplete(OK)
+          case false =>
+            val errors = responses.filterNot(_.status.isSuccess) map { e => (e, ErrorReport.tryUnmarshal(e)) }
+            val errorReports = errors collect { case (_, Success(report)) => report }
+            val missingReports = errors collect { case (originalError, Failure(_)) => originalError }
+            val errorMessage = {
+              val baseMessage = "%d entities deleted, with %d failures.  Errors: %s".format(entities.size, errors.size, errors mkString ",")
```
dmohs: nit: line length (maybe adjust your editor settings?)

dmohs:
:+1: Looks good to me. Fix nits if you want. Merge when ready.

src/main/scala/org/broadinstitute/dsde/firecloud/core/DeleteEntities.scala
rushtong: Yeah, it bugs me too. I was hoping to avoid changing the code this much - the original code suffers from the same problem and I wanted to focus this task on just the actor refactoring. I'll make some tweaks across the board and squash.

rushtong:
CI checks pass, merging.

PR #163, opened 2015-11-25
Log to syslog in supported environments

dmohs:


rushtong:
:+1: Looks good. Had no effect with local (I assume there's no log driver set up for local env) and did the right thing for dev.

PR #164, opened 2015-11-25
Fix json deserialization problem.

rushtong:
The code was only looking for certain known columns to try and deserialize. Updated so that it will always look and therefore handle any unknown cases.


dmohs:
:+1:

PR #165, opened 2015-11-30
store refresh token to rawls, not access token

davidangb:
one-word change. Discovered by Blue Team and discussed over email.


rushtong:
:+1:

PR #166, opened 2015-11-30
fix GCS scope for centralized swagger

davidangb:
I tried (and failed) to use the shorthand for the scope; this fixes it.


rushtong:
Like the last version, I couldn't get this to run locally. However, the changes look fine - just hoping this works on dev.

PR #167, opened 2015-11-30
DSDEEPB-2111: Unable to load workspaces with space in name

MattPutnam:
I'm not 100% solid on how the `passthroughAllPaths` directive works, but this seems to fix it and doesn't break the other usages.


rushtong:
:+1:

PR #168, opened 2015-12-01
Remove now unused client id and secrets from proxy compose

abaumann:


davidangb:
Are these also removed from the Vault secrets?

abaumann:
yes they are

abaumann:
ooo i need to do local

abaumann:
yes all set now

davidangb:
:+1:

PR #169, opened 2015-12-01
Add Rawls User Registration

rushtong:
This PR corresponds to this one in UI: https://github.com/broadinstitute/firecloud-ui/pull/317

* To test, you will need to create a new user account in gmail. There is a limit as to how many times google will let you use your phone for sms verification codes - please test wisely.
* There needs to be a follow-on story for better error display in the UI. When a new user is registered, they are not enabled in LDAP so get big red Unauthorized messages in the workspace tab.

Also addresses: https://broadinstitute.atlassian.net/browse/DSDEEPB-2181


src/test/scala/org/broadinstitute/dsde/firecloud/service/UserServiceSpec.scala
```diff
@@ -137,10 +130,9 @@ class UserServiceSpec extends ServiceSpec with UserService {
       }
     }
 
-    // Testing the only non-passthrough routes in UserService so we are testing for correct status
     "when POST-ting an incomplete profile" - {
       "BadRequest response is not returned" in {
```
davidangb: nitpick: should this be "BadRequest **is** returned"?

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -0,0 +1,130 @@
...
+
+  implicit val system = context.system
+  import system.dispatcher
+  val log = Logging(system, getClass)
+
+  override def receive: Receive = {
+
+    case UpdateProfile(userInfo: UserInfo, profile: Profile) =>
+      val parent = context.parent
+      val pipeline = authHeaders(requestContext) ~> sendReceive
```
davidangb: looks like you could define pipeline at the class level, once

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  /**
+    * Rawls can come back with a conflict (500 wrapping a 409) if the user exists either in
+    * Rawls or LDAP when registering. That is not a real error so we can consider that a success.
+    *
+    * @param postResponse The HttpResponse
+    * @return A conflict or not
+    */
+  def isConflict(postResponse: HttpResponse): Boolean = {
```
davidangb: Is there a corner case where the user exists in LDAP but not in rawls, where the isConflict will return false because the response entity doesn't contain "409 Conflict"? Maybe have to look for either "409 Conflict" or "LDAP: error code 68"?

I wish there were a more precise way to check for this case, but given the response from rawls, I'm not seeing it.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -0,0 +1,130 @@
...
+              Some(FireCloudKeyValue(Some(key), Some(value)))
+          ))
+        }
+    }
+    Future sequence propertyPosts.toList
+  }
+
+  def checkUserInRawls(requestContext: RequestContext): Future[PerRequestMessage] = {
+    val pipeline = authHeaders(requestContext) ~> sendReceive
+    pipeline { Get(UserService.rawlsGetUserURL) } flatMap { response =>
```
davidangb: you're using flatMap here, but map at line 107. Is there a reason (nested futures)? If not, I suggest standardize on map.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
rushtong: I had considered that, but I'm not clear on if that approach is thread-safe enough. I'll give it a shot and see how the tests run.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
rushtong: If that does happen, then I think it would have to be dealt with on the Rawls side. I don't see a way to testing that case since we never actually talk to LDAP. I only know about the 409 issue from testing it with pre-existing users.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
rushtong: There is a reason. Map-ping over a future will give you another future back which works well if you have a single 'stack' of futures to deal with. If there are conditional or nested futures that need to be made, then you get this ugly `Future[Future[PerRequest]]` thing when working in a map. To avoid that, you have to flatMap instead. I think we need to have an in-person code-discussion for this scenario.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
davidangb: oooh if that's not thread-safe then I have a problem in AgoraPermissionActor if not elsewhere :(

davidangb:
Looks good to me, and works well for me too. Minor cleanup comments inline, but this is pretty much ready to go. Back to you, Greg, to see how much of the cleanup is feasible.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
rushtong: I refactored the other actor classes, but not the OAuth one yet. Have a look at those. In general, where we are using 
`future onComplete ...`, we should instead be constructing a Future[PerRequest] somewhere and then piping that to the context.parent. The problem we had was that the context was not fixed and in between actor calls, the context as seen by one future block could be a different context that initially created it. Not exactly a thread-safety issue, but a similar effect. 

Looking at pipeline, I think it's OK if we fix it as a val at the beginning of the case class and then pass it around. That way the requestContext used in the various futures is guaranteed to be the same one that initiated the actor call.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
davidangb: Nope, all good - I thought it might be nested futures, no need to discuss :)

PR #170, opened 2015-12-02
Minor refactoring

rushtong:
* Trimmed down the TSV Formatter so that it doesn't need to use a mutable seq any longer
* Consolidate the use of the Rawls `/register/user' path since it is the same for get and post.


davidangb:
looks good works good. Love getting rid of the mutables!

PR #171, opened 2015-12-03
tweak spray client timeout/connection params

davidangb:
See http://kamon.io/teamblog/2014/11/02/understanding-spray-client-timeout-settings/ and http://spray.io/documentation/1.2.3/spray-can/configuration/ for explanations. Goal of this is twofold:

1) increase timeouts for things like TSV uploads (DSDEEPB-1747). I don't think the timeouts will be completely addressed by this change, because we also need changes to Apache proxy timeouts.

2) increase performance under simultaneous load a bit, by allowing more simultaneous connections to rawls.

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -27,6 +27,46 @@ spray.can {
       max-content-length = 50m
     }
   }
+  spray.can {
+    client {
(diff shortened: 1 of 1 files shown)


rushtong:
:+1:

PR #172, opened 2015-12-07
Passthrough for getting the Rawls version of user status

rushtong:
This is in support of DSDEEPB-2185 and is a minor update to support the needed UI functionality.


davidangb:
:+1:

PR #174, opened 2015-12-11
use appropriate Vault value for google client id

davidangb:
fix a regression from d5945a9153f71d400f1540be3ecf5dbae59e5a76.


dmohs:
:+1:

PR #175, opened 2015-12-11
add Bearer token into swagger securityDefinitions

davidangb:
This fixes local swagger for me. When we (I) added securityDefinitions for oauth, it broke token-injection. This should fix that.


dmohs:
:+1: Fixes local swagger for me!

PR #176, opened 2015-12-17
add a stable 'me' endpoint for checking user status

davidangb:
In support of the UI's checking/handling of user auth/registration status, this adds an endpoint at /me (i.e. /service/me) that:
* always returns json
* always returns 200, 401, 403, 404, or 500

Hopefully this helps keep the UI implementation simpler.

This is a proof-of-concept PR and I'm expecting folks might have feedback!

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/UserService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/UserService.scala
@@ -8,11 +8,10 @@ import org.broadinstitute.dsde.firecloud.model._
 import org.broadinstitute.dsde.firecloud.utils.StandardUserInfoDirectives
 import org.slf4j.LoggerFactory
 import spray.client.pipelining._
-import spray.http.{StatusCode, HttpRequest, HttpCredentials}
+import spray.http.{StatusCode, HttpCredentials}
 import spray.http.HttpHeaders.Authorization
 import spray.http.StatusCodes._
 import spray.httpx.SprayJsonSupport._
-import spray.json.DefaultJsonProtocol._
 import spray.httpx.unmarshalling._
 import spray.routing._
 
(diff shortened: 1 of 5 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/UserService.scala
```diff
@@ -7,9 +7,17 @@ import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
 import org.broadinstitute.dsde.firecloud.model._
 import org.broadinstitute.dsde.firecloud.utils.StandardUserInfoDirectives
 import org.slf4j.LoggerFactory
+import spray.client.pipelining._
+import spray.http.{StatusCode, HttpRequest, HttpCredentials}
```
rushtong: `HttpRequest` is not used

src/main/scala/org/broadinstitute/dsde/firecloud/service/UserService.scala
```diff
@@ -7,9 +7,17 @@ import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
...
+import spray.http.HttpHeaders.Authorization
+import spray.http.StatusCodes._
 import spray.httpx.SprayJsonSupport._
+import spray.json.DefaultJsonProtocol._
```
rushtong: This import is also not needed.

rushtong:
It would be nice to have a test for the new route to keep coverage up.

rushtong:
:+1: Code looks good, works as expected - handing back to @davidangb for feedback.

rushtong:
FYI, there's a new story for this: https://broadinstitute.atlassian.net/browse/DSDEEPB-2462

davidangb:
okay, @rushtong back to you, I think I addressed all your feedback (and rebased to develop due to merge conflicts)

rushtong:
:+1: Merging.

PR #177, opened 2015-12-18
DSDEEPB-1926, DSDEEPB-2279: JWT from NIH

davidangb:
This adds an endpoint at /api/nih/callback that accepts a JWT, then:
* validates the JWT
* extracts the NIH username from the JWT
* records the current time and NIH username in the user's profile

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/NIHService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/NIHService.scala
@@ -1,0 +1,65 @@
+package org.broadinstitute.dsde.firecloud.service
+
+import akka.actor.Props
+import authentikat.jwt._
+import org.broadinstitute.dsde.firecloud.FireCloudConfig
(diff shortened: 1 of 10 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/NIHService.scala
```diff
@@ -0,0 +1,65 @@
...
+            // validate the token
+            val isValid = JsonWebToken.validate(jwt, FireCloudConfig.Shibboleth.signingKey)
+
+            if (!isValid) {
+              requestContext.complete(StatusCodes.BadRequest)
+            } else {
+              // the NIH JWT is nonstandard. The claims portion of the token *should* be json, but is in fact
+              // a simple string. So, libraries tend to fail when working with it. Extract it manually.
+              // we shouldn't have to check for bad/missing parts of the token, because we've already validated it.
+              val claim = jwt.split("\\.")(1)
```
dmohs: Shit. Sorry. I will fix this.

dmohs:
:+1:

PR #178, opened 2015-12-21
endpoint for the current user's NIH link status

davidangb:
adds an endpoint at /api/nih/status that returns information about the current user's link status.

Response looks like this:
{
  "descriptionSinceLastLink": "4 hours ago",
  "linkedNihUsername": "CBIRGER_SO",
  "secondsSinceLastLink": 15436,
  "lastLinkTime": 1450803258,
  "loginRequired": false,
  "linkExpireTime": 1453395258,
  "isDbgapAuthorized": false
}

Most importantly, has a "loginRequired" key so the UI can determine whether or not to ask the user to refresh their link.

Still left to do:
* call rawls to actually enable users for TCGA access; this is blocked on the API existing in rawls
* properly set the linkExpireTime during standard OAuth (non-NIH) login


davidangb:
Okay, I've added a bunch more commits; this is ready for review. Note the TODOs listed in the description.

rushtong:
This looks good as far as I can test it. Both unauthorized and Not Found work as expected. 
Code looks right - I can't see any missing pieces to the logic.
Needs a rebase, but should be good to go after that.

rushtong:
My mistake, this doesn't need re-basing.

PR #179, opened 2015-12-21
move shibboleth signing secret in vault

davidangb:
In Vault, I've added the signing key to the new common location for local and dev. I've also removed it from local, but NOT from dev (need to coordinate with the shib service before removing it there)


dmohs:
:+1:

PR #180, opened 2016-01-05
DSDEEPB-2403: FireCloud OAuth login updates NIH link expiration time

davidangb:
When logging in to FireCloud, if the current user:
* has a linked NIH account
* has not re-linked with NIH within the last 24 hours
* has a NIH link expiration of more than 24 hours
then set the user's NIH link expiration to be 24 hours.

The implementation of this PR is, at first glance (and maybe the fifth glance), unconventional. The OAuth login would need to query Thurloe for NIH link keys, then inspect those keys and act appropriately. We already had an endpoint that queried and inspected those keys ... so I added the check-for-expiration feature into that code path. I like this approach also because it makes it trivially easy to enable the check-for-expiration feature in the /api/nih/status endpoint; in turn this may make it easy to call the check more often.

ProfileClient.scala has a bunch of code that moved around, so the diff shows bigger than it really is. I'll try to comment inline with details.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
@@ -3,15 +3,13 @@ package org.broadinstitute.dsde.firecloud.core
 import akka.actor.{Actor, Props}
 import akka.event.Logging
 import akka.pattern.pipe
-import com.ocpsoft.pretty.time.PrettyTime
 
-import org.broadinstitute.dsde.firecloud.core.ProfileClient.{GetNIHStatus, UpdateNIHLink, UpdateProfile}
+import org.broadinstitute.dsde.firecloud.core.ProfileClient._
 import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
 import org.broadinstitute.dsde.firecloud.model._
 import org.broadinstitute.dsde.firecloud.service.{UserService, FireCloudRequestBuilding}
 import org.broadinstitute.dsde.firecloud.service.PerRequest.{PerRequestMessage, RequestComplete}
 import org.broadinstitute.dsde.firecloud.utils.DateUtils
-import org.joda.time.{Hours, DateTime}
 
 import spray.client.pipelining._
 import spray.http.{StatusCodes, HttpRequest, HttpResponse}
@@ -20,7 +18,6 @@ import spray.httpx.SprayJsonSupport._
 import spray.httpx.unmarshalling._
 import spray.routing.RequestContext
 
-
 import scala.concurrent.Future
 import scala.util.{Failure, Success}
 
@@ -28,6 +25,7 @@ object ProfileClient {
   case class UpdateProfile(userInfo: UserInfo, profile: BasicProfile)
   case class UpdateNIHLink(userInfo: UserInfo, nihLink: NIHLink)
   case class GetNIHStatus(userInfo: UserInfo)
+  case class GetAndUpdateNIHStatus(userInfo: UserInfo)
   def props(requestContext: RequestContext): Props = Props(new ProfileClientActor(requestContext))
 }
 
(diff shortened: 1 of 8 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -84,77 +82,101 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
...
+      response.status match {
+        case OK =>
+          val profileEither = response.entity.as[ProfileWrapper]
+          profileEither match {
+            case Right(profileWrapper) =>
+              val profile = Profile(profileWrapper)
+              profile.linkedNihUsername match {
+                case Some(nihUsername) =>
+                  // we have a linked profile.
+
```
davidangb: most of the getProfile function is a cut-and-paste from the previous GetNIHStatus case. However, note I've removed the check on profile.isDbgapAuthorized, which added very little. See DSDEEPB-2404 for more context.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -84,77 +82,101 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
...
+                  // if we have no record of the user logging in or the user's expire time, default to 0
+                  val lastLinkSeconds = profile.lastLinkTime.getOrElse(0L)
+                  var linkExpireSeconds = profile.linkExpireTime.getOrElse(0L)
```
davidangb: I've also removed the previous code's
```(profile.lastLinkTime, profile.linkExpireTime) match { ... }```
and replaced it with these, to simplify branching.

File: src/test/scala/org/broadinstitute/dsde/firecloud/service/NIHServiceSpec.scala [hunk 3]
```diff
@@ -149,25 +136,6 @@ class NIHServiceSpec extends ServiceSpec with NIHService {
...
-          linkExpireTime = Some(linkExpireTime),
-          isDbgapAuthorized = Some(true)
-        )
-        Get(targetUri) ~> dummyUserIdHeaders(uniqueId) ~> sealRoute(routes) ~> check {
-          status should equal(OK)
-          val nihStatus = responseAs[NIHStatus]
-          nihStatus.loginRequired shouldBe(true)
-        }
-      }
-    }
```
davidangb: both of these tests targeted functionality that has changed/removed.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -84,77 +82,101 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
...
+
+                  val howOld = DateUtils.hoursSince(lastLinkSeconds)
+                  val howSoonExpire = DateUtils.secondsSince(linkExpireSeconds)
+
+                  // if the user's link has expired, the user must re-link.
+                  // NB: we use a separate val here in case we need to change the logic. For instance, we could
+                  // change the logic later to be "link has expired OR user hasn't logged in within 24 hours"
+                  val loginRequired = (howSoonExpire >= 0)
```
davidangb: see the previous code's TODO: we no longer need to check last login time, and we can rely solely on expiration time.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -69,9 +69,10 @@ object HttpGoogleServicesDAO {
...
       case x => Some(x)
     }
-    val idToken = gcsTokenResponse.getIdToken match {
-      case null => None
-      case x => Some(x)
+    val List(idToken:Option[String], subjectId:Option[String]) = gcsTokenResponse.getIdToken match {
+      case null => List(None, None)
+      case x =>
+        List( Some(x), Some(gcsTokenResponse.parseIdToken().getPayload.getSubject) )
     }
```
rushtong: I'm pretty sure the `List`s here are extraneous - I haven't seen this exact form of multiple assignment. This could be 
```
    val (idToken:Option[String], subjectId:Option[String]) = gcsTokenResponse.getIdToken match {
      case null => (None, None)
      case x =>
        (Some(x), Some(gcsTokenResponse.parseIdToken().getPayload.getSubject))
    }
```

src/main/scala/org/broadinstitute/dsde/firecloud/utils/DateUtils.scala
```diff
@@ -26,6 +26,10 @@ object DateUtils {
     Hours.hoursBetween(dtFromSeconds(seconds), nowDateTime).getHours
   }
 
+  def hoursUntil(seconds: Long): Int = {
+    Hours.hoursBetween(nowDateTime, dtFromSeconds(seconds)).getHours
+  }
+
```
rushtong: I don't see any usages for this - I'm guessing this is for upcoming work?

rushtong:
:+1: Looks good and functions as expected. Merging.

PR #181, opened 2016-01-05
fix misconfiguration in application.conf timeouts

davidangb:
this does not fully fix the issue for the end user.

There was a simple misconfiguration in our application.conf - previously, I was trying to set spray.can.spray.can.client timeouts, not spray.can.client.

After this change is in place, the bug in question now gets 504 timeouts after 60 seconds. This is probably a timeout from the Apache proxy and needs to be fixed by devops - but this PR is a prerequisite.

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -27,45 +27,43 @@ spray.can {
       max-content-length = 50m
     }
   }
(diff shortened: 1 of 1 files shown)


rushtong:
:+1: merging

PR #184, opened 2016-01-08
Populate dbgap authorization from rawls instead of thurloe

rushtong:
Implementation of retrieving group information from Rawls.

Testing Notes:
* Use Advanced Rest Client to navigate to `https://local.broadinstitute.org/service/api/nih/status` with a valid auth token to check that you are getting a populated NIH user status. 
* Until there are actual groups set up through Rawls, the dbgap status will always be false.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
@@ -3,6 +3,7 @@ package org.broadinstitute.dsde.firecloud.core
 import akka.actor.{Actor, Props}
 import akka.event.Logging
 import akka.pattern.pipe
+import org.broadinstitute.dsde.firecloud.FireCloudConfig
 
 import org.broadinstitute.dsde.firecloud.core.ProfileClient._
 import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
@@ -12,7 +13,7 @@ import org.broadinstitute.dsde.firecloud.service.PerRequest.{PerRequestMessage, 
 import org.broadinstitute.dsde.firecloud.utils.DateUtils
 
 import spray.client.pipelining._
-import spray.http.{StatusCodes, HttpRequest, HttpResponse}
+import spray.http.{StatusCode, StatusCodes, HttpRequest, HttpResponse}
 import spray.http.StatusCodes._
 import spray.httpx.SprayJsonSupport._
 import spray.httpx.unmarshalling._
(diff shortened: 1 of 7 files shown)


davidangb:
@jmthibault79 @rushtong please coordinate between this PR and #185 ? Both PRs add the dbGapAuthorizedUsers group name to orchestration, but one does it via Vault/env vars, and the other does it via application.conf. (I think it will have to happen via Vault/env vars, so the group name can be different in dev/staging/alpha/prod)

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -176,6 +168,32 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
...
+        case x if x == OK => true
+        case _ => false
+      }
+      RequestComplete(OK,
+        NIHStatus(
+          loginRequired,
+          linkedNihUsername = profile.linkedNihUsername,
+          isDbgapAuthorized = Some(authorized),
+          lastLinkTime = Some(profile.lastLinkTime.getOrElse(0L)),
+          linkExpireTime = Some(profile.linkExpireTime.getOrElse(0L)),
```
davidangb: linkExpireTime is wrong here. Note that above, linkExpireSeconds is a var. If the previous code determines that the user's link expire time should be reset, it persists the new expire time to Thurloe and updates the var. Here, you should be using the updated var, not the original value from profile.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -176,6 +168,32 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
...
+          descriptionSinceLastLink = Some(DateUtils.prettySince(profile.lastLinkTime.getOrElse(0L))),
+          descriptionUntilExpires = Some(DateUtils.prettySince(profile.linkExpireTime.getOrElse(0L)))
```
davidangb: if you're calculating the pretty time here, it can be removed from line ~148

jmthibault79:
Regarding the dbGapAuthorizedUsers group name in Rawls, we discussed in Blue Team HipChat whether it was necessary to have unique group names per environment.  We do not (can discuss in more detail if you like) so I decided that an application.conf setting would be enough.

jmthibault79:
I'm not seeing any other obvious conflicts with #185 - thanks for the heads up!

davidangb:
Okay, this is working pretty well for me, close to ready, with a few changes:
1) See inline about coordinating where the dbGapAuthorizedUsers group name lives. If it lives in Vault, you can/should update at least local and dev yourself.
2) See inline for discussion of link-expire-time as a var.
3) with isDbgapAuthorized populated from rawls/Google groups, I'd like to remove isDbgapAuthorized from Thurloe altogether to avoid any confusion or mis-coding. I'm fine if this happens as a followup story.

thanks!

rushtong:
Will address feedback and use the Blue team approach for the group configuration.

rushtong:
I created https://broadinstitute.atlassian.net/browse/DSDEEPB-2512 as a follow-on to this one for cleaning up the "isDbgapAuthorized" key.

rushtong:
@davidangb - PR is updated and ready for re-review.

davidangb:
I've updated DSDEEPB-2512 to reflect my concerns. Thanks for the code updates; looks good. Merging!

PR #185, opened 2016-01-08
nih/sync endpoint to sync the NIH TCGA/DBGAP whitelist

jmthibault79:
~~Please comment but note that it is completely untested.~~

~~The main commit is "nih/sync endpoint to sync the NIH TCGA/DBGAP whitelist" and the others all involve setting config files.~~

DSDEEPB-2384

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -53,6 +53,7 @@ libraryDependencies ++= {
     "com.typesafe.akka"   %%  "akka-testkit"  % akkaV     % "test",
     "com.typesafe.akka"   %%  "akka-slf4j"    % akkaV,
     ("com.google.api-client" % "google-api-client" % "1.20.0").exclude("com.google.guava", "guava-jdk5"),
+    "com.google.apis" % "google-api-services-storage" % "v1-rev30-1.20.0",
     "joda-time"            % "joda-time"      % "2.9.1",
     "org.joda"             % "joda-convert"   % "1.8.1",
     "com.jason-goodwin"   %% "authentikat-jwt" % "0.4.1",
--- a/src/main/config/firecloud-orchestration-compose.yaml.ctmpl
+++ b/src/main/config/firecloud-orchestration-compose.yaml.ctmpl
@@ -15,6 +15,13 @@ app:
     - FIRECLOUD_URL_ROOT={{.Data.env_firecloud_url_root}}
     - GOOGLE_SECRET_JSON={{.Data.env_google_secret_json}}
     - ADMIN_REFRESH_TOKEN={{.Data.env_admin_refresh_token}}
+    - REFRESH_TOKEN_SECRET_JSON={{.Data.env_refresh_token_secret_json}}
+    - ORCHESTRATION_PEM=/etc/firecloud-account.pem
+    {{$pem_file := printf "secret/dsde/%s/common/firecloud-account.pem" $environment}}
+    {{with vault $pem_file}}
+    - ORCHESTRATION_PEM_CLIENT_ID={{.Data.client_email}}
+    {{end}}
+    - NIH_WHITELIST_BUCKET={{.Data.env_nih_whitelist_bucket}}
     {{$signingkey := printf "secret/dsde/%s/common/signing-secret" $environment}}
     {{with vault $signingkey}}
     - JWT_SIGNING_KEY={{.Data.value}}
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -91,13 +91,19 @@ workspace {
   workspacesEntitiesCopyPath="/workspaces/entities/copy"
   submissionsPath="/workspaces/%s/%s/submissions"
   submissionsIdPath="/workspaces/%s/%s/submissions/%s"
-  dbGapAuthorizedUsersGroup="dbGapAuthorizedUsers"
+  overwriteGroupMembershipPath="/admin/groups/%s/members"
 }
 
 userprofile {
   authPrefix = "/api"
   setKey="/thurloe"
-  getAll="/thurloe/%s"
   get="/thurloe/%s/%s"
+  getAll="/thurloe/%s"
+  getQuery="/thurloe?"
   delete="/thurloe/%s/%s"
+}
+
+nih {
+  whitelistFile = "users.txt"
+  rawlsGroupName = "dbGapAuthorizedUsers"
 }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
@@ -1,11 +1,13 @@
 package org.broadinstitute.dsde.firecloud.core
 
-import akka.actor.{Actor, Props}
+import akka.actor.{ActorRefFactory, Actor, Props}
 import akka.event.Logging
 import akka.pattern.pipe
 import org.broadinstitute.dsde.firecloud.FireCloudConfig
 
 import org.broadinstitute.dsde.firecloud.core.ProfileClient._
+import org.broadinstitute.dsde.firecloud.dataaccess.HttpGoogleServicesDAO
+import spray.json.DefaultJsonProtocol._
 import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
 import org.broadinstitute.dsde.firecloud.model._
 import org.broadinstitute.dsde.firecloud.service.{UserService, FireCloudRequestBuilding}
@@ -13,20 +15,23 @@ import org.broadinstitute.dsde.firecloud.service.PerRequest.{PerRequestMessage, 
 import org.broadinstitute.dsde.firecloud.utils.DateUtils
 
 import spray.client.pipelining._
-import spray.http.{StatusCode, StatusCodes, HttpRequest, HttpResponse}
+import spray.http.{HttpRequest, HttpResponse}
 import spray.http.StatusCodes._
 import spray.httpx.SprayJsonSupport._
 import spray.httpx.unmarshalling._
 import spray.routing.RequestContext
 
-import scala.concurrent.Future
-import scala.util.{Failure, Success}
+import scala.concurrent.{ExecutionContext, Future}
+import scala.io.Source
+import scala.util.{Try, Failure, Success}
 
 object ProfileClient {
   case class UpdateProfile(userInfo: UserInfo, profile: BasicProfile)
   case class UpdateNIHLink(userInfo: UserInfo, nihLink: NIHLink)
   case class GetNIHStatus(userInfo: UserInfo)
   case class GetAndUpdateNIHStatus(userInfo: UserInfo)
+  case object SyncWhitelist
+
   def props(requestContext: RequestContext): Props = Props(new ProfileClientActor(requestContext))
 }
 
@@ -89,7 +94,10 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
     case GetAndUpdateNIHStatus(userInfo: UserInfo) =>
       getProfile(userInfo, true)
       // do NOT pipe the response back to the parent!
-  }
+
+    case SyncWhitelist =>
+      context.parent ! syncWhitelist
+   }
 
   def getProfile(userInfo: UserInfo, updateExpiration: Boolean): Future[PerRequestMessage] = {
     val pipeline = addCredentials(userInfo.accessToken) ~> sendReceive
@@ -162,7 +170,6 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
       // unexpected error retrieving profile
       case e: Throwable => RequestCompleteWithErrorReport(InternalServerError, e.getMessage)
     }
-
   }
 
   def getNIHStatusResponse(pipeline: WithTransformerConcatenation[HttpRequest, Future[HttpResponse]],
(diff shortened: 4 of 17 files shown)


build.sbt
```diff
@@ -53,6 +53,7 @@ libraryDependencies ++= {
     "com.typesafe.akka"   %%  "akka-testkit"  % akkaV     % "test",
     "com.typesafe.akka"   %%  "akka-slf4j"    % akkaV,
     ("com.google.api-client" % "google-api-client" % "1.20.0").exclude("com.google.guava", "guava-jdk5"),
+    ("com.google.apis" % "google-api-services-storage" % "v1-rev30-1.20.0").exclude("com.google.guava", "guava-jdk5"),
```
abaumann: new thing that popped up for you?  agora had a problem with guava where they had to exclude because the builds were finding duplicate dependency files

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -88,7 +94,23 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
     case GetAndUpdateNIHStatus(userInfo: UserInfo) =>
       getProfile(userInfo, true)
       // do NOT pipe the response back to the parent!
-  }
+
+    case SyncWhitelist =>
+      val (bucket, file) = (FireCloudConfig.Nih.whitelistBucket, FireCloudConfig.Nih.whitelistFile)
+      val whitelist = Source.fromInputStream(HttpGoogleServicesDAO.getBucketObjectAsInputStream(bucket, file)).getLines().toSet
```
abaumann: what about the case when the whitelist does not exist - what happens here?

build.sbt
jmthibault79: No, this is copy/paste from Rawls.  I have not tested sbt assembly yet - I'll do that now.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
jmthibault79: Blows up, probably.  I'll work on a fix.

What have decided as the correct behavior: ignore, ignore X times and then remove access to all, or remove access to all right away?

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -258,4 +279,41 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
...
   }
 
+  // get a mapping of FireCloud user name to NIH User name, for only those Thurloe users with a non-expired NIH link
+  def getCurrentNihUsernameMap: Future[Map[String, String]] = {
+
+    val pipeline = addCredentials(OAuth2BearerToken(HttpGoogleServicesDAO.getAdminUserAccessToken)) ~> sendReceive
+    def getAllUserValuesForKey(key: String): Future[Map[String, String]] = pipeline {
+      Get(s"${UserService.remoteGetQueryURL}key=${key}")
+    } map { response =>
+      if (response.status != OK) throw new Exception(response.toString)   // TODO: better error handling
```
jmthibault79: Is there something like a standard OrchestrationException that can be thrown here?

Alternatively, something involving ErrorReport and PerRequest.

build.sbt
jmthibault79: Seems to work fine either way.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -258,4 +266,57 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
...
+
+    val usernames = getAllUserValuesForKey("linkedNihUsername")
+    val expiretimes = getAllUserValuesForKey("linkExpireTime")
+
+    // compose futures
+    for {
+      usernameKvp <- usernames
+      expirationKvp <- expiretimes
+    } yield {
+      // exclude from results: link expiration is not in Thurloe, value is not a long, value is in the past
```
dvoet: I am not entirely certain that the below works. At the very least all the flatMaps make it hard to follow. What I would probably do is first find all the non-expired fc users:

val validFcUsers = expirationKvp.map{ case (fcUser, expStr) => fcUser -> Try { expStr.toLong }.toOption }.collect {
  case (fcUser, Some(exp: Long)) if DateUtils.now > exp => fcUser
}

Then filter usernameKvp with the results from above:
usernameKvp.filter { case (fcUser, nihUser) => validFcUsers.contains(fcUser) } values

(maybe that all compiles)

davidangb:
fyi on the Travis errors - we have intermittent Travis failues. Just re-run that test once or a few times until it passes and your PR is green (*sigh*)

src/main/config/firecloud-orchestration-compose.yaml.ctmpl
```diff
@@ -15,6 +15,12 @@ app:
     - FIRECLOUD_URL_ROOT={{.Data.env_firecloud_url_root}}
     - GOOGLE_SECRET_JSON={{.Data.env_google_secret_json}}
     - ADMIN_REFRESH_TOKEN={{.Data.env_admin_refresh_token}}
+    - ORCHESTRATION_PEM=/etc/firecloud-account.pem
+    {{$pem_file := printf "secret/dsde/%s/common/firecloud-account.pem" $environment}}
+    {{with vault $pem_file}}
+    - ORCHESTRATION_PEM_CLIENT_ID={{.Data.client_id}}
+    {{end}}
+    - NIH_WHITELIST_BUCKET={{.Data.env_nih_whitelist_bucket}}
```
davidangb: is this populated in Vault yet?

src/main/resources/application.conf
```diff
@@ -86,12 +86,19 @@ workspace {
...
-  getAll="/thurloe/%s"
   get="/thurloe/%s/%s"
+  getAll="/thurloe/%s"
+  getQuery="/thurloe?"
   delete="/thurloe/%s/%s"
 }
+
+nih {
+  whitelistFile = "users.txt"
+  rawlsGroupName = "dbGapAuthorizedUsers"
```
davidangb: this conflicts with #184 but, with the addition of the filename, I like this PR's version better

rushtong: I like this version better too. Is it worth it to address in this PR or should we add another story for refactoring the other version?

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
davidangb: nope, nothing standard - use an ErrorReport if you can, or create a new Exception class.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
jmthibault79: updated to do this

src/main/config/firecloud-orchestration-compose.yaml.ctmpl
jmthibault79: No.

jmthibault79: The PEM stuff *is* in Vault.  AFAIK we have not decided on a proper bucket name yet.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -2,9 +2,12 @@ package org.broadinstitute.dsde.firecloud.dataaccess
 
 import java.io.StringReader
 
-import com.google.api.client.googleapis.auth.oauth2.{GoogleAuthorizationCodeFlow, GoogleClientSecrets, GoogleTokenResponse}
```
davidangb: uh oh now you've seen how much I bastardized rawls' version of HttpGoogleServicesDAO!

davidangb:
code structure looks good overall. Let me know when it's ready for testing. Also, looks like this needs a rebase against develop.

davidangb:
Can you factor out any of the methods so they are testable, either on their own or with mock services for Thurloe, etc? Specifically I'm thinking about the code that reads the whitelist, queries Thurloe, and generates the list of valid users.

jmthibault79:
rebased.  will refactor after my manual testing works.

jmthibault79:
@abaumann brought up a good point: this needs to move outside `/api` and/or somewhere that doesn't require calling it "as a user".  I am not familiar with how to do this.

src/main/config/firecloud-orchestration-compose.yaml.ctmpl
```diff
@@ -15,6 +15,13 @@ app:
     - FIRECLOUD_URL_ROOT={{.Data.env_firecloud_url_root}}
     - GOOGLE_SECRET_JSON={{.Data.env_google_secret_json}}
     - ADMIN_REFRESH_TOKEN={{.Data.env_admin_refresh_token}}
+    - REFRESH_TOKEN_SECRET_JSON={{.Data.env_refresh_token_secret_json}}
```
davidangb: in Vault's dev env, env_refresh_token_secret_json and env_google_secret_json are equivalent. If they are intended to be the same, let's just use one? If they are intended to be different, Vault dev needs updating.

jmthibault79: It's not clear to me how all the different credentials interact, so I would like to keep them separate for today, and revisit post-release.

jmthibault79:
Please re-review.

Some basic unit tests, and verified to work manually.

src/test/scala/org/broadinstitute/dsde/firecloud/service/NIHWhitelistSpec.scala
```diff
@@ -0,0 +1,47 @@
...
+  val currentUsernames1 = Map("fcSubjectId2" -> "nihUsername2")
+
+  val expiretimes2 = Future { Map("fcSubjectId1" -> DateUtils.nowMinus24Hours.toString) }
+  val expiretimes3 = Future { Map("fcSubjectId1" -> DateUtils.nowMinus24Hours.toString, "fcSubjectId2" -> "not a number") }
+
+  "NihWhitelistSpec" - {
+    "when handling expired and unexpired users" - {
+      "should only include unexpired users" in {
+        assertResult(currentUsernames1) {
+          Await.result(NIHWhitelistUtils.filterForCurrentUsers(usernames, expiretimes1), 1 second)
```
davidangb: These give compile warnings when running tests, pretty simple fix:

```
[warn] /Users/davidan/work/src/firecloud-orchestration/src/test/scala/org/broadinstitute/dsde/firecloud/service/NIHWhitelistSpec.scala:24: postfix operator second should be enabled
[warn] by making the implicit value scala.language.postfixOps visible.
[warn] This can be achieved by adding the import clause 'import scala.language.postfixOps'
[warn] or by setting the compiler option -language:postfixOps.
[warn] See the Scala docs for value scala.language.postfixOps for a discussion
[warn] why the feature should be explicitly enabled.
[warn]           Await.result(NIHWhitelistUtils.filterForCurrentUsers(usernames, expiretimes1), 1 second)
[warn]                                                                                            ^
[warn] /Users/davidan/work/src/firecloud-orchestration/src/test/scala/org/broadinstitute/dsde/firecloud/service/NIHWhitelistSpec.scala:32: postfix operator second should be enabled
[warn] by making the implicit value scala.language.postfixOps visible.
[warn]           Await.result(NIHWhitelistUtils.filterForCurrentUsers(usernames, expiretimes2), 1 second)
[warn]                                                                                            ^
[warn] /Users/davidan/work/src/firecloud-orchestration/src/test/scala/org/broadinstitute/dsde/firecloud/service/NIHWhitelistSpec.scala:40: postfix operator second should be enabled
[warn] by making the implicit value scala.language.postfixOps visible.
[warn]           Await.result(NIHWhitelistUtils.filterForCurrentUsers(usernames, expiretimes3), 1 second)
[warn]                                                                                            ^
[warn] three warnings found
```

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -273,4 +283,64 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
       postResponse.entity.asString.contains(Conflict.reason)
   }
 
+  def syncWhitelist: Unit = {
+    val (bucket, file) = (FireCloudConfig.Nih.whitelistBucket, FireCloudConfig.Nih.whitelistFile)
+    val whitelist = Source.fromInputStream(HttpGoogleServicesDAO.getBucketObjectAsInputStream(bucket, file)).getLines().toSet
```
davidangb: here, or maybe in HttpGoogleServicesDAO, trap and catch any exceptions in reading the whitelist from the bucket? My concern here is that if we let the raw exception bubble up, we might be exposing sensitive information (bucket name, or even worse some of the secret credentials) in a stack trace visible by the end user.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -273,4 +283,64 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
...
+    val url = FireCloudConfig.Rawls.overwriteGroupMembershipUrlFromGroupName(FireCloudConfig.Nih.rawlsGroupName)
+
+    memberList flatMap { members => pipeline { Put(url, members) } }
+  }
+
 }
+
+object NIHWhitelistUtils {
+  def getAllUserValuesForKey(key: String)(implicit arf: ActorRefFactory, ec: ExecutionContext): Future[Map[String, String]] = {
+    val pipeline = addCredentials(OAuth2BearerToken(HttpGoogleServicesDAO.getAdminUserAccessToken)) ~> sendReceive
```
davidangb: future refactoring ... this code:
```
addCredentials(OAuth2BearerToken(HttpGoogleServicesDAO.getAdminUserAccessToken))
```
could/should be encapsulated into a custom RequestTransformer in FireCloudRequestBuilding, something like "addAdminCredentials"

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -100,4 +110,29 @@ object HttpGoogleServicesDAO {
 
   def randomString(length: Int) = scala.util.Random.alphanumeric.take(length).mkString
 
+  def getAdminUserAccessToken = {
+    val googleCredential = new GoogleCredential.Builder().setTransport(httpTransport)
+      .setJsonFactory(jsonFactory)
+      .setClientSecrets(refreshTokenClientSecrets)
+      .build().setFromTokenResponse(new TokenResponse().setRefreshToken(adminUserRefreshToken))
+
+    googleCredential.refreshToken()
```
davidangb: is this line necessary?

src/test/scala/org/broadinstitute/dsde/firecloud/service/NIHWhitelistSpec.scala
```diff
@@ -0,0 +1,47 @@
...
+        }
+      }
+    }
+
+    "when handling users with no expiration times" - {
```
davidangb: nitpick - this tests users who are expired, not users with no expiration times

jmthibault79: The purpose of this test is to show what happens when a user is in `usernames` but not in `expiretimes`.  It also shows that an expired user is not returned, but this is true for all of the tests.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
jmthibault79: Yes.  This is the line that actually generates the new access token.

jmthibault79:
Addressed review comments

davidangb:
okay, @jmthibault79 I'm giving this a big ol' :+1: . Please squash and rebase if needed. Feel free to merge on your own.

PR #186, opened 2016-01-11
DSDEEPB-2248: use version number for Thurloe profile instead of true/false

davidangb:
now, when we save a profile to Thurloe, we store a version number instead of a true/false indicator. This allows us to change the profile fields over time.


rushtong:
:+1: Looks good, merging.

PR #187, opened 2016-01-11
DSDEEPB-2266: Passthrough for /inputsOutputs endpoint

MattPutnam:


src/main/resources/swagger/api-docs.yaml
```diff
@@ -136,6 +136,38 @@ paths:
...
+      produces:
+        - application/json
+      parameters:
+        - name: methodName
+          description: name of Method to look up
+          required: true
+          type: string
+          in: body
+          schema:
+            $ref: MethodID
```
davidangb: not your doing originally, but go down to the definition for MethodId around line 2142, and add default: 0 to the methodVersion. Without the default value, swagger won't show this key when pre-filling the payload. (with methodVersion missing from the payload, swagger returns no response; that was my problem)

davidangb:
Looks good, with one minor swagger change please. Feel free to make the change and merge yourself, or make the change and pass back to me if you'd rather.

MattPutnam:
@davidangb Can you do a quick check on the swagger?  I'm having trouble getting it to generate.

davidangb:
:+1:

PR #188, opened 2016-01-12
handle larger responses from rawls

dvoet:


davidangb:
What happens in the UI when rawls is returning 50m of data? I suspect the UI will get swamped ... which is better, swamping or throwing an error?

dvoet:
I don't actually consider this issue fixed. I left DSDEEPB-2488 open. But if we allow 50m uploads we have to be able to render a similar size. Even 50m is not enough. We really need a different architecture (server side pagination?) to handle arbitrarily large workspace sizes.

This interim fix allows users to deal with larger workspaces sizes.

PR #192, opened 2016-01-15
add a catchall case to mappedPropVals.propertyValueMap

davidangb:


src/main/scala/org/broadinstitute/dsde/firecloud/model/Profile.scala
```diff
@@ -132,6 +132,7 @@ trait mappedPropVals {
         f.get(this) match {
           case x: String => f.getName -> x
           case y: Option[_] => f.getName -> y.asInstanceOf[Option[_]].getOrElse("").toString
+          case z => f.getName -> z.toString
```
MattPutnam: Maybe the explicitness is nice, but if you put this at the top you would eliminate the need for the String case.

rushtong: That would catch Option too which is bad. You'd get "Some(whatever)" if you did.

davidangb: it has to go below the Option[_], but you're right it could go above String and eliminate String. I'd like to refactor this later, though, so I'd rather leave it ugly for now.

rushtong:
:+1: Tested - this doesn't fail on the long values any more.

PR #195, opened 2016-02-02
Refactor workspace tests

rushtong:
Significant refactor to workspace service tests and entity client
- Updated tests to reflect passthrough testing approach
- Refactored EntityClient to use the `Future[PerRequestMessage]` approach
- Removed a lot of unused code from EntityClient that is implemented as passthrough requests elsewhere
- TSV tests were left as-is, for the most part, to keep the original intent intact


src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceServiceSpec.scala
```diff
@@ -1,250 +1,339 @@
...
+  private final val workspacesPath = workspacesRoot + "/%s/%s".format(workspace.namespace.get, workspace.name.get)
+  private final val methodconfigsPath = workspacesRoot + "/%s/%s/methodconfigs".format(workspace.namespace.get, workspace.name.get)
+  private final val updateAttributesPath = workspacesRoot + "/%s/%s/updateAttributes".format(workspace.namespace.get, workspace.name.get)
+  private final val aclPath = workspacesRoot + "/%s/%s/acl".format(workspace.namespace.get, workspace.name.get)
+  private final val clonePath = workspacesRoot + "/%s/%s/clone".format(workspace.namespace.get, workspace.name.get)
+  private final val lockPath = workspacesRoot + "/%s/%s/lock".format(workspace.namespace.get, workspace.name.get)
+  private final val unlockPath = workspacesRoot + "/%s/%s/unlock".format(workspace.namespace.get, workspace.name.get)
+
+  private final val workspaceBasePath = FireCloudConfig.Rawls.authPrefix + FireCloudConfig.Rawls.workspacesPath
+  private final val tsvImportPath = "/" + workspacesRoot + "/%s/%s/importEntities".format(workspace.namespace.get, workspace.name.get)
```
davidangb, line 38: why is this one different (starts with a slash, other paths do not)?

davidangb:
One very minor question inline - otherwise, rebase/squash and this is ready to go. Good stuff!

src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceServiceSpec.scala
rushtong, line 38: Short answer is, without the slash, the call fails. I didn't dig into why, but I can do that.

rushtong, line 38: Ah - I do know the difference. 
The urls without "/" are all mock urls to Rawls. 
All of the urls with the "/" (or just the tsvImportPath path) is a local url and is what's used for calls to Orchestration. 
From a code organization point of view, this could be better laid out.

rushtong:
@davidangb - PR is rebased and squashed.

PR #199, opened 2016-02-08
Turn on stackdriver status in proxy

coreone:


rushtong:
@coreone - How would one go about testing this locally?

coreone:
@rushtong I'm not sure you need to test anything with this, as this settings just changes how the proxy operates.  As long as your application still works with the current openidc-proxy proxy in front of it, then it should be good.  Whether the stackdriver virtual host we have added works is tested through Stackdriver (i.e. if statistics show up, it works.  If they don't, we will have to investigate why).

davidangb:
this needs to move to the new configuration-management method, i.e. it needs to move to firecloud-develop.

coreone:
@davidangb Just made the following PR.  If it is accepted and we start deploying using the new method, this PR can just be closed and the branch deleted.

https://github.com/broadinstitute/firecloud-develop/pull/5

davidangb:
closing, per verbal discussion with Andrew, in favor of broadinstitute/firecloud-develop#5

PR #201, opened 2016-02-16
Implement membership file download

rushtong:
Update tsv formatter to handle membership files.
Update export actor to zip up membership and entity file for download.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/core/ExportEntitiesByType.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/core/ExportEntitiesByType.scala
@@ -1,5 +1,8 @@
 package org.broadinstitute.dsde.firecloud.core
 
+import java.io.ByteArrayOutputStream
+import java.util.zip.{ZipEntry, ZipOutputStream}
+
 import akka.actor.{Actor, Props}
 import akka.event.Logging
 import akka.pattern.pipe
@@ -7,7 +10,7 @@ import akka.pattern.pipe
 import org.broadinstitute.dsde.firecloud.core.ExportEntitiesByType.ProcessEntities
 import org.broadinstitute.dsde.firecloud.core.GetEntitiesWithType.EntityWithType
 import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
-import org.broadinstitute.dsde.firecloud.model.RequestCompleteWithErrorReport
+import org.broadinstitute.dsde.firecloud.model.{RequestCompleteWithErrorReport, ErrorReport}
 import org.broadinstitute.dsde.firecloud.service.FireCloudRequestBuilding
 import org.broadinstitute.dsde.firecloud.service.PerRequest.{RequestComplete, RequestCompleteWithHeaders}
 import org.broadinstitute.dsde.firecloud.utils.TSVFormatter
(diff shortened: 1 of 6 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/core/ExportEntitiesByType.scala
```diff
@@ -34,27 +38,72 @@ class ExportEntitiesByTypeActor(requestContext: RequestContext) extends Actor wi
   val log = Logging(system, getClass)
 
   override def receive: Receive = {
-    case ProcessEntities(encodedUrl: String, filename: String, entityType: String) =>
+    case ProcessEntities(baseEntitydUrl: String, filename: String, entityType: String) =>
```
jmthibault79: typo in variable name

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
```diff
@@ -2,14 +2,34 @@ package org.broadinstitute.dsde.firecloud.utils
...
+import spray.json.JsValue
+
 object TSVFormatter {
 
-  def makeTsvString(entities: List[EntityWithType], entityType: String) = {
+  def makeMembershipTsvString(entities: List[EntityWithType], entityType: String): String = {
+    val headers: immutable.IndexedSeq[String] = immutable.IndexedSeq("membership:" + entityType + "_id", entityType.replace("_set", "_id"))
+    val rows: List[IndexedSeq[String]] = entities filter { _.entityType == entityType } flatMap { entity =>
+      entity.attributes.getOrElse(Map.empty) flatMap { m =>
+        m._2.convertTo[List[JsValue]] map { jsValue =>
```
jmthibault79, line 18: Could you explain the logic here?

jmthibault79:
Verified locally that it works.  A few comments.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
rushtong, line 18: The json attributes in a membership object (i.e. pair_set, sample_set, participant_set) represent rows as opposed to the root types, where the attributes represent columns. So, for each set attribute, we need to get all of the elements out of it and turn each one into a row. Here, `m._2` is the attribute list. We unmarshal that into a list and then work with each value. 

Here is a pair_set example:

```
[{
  "name": "pair_set_1",
  "entityType": "pair_set",
  "attributes": {
    "pairs": [{
      "entityType": "pair",
      "entityName": "1"
    }, {
      "entityType": "pair",
      "entityName": "2"
    }]
  }
}]
```

src/test/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatterSpec.scala
```diff
@@ -14,91 +15,104 @@ class TSVFormatterSpec extends FreeSpec with ScalaFutures with Matchers with Ins
...
-          "control_sample_id" -> """{"entityType": "sample", "entityName": "HCC1143_BL"}""".parseJson,
-          "participant_id" -> """{"entityType":"participant_set","entityName":"subject_HCC1143"}""".parseJson,
-          "gender" -> "male".toJson,
-          "header_1" -> MockUtils.randomAlpha().toJson,
-          "header_2" -> MockUtils.randomAlpha().toJson
-        )
+    "Participant tests should pass for" - {
+
+      "Entity Data" in {
+        val participantSetAtts = {
```
jmthibault79: This should represent attributes for a single participant, but is named like a participant set, and contains attributes appropriate for pairs.

rushtong: It's test data - the values don't mean anything. I'll update it though.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
jmthibault79, line 18: ok thanks

src/test/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatterSpec.scala
jmthibault79: Thanks, I appreciate it.  My branch makes these tests more strict, so this helps me.

jmthibault79:
I'm happy.  Rebase and squash and we're good to go.

rushtong:
Merging.

PR #202, opened 2016-02-17
GAWB-80: Try to insert timestamp into JSON error responses

MattPutnam:


src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
```diff
@@ -43,23 +43,39 @@ class FireCloudServiceActor extends HttpServiceActor with FireCloudDirectives {
...
+      try {
+        import spray.json._
+        import spray.json.DefaultJsonProtocol._
+        val dataMap = response.entity.asString.parseJson.convertTo[Map[String, JsValue]]
+        val withTimestamp = dataMap + ("timestamp" -> JsNumber(System.currentTimeMillis()))
+        response.withEntity(HttpEntity(withTimestamp.toJson.prettyPrint))
+      } catch {
+        // usually a failure to parse, if the response isn't JSON (e.g. HTML responses from Google)
+        case e: Exception => response
+      }
```
rushtong, line 57: Here's another (scala-y) way you can write this try/catch block:

```
      val dataMap = Try(response.entity.asString.parseJson.convertTo[Map[String, JsValue]])
      dataMap match {
        case scala.util.Success(data) =>
          val withTimestamp = data + ("timestamp" -> JsNumber(System.currentTimeMillis()))
          response.withEntity(HttpEntity(withTimestamp.toJson.prettyPrint))
        case scala.util.Failure(f) => response
      }
```

rushtong:
:+1:

PR #207, opened 2016-02-29
Update *-set tsv downloads

rushtong:
This PR fixes how sets are downloaded. Previous implementation was incorrect in that set downloads would create a membership file from the set data and an entity file from the other entity data, and not the set data. For example, it would grab the **sample_set** data to make a membership file and it would **also** grab the **sample** data to make an entity file. That is incorrect so this version creates an entity file and membership file from the same set entity type (was trying to pull two different entity types before). 

This also fixes a bug where empty values were being ignored. If two sample_sets in a workspace had a different group of top level attributes, each row would only contain the values in the sample_set so there would be a mis-match between the number of columns in the header and each row.

All of this is in the same area of the code so I updated both at the same time. 

To test, one can upload a large amount of data into a workspace, download and try to re-import into a new workspace. Data in https://broadinstitute.atlassian.net/browse/GAWB-286 will be helpful in doing so.

Addresses:
- https://broadinstitute.atlassian.net/browse/GAWB-1
- https://broadinstitute.atlassian.net/browse/GAWB-322
# 
- [x] **LR**: sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/core/ExportEntitiesByType.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/core/ExportEntitiesByType.scala
@@ -10,7 +10,7 @@ import akka.pattern.pipe
 import org.broadinstitute.dsde.firecloud.core.ExportEntitiesByType.ProcessEntities
 import org.broadinstitute.dsde.firecloud.core.GetEntitiesWithType.EntityWithType
 import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
-import org.broadinstitute.dsde.firecloud.model.{RequestCompleteWithErrorReport, ErrorReport}
+import org.broadinstitute.dsde.firecloud.model.{ModelSchema, RequestCompleteWithErrorReport, ErrorReport}
 import org.broadinstitute.dsde.firecloud.service.FireCloudRequestBuilding
 import org.broadinstitute.dsde.firecloud.service.PerRequest.{RequestComplete, RequestCompleteWithHeaders}
 import org.broadinstitute.dsde.firecloud.utils.TSVFormatter
@@ -24,7 +24,7 @@ import spray.json.DefaultJsonProtocol._
 import spray.routing.RequestContext
 
 import scala.concurrent.Future
-import scala.util.{Failure, Success}
+import scala.util.{Try, Failure, Success}
 
 object ExportEntitiesByType {
   case class ProcessEntities(baseEntityUrl: String, filename: String, entityType: String)
@@ -40,28 +40,18 @@ class ExportEntitiesByTypeActor(requestContext: RequestContext) extends Actor wi
   override def receive: Receive = {
     case ProcessEntities(baseEntityUrl: String, filename: String, entityType: String) =>
       val pipeline = authHeaders(requestContext) ~> sendReceive
-
-      // If we have a set type of entity, we need to make multiple calls for each data set.
-      val entityUrls = entityType match {
-        case x if x.endsWith("_set") => List(baseEntityUrl + "/" + entityType, baseEntityUrl + "/" + x.replace("_set", ""))
-        case _ => List(baseEntityUrl + "/" + entityType)
-      }
-
-      val urlsFuture: Future[List[HttpResponse]] = Future sequence { entityUrls map { url => pipeline { Get(url) } } }
-
-      urlsFuture map {
-        responses =>
-          val allSucceeded = responses.forall(_.status == OK)
-          allSucceeded match {
-            case true =>
-              val entities = responses.flatMap(unmarshal[List[EntityWithType]].apply)
-              entityType match {
-                case x if x.endsWith("_set") =>
-                  val rootType = x.replace("_set", "")
-                  val zippedFileName = x + ".zip"
-                  val rootData = TSVFormatter.makeEntityTsvString(entities, rootType)
-                  val membershipData = TSVFormatter.makeMembershipTsvString(entities, entityType)
-                  val zipBytes: Array[Byte] = getZipBytes(filename, membershipData, rootType, rootData)
+      pipeline { Get(baseEntityUrl + "/" + entityType) } map {
+        response =>
+          response.status match {
+            case OK =>
+              val entities = unmarshal[List[EntityWithType]].apply(response)
+              ModelSchema.getCollectionMemberType(entityType) match {
+                case Success(collectionType) if collectionType.isDefined =>
+                  val collectionMemberType = ModelSchema.getPlural(collectionType.get)
+                  val entityData = TSVFormatter.makeEntityTsvString(entities, entityType)
+                  val membershipData = TSVFormatter.makeMembershipTsvString(entities, entityType, collectionMemberType.get)
+                  val zipBytes: Array[Byte] = getZipBytes(entityType, membershipData, entityData)
+                  val zippedFileName = entityType + ".zip"
                   RequestCompleteWithHeaders(
                     HttpEntity(ContentTypes.`application/octet-stream`, zipBytes),
                     HttpHeaders.`Content-Disposition`.apply("attachment", Map("filename" -> zippedFileName)))
@@ -73,34 +63,21 @@ class ExportEntitiesByTypeActor(requestContext: RequestContext) extends Actor wi
                     HttpHeaders.`Content-Type`(`text/plain`))
               }
             case _ =>
-              val errors = responses.filterNot(_.status == OK) map { e => (e, ErrorReport.tryUnmarshal(e)) }
-              val errorReports = errors collect { case (_, Success(report)) => report }
-              val missingReports = errors collect { case (originalError, Failure(_)) => originalError }
-              val errorMessage = {
-                val baseMessage = "%d failures out of %d attempts retrieving entity urls.  Errors: %s"
-                  .format(errors.size, responses.size, errors mkString ",")
-                if (missingReports.isEmpty) baseMessage
-                else {
-                  val supplementalErrorMessage = "Additionally, %d of these failures did not provide error reports: %s"
-                    .format(missingReports.size, missingReports mkString ",")
-                  baseMessage + "\n" + supplementalErrorMessage
-                }
-              }
-              RequestCompleteWithErrorReport(InternalServerError, errorMessage, errorReports)
+              RequestCompleteWithErrorReport(response.status, response.entity.asString)
           }
       } recover { case e:Throwable =>  RequestCompleteWithErrorReport(InternalServerError, e.getMessage) } pipeTo context.parent
     case _ =>
       Future(RequestComplete(StatusCodes.BadRequest)) pipeTo context.parent
   }
 
-  private def getZipBytes(filename: String, membershipData: String, rootType: String, rootData: String): Array[Byte] = {
+  private def getZipBytes(entityType: String, membershipData: String, entityData: String): Array[Byte] = {
     val bos = new ByteArrayOutputStream()
     val zos = new ZipOutputStream(bos)
-    zos.putNextEntry(new ZipEntry(filename))
+    zos.putNextEntry(new ZipEntry(entityType + "_membership.txt"))
     zos.write(membershipData.getBytes)
     zos.closeEntry()
-    zos.putNextEntry(new ZipEntry(rootType + ".txt"))
-    zos.write(rootData.getBytes)
+    zos.putNextEntry(new ZipEntry(entityType + "_entity.txt"))
+    zos.write(entityData.getBytes)
     zos.closeEntry()
     zos.finish()
     bos.toByteArray
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
@@ -4,27 +4,36 @@ import org.broadinstitute.dsde.firecloud.core.GetEntitiesWithType.EntityWithType
 import org.broadinstitute.dsde.firecloud.model.ModelSchema
 
 import scala.collection.immutable
-import scala.util.Try
+import scala.util.{Success, Try}
 
 import spray.json.DefaultJsonProtocol._
 import spray.json.JsValue
 
 object TSVFormatter {
 
-  def makeMembershipTsvString(entities: List[EntityWithType], entityType: String): String = {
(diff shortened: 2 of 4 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/core/ExportEntitiesByType.scala
```diff
@@ -55,13 +55,13 @@ class ExportEntitiesByTypeActor(requestContext: RequestContext) extends Actor wi
...
               val entities = responses.flatMap(unmarshal[List[EntityWithType]].apply)
-              entityType match {
-                case x if x.endsWith("_set") =>
-                  val rootType = x.replace("_set", "")
-                  val zippedFileName = x + ".zip"
-                  val rootData = TSVFormatter.makeEntityTsvString(entities, rootType)
-                  val membershipData = TSVFormatter.makeMembershipTsvString(entities, entityType)
-                  val zipBytes: Array[Byte] = getZipBytes(filename, membershipData, rootType, rootData)
+              ModelSchema.getCollectionMemberType(entityType) match {
+                case Success(collectionType) if collectionType.isDefined =>
```
helgridly, line 49: Is it valid to fall back to `case _` if Failure() or `!collectionType.isDefined`?

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
```diff
@@ -4,26 +4,36 @@ import org.broadinstitute.dsde.firecloud.core.GetEntitiesWithType.EntityWithType
...
-    val rows: List[IndexedSeq[String]] = entities filter { _.entityType == entityType } flatMap { entity =>
-      entity.attributes.getOrElse(Map.empty) flatMap { m =>
-        m._2.convertTo[List[JsValue]] map { jsValue =>
-          val cellValue: String = jsValue match {
-            case x if Try(x.asJsObject.fields.contains("entityName")).isSuccess =>
-              cleanValue(x.asJsObject.fields.getOrElse("entityName", x))
-            case _ =>
-              ""
+    val rows: List[IndexedSeq[String]] = entities.
+      filter { _.entityType == entityType }.
```
helgridly: Nit: What's the style guide here?

```
entities.
filter {
```

or

```
entities
.filter {
```

?

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
```diff
@@ -72,10 +112,19 @@ object TSVFormatter {
         }
         columnPosition -> cellValue
     }
-    // The rowMap manipulation sorts the position-value map by the key, converts it to a seq of
-    // tuples, pulls out the second element of the tuple (column value) resulting in a sorted seq
-    // of the original map values.
-    entity.name +: rowMap.toSeq.sortBy(_._1).map(_._2).toIndexedSeq
+    // If there are entities that don't have a value for which there is a known header, that will
+    // be missing in the row. Fill up those positions with empty strings in that case.
+    val completedRowMap: IndexedSeq[(Int, String)] =
```
helgridly, line 114: Example?

src/test/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatterSpec.scala
```diff
@@ -15,6 +17,28 @@ class TSVFormatterSpec extends FreeSpec with ScalaFutures with Matchers with Ins
 
   "TSVFormatter" - {
 
+    "Sparse data fields should pass for" - {
```
helgridly, line 20: Is this the test that would have failed had it existed previously, and now does not fail?

src/main/scala/org/broadinstitute/dsde/firecloud/core/ExportEntitiesByType.scala
```diff
@@ -55,13 +55,13 @@ class ExportEntitiesByTypeActor(requestContext: RequestContext) extends Actor wi
...
             case true =>
               val entities = responses.flatMap(unmarshal[List[EntityWithType]].apply)
-              entityType match {
-                case x if x.endsWith("_set") =>
-                  val rootType = x.replace("_set", "")
-                  val zippedFileName = x + ".zip"
-                  val rootData = TSVFormatter.makeEntityTsvString(entities, rootType)
-                  val membershipData = TSVFormatter.makeMembershipTsvString(entities, entityType)
-                  val zipBytes: Array[Byte] = getZipBytes(filename, membershipData, rootType, rootData)
+              ModelSchema.getCollectionMemberType(entityType) match {
```
abaumann, line 48: this PR doesn't modify the code at line 46 above - so I commented here: is what line 46 is doing above still needed?  Isn't that from what it used to do before to get both the set and all of the entities of that set type?

src/main/scala/org/broadinstitute/dsde/firecloud/core/ExportEntitiesByType.scala
```diff
@@ -93,14 +93,14 @@ class ExportEntitiesByTypeActor(requestContext: RequestContext) extends Actor wi
...
+  private def getZipBytes(entityType: String, membershipData: String, entityData: String): Array[Byte] = {
     val bos = new ByteArrayOutputStream()
     val zos = new ZipOutputStream(bos)
-    zos.putNextEntry(new ZipEntry(filename))
+    zos.putNextEntry(new ZipEntry(entityType + "_membership.txt"))
     zos.write(membershipData.getBytes)
     zos.closeEntry()
-    zos.putNextEntry(new ZipEntry(rootType + ".txt"))
-    zos.write(rootData.getBytes)
+    zos.putNextEntry(new ZipEntry(entityType + "_entity.txt"))
```
abaumann, line 79: I'm not sure if the _entity is necessary on the file name, but it is confusing to need two files per set, so maybe it's better to add that to the end instead of just calling it "sample_set.txt"

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
abaumann: the latter, I'd hope

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
```diff
@@ -4,26 +4,36 @@ import org.broadinstitute.dsde.firecloud.core.GetEntitiesWithType.EntityWithType
...
+          // All other top-level properties are not necessary and are only used for the data load file.
+          filter { attribute => attribute._1.equals(collectionMemberType) }.
+          flatMap { m =>
+            Try(m._2.convertTo[List[JsValue]]) match {
+              case Success(values) =>
+                values map { jsValue =>
+                  val cellValue: String = jsValue match {
+                    case x if Try(x.asJsObject.fields.contains("entityName")).isSuccess =>
+                      cleanValue(x.asJsObject.fields.getOrElse("entityName", x))
+                    case _ =>
```
abaumann, line 29: What case is this here?  If it's a collection attribute called something like "samples", but then the referenced samples doesn't have entityName/entityType fields, we can't write an empty sample id, I think we should throw an exception here

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
jmthibault79: Pretty sure the first is obligatory, or the meaning changes.

src/test/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatterSpec.scala
```diff
@@ -144,7 +168,10 @@ class TSVFormatterSpec extends FreeSpec with ScalaFutures with Matchers with Ins
     val headers = lines.head.split("\t")
 
     // Make sure that all of the post-rename values exist in the list of headers from the TSV file
-    forAll (headerRenamingMap.values.toList) { x => headers should contain(x) }
+    // for all non-set types
```
abaumann, line 182: I'm wondering if we need to also test that each line has the same number of values as columns, given this fix deals with sparse data?

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
```diff
@@ -32,17 +42,31 @@ object TSVFormatter {
...
       flatMap(_.keySet).
       distinct.
       map { key => headerRenamingMap.getOrElse(key, key) }.
       toIndexedSeq
-    val rows: immutable.IndexedSeq[IndexedSeq[String]] = entities.filter { _.entityType == entityType }
+    val rows: immutable.IndexedSeq[IndexedSeq[String]] = filteredEntities.filter { _.entityType == entityType }
       .map { entity =>
         makeRow(entity, headers, headerRenamingMap)
       }.toIndexedSeq
+    println("Header Size: " + headers.size)
```
jmthibault79: stray printlns

src/main/scala/org/broadinstitute/dsde/firecloud/core/ExportEntitiesByType.scala
rushtong, line 48: I missed that - you're right, it is not necessary - I'll remove that. I can probably clean up the futures here too.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ExportEntitiesByType.scala
rushtong, line 49: I wavered on what to do here. If the model schema ever changes, there could be the case where a type may not be discoverable. I chose to handle it anyway, but can be persuaded to throw an exception in the undefined case.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ExportEntitiesByType.scala
rushtong, line 79: I was erring on the side of clarity since this is an entity file and not an update file.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
rushtong: I think Joel is right - I was getting weird compilation errors when trying to condense it in various ways. This was what worked.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
rushtong, line 29: There are two cases - a plain name-value pair or a json entity object (i.e. sample, participant, etc.) with an `entityName`. The _ case captures the plain name-value pair case.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
rushtong: Argh! Thanks Joel!

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
rushtong, line 114: There is an example in the test case. Are you saying you want to see that in a code comment?

src/test/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatterSpec.scala
rushtong, line 20: Yes - that is true.

src/test/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatterSpec.scala
rushtong, line 182: You're right, we do. I'll add that.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
abaumann, line 29: But it filters only on attributes that are collectionMemberType, so wouldn't this mean it's supposed to be a set attribute only (containing json entity objects as you say)?  Maybe I'm missing something but it seems like if we somehow hit this case, we would get "" for the id of a set member?

rushtong, line 29: Actually, I think I mis-spoke in my previous comment here. There are two cases when iterating over top-level attributes of an entity. In this case, this only should ever hit the `case x` case. For example, here is what we are looking at in this case:

```
"samples": [{
      "entityType": "sample",
      "entityName": "COAD-4N-A93T-NB"
    }, {
      "entityType": "sample",
      "entityName": "COAD-AA-3489-TP"
    }...
```

We should never really see the `case _` but I didn't want to throw away this field in the case the data was somehow not what is expected.

src/test/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatterSpec.scala
rushtong, line 182: Added. See lines 179-181 in this file.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
abaumann, line 29: What will happen here if we see this case (unexpectedly), is we will get an entity_id entry in that column that is "", and you won't be able to upload the TSV that was created.  I think here we should throw an exception and fail to make TSV rather than fail upon reimport

rushtong, line 29: Makes sense - will do.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
jmthibault79: An alternate style (common in Rawls) is to make sure the `{` is usually on the first line.

`entities.filter {`

davidangb:
signing off as tech lead for high-level implementation ... relying on others for lower-level syntax + nitpicks. :+1: !

rushtong:
@abaumann - PR feedback applied, back to you.

abaumann:
:+1:

PR #208, opened 2016-02-29
GAWB-102: use http gzip compression for entity lists

davidangb:
For entity lists - the `workspaces / Segment / Segment / entities_with_type` endpoint that powers the Data tab inside a workspace - enable http gzip encoding.

We don't want to enable gzip for all endpoints, because it adds processing overhead for little-to-no benefit on small responses. However, this endpoint tends to get pretty big, so gzip is worth it.

I have seen 18MB responses from this endpoint compress to under .5 MB.

A good follow-on PR would enable gzip encoding between rawls and orchestration. This only enables it between orchestration and the browser.
# 
- [x] **LR**: sign off


helgridly:
_hyperventilates with happiness_

davidangb:
this does not complete GAWB-102; it is just a super quick fix along the path of that story.

dmohs:
:+1: I am happy with this, but if it works I suggest we wrap nearly every response with this thing.

davidangb:
I disagree - we should not wrap nearly every response with this. For small payloads, the reduction in size is negligible (and for very small payloads, the gzip dictionary actually ADDS size), and we add compression/decompression overhead on both the server and the browser. But we can add it to more places.

dmohs:
@davidangb I retract my suggestion. I agree with you.

PR #209, opened 2016-03-01
GAWB-283: Orch work for clone-with-realm

MattPutnam:
- Refactored create/clone code
- Added response map to append "isProtected" field onto response for the single workspace GET.  Appending the same field on the "get all workspaces" will be done for GAWB-284
- [ ] **Tech lead**:  sign off
- [ ] **LR**: sign off

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -105,5 +105,6 @@ userprofile {
 
 nih {
   whitelistFile = "users.txt"
-  rawlsGroupName = "dbGapAuthorizedUsers"
+  rawlsGroupName = "testRealm"
+//  rawlsGroupName = "dbGapAuthorizedUsers"
 }
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -105,6 +105,5 @@ userprofile {
 
 nih {
   whitelistFile = "users.txt"
-  rawlsGroupName = "testRealm"
-//  rawlsGroupName = "dbGapAuthorizedUsers"
+  rawlsGroupName = "dbGapAuthorizedUsers"
 }
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -524,6 +524,13 @@ paths:
           required: true
           type: string
           in: path
+        - name: workspace
+          description: Cloned workspace information
+          required: true
+          type: string
+          in: body
+          schema:
+            $ref: WorkspaceIngest
       responses:
         201:
           description: Successful Request
@@ -1957,6 +1964,7 @@ definitions:
     required:
       - namespace
       - name
+      - attributes
     properties:
       namespace:
         type: string
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/model/Workspace.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/model/Workspace.scala
@@ -1,5 +1,7 @@
 package org.broadinstitute.dsde.firecloud.model
 
+import org.broadinstitute.dsde.firecloud.FireCloudConfig
+
 case class WorkspaceName(
   namespace: Option[String] = None,
   name: Option[String] = None)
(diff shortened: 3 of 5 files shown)


src/main/resources/application.conf
```diff
@@ -105,5 +105,6 @@ userprofile {
 
 nih {
   whitelistFile = "users.txt"
-  rawlsGroupName = "dbGapAuthorizedUsers"
+  rawlsGroupName = "testRealm"
+//  rawlsGroupName = "dbGapAuthorizedUsers"
```
jmthibault79: Let's fix this in Rawls instead of faking it

rushtong: Agreed.

MattPutnam: Do we want to do that before merging this PR?

src/main/resources/swagger/api-docs.yaml
```diff
@@ -524,6 +524,13 @@ paths:
           required: true
           type: string
           in: path
+        - name: workspace
+          description: Cloned workspace information
+          required: true
+          type: string
+          in: body
+          schema:
+            $ref: WorkspaceIngest
```
rushtong, line 533: I'm not clear on why we need to have a full WorkspaceIngest as the body of the post. If we're cloning a ws, there are only two things that can happen wrt realms. 
1. The source ws is already protected so the clone has to be as well. 
2. The option to make the clone protected, regardless of the source ws. 
Isn't it sufficient to just add a single parameter, something like `isProtected` to the POST?

src/main/resources/application.conf
rushtong: I think we do. IIRC, `dbGapAuthorizedUsers` is something that should exist in dev since it was part of the NIH whitelist stuff.

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -24,20 +26,22 @@ trait WorkspaceService extends HttpService with PerRequestCreator with FireCloud
   lazy val log = LoggerFactory.getLogger(getClass)
   lazy val rawlsWorkspacesRoot = FireCloudConfig.Rawls.workspacesUrl
 
+  def transformSingleWorkspaceRequest(entity: HttpEntity): HttpEntity = {
+    entity.as[RawlsWorkspaceResponse] match {
+      case Right(rwr) => new UIWorkspaceResponse(rwr).toJson.prettyPrint
```
rushtong, line 31: I think it is more idiomatic to use Success/Failure instead of Right/Left. 
Also, no need for prettyPrint, toJson should be good.

src/main/resources/swagger/api-docs.yaml
MattPutnam, line 533: This is matching Rawls functionality.  Rawls was updated to take the full ingest object and do an attribute update as part of the clone.  Not yet in the UI, but this would allow the user to update the description at clone time as well.  It will be the UI's responsibility to make sure we're not allowing nonsensical realms (or we'll get an error from Rawls).

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
MattPutnam, line 31: entity.as[T] is returning an Either.  I agree that it's kinda weird, do you know a better way to attempt the unmarshalling?

And as-is, it's not liking just `.toJson`, it's needing it as a string in order to convert back to HttpEntity.  Maybe there's more implicit magic I'm missing.

src/main/resources/application.conf
MattPutnam: Rawls is still giving me a 500 with "Cannot find Group corresponding to RawlsGroupName(dbGapAuthorizedUsers)"

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
rushtong, line 31: Ah, you're right, just `new UIWorkspaceResponse(rwr).toJson.toString` should work.

src/main/resources/application.conf
rushtong: That will be the case until it gets re-added in. An admin in former-Blue should be able to get that done.

src/main/resources/swagger/api-docs.yaml
rushtong, line 533: Good, thanks.

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
MattPutnam, line 31: That works, but then it's compact printed.

src/main/scala/org/broadinstitute/dsde/firecloud/model/Workspace.scala
```diff
@@ -14,14 +16,72 @@ case class WorkspaceEntity(
 case class WorkspaceCreate(
   namespace: String,
   name: String,
-  attributes: Option[Map[String, String]] = None,
+  attributes: Map[String, String],
```
rushtong, line 19: Aren't attributes optional?

rushtong, line 19: Never mind. I see in Rawls that they are not. They can be empty though.

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
rushtong, line 31: With respect to unmarshalling, you're doing the right thing. You could also use spray json to do this, but it's uglier

```
val convertedEntity = entity.toString.parseJson.convertTo[RawlsWorkspaceResponse]
```

rushtong:
:+1: Looks good except for the actual group name. When that is in rawls dev, this can be merged.

PR #210, opened 2016-03-02
Add 'isProtected' mapping to list workspaces

MattPutnam:
Not part of a current JIRA story, but this will be needed for some UI refinements that will eventually be needed
- [x] **LR**: sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
@@ -8,6 +8,7 @@ import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
 import org.broadinstitute.dsde.firecloud.{EntityClient, FireCloudConfig}
 import org.slf4j.LoggerFactory
 import spray.json._
+import spray.json.DefaultJsonProtocol._
 import spray.http.{HttpMethods, HttpEntity}
 import spray.httpx.unmarshalling._
 import spray.httpx.SprayJsonSupport._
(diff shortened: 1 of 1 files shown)


rushtong:
I'm not sure why this needs to be in orchestration. Doesn't the workspace object from Rawls contain a realm that could be referenced in the UI?

MattPutnam:
The UI doesn't know which realm corresponds to the "protected" realm.  We could add that, but then we have config stuff in 2 places.  David An also thought it would be good to not rely on that from the UI side of things.

rushtong:
OK, I can see that perspective. Not sure I totally agree, but it's a minor point.

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -35,10 +36,21 @@ trait WorkspaceService extends HttpService with PerRequestCreator with FireCloud
...
 
+  def transformListWorkspaceRequest(entity: HttpEntity): HttpEntity = {
+    entity.as[List[RawlsWorkspaceResponse]] match {
+      case Right(lrwr) => lrwr.map(new UIWorkspaceResponse(_)).toJson.prettyPrint
+      case Left(error) =>
+        log.error("Unable to unmarshal entity -- " + error.toString)
+        entity
+    }
+  }
+
```
rushtong, line 47: Now that we have two methods that do roughly the same thing (this and transformSingleWorkspaceRequest) - it might be nice to refactor those into `FireCloudDirectiveUtils`

rushtong:
:+1: One minor comment, but looks good.

PR #212, opened 2016-03-09
Add the option of requesting compression for outgoing calls

rushtong:
Use GZIP to request compressed calls on selected rawls endpoints

Notes:
- Addresses GAWB-342: https://broadinstitute.atlassian.net/browse/GAWB-342
- To test the main functionality, follow the set and non-set entity download links
- There are three primary paths for external requests:
  - passthrough
  - passthroughAllPaths
  - externalHttpPerRequest
    the framework.
- The specific calls that make use of compression on outgoing calls are in 
  - EntityService
  - ExportEntitiesByType
  - GetEntitiesWithType
# 
- [x] **LR**: sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/HttpClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/HttpClient.scala
@@ -2,6 +2,8 @@ package org.broadinstitute.dsde.firecloud
 
 import java.text.SimpleDateFormat
 import org.broadinstitute.dsde.firecloud.model.RequestCompleteWithErrorReport
+import spray.httpx.encoding.Gzip
+import spray.http.HttpEncodings._
 
 import scala.concurrent.Future
 import scala.util.{Failure, Success}
@@ -9,7 +11,7 @@ import scala.util.{Failure, Success}
 import akka.actor.{Actor, Props}
 import org.slf4j.{Logger, LoggerFactory}
 import spray.client.pipelining._
-import spray.http.HttpHeaders.Cookie
+import spray.http.HttpHeaders.{`Accept-Encoding`, Cookie}
 import spray.http._
 import spray.routing.RequestContext
 import spray.http.StatusCodes._
@@ -20,7 +22,7 @@ import org.broadinstitute.dsde.firecloud.service.PerRequest.RequestCompleteWithH
 
 object HttpClient {
 
-  case class PerformExternalRequest(request: HttpRequest)
+  case class PerformExternalRequest(requestCompression: Boolean, request: HttpRequest)
 
   def props(requestContext: RequestContext): Props = Props(new HttpClient(requestContext))
 
@@ -41,17 +43,24 @@ class HttpClient (requestContext: RequestContext) extends Actor
 
   override def receive: Receive = {
 
-    case PerformExternalRequest(externalRequest: HttpRequest) =>
-      createResponseFutureFromExternalRequest(requestContext, externalRequest)
+    case PerformExternalRequest(requestCompression: Boolean, externalRequest: HttpRequest) =>
+      createResponseFutureFromExternalRequest(requestCompression, requestContext, externalRequest)
 
   }
 
   def createResponseFutureFromExternalRequest(
+      requestCompression: Boolean,
       requestContext: RequestContext,
       externalRequest: HttpRequest): Unit = {
     val pipeline: HttpRequest => Future[HttpResponse] =
-      authHeaders(requestContext) ~> addHeader(Cookie(requestContext.request.cookies)) ~>
-        logRequest(log) ~> sendReceive
+      requestCompression match {
+        case true =>
+          authHeaders(requestContext) ~> addHeaders(Cookie(requestContext.request.cookies), `Accept-Encoding`(gzip)) ~>
+            logRequest(log) ~> sendReceive ~> decode(Gzip)
+        case _ =>
+          authHeaders(requestContext) ~> addHeader(Cookie(requestContext.request.cookies)) ~>
+            logRequest(log) ~> sendReceive
+      }
     pipeline(externalRequest) onComplete {
       case Success(response) =>
         log.debug("Got response: " + response)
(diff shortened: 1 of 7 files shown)


rushtong:
Back to me for some re-work/re-testing

rushtong:
@davidangb - back to you for re-review.

src/main/scala/org/broadinstitute/dsde/firecloud/HttpClient.scala
```diff
@@ -41,17 +43,24 @@ class HttpClient (requestContext: RequestContext) extends Actor
...
+      requestCompression: Boolean,
       requestContext: RequestContext,
       externalRequest: HttpRequest): Unit = {
     val pipeline: HttpRequest => Future[HttpResponse] =
-      authHeaders(requestContext) ~> addHeader(Cookie(requestContext.request.cookies)) ~>
-        logRequest(log) ~> sendReceive
+      requestCompression match {
+        case true =>
+          authHeaders(requestContext) ~> addHeaders(Cookie(requestContext.request.cookies), `Accept-Encoding`(gzip)) ~>
+            logRequest(log) ~> sendReceive ~> decode(Deflate)
```
davidangb: does `decode(Deflate)` work? Should this be `decode(Gzip)` instead? (here and elsewhere)

rushtong: It worked in my tests. Here is the doc example: http://spray.io/documentation/1.2.2/spray-client/#usage

rushtong: Re-reading the docs for http://spray.io/documentation/1.2.2/spray-routing/encoding-directives/compressResponseIfRequested/ makes me think you're right though, this shouldn't work.

src/main/scala/org/broadinstitute/dsde/firecloud/HttpClient.scala
```diff
@@ -2,6 +2,8 @@ package org.broadinstitute.dsde.firecloud
 
 import java.text.SimpleDateFormat
 import org.broadinstitute.dsde.firecloud.model.RequestCompleteWithErrorReport
+import spray.httpx.encoding.{Gzip, Deflate}
```
davidangb: Deflate import is no longer needed

src/main/scala/org/broadinstitute/dsde/firecloud/core/ExportEntitiesByType.scala
```diff
@@ -10,21 +10,24 @@ import akka.pattern.pipe
...
 import org.broadinstitute.dsde.firecloud.utils.TSVFormatter
 
 import spray.client.pipelining._
+import spray.http.HttpEncodings._
+import spray.http.HttpHeaders.`Accept-Encoding`
 import spray.http.MediaTypes._
 import spray.http.StatusCodes._
 import spray.http._
 import spray.httpx.SprayJsonSupport._
+import spray.httpx.encoding.{Gzip, Deflate}
```
davidangb: same here - Deflate import not needed

src/main/scala/org/broadinstitute/dsde/firecloud/core/GetEntitiesWithType.scala
```diff
@@ -13,21 +11,19 @@ import org.broadinstitute.dsde.firecloud.service.{FireCloudDirectiveUtils, FireC
 import org.broadinstitute.dsde.firecloud.service.PerRequest.{PerRequestMessage, RequestComplete}
 
 import spray.client.pipelining._
+import spray.http.HttpEncodings._
+import spray.http.HttpHeaders.`Accept-Encoding`
 import spray.http.HttpResponse
 import spray.http.StatusCodes._
 import spray.httpx.SprayJsonSupport._
+import spray.httpx.encoding.{Gzip, Deflate}
```
davidangb: Deflate import not necessary

src/main/scala/org/broadinstitute/dsde/firecloud/service/FireCloudDirectives.scala
```diff
@@ -22,37 +22,47 @@ trait FireCloudDirectives extends spray.routing.Directives with PerRequestCreato
...
-    val outMethod = new RequestBuilder(inMethod)
+    generateExternalHttpPerRequestForMethod(requestCompression = false, unencodedPath, inMethod)
+  } reduce (_ ~ _)
 
-    val path = encodeUri(unencodedPath)
+  def passthrough(requestCompression: Boolean, unencodedPath: String, methods: HttpMethod*) = methods map { inMethod =>
+    generateExternalHttpPerRequestForMethod(requestCompression, unencodedPath, inMethod)
+  } reduce (_ ~ _)
+
+  def passthroughAllPaths(ourEndpointPath: String, targetEndpointUrl: String, requestCompression: Option[Boolean] = Some(false)) = pathPrefix(ourEndpointPath) {
```
davidangb: could you make the requestCompression argument be `requestCompression: Boolean = false`, instead of an `Option[Boolean]`? Then you could remove the match/case, and just pass the value through. Since the `passthrough()` signature that omits the requestCompression argument defaults to false anyway, this should be equivalent functionality and less code.

rushtong: That is less code, thanks - good idea.

davidangb:
I've made a few syntax comments inline (and you're responding). Otherwise, syntax looks good.

As far as overall approach, I'm waffling - which is better?
- the approach you took, which makes requestCompression an argument for outgoing calls. This allows flexibility on different service calls, but adds complexity to the code and requires touching every service call to enable compression
- a global approach, which adds `addHeader(`Accept-Encoding`(gzip)) ~> sendReceive ~> decode(Gzip)` to _every_ outgoing call, and relies on the target service to enable/disable compression. This is brute force, but keeps the codebase simple, and doesn't require any updates to enable compression over time

rushtong:
I agree that the two approaches are waffle-able - but I come down more on the per-service approach so that concerns are addressed as needed instead of all-at-once. I don't have strong feelings on this - I could be persuaded the other way.

davidangb:
okay, looks good, and I'm fine with your approach. I'm giving this a :+1: but please don't merge - this is blocked on infosec just like my other PRs.

davidangb:
@rushtong back to you - this is now unblocked and approved - do all the merging stuff! This will need the rawls config changes (and a rawls redeploy) to take effect.

PR #218, opened 2016-04-27
GAWB-553: orchestration cookies (and logging optimization)

davidangb:
(no this isn't in our sprint. It really bugged me and I went to investigate it and found out afterwards Joel filed a ticket for it.)

Two changes in here, in separate commits:

1) Orchestration's HttpClient, used for passthroughs and elsewhere, no longer forwards cookies on to underlying services. I cannot think of a reason that we need to send cookies. Most importantly, the particular code formation that was in use ALWAYS added a Cookie: header, even if it had an empty value ... which it did most/all the time. This resulted in an invalid cookie, which in turn caused rawls (and probably agora, thurloe, etc) to spew messages like:

```
rawls [WARN] [04/26/2016 21:30:52.401] [rawls-akka.actor.default-dispatcher-8] [akka://rawls/user/IO-HTTP/listener-0/0] Illegal request header: Illegal 'Cookie' header: Unexpected end of input, expected $timesCookie (line 1, pos 1):
```

over and over and over. I tried to dig through history to find out why we added cookies. It appears to have been added as part of #74, which ... had no reviewers???? This was ~7 months ago.

2) While I was in this class, I made a slight tweak to the logRequest RequestTransformer. Previously, we would ALWAYS call `request.toString`, then perform string concat for `"Sending request: " + request`, and send the resultant string into the logger. The logger would then perform a level check, and since we don't run at debug level except in local-dev cases, the logger would then toss out the string; all the string concatenation and manipulation was wasted. Now, we perform the logger level check up front, and short-circuit, avoiding all the string manipulation. I would not normally worry about this level of micro-optimization, but (I was in the code anyway and) the request object is a pretty big object to serialize to string, and we were doing this for EVERY http request.
- [x] **LR**: sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/HttpClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/HttpClient.scala
@@ -73,7 +73,7 @@ class HttpClient (requestContext: RequestContext) extends Actor
 
 trait LogRequestBuilding extends spray.httpx.RequestBuilding {
   def logRequest(log: Logger): RequestTransformer = { request =>
-    log.debug("Sending request: " + request)
+    if (log.isDebugEnabled) {log.debug("Sending request: " + request)}
     request
   }
 }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/HttpClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/HttpClient.scala
@@ -55,11 +55,9 @@ class HttpClient (requestContext: RequestContext) extends Actor
     val pipeline: HttpRequest => Future[HttpResponse] =
       requestCompression match {
         case true =>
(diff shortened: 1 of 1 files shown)


dmohs:
👍 This is great and I agree with your optimization.

jmthibault79:
Looks good to me.  One concern: does this affect the "Cookie Authed" endpoints?

davidangb:
This should not affect the cookie-authed endpoints. Those endpoints accept a cookie from the browser, translate the cookie to an Authorization: header, and send the auth header to rawls. My change above only affects whether or not we send cookies from orch to rawls. Nonetheless, I will test manually and make sure!

jmthibault79:
👍

davidangb:
tested TSV download and GCS object download, which are cookie-authed endpoints, and they work fine for me locally. @jmthibault79 anything else on this PR?

PR #223, opened 2016-06-02
GAWB-531: create mock NIH endpoint for testing linkages

davidangb:
Please see GAWB-531 for important info that doesn't belong in this PR.

This PR creates an endpoint that we can use in local/dev to test linking accounts with NIH. It provides capability - via presence/absence of config keys - to disable the endpoint in prod and other environments.

This PR is incomplete on its own! It requires additional PRS for config:
- https://github.com/broadinstitute/firecloud-develop/pull/88 (dev)
- https://github.com/broadinstitute/firecloud-develop/pull/87 (local)

This endpoint Intentionally left out of swagger; it will not work in production.

Finally - this does NOT complete GAWB-531; it only solves half of it. So I'm not updating the jira ticket.
- [x] **Submitter**: Rebase to develop. DO NOT SQUASH
- [x] **Submitter**: Make sure Swagger is updated if API changes
- [x] **Submitter**: Make sure documentation for code is complete
- [x] **Submitter**: Review code comments; remove done TODOs, create stories for remaining TODOs
- [x] **Submitter**: Include the JIRA issue number in the PR description
- [x] **Submitter**: Add description or comments on the PR explaining the hows/whys (if not obvious)
- [x] **Submitter**: Tell tech lead that the PR exists if s/he wants to look at it
- [x] **Submitter**: Anoint a lead reviewer (LR). **Assign PR to LR**
- [x] **LR**: Initial review by LR and others.
- [ ] Comment / review / update cycle:
  - Rest of team may comments on PR at will
  - **LR assigns to submitter** for feedback fixes
  - Submitter updates documentation as needed
  - Submitter rebases to develop again if necessary
  - Submitter makes further commits. DO NOT SQUASH. **Reassign to LR** for further feedback
- [x] **Tech lead**:  sign off
- [x] **LR**: sign off
- [x] **Assign to submitter** to finalize
- [ ] **Submitter**: Squash commits, rebase if necessary
- [ ] **Submitter**: Verify all tests go green, including CI tests
- [ ] **Submitter**: Merge to develop 
- [ ] **Submitter**: Delete branch after merge
- [ ] **Submitter**: Check configuration files in Jenkins in case they need changes
- [ ] **Submitter**: **Test this change works on dev environment after deployment**. YOU own getting it fixed if dev isn't working for ANY reason!
- [ ] **Submitter**: Verify swagger UI on dev server still works after deployment
- [ ] **Submitter**: Inform other teams of any API changes via hipchat and/or email
[...]

PR #225, opened 2016-06-17
Update nih link callback to be a better per-request citizen

rushtong:
Addresses https://broadinstitute.atlassian.net/browse/GAWB-695
This has a companion PR in UI: https://github.com/broadinstitute/firecloud-ui/pull/456

The intention of this PR is to fix the nih link callback so that it does the white-list sync _and then_ handles the update in Thurloe for the user's nih profile information. 

Testing:
- See companion PR in UI
- There is a branch not in dev at the moment for handling a mock NIH link endpoint. See #223 for that. This PR needs #223 for the UI piece to work correctly.
- To ensure that the NIH linking response is the one that is actually returned by the ProfileClient, you can add 
  some custom content to the returned callback response in ProfileClient, line 102: `case true => Future(RequestComplete(OK))` and see that in the network panel from the UI call.
# 
- [ ] **Submitter**: Anoint a lead reviewer (LR). **Assign PR to LR**
- [ ] **LR**: Initial review by LR and others.
- [ ] **Tech lead**:  sign off
- [ ] **LR**: sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
@@ -2,6 +2,7 @@ package org.broadinstitute.dsde.firecloud.core
 
 import akka.actor.{ActorRefFactory, Actor, Props}
 import akka.event.Logging
+import akka.pattern.ask
 import akka.pattern.pipe
 import org.broadinstitute.dsde.firecloud.FireCloudConfig
 
@@ -22,12 +23,13 @@ import spray.httpx.unmarshalling._
 import spray.routing.RequestContext
 
 import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.duration._
 import scala.io.Source
 import scala.util.{Try, Failure, Success}
 
 object ProfileClient {
   case class UpdateProfile(userInfo: UserInfo, profile: BasicProfile)
-  case class UpdateNIHLink(userInfo: UserInfo, nihLink: NIHLink)
+  case class UpdateNIHLinkAndSyncSelf(userInfo: UserInfo, nihLink: NIHLink)
   case class GetNIHStatus(userInfo: UserInfo)
   case class GetAndUpdateNIHStatus(userInfo: UserInfo)
   case object SyncWhitelist
@@ -87,8 +89,10 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
 
       profileResponse pipeTo context.parent
 
-    case UpdateNIHLink(userInfo: UserInfo, nihLink: NIHLink) =>
+    case UpdateNIHLinkAndSyncSelf(userInfo: UserInfo, nihLink: NIHLink) =>
       val parent = context.parent
+      // Complete syncWhitelist and ignore as neither success nor failure are useful to the client
+      ask(self, ProfileClient.SyncSelf(userInfo))(1.minute)
       val pipeline = authHeaders(requestContext) ~> sendReceive
       val profilePropertyMap = nihLink.propertyValueMap
       val propertyUpdates = updateUserProperties(pipeline, userInfo, profilePropertyMap)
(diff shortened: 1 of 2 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/NIHService.scala
```diff
@@ -48,19 +48,15 @@ trait NIHService extends HttpService with PerRequestCreator with FireCloudDirect
                 val lastLinkTime = DateUtils.now
                 val linkExpireTime = DateUtils.nowPlus30Days
 
-                // update this user's dbGaP access in rawls, assuming the user exists in the whitelist.
-                // this call is silent/async - if it fails, we continue on, and the user's access will
-                // be updated the next time sync_whitelist runs; the user will have to wait for that.
-                val groupUpdate = perRequest(requestContext, Props(new ProfileClientActor(requestContext)),
-                  ProfileClient.SyncSelf(userInfo))
-
```
rushtong, line 56: In the original code, this is one of two PerRequests that are fired off. Whichever one finishes first is the one that the client receives.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -87,8 +89,10 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
 
       profileResponse pipeTo context.parent
 
-    case UpdateNIHLink(userInfo: UserInfo, nihLink: NIHLink) =>
+    case UpdateNIHLinkAndSyncSelf(userInfo: UserInfo, nihLink: NIHLink) =>
       val parent = context.parent
+      // Complete syncWhitelist and ignore as neither success nor failure are useful to the client
+      ask(self, ProfileClient.SyncSelf(userInfo))(1.minute)
```
davidangb, line 95: does this wait for the ask to complete? i.e. does this need Await.ready? I haven't tried it, but just from reading the code I wonder ...

rushtong, line 95: From what I understand (http://doc.akka.io/docs/akka/current/scala/futures.html), `ask` is the non-blocking alternative to `Await`. I am pretty sure we do _not_ want to introduce a blocking call like Await here.

rushtong, line 95: I'm going to do a little more testing around this just to be 100% sure.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -332,7 +336,9 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
     RequestComplete(NoContent)
   }.recover {
     // intentionally quash errors so as not to leak sensitive data
-    case _ => RequestCompleteWithErrorReport(InternalServerError, "Error synchronizing NIH whitelist")
+    case _ =>
+      log.error("Error synchronizing NIH whitelist")
+      RequestCompleteWithErrorReport(InternalServerError, "Error synchronizing NIH whitelist")
```
rushtong, line 341: @davidangb - I'm seeing this error case trigger regularly while testing. I don't see why yet, so will be digging into that problem.
Edit: the errors are 401 unauthorized.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
davidangb, line 95: whether or not we block a thread, I think this needs to ensure the call completes. Otherwise, it raises the possibility that the call to put the user in the right group hasn't finished by the time we redirect the user ... which is one of the reasons the user can experience the bug.

Summary: I don't care about blocking ... I do care that both calls complete before the user gets a response. I should have been more clear.

rushtong, line 95: I agree - and that was the intent here. After some more testing, I'm not convinced that the ask pattern is right. I am trying another approach.

rushtong:
Closing this PR as the current approach is not doing what we want.

PR #228, opened 2016-06-22
GAWB-613: send thurloe notification to user upon account creation

MatthewBemis:
This relates to the following PRs:

rawls: https://github.com/broadinstitute/rawls/pull/459 (approved)
thurloe: https://github.com/broadinstitute/thurloe/pull/31 (approved)
firecloud-develop: https://github.com/broadinstitute/firecloud-develop/pull/93 and https://github.com/broadinstitute/firecloud-develop/pull/94
- [ ] **LR**: sign off

```

```

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
@@ -84,7 +84,8 @@ object FireCloudConfig {
   }
 
   object Notification {
-    lazy val activationTemplateId = sys.env.get("ACTIVATION_TEMPLATE_ID").getOrElse("dummy")
+    private val notification = config.getConfig("notification")
+    lazy val activationTemplateId = sys.env.get("ACTIVATION_TEMPLATE_ID").getOrElse(notification.getString("activationTemplateId"))
   }
 
 }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
@@ -65,6 +65,7 @@ object FireCloudConfig {
     lazy val getAll = profile.getString("getAll")
     lazy val getQuery = profile.getString("getQuery")
     lazy val delete = profile.getString("delete")
+    lazy val postNotify = profile.getString("postNotify")
   }
 
   object FireCloud {
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
@@ -83,4 +83,8 @@ object FireCloudConfig {
     lazy val rawlsGroupName = nih.getString("rawlsGroupName")
   }
 
+  object Notification {
+    lazy val activationTemplateId = sys.env.get("ACTIVATION_TEMPLATE_ID")
+  }
+
 }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
@@ -84,7 +84,7 @@ object FireCloudConfig {
   }
 
   object Notification {
-    lazy val activationTemplateId = sys.env.get("ACTIVATION_TEMPLATE_ID").get
+    lazy val activationTemplateId = sys.env.get("ACTIVATION_TEMPLATE_ID").getOrElse("dummy")
   }
 
 }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
@@ -84,7 +84,7 @@ object FireCloudConfig {
   }
 
   object Notification {
-    lazy val activationTemplateId = sys.env.get("ACTIVATION_TEMPLATE_ID")
+    lazy val activationTemplateId = sys.env.get("ACTIVATION_TEMPLATE_ID").get
   }
 
 }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
@@ -74,7 +74,7 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
             }
             completionUpdate.flatMap { response =>
               response.status match {
-                case x if x.isSuccess => checkUserInRawls(pipeline, requestContext)
+                case x if x.isSuccess => checkUserInRawls(pipeline, requestContext, userInfo.getUniqueId)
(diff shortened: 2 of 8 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -258,13 +258,13 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
 
   def checkUserInRawls(
     pipeline: WithTransformerConcatenation[HttpRequest, Future[HttpResponse]],
-    requestContext: RequestContext): Future[PerRequestMessage] = {
+    requestContext: RequestContext, userId: String): Future[PerRequestMessage] = {
```
MatthewBemis, line 261: do I have to pass userId into checkUserInRawls? is there a way to pull it out of the requestContext? it's needed for the call to thurloe if the user is new, but i would assume that this id already exists somewhere in the request

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -275,10 +275,14 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
...
     pipeline { Post(UserService.rawlsRegisterUserURL) } map { response =>
         response.status match {
           case x if x.isSuccess || isConflict(response) =>
+            //sendNotification is purposely detached from the sequence. we don't want to block
+            //the completion of registration by waiting for sendgrid to succeed or fail the registration
+            // if sendgrid fails. unfortunately, if this call fails then ¯\_(ツ)_/¯
+            sendNotification(pipeline, ActivationNotification(userId))
```
MatthewBemis, line 285: this is how i think it should work. disagree?

i do log when this fails.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
davidangb, line 261: The spray pipeline is created at line 64, and includes the Authorization: header from the original request. That pipeline is passed into checkUserInRawls (line 260). So, when orch calls rawls, it passes the Authorization: header to rawls. Then, the rawls proxy can identify the user, just from that header.

src/main/scala/org/broadinstitute/dsde/firecloud/model/Notification.scala
```diff
@@ -0,0 +1,16 @@
...
+ */
+sealed trait Notification {
+  val userId: String
+  val notificationId: String
+  def toMap: Map[String, String]
+}
+
+case class ActivationNotification(recipientUserId: String) extends Notification {
+  val userId = recipientUserId
+  val notificationId = "8c646115-d58b-45c2-a8ea-a4f5d47d5d2e"
```
davidangb: What is this UUID? Is this a hardcoded sendGrid ID? Maybe this should be in config.

MatthewBemis: yeah, on second thought it should totally be in a config. these UUIDs will be different per environment (or at least different for prod) because it will be using a different sendgrid account. it looks like firecloud-orch only uses environment variables and application.conf which is the same per env. so i'm assuming this should be a new config populated per environment in fc-develop?

edit: and this uuid is also for my personal sendgrid account...so it shouldn't be in here anyway

davidangb: yes, orchestration defines per-env settings via docker-compose in firecloud-develop.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
rushtong, line 285: I agree, it is an unrecoverable error from the client's perspective and should be logged, not passed through.

src/main/scala/org/broadinstitute/dsde/firecloud/model/Notification.scala
rushtong: 👍  for moving this value to config.

rushtong:
👍 on code review. Looking at functional testing now.

rushtong:
👍  on testing. I was able to stand up ui+orch+rawls and see the right thurloe endpoint get called. That's as far as I can get until the Thurloe changes are in place.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
MatthewBemis, line 261: ahh, i see. so when I call the checkUserInRawls function in line 77, it doesn't yet know what the userId is for this Authorization header because it hasn't passed through the proxy. so i would assume this means it's ok if I pass in userId here instead of deriving it from the pipeline?

rushtong, line 261: By the time the request has made it into orchestration, it has passed through the orchestration proxy so the userInfo object _should_ be fully populated.

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
```diff
@@ -82,4 +83,8 @@ object FireCloudConfig {
     lazy val rawlsGroupName = nih.getString("rawlsGroupName")
   }
 
+  object Notification {
+    lazy val activationTemplateId = sys.env.get("ACTIVATION_TEMPLATE_ID").getOrElse("dummy")
```
MatthewBemis: I don't plan on this getOrElse being in the final code

MatthewBemis:
This is ready for re-review. I played around a bit with build.sbt to try and find a way to default to a test.conf during testing which would inherit application.conf and add the notification configs. The only way I could see to do this was to introduce the entire Play library.

src/main/scala/org/broadinstitute/dsde/firecloud/model/Profile.scala
```diff
@@ -10,6 +10,11 @@ case class ThurloeKeyValue(
   userId: Option[String] = None,
   keyValuePair: Option[FireCloudKeyValue] = None)
 
+case class ThurloeNotification(
+  userId: String,
+  notificationId: String,
+  substitutions: Map[String, String])
+
```
dvoet: @MatthewBemis we had talked through a more strongly typed way of modeling notifications, what happened to that?

davidangb:
I like the notification handling. Potentially set the value in application.conf to null instead of the empty string? Either way is fine.

As you said, next step is to use a test.conf file. I don't think you'll need all of Play - probably going to be something like this: http://stackoverflow.com/questions/15399161/how-do-i-specify-a-config-file-with-sbt-0-12-2-for-sbt-test

MatthewBemis:
@davidangb back to you. i had tried that stack overflow solution before as well and couldn't get it to work even with some tinkering. i guess if you just make another conf named "application.conf" and put it in the test package, it will automatically use that without any build.sbt changes

davidangb:
I'd love to see a unit test directly against ProfileClient.registerUserInRawls, especially one that checks its side effect of sending a notification, but I won't hold the PR for that. Giving this a 👍 .

PR #233, opened 2016-07-08
GAWB-843: throw exception if rawls cannot update whitelist membership

MatthewBemis:
- [x] **LR**: sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
@@ -342,7 +342,7 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
 
     memberList flatMap { members => pipeline { rawlsRequestMethod(url, members) } } map { response =>
       if(response.status.isFailure)
-        throw new Exception("Error synchronizing NIH whitelist")
+        RequestCompleteWithErrorReport(InternalServerError, "Error synchronizing NIH whitelist")
       else RequestComplete(NoContent)
     } map { _ => RequestComplete(NoContent) } // don't leak any sensitive data
   }.get recover {
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
@@ -343,8 +343,8 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
     memberList flatMap { members => pipeline { rawlsRequestMethod(url, members) } } map { response =>
       if(response.status.isFailure)
         RequestCompleteWithErrorReport(InternalServerError, "Error synchronizing NIH whitelist")
-      else RequestComplete(NoContent)
-    } map { _ => RequestComplete(NoContent) } // don't leak any sensitive data
+      else RequestComplete(NoContent) // don't leak any sensitive data
+    }
   }.get recover {
     // intentionally quash errors so as not to leak sensitive data
     case _ => RequestCompleteWithErrorReport(InternalServerError, "Error synchronizing NIH whitelist")
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
@@ -88,7 +88,7 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
 
     case UpdateNIHLinkAndSyncSelf(userInfo: UserInfo, nihLink: NIHLink) =>
       val parent = context.parent
-      val syncWhiteListResult = Future(syncWhitelist(Some(userInfo.getUniqueId)))
+      val syncWhiteListResult = syncWhitelist(Some(userInfo.getUniqueId))
       val pipeline = authHeaders(requestContext) ~> sendReceive
       val profilePropertyMap = nihLink.propertyValueMap
       val propertyUpdates = updateUserProperties(pipeline, userInfo, profilePropertyMap)
@@ -114,9 +114,9 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
       // do NOT pipe the response back to the parent!
 
     case SyncWhitelist =>
-      context.parent ! syncWhitelist()
+      syncWhitelist() pipeTo context.parent
 
-   }
+  }
 
   def getProfile(userInfo: UserInfo, updateExpiration: Boolean): Future[PerRequestMessage] = {
     val pipeline = addCredentials(userInfo.accessToken) ~> sendReceive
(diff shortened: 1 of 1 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -340,14 +340,15 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
       case None => Put
     }
 
-    memberList flatMap { members => pipeline { rawlsRequestMethod(url, members) } }
-
-    // don't leak any sensitive data
-    RequestComplete(NoContent)
-  }.recover {
+    memberList flatMap { members => pipeline { rawlsRequestMethod(url, members) } } map { response =>
```
MatthewBemis, line 343: this was the ideal approach but I don't think it's going to work well. this call will probably time out 100% of the time it's called. it needs to wait for rawls to issue all of the google group updates. for n changes to the whitelist, there will be n calls to google just to update the dbGapAuthorizedUsers group. Then, for each of the n users, we need to issue a call to google to update the intersection groups for each protected workspace that they're a member of. rawls also needs to update all of the groups in the database. this is probably the slowest endpoint we have.

an alternative approach is to undo these changes and go back to issuing the rawls call on the side and then return some status code like Accepted. in the background, we'll log whether or not the call to rawls succeeded. regardless of the approach, we also need to be aware of the fact that the call to rawls itself will likely timeout too.

thoughts?

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -340,14 +340,15 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
...
+      if(response.status.isFailure)
+        throw new Exception("Error synchronizing NIH whitelist")
```
davidangb: use `RequestCompleteWithErrorReport(InternalServerError ...` here instead?

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
jmthibault79, line 343: I think something like the alternative you mention will be necessary, because this could take ... say 10 minutes, ballpark?  Also note that we need to consider the "Power Goes Out" situation, so that implies to me something like a PendingWhitelistSyncTable and WhitelistSyncMonitor.

davidangb, line 343: If the call to rawls is truly bound to time out, then I think orchestration should return Accepted, and rawls should log success/failure.

The goal of all this work is that we are able to detect errors in the whitelist synchronization. We do this (as you know) via loggly alerts - those can look at any log. So as long as rawls logs failures, we can alert on the rawls failures - which may or may not require updates to the loggly alert criteria.

Orchestration can/should still log failures with everything except the call to rawls, of course.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
rushtong: If we throw any kind of error here, we'll be percolating a non-fatal exception back down to the user. That error will be useless to the user. I think we should log it and catch it as flagged alert.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
davidangb, line 343: Do we really need to handle power-goes-out for this call? Since it's a cron job that runs every ... 6 hours? 12 hours? It should run again after the power comes back on. Are you worried about a half-completed synchronization?

jmthibault79, line 343: Maybe we don't.  You're right that any problems will be corrected in 6 (I think) hours.

MatthewBemis, line 343: @davidangb ok, so it sounds like the bulk of the changes for this ticket should actually occur in rawls and not orch. rawls has no idea what it means to sync a whitelist so it would need to learn.

davidangb, line 343: @MatthewBemis erm, you're right. Rawls probably shouldn't know about whitelist synching. Hmm. This is a good question. Thinking ...

dvoet, line 343: While I do believe this is a slow call I have not seen it time out in production. And even if timeouts happen the background process goes on and the next sync will be a success. But that is only for this use case and it will only get worse.

So first, I think we should understand when this thing will timeout because it also will be a problem for changing permissions on a workspace. If we do need something asynchronous, we should establish a pattern for it that includes monitoring because it is not the only thing we will want it for.

I don't think that rawls needs to understand white lists. What we would implement is asynchronous group membership changes.

Note that I have already added the loggly alert for the rawls call. With or without this change we should still get the proper alerting.

MatthewBemis, line 343: So to follow up on this: do we need this or not? It seems like the current implementation isn't necessarily wrong. Yes, it does seem weird that orchestration would return success even if the call to rawls eventually failed, but there's nothing that needs to rely on this status code from orch if the call to rawls has loggly alerts setup. Loggly would just be alerting from 1 layer instead of 2.

davidangb, line 343: Can we get an answer on whether or not these calls time out in practice?

I still think we need this. Even if the call to rawls could time out, the call to rawls _COULD_ return an error quickly. In that case, orchestration should report that error.

MatthewBemis, line 343: Good point. Doug says he hasn't seen it time out in production but I'm seeing it timeout in dev every single time, even with retries on Google calls disabled. I'll do some more investigation to see what's causing the difference.

MatthewBemis, line 343: @davidangb @dvoet After investigating I don't see how this call to rawls will ever NOT timeout (unless it returns an error quickly like you said). When I tested this yesterday with between 80-90 protected workspaces in dev, it was generating ~23,000 queries to update all of the rawls groups, taking a total of ~38 minutes (granted, local is a lot slower than dev...). The dev DB was in a horrible state for testing whitelist syncing so today I cleaned it all up and made just 1 protected workspace. Syncing the whitelist now doesn't time out, but it generates about 280 queries- which is in line with the 23k @ 80 workspaces I saw yesterday. 

I tested a few other things:

adding a user as a reader on a non-protected workspace: ~135 queries
adding a user as a reader on a protected workspace: ~870 queries

So I think this should spawn another tech debt story which is to refactor the group update queries. As for this PR (and GAWB-843)- I'm happy to clean this up based on the other comments.

I also attached the Slick log output for updating the whitelist with 1 protected workspace.

[whitelist_sync_output.txt](https://github.com/broadinstitute/firecloud-orchestration/files/362086/whitelist_sync_output.txt)

davidangb, line 343: for clarity - are your numbers DB queries or API calls to Google? I'm assuming DB queries. Just wanna make sure we're not hammering Google like that. Agreed this needs a tech debt story to clean up.

MatthewBemis, line 343: In this case it's DB queries, but there's also a chance that we'll be hammering Google too if there's a huge change in the whitelist. For example, there are about 30 controlled access workspaces that every dbGapAuthorizedUser is given access to. If the whitelist updates and we see that we need to remove 50 users, we'll end up making 30x50 calls to Google to remove each individual from each workspace's read intersection group. I haven't yet seen a way to batch update Google groups :(

MatthewBemis:
@davidangb back to you with a few minor adjustments. i also need to create that tech debt ticket before i consider this done

davidangb:
👍

PR #234, opened 2016-07-11
Load balancing endpoints

rushtong:
This PR addresses GAWB-807 but does not complete it.
https://broadinstitute.atlassian.net/browse/GAWB-807

Added two endpoints required for load balancing. We need a normal health check to ensure system functionality and we need an unhealthy check to force the system into "maintenance mode". By switching the configured LB health check to be one or the other, we can force the switch. These endpoints are not behind `/api` so that the LBs do not have to deal with OAuth. 

**Question for reviewer**: Do we need these to be defined in swagger? My inclination is no since these are purely internally used endpoints for the load balancer.
# 
- [x] **LR**: sign off

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1819,6 +1819,26 @@ paths:
             - profile
             - https://www.googleapis.com/auth/devstorage.full_control
         - bearer:
+  /health:
+    get:
+      tags:
+      - Status
+      operationId: health
+      summary: |
+        A health status endpoint for load balancing purposes
+      responses:
+        200:
+          description: Success
+  /error:
+    get:
+      tags:
+      - Status
+      operationId: error
+      summary: |
+        An error status endpoint for load balancing purposes
+      responses:
+        503:
+          description: Service Unavailable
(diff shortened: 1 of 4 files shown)


MatthewBemis:
The only thing I have to say about the swagger question is that we're already listing an internal endpoint- /sync_whitelist. I agree with you, though, and think that listing internal endpoints is kind of weird for users. I feel like there's a reason we listed sync_whitelist but I can't remember what it was... maybe it was for us developers to be able to see all endpoints easily? 

:+1: pending a more definitive swagger answer

davidangb:
👍 for code. Nitpick: "health" is a noun and "unhealthy" is an adjective. Could you make these consistent? I like keeping "/health" for the up-check, since that's consistent with UI (https://github.com/broadinstitute/firecloud-ui/pull/460/files). They don't even have to be pure antonyms - /health and /error, maybe.

Swagger is a mess. I say go ahead and document them. End users shouldn't need these, but I'm all for transparency. It won't hurt to document.

rushtong:
👍  on the feedback - will update and re-submit.

davidangb:
👍

MatthewBemis:
:+1:

PR #246, opened 2016-08-30
GAWB-1004: Fix bug in NIH linking

MatthewBemis:
This fixes the following bug:

A user who links for the first time OR is re-linking after their access has expired will not be given immediate access to TCGA data. Re-linking a second time or waiting for the entire whitelist to sync would grant the user access.

When I dug into the code I realized that when doing individual-user linking, orchestration was asking thurloe for the expiration date of the current user's linkage. Since the date it would get from thurloe was either expired or didn't exist, orchestration would exclude them from the dbGapAuthorizedUsers group-update call to rawls. Orch would then update the expiration date in thurloe, so the user would be in a weird state where their NIH access is _technically_ active, but they haven't been granted any of the powers associated with it. Since their expiration date was now set to a date that would be considered active by orchestration, when they would link a second time (or wait for full whitelist syncing to occur), they would be given access.
  - Authentication
  - Authorization
  - Encryption
  - Audit trails
- Review cycle:
  - LR reviews
- [x] **TL** sign off
- [ ] **LR** sign off

…rs re-linking

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
@@ -321,11 +321,24 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
 
     // note: everything after this point involves Futures
 
-    val nihEnabledFireCloudUsers = NIHWhitelistUtils.getCurrentNihUsernameMap(subjectId) map { mapping =>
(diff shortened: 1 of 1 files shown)


helgridly:
I will :+1: this on style but I don't know enough about NIH linking to check this for correctness.

davidangb:
👍 great catch, thanks for the fix!

additional reading: http://blog.originate.com/blog/2014/06/15/idiomatic-scala-your-options-do-not-match/ ... no action necessary for this PR, but it made me think of that blog post (via Jeff Gentry) and I figured I'd pass it along.

PR #248, opened 2016-09-01
GAWB-986: Library attribute definitions

davidangb:
The bulk of this story is the src/main/resources/library/attribute-definitions.json file. This contains a first-draft/best-guess at aligning the attribute definitions from the two attachments to GAWB-916 (which don't always agree). Goal of this story is to create a draft, and allow edits later.

The format of that file is JSON Schema (http://json-schema.org/). By adhering to that standard, we inherit a bunch of syntax - we also enable the future ability to use attribute-definitions.json to validate the data we collect. Note that I've added a few custom fields on top of JSON Schema: "propertyOrder" to define how the order in which the UI renders fields, "inputHint" to provide end-user guidance in the UI. Everything else is standard. _TODO someday: create a JSON schema to validate these extensions._

The UI can read these definitions directly from orchestration, at /schemas/library-attributedefinitions-v1.

There's also a bit of refactoring in orchestration around reading resource files.

(This PR needs a rebase!)
  - Authentication
  - Authorization
  - Encryption
  - Audit trails
- Review cycle:
  - LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
@@ -39,6 +39,7 @@ class FireCloudServiceActor extends HttpServiceActor with FireCloudDirectives {
   val userService = new UserService with ActorRefFactoryContext
   val nihSyncService = new NIHSyncService with ActorRefFactoryContext
   val healthService = new HealthService with ActorRefFactoryContext
+  val libraryService = new LibraryService with ActorRefFactoryContext
 
   lazy val log = LoggerFactory.getLogger(getClass)
   val logRequests = mapInnerRoute { route => requestContext =>
@@ -73,6 +74,7 @@ class FireCloudServiceActor extends HttpServiceActor with FireCloudDirectives {
         userService.routes ~
         nihSyncService.routes ~
         healthService.routes ~
+        libraryService.routes ~
         pathPrefix("api") {
           routes
         } ~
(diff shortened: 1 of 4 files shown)


dvoet:
I see that we are providing "user friendly" metadata labels (e.g. Dataset Depositor for datasetDepositor). I am concerned about this because showing only these on the UI will lead to problems when people want to write method config expressions that use the values.  How will the user know they need to use datasetDepositor to reference the thing in the UI shown as "Dataset Depositor"?

davidangb:
@dvoet I haven't heard any use cases yet for people writing expressions against this metadata ... that said, I don't see any reason why we CAN'T display both the friendly name and the key.

dvoet:
@davidangb I am just raising the issue... I don't want to dictate solutions... this time.

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
```diff
@@ -98,8 +100,8 @@ class FireCloudServiceActor extends HttpServiceActor with FireCloudDirectives {
             serveIndex()
           } ~
             pathSuffix("api-docs") {
-              complete {
-                getResourceFileContents("swagger/api-docs.yaml")
+              withResourceFileContents("swagger/api-docs.yaml") { apiDocs =>
```
dmohs, line 103: Nice.

dmohs:
👍

PR #249, opened 2016-09-01
Add header to outgoing calls

rushtong:
This PR adds a predefined custom header and value to outgoing calls.
Required for https://broadinstitute.atlassian.net/browse/GAWB-799

The following PRs are related:
https://github.com/broadinstitute/firecloud-develop/pull/134
https://github.com/broadinstitute/firecloud-develop/pull/135
https://github.com/broadinstitute/thurloe/pull/35
  - Authentication
  - Authorization
  - Encryption
  - Audit trails
- Review cycle:
  - LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/FireCloudRequestBuilding.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/FireCloudRequestBuilding.scala
@@ -1,12 +1,15 @@
 package org.broadinstitute.dsde.firecloud.service
 
+import org.broadinstitute.dsde.firecloud.FireCloudConfig
 import org.broadinstitute.dsde.firecloud.dataaccess.HttpGoogleServicesDAO
-import spray.http.HttpHeaders.{RawHeader, Authorization}
+import spray.http.HttpHeaders.{Authorization, RawHeader}
 import spray.http._
 import spray.routing.RequestContext
 
 
 trait FireCloudRequestBuilding extends spray.httpx.RequestBuilding {
+
+  val fireCloudHeader = HttpHeaders.RawHeader("X-FireCloud-Id", FireCloudConfig.FireCloud.fireCloudId)
 
   // TODO: would be much better to make requestContext implicit, so callers don't have to always pass it in
   // TODO: this could probably be rewritten more tersely in idiomatic scala - for instance, don't create
@@ -20,10 +23,10 @@ trait FireCloudRequestBuilding extends spray.httpx.RequestBuilding {
 
     authorizationHeader match {
       // if we have authorization credentials, apply them to the outgoing request
-      case Some(c) => addCredentials(c)
+      case Some(c) => addCredentials(c) ~> addFireCloudCredentials
       // else, noop. But the noop needs to return an identity function in order to compile.
       // alternately, we could throw an error here, since we assume some authorization should exist.
-      case None => (r: HttpRequest) => r
+      case None => (r: HttpRequest) => r ~> addFireCloudCredentials
     }
 
   }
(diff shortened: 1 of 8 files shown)


davidangb:
I think this needs .travis.yml updates to handle the new `FIRE_CLOUD_ID` env var.

davidangb:
The passthrough at UserService line 164 needs the header, too, yes? (https://github.com/broadinstitute/firecloud-orchestration/blob/develop/src/main/scala/org/broadinstitute/dsde/firecloud/service/UserService.scala#L164) Or does that one feel safe, since we are forcefully injecting the current user's id?

src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
```diff
@@ -93,7 +93,7 @@ trait OAuthService extends HttpService with PerRequestCreator with FireCloudDire
 
             // we can't use the standard externalHttpPerRequest here, because the requestContext doesn't have the
             // access token yet; we have to add the token manually
-            val pipeline = addCredentials(OAuth2BearerToken(accessToken)) ~> sendReceive
+            val pipeline = addFireCloudCredentials ~> addCredentials(OAuth2BearerToken(accessToken)) ~> sendReceive
```
davidangb: I believe this is only used to talk to rawls, so the FC credentials are not needed here.

src/main/scala/org/broadinstitute/dsde/firecloud/service/FireCloudRequestBuilding.scala
```diff
@@ -41,6 +44,8 @@ trait FireCloudRequestBuilding extends spray.httpx.RequestBuilding {
   }
 
   // with great power comes great responsibility!
-  def addAdminCredentials = addCredentials(OAuth2BearerToken(HttpGoogleServicesDAO.getAdminUserAccessToken))
+  def addAdminCredentials = addCredentials(OAuth2BearerToken(HttpGoogleServicesDAO.getAdminUserAccessToken)) ~> addFireCloudCredentials
```
davidangb: I believe that addAdminCredentials is used twice in the orch codebase, both in ProfileClient.

First usage talks to Rawls. This usage needs the admin creds, but does NOT need the Thurloe shared secret.
Second usage talks to Thurloe. This usage needs the shared secret, but NOT the admin creds.

I suggest leaving addAdminCredentials as-is (i.e. do not include addFireCloudCredentials), and then change the second usage - the one that talks to Thurloe - to remove addAdminCredentials and replace it with addFireCloudCredentials.

davidangb:
Could you also modify the mock server in UserServiceSpec to check for the presence of the shared-secret header (and its mock value "123") on all relevant endpoints?

done with first-round review!

rushtong:
Good catch on `UserService.remoteGetAllURL`, I'm pretty sure the UI isn't calling that endpoint. It does need to be handled since the user's ID isn't enough - we also need to add the right header. As these cases expand, I'm less and less happy with handling them in individual cases and wonder if a better approach is to add it to all requests regardless of destination.

src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
rushtong: That pipeline is re-used to make a call to Thurloe somewhere below. The profile page definitely breaks without the credentials here.

rushtong: I'm wrong about this - it is not needed in `OAuthService`.

rushtong:
@davidangb - Ready for re-review. Travis isn't happy for something unrelated so I'm looking into why.

davidangb:
Code looks 👍 ... I haven't tested beyond unit tests.

PR #251, opened 2016-09-06
GAWB-992: use GCS signed urls for downloading

davidangb:
Related PRs are already merged:
broadinstitute/firecloud-develop#138
broadinstitute/firecloud-develop#139

Previously, when downloading a file from GCS, we proxied the entire file through PerRequest in orchestration. For large files, this was hugely inefficient, and according to the jira ticket could crash the orchestration service.

Now, when the user requests a download, we 1) request the user's identity from Google, using their token; 2) perform a final check to see if the user has permission access to the file. If the user is valid according to Google, and has permission to the file, we immediately sign a URL and redirect the browser to that signed URL. The URL is time-limited to two\* minutes for safety - and since we immediately redirect to it, two minutes should be more than enough for the current user. Empirical testing shows that the time limit is only relevant for _starting_ the download; once the download is going, it won't stop midway because of the signing expiration.

We log (in the orchestration log, which gets pushed to loggly) all download attempts, both successes and failures.

This has FISMA/security impact. If the current user shares the pre-signed url with somebody else, that other person is able to download the file within the two-minute time period, even if they would not otherwise have permission to that file.

This PR also fixes GAWB-988.

_* original PR had a five-minute expiration; I changed this to two minutes in the fourth commit._
  - Authentication
  - Authorization
  - Encryption
  - Audit trails
- Review cycle:
  - LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
@@ -122,6 +122,19 @@ object HttpGoogleServicesDAO {
       .build()
   }
 
+  def getRawlsServiceAccountAccessToken = {
+    val googleCredential = new GoogleCredential.Builder()
+      .setTransport(httpTransport)
+      .setJsonFactory(jsonFactory)
+      .setServiceAccountId(rawlsPemFileClientId)
+      .setServiceAccountScopes(storageReadOnly)
+      .setServiceAccountPrivateKeyFromPemFile(new java.io.File(rawlsPemFile))
+      .build()
+
+    googleCredential.refreshToken()
+    googleCredential.getAccessToken
+  }
+
   def getBucketObjectAsInputStream(bucketName: String, objectKey: String) = {
     val storage = new Storage.Builder(httpTransport, jsonFactory, getBucketServiceAccountCredential).setApplicationName("firecloud").build()
     storage.objects().get(bucketName, objectKey).executeMediaAsInputStream
@@ -161,5 +174,6 @@ object HttpGoogleServicesDAO {
     val gcsStatUrl = "https://www.googleapis.com/storage/v1/b/%s/o/%s"
     gcsStatUrl.format(bucketName, java.net.URLEncoder.encode(objectKey,"UTF-8"))
   }
+
 
 }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
@@ -155,4 +155,9 @@ object HttpGoogleServicesDAO {
       "&Signature=" + java.net.URLEncoder.encode(java.util.Base64.getEncoder.encodeToString(signedBytes), "UTF-8")
   }
 
+  def getObjectResourceUrl(bucketName: String, objectKey: String) = {
+    val gcsStatUrl = "https://www.googleapis.com/storage/v1/b/%s/o/%s"
+    gcsStatUrl.format(bucketName, java.net.URLEncoder.encode(objectKey,"UTF-8"))
+  }
+
 }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
@@ -132,10 +132,10 @@ object HttpGoogleServicesDAO {
     val verb = "GET"
     val md5 = ""
     val contentType = ""
-    val expire = (System.currentTimeMillis() /1000) + 300 // expires 5 minutes (300 seconds) from now
+    val expireSeconds = (System.currentTimeMillis() / 1000) + 120 // expires 2 minutes (120 seconds) from now
     val objectPath = s"/$bucketName/$objectKey"
 
-    val signableString = s"$verb\n$md5\n$contentType\n$expire\n$objectPath"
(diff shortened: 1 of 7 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -123,4 +125,39 @@ object HttpGoogleServicesDAO {
...
   }
 
+  // create a GCS signed url as per https://cloud.google.com/storage/docs/access-control/create-signed-urls-program
+  def getSignedUrl(bucketName: String, objectKey: String) = {
+
+    // generate the string-to-be-signed
+    val verb = "GET"
+    val md5 = ""
+    val contentType = ""
+    val expire = (System.currentTimeMillis() /1000) + 300 // expires 5 minutes (300 seconds) from now
```
dmohs: Nit: spacing. Also, I find it helpful to include the unit for integer durations, so I'd call this `expireSeconds`. Your option.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -123,4 +125,39 @@ object HttpGoogleServicesDAO {
...
+      .setServiceAccountPrivateKeyFromPemFile(new java.io.File(rawlsPemFile))
+    val privateKey = builder.getServiceAccountPrivateKey
+
+    // sign the string
+    val signature = java.security.Signature.getInstance("SHA256withRSA")
+    signature.initSign(privateKey)
+    signature.update(signableString.getBytes("UTF-8"))
+    val signedBytes = signature.sign()
+
+    // assemble the final url
```
dmohs, line 171: Nit: obvious. Suggest removing comment.

dmohs:
👍  Security-wise, I see no risks beyond what has already been stated.

However, we can use the functionality introduced here to improve our security posture in general. As a follow-up to this work, we can provide a separate API call for getting a signed URL that uses the `Authorization` header instead of a cookie. Once we update the UI to use that endpoint, we can likely stop storing the token in a cookie altogether.

davidangb:
updated spacing and variable naming. I left the obvious comment in place, because it's consistent with the stucture of that method - every other block is commented.

I also changed the expiration time to two minutes!

I don't think we can have a follow-on story to use the Authorization: header - at least not without significant work. The big problem here is that the download (or for files like PDFs, the opening) of the target file is in a `target="_blank"` new window. We don't have a way, due to browser limitations, to tell that new window to send an Authorization: header.

dvoet:
regarding the 2 minute expiration and starting the download - as downloads become larger it becomes more likely that they will fail which may lead us to multi-part downloads and then 2 minutes will be a problem... but that may be for another day. However, since the whole point of this is to download large files, maybe it is today.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -123,4 +127,39 @@ object HttpGoogleServicesDAO {
...
+    // use GoogleCredential.Builder to parse the private key from the pem file
+    val builder = new GoogleCredential.Builder()
+      .setServiceAccountPrivateKeyFromPemFile(new java.io.File(rawlsPemFile))
+    val privateKey = builder.getServiceAccountPrivateKey
+
+    // sign the string
+    val signature = java.security.Signature.getInstance("SHA256withRSA")
+    signature.initSign(privateKey)
+    signature.update(signableString.getBytes("UTF-8"))
+    val signedBytes = signature.sign()
```
dvoet, line 169: Is there no way to sign as the user? The obvious problem is security which you have jumped through some hoops deal with. The other problem is that this won't work if the rawls service account does not have access while the user does (protected TCGA data may fall into this category, I am not sure).

src/main/scala/org/broadinstitute/dsde/firecloud/service/CookieAuthedService.scala
```diff
@@ -37,16 +40,41 @@ trait CookieAuthedService extends HttpService with PerRequestCreator with FireCl
...
+            // exception to the user, just as if the user tried to access Google directly.
+            val objectReq = Get( HttpGoogleServicesDAO.getObjectResourceUrl(bucket, obj.toString) )
+            val objectPipeline = authHeaders(requestContext) ~> sendReceive
+
+            val userReq = Get( HttpGoogleServicesDAO.profileUrl )
+            val userPipeline = authHeaders(requestContext) ~> sendReceive
+
+            val objectStr = s"gs://$bucket/$obj"
+
+            userPipeline{userReq} map { userResponse =>
```
dmohs: Is the reason for checking for a valid user so we can log invalid access requests? If so, it would be helpful to see that in a comment too.

dmohs:
One comment suggestion. No need to re-review.

We could fetch the GCS URL and make that the `href` attribute of the link, so it would work in the target="_blank" case.

Re Doug's comment: Can we move forward with this and see how it goes? Even if it prevents downloading of certain files, not having the server crash is pretty important.

davidangb:
@dmohs, @dvoet : I've added functionality to check if the service account has permission to sign the url, and I've refactored. The second-to-last commit adds the functionality, and the last commit should be pure refactoring.

In the case where the user has permission, but the service account doesn't, we revert to using a direct link - that's the link that could potentially use the wrong identity.

David Mohs - we can't fetch the link and use it in the `href=`. In that model, we have no control over the delay between generating the link and using the link - which means our signed urls would be prone to expire.

dvoet:
👍

dmohs:
👍  Very reasonable handling of a difficult situation.

PR #252, opened 2016-09-15
GAWB-1063: isCurator passthrough

davidangb:
adds a new /api/library/user/role/curator API in orchestration (yeah, that's a wordy url), plus a simple test for the passthrough route definition.

as a rider, adds tests to validate JSON and JSON Schema for the previously-committed attribute definitions.
  - Authentication
  - Authorization
  - Encryption
  - Audit trails
- Review cycle:
  - LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -2024,6 +2024,29 @@ paths:
             - email
             - profile
             - https://www.googleapis.com/auth/devstorage.full_control
+  '/api/library/user/role/curator':
+    get:
+      responses:
+        '200':
+          description: User is a curator
+        '404':
+          description: User is not a curator
+        '500':
+          description: Rawls Internal Error
+          schema:
+            $ref: '#/definitions/ErrorReport'
+      description: "gets the curator status of a user"
+      tags:
+        - Profile
+      summary: Query to see if you're a library curator
+      operationId: isCurator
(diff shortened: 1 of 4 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -2859,3 +2882,52 @@ definitions:
...
+      - causes
+      - stackTrace
+    properties:
+      source:
+        type: string
+        description: service causing error
+      message:
+        type: string
+        description: what went wrong
+      statusCode:
```
skwalker, line 2900: I'm guessing it doesn't matter, but why is statusCode not required when source, message, causes, stackTrace all are?

skwalker (approved):
👍 looks good!

src/main/resources/swagger/api-docs.yaml
davidangb: Hm - good question. I copy/pasted this directly from rawls; since this endpoint is a passthrough, it's going to return whatever rawls does.

In Rawls, ErrorReport.statusCode is an `Option[StatusCode]` ... maybe because we can cause an error internally (e.g. database connection failure) without an HTTP status code?

skwalker: Hmm ok, I guess that makes sense

PR #253, opened 2016-09-19
Fix whitelist_sync regression

MatthewBemis:
Orch could no longer make the call to thurloe to grab the list of currently linked users because this header was dropped. No ticket because it was quicker to fix it.
  - Authentication
  - Authorization
  - Encryption
  - Audit trails
- Review cycle:
  - LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
@@ -367,7 +367,7 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
 
 object NIHWhitelistUtils extends FireCloudRequestBuilding {
   def getAllUserValuesForKey(key: String, userId: Option[String] = None)(implicit arf: ActorRefFactory, ec: ExecutionContext): Future[Map[String, String]] = {
-    val pipeline = addFireCloudCredentials ~> sendReceive
+    val pipeline = addAdminCredentials ~> addFireCloudCredentials ~> sendReceive
 
     val queryParams = userId match {
       case Some(sub) => Map("key"->key, "userId"->sub)


davidangb:
I'm all for fixing regressions, but ... if the pipeline in question is querying Thurloe, why does it need admin credentials?

MatthewBemis:
@davidangb It doesn't necessarily need _admin_ credentials, it just needs _some_ credentials, otherwise it gets 401'd. I was just putting back what we had before it was dropped here: https://github.com/broadinstitute/firecloud-orchestration/pull/249/files#diff-dfeeee9107bdde9bfc8b72cef89e6bebR370

I have no idea why I said rawls<->orch communication was the issue in the description. I looked at that as being the source of the issue too so I think I just got mixed up 😧

rushtong:
@MatthewBemis - that pipeline does have user credentials. It seems like that's not enough?

rushtong:
@MatthewBemis - I misread that, sorry. I think if it only needs _some_ credentials, we should add the user's.

davidangb:
There may be two use cases here: 1) the current user is linking with NIH and runs through this code path; 2) the whitelist cron is running. In the latter case, there is no user, right? Ideally we'd use the user's creds if we have them, and admin otherwise.

MatthewBemis:
@davidangb yes that sounds right to me. 

Do we want to take this PR as is since it gets us back to the original working state, and file a ticket to take the cleaner approach in the future? Or do we want to implement the ideal approach now?

davidangb:
I say accept as-is, file a future ticket. (edit: "say" not "saw")

davidangb:
( if it wasn't clear, previous comment was a 👍 )

MatthewBemis:
Ticket filed as GAWB-1071

PR #257, opened 2016-09-30
GAWB-1057:  add publish endpoint to orchestration

skwalker:
  - Authentication
  - Authorization
  - Encryption
  - Audit trails
- Review cycle:
  - LR reviews
- [x] **TL** sign off
- [x] **LR** sign off


src/main/scala/org/broadinstitute/dsde/firecloud/webservice/LibraryApiService.scala
```diff
@@ -49,15 +49,24 @@ trait LibraryApiService extends HttpService with FireCloudRequestBuilding
...
               }
             }
+          } ~
+          path(Segment / Segment / "set-is-published" ) { (namespace, name) =>
+            parameters('value) { value =>
+              post { requestContext =>
+                perRequest(requestContext,
+                  LibraryService.props(libraryServiceConstructor, userInfo),
+                  LibraryService.SetPublished(namespace, name, value.toBoolean))
+              }
```
davidangb, line 72: Use different request method - POST and DELETE, for instance - to denote publish vs. unpublish instead of passing a `value` parameter. It's a bit more RESTful that way.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/LibraryApiService.scala
```diff
@@ -49,15 +49,24 @@ trait LibraryApiService extends HttpService with FireCloudRequestBuilding
...
+          path(Segment / Segment / "metadata") { (namespace, name) =>
+            put {
+              entity(as[String]) { rawAttrsString => requestContext =>
+                perRequest(requestContext,
+                  LibraryService.props(libraryServiceConstructor, userInfo),
+                  LibraryService.UpdateAttributes(namespace, name, rawAttrsString.toString.parseJson))
               }
             }
+          } ~
+          path(Segment / Segment / "set-is-published" ) { (namespace, name) =>
```
davidangb: how about just "publish"

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -95,4 +97,20 @@ class LibraryService (protected val userInfo: UserInfo, val rawlsDAO: RawlsDAO)(
...
+  def setWorkspaceIsPublished(ns: String, name: String, value: Boolean): Future[PerRequestMessage] = {
+    rawlsDAO.getWorkspace(ns, name) flatMap { workspaceResponse =>
+      // verify owner on workspace
+      if (!workspaceResponse.accessLevel.contains("OWNER")) {
+        Future(RequestCompleteWithErrorReport(Forbidden, "must be an owner"))
+      } else {
+        val operations: Seq[AttributeUpdateOperation] = Seq()
+        // either add or remove the library attribute
+        if (value) operations :+ AddUpdateAttribute("library", AttributeString("published"))
+        else operations :+ RemoveAttribute("library")
```
davidangb: you're creating an empty Seq, then creating a new copy of that Seq with a value appended. It would be more efficient to create it once. Does this work?

```
val operations: Seq[AttributeUpdateOperation] = if (value)
  Seq(AddUpdateAttribute("library", AttributeString("published")))
else
  Seq(RemoveAttribute("library"))
```

also, add a TODO comment here - once we support boolean attributes, it would be better for the value to be true/false instead of the string "published". Unless you think we need different states for publishing, e.g. "pending, published, approved, etc"?

davidangb (commented):
some feedback inline, overall this is the right approach!

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/LibraryApiService.scala
dmohs: I agree with changing this since I agree with your next suggestion, but I suggested "published" because it seems confusing to me to POST (create) or DELETE a verb.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
skwalker: Creating one seq instead of doing it twice makes sense, thanks!

I think true/false sounds better to use for this case (especially if it's just a POST/DELETE with no additional parameters). 
Should there be a ticket for supporting boolean attributes?

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/LibraryApiService.scala
```diff
@@ -49,15 +49,27 @@ trait LibraryApiService extends HttpService with FireCloudRequestBuilding
...
+              perRequest(requestContext,
+                LibraryService.props(libraryServiceConstructor, userInfo),
+                LibraryService.SetPublishAttribute(namespace, name, true))
+            }
+            delete { requestContext =>
+              perRequest(requestContext,
+                LibraryService.props(libraryServiceConstructor, userInfo),
+                LibraryService.SetPublishAttribute(namespace, name, false))
+            }
+          }
```
davidangb, line 74: these three endpoints could be consolidated a bit, moving the namespace/name to the outside. Something like:

```
path(Segment / Segment) { (namespace, name) =>
  path("metadata") {
    put {
      ...
    }
  } ~
  path("published") {
    post {
      ...
    } ~
    delete {
      ...
    }
  }
}


```

src/main/resources/swagger/api-docs.yaml
```diff
@@ -2084,6 +2084,73 @@ paths:
             - email
             - profile
             - https://www.googleapis.com/auth/devstorage.full_control
+  '/api/library/{workspaceNamespace}/{workspaceName}/published':
+    put:
```
dmohs: put?

src/main/resources/swagger/api-docs.yaml
```diff
@@ -2084,6 +2084,73 @@ paths:
...
+          description: Workspace Namespace
+          required: true
+          type: string
+          in: path
+        - name: workspaceName
+          description: Workspace Name
+          required: true
+          type: string
+          in: path
+      responses:
```
dmohs, line 2105: Looks like this can return a "forbidden" as well. Also, since pretty much any endpoint can return a 500, is that worth including in our API docs for every single URL?

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/LibraryApiService.scala
```diff
@@ -49,15 +49,27 @@ trait LibraryApiService extends HttpService with FireCloudRequestBuilding
...
+          path(Segment / Segment / "metadata") { (namespace, name) =>
+            put {
+              entity(as[String]) { rawAttrsString => requestContext =>
+                perRequest(requestContext,
+                  LibraryService.props(libraryServiceConstructor, userInfo),
+                  LibraryService.UpdateAttributes(namespace, name, rawAttrsString.toString.parseJson))
               }
             }
+          } ~
+          path(Segment / Segment / "published" ) { (namespace, name) =>
```
dmohs: nit: spacing

dmohs (approved):
Looks good to me.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -42,11 +43,13 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
 
   override def receive = {
     case UpdateAttributes(ns: String, name: String, attrs: JsValue) => asCurator {updateAttributes(ns, name, attrs)} pipeTo sender
+    case SetPublishAttribute(ns: String, name: String, value: Boolean) => asCurator {setWorkspaceIsPublished(ns, name, value)} pipeTo sender
```
davidangb, line 46: see #261 ; this needs to be `pipeTo context.parent`. I realize you are just copying existing code.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
```diff
@@ -37,4 +37,13 @@ trait LibraryServiceSupport {
     (removeOperations ++ updateOperations)
   }
 
+  def updatePublishAttribute(value: Boolean): Seq[AttributeUpdateOperation] = {
+    // TODO: publish attribute can just be a boolean once we support boolean attributes
+    val operations: Seq[AttributeUpdateOperation] =
+    if (value) Seq(AddUpdateAttribute("library:published", AttributeString("true")))
+    else Seq(RemoveAttribute("library:published"))
+
+    operations
```
davidangb: fewer lines of code: this could be just

```
  def updatePublishAttribute(value: Boolean): Seq[AttributeUpdateOperation] = {
    // TODO: publish attribute can just be a boolean once we support boolean attributes
    if (value) Seq(AddUpdateAttribute("library:published", AttributeString("true")))
    else Seq(RemoveAttribute("library:published"))
  }
```

davidangb (commented):
two small changes inline, one (sender -> context.parent) is pretty important. Otherwise looks good!

If these are the only changes you make, go ahead and squash&merge. If you make other changes, I'll re-review.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
davidangb: update: uhhh, nope, I was wrong. sender is fine - please ignore.

PR #269, opened 2016-10-17
Namespace Permissions Passthroughs

rushtong:
This PR addresses [GAWB-156](https://broadinstitute.atlassian.net/browse/GAWB-156). 

We've added parallel calls to configurations/methods to get namespace permissions for each type of Agora entity. We need two basic calls to support the UI:
- `GET` ACLs for a specific namespace
- `POST` ACLs for a specific namespace

Agora also provides support for `DELETE` and `PUT` but those are unneeded - the `POST` endpoint handles insert, update, and delete. It should likely be changed to PUT since that's what PUT is supposed to do.

The individual commits are a mess - I suggest looking at the end result.

---
  - Authentication
  - Authorization
  - Encryption
  - Audit trails
- Review cycle:
  - LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -356,6 +356,138 @@ paths:
             - email
             - profile
             - https://www.googleapis.com/auth/devstorage.full_control
+  /api/methods/{namespace}/permissions:
+    get:
+      tags:
+        - Method Repository
+      operationId: getMethodNamespaceACL
+      summary: get ACL permissions on a Method Repository Method Namespace
+      produces:
+        - application/json
+      parameters:
+        - name: namespace
+          description: Method namespace
+          required: true
+          type: string
+          in: path
+      responses:
+        200:
+          description: List of indicated method namespace ACL permissions
+          schema:
+            $ref: '#/definitions/MethodConfigACL'
+        403:
+          description: user does not have correct permissions to manage namespace permissions or the indicated namespace was not found
+        500:
+          description: Internal Server Error
+      security:
+        - googleoauth:
+            - openid
+            - email
+            - profile
+            - https://www.googleapis.com/auth/devstorage.full_control
+    post:
+      tags:
+        - Method Repository
+      operationId: setMethodNamespaceACL
+      summary: set ACL permissions on a Method Repository Method Namespace
+      produces:
+        - application/json
+      parameters:
+        - name: namespace
+          description: Method namespace
+          required: true
+          type: string
+          in: path
+        - name: payload
+          description: the ACLs to upsert
+          required: true
+          in: body
+          schema:
+            $ref: '#/definitions/MethodConfigACL'
+      responses:
+        200:
+          description: the indicated method namespace ACL permissions once the posted changes have been applied
+          schema:
+            $ref: '#/definitions/MethodConfigACL'
+        400:
+          description: Invalid content format
+        403:
+          description: User does not have correct permissions to manage namespace permissions or the indicated namespace was not found
+        500:
+          description: Internal Server Error
+      security:
+        - googleoauth:
+            - openid
+            - email
+            - profile
+            - https://www.googleapis.com/auth/devstorage.full_control
+  /api/configurations/{namespace}/permissions:
+    get:
+      tags:
+        - Method Repository
+      operationId: getConfigNamespaceACL
+      summary: get ACL permissions on a Method Repository Configuration Namespace
+      produces:
+        - application/json
+      parameters:
+        - name: namespace
+          description: Method Configuration namespace
+          required: true
+          type: string
+          in: path
+      responses:
+        200:
+          description: List of the indicated configuration namespace ACL permissions
+          schema:
+            $ref: '#/definitions/MethodConfigACL'
+        403:
+          description: user does not have correct permissions to manage namespace permissions or the indicated namespace was not found
+        500:
+          description: Internal Server Error
+      security:
+        - googleoauth:
+            - openid
+            - email
+            - profile
+            - https://www.googleapis.com/auth/devstorage.full_control
+    post:
+      tags:
+        - Method Repository
+      operationId: setConfigNamespaceACL
+      summary: set ACL permissions on a Method Repository Configuration Namespace
+      produces:
+        - application/json
+      parameters:
+        - name: namespace
+          description: Method Configuration namespace
+          required: true
+          type: string
+          in: path
+        - name: payload
+          description: the ACLs to upsert
+          required: true
+          in: body
+          schema:
+            $ref: '#/definitions/MethodConfigACL'
+      responses:
+        200:
+          description: the indicated configuration ACL once the posted changes have been applied
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/MethodConfigACL'
+        400:
+          description: Invalid content format
+        403:
+          description: User does not have correct permissions to manage namespace permissions or the indicated namespace was not found
+        500:
+          description: Internal Server Error
+      security:
+        - googleoauth:
+            - openid
+            - email
+            - profile
+            - https://www.googleapis.com/auth/devstorage.full_control
   /api/workspaces/{workspaceNamespace}/{workspaceName}:
     get:
       tags:
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpAgoraDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpAgoraDAO.scala
@@ -1,0 +1,29 @@
+package org.broadinstitute.dsde.firecloud.dataaccess
+
+import akka.actor.ActorSystem
+import com.typesafe.config.Config
+import org.broadinstitute.dsde.firecloud.FireCloudConfig
+import org.broadinstitute.dsde.firecloud.model.MethodRepository.AgoraPermission
+import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
+import org.broadinstitute.dsde.firecloud.model.UserInfo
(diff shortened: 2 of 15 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/model/MethodRepository.scala
```diff
@@ -62,9 +62,9 @@ object MethodRepository {
     val ListAll = List("All")
   }
 
-  def validateEmail(email:String): Boolean = {
+  def validatePublicOrEmail(email:String): Boolean = {
     // TODO: real email address validation!
-    !email.trim.isEmpty
+    email.trim.equalsIgnoreCase("public") || !email.trim.isEmpty
```
davidangb: isn't `!email.trim.isEmpty` a superset of `email.trim.equalsIgnoreCase("public")`?

rushtong: Why yes, yes it is :-)

src/main/scala/org/broadinstitute/dsde/firecloud/service/NamespaceService.scala
```diff
@@ -0,0 +1,74 @@
...
+import spray.httpx.SprayJsonSupport
+import spray.httpx.SprayJsonSupport._
+import spray.json.DefaultJsonProtocol._
+import spray.json._
+import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
+
+import scala.concurrent.{ExecutionContext, Future}
+
+object NamespaceService {
+  case class GetPermissions(ns: String, entity: String)
```
skwalker: Should this be part of a sealed trait like in library service?

skwalker:
👍  minor question, but overall looks great!

src/main/scala/org/broadinstitute/dsde/firecloud/service/NamespaceService.scala
rushtong: Probably a good idea. That lets the compiler know if some cases are not handled within a match block.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -356,6 +356,138 @@ paths:
...
+        403:
+          description: User does not have correct permissions to manage namespace permissions or the indicated namespace was not found
+        500:
+          description: Internal Server Error
+      security:
+        - googleoauth:
+            - openid
+            - email
+            - profile
+            - https://www.googleapis.com/auth/devstorage.full_control
```
davidangb, line 490: this may be a dumb question, but ... are method namespaces and config namespaces different things? Do we need separate endpoints? Could we have a single endpoint like `/api/namespace/{namespace}/permissions`?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpAgoraDAO.scala
```diff
@@ -0,0 +1,28 @@
...
+import spray.json.DefaultJsonProtocol._
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class HttpAgoraDAO(implicit val system: ActorSystem, implicit val executionContext: ExecutionContext)
+  extends AgoraDAO with RestJsonClient {
+
+  private def getNamespaceUrl(ns: String, entity: String): String = {
+    s"${FireCloudConfig.Agora.authUrl}/$entity/$ns/permissions"
+  }
```
davidangb, line 21: Instead of HttpAgoraDAO reading from FireCloudConfig, what if HttpAgoraDAO accepted a Config object as part of its constructor, and FireCloudServiceActor passed the config to it? (push-injection vs. pull)

I realize I didn't do this as part of HttpRawlsDAO, and I know you modeled this after my work.

This is just a thought - not necessary to resolve for this PR.

src/main/scala/org/broadinstitute/dsde/firecloud/model/MethodRepository.scala
```diff
@@ -62,9 +62,9 @@ object MethodRepository {
     val ListAll = List("All")
   }
 
-  def validateEmail(email:String): Boolean = {
+  def validatePublicOrEmail(email:String): Boolean = {
     // TODO: real email address validation!
-    !email.trim.isEmpty
+    !email.trim.isEmpty || email.trim.equalsIgnoreCase("public")
```
davidangb: per previous comment ... leave off the second part of this.

if `!email.trim.isEmpty` is false, `email.trim.equalsIgnoreCase("public")` can never be true, and it's just extra work for the jvm to calculate.

src/test/scala/org/broadinstitute/dsde/firecloud/service/BaseServiceSpec.scala
```diff
@@ -0,0 +1,13 @@
...
+import org.broadinstitute.dsde.firecloud.dataaccess._
+
+class BaseServiceSpec extends ServiceSpec {
+
+  val agoraDAO:AgoraDAO = new MockAgoraDAO
+  val rawlsDAO:RawlsDAO = new MockRawlsDAO
+  val searchDAO:SearchDAO = new MockSearchDAO
+  val app:Application = new Application(agoraDAO, rawlsDAO, searchDAO)
+
+}
```
davidangb, line 13: nice!

src/test/scala/org/broadinstitute/dsde/firecloud/service/NamespaceApiServiceSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+
+    "when calling GET on a namespace permissions path" - {
+      "a valid list of FireCloud permissions is returned" in {
+        urls map {
+          url =>
+            Get(url) ~> dummyUserIdHeaders("1234") ~> sealRoute(namespaceRoutes) ~> check {
+              status should equal(OK)
+              val permission = responseAs[List[FireCloudPermission]]
+              permission shouldNot be (None)
+            }
```
davidangb, line 35: because you control the response inside MockAgoraDAO, you can - and I think should - make the test more strict:

```
            Get(url) ~> dummyUserIdHeaders("1234") ~> sealRoute(namespaceRoutes) ~> check {
              status should equal(OK)
              val permission = responseAs[List[FireCloudPermission]]
              permission should be (Seq(FireCloudPermission("test-user@broadinstitute.org", ACLNames.Owner)))
            }
```

you could also make the test values in MockAgoraDAO public and read them, instead of copying/hardcoding them here.

src/main/scala/org/broadinstitute/dsde/firecloud/service/NamespaceService.scala
```diff
@@ -0,0 +1,76 @@
...
+    val agoraPermissionsPosted = agoraDAO.postNamespacePermissions(ns, entity, agoraPermissionsToPost)
+    delegatePermissionsResponse(agoraPermissionsPosted)
+  }
+
+  private def delegatePermissionsResponse(agoraPerms: Future[List[AgoraPermission]]): Future[PerRequestMessage] = {
+    agoraPerms map {
+      perms =>
+        val fcPermissions = convertPermissions(perms)
+        RequestComplete(OK, fcPermissions)
+    } recover {
```
davidangb, line 60: you could make this a lot terser:

```
    agoraPerms map {
      perms =>
        RequestComplete(OK, perms map AgoraPermissionHandler.toFireCloudPermission)
    } recover {
```

and then delete `convertPermissions` entirely.

Totally willing to hear arguments about readability.

davidangb (changes requested):
see inline comments - otherwise works well and looks good!

src/main/resources/swagger/api-docs.yaml
rushtong: They are different endpoints in Agora - but I don't know how real that is. I can add a story to investigate and consolidate if possible.

src/main/scala/org/broadinstitute/dsde/firecloud/model/MethodRepository.scala
rushtong: I really didn't like getting rid of the intention behind the method so I added an email check and tests to cover the different cases.

src/main/scala/org/broadinstitute/dsde/firecloud/service/NamespaceService.scala
rushtong: You're right - there was a lot of refactoring to handle errors and convertPermissions was doing a lot more before. It should be removed altogether.

src/test/scala/org/broadinstitute/dsde/firecloud/service/NamespaceApiServiceSpec.scala
rushtong: 👍

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpAgoraDAO.scala
rushtong: You did implement that in the ES dao - it makes sense that we should continue the pattern.

rushtong: On second thought, loading a config would mean having to deconstruct that config the same way we're doing it in FireCloudConfig, but again in the dao class. I can see the case for constructing it with specific elements, but not the whole config.

rushtong:
@davidangb - ready for re-review

davidangb:
See #275 - those are my comments on this PR. Assuming you accept that PR, 👍 to merge! If you don't like #275, let's discuss.

davidangb (approved):

PR #270, opened 2016-10-17
GAWB-1122: library validation

davidangb:
This PR enables schema validation for the library update-metadata endpoint. When this PR is merged, the UI library wizard will start reporting errors, because the wizard doesn't send anything other than strings.

At a high level, this PR does the following:
- adds the "library:" prefix to the schema definition.
- refactors how we read the schema from disk
- adds json-schema validation to the put-library-metadata endpoint
- adds a bunch of unit tests

In the unit tests, I work against a copy of our current schema. This allows us to change the runtime schema without having to keep the unit tests updated.

This PR needs to be merged in sync with broadinstitute/firecloud-ui#530, otherwise the UI will be adding "library:library:foo" attributes.

The first batch of separate commits in this PR are decent, if you want to read them individually. The last three are messier cleanup after manually rebasing on top of develop.
  - Authentication
  - Authorization
  - Encryption
  - Audit trails
- Review cycle:
  - LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
@@ -5,11 +5,14 @@ import akka.pattern._
 import com.typesafe.scalalogging.slf4j.LazyLogging
 import org.broadinstitute.dsde.firecloud.Application
 import org.broadinstitute.dsde.firecloud.dataaccess.{RawlsDAO, SearchDAO}
+
 import org.broadinstitute.dsde.firecloud.model.Attributable.AttributeMap
-import org.broadinstitute.dsde.firecloud.model.{AttributeName, Document, RequestCompleteWithErrorReport, UserInfo}
+import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol.impRawlsWorkspace
+import org.broadinstitute.dsde.firecloud.model.{AttributeName, Document, ErrorReport, RequestCompleteWithErrorReport, UserInfo}
 import org.broadinstitute.dsde.firecloud.service.LibraryService._
 import org.broadinstitute.dsde.firecloud.service.PerRequest.{PerRequestMessage, RequestComplete}
 import org.broadinstitute.dsde.firecloud.utils.RoleSupport
+import org.everit.json.schema.ValidationException
 import org.slf4j.LoggerFactory
 import spray.http.StatusCodes._
 import spray.httpx.SprayJsonSupport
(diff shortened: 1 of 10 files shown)


src/test/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSpec.scala
```diff
@@ -267,5 +294,87 @@ class LibraryServiceSpec extends FreeSpec with LibraryServiceSupport {
...
       }
     }
+    "in its runtime schema definition" - {
+      "has valid JSON" in {
+        val fileContents = FileUtils.readAllTextFromResource("library/attribute-definitions.json")
+        val jsonVal:Try[JsValue] = Try(fileContents.parseJson)
+        assert(jsonVal.isSuccess, "Schema should be valid json")
+      }
+      "has valid JSON Schema" in {
+        val schemaStream = new URL("http://json-schema.org/draft-04/schema").openStream()
```
rushtong, line 304: I worry that referencing an external, unmanaged resource will introduce some flakiness to the test. I suppose if it becomes an issue, we can add a local copy of the schema to use.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -61,15 +62,24 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
...
+              ve.getMessage + ". See causes for details."
+            else
+              ve.getMessage
+            Future(RequestCompleteWithErrorReport(BadRequest, userMsg, errorReports))
+          case Failure(e) =>
+            Future(RequestCompleteWithErrorReport(BadRequest, BadRequest.defaultMessage, e))
+          case Success(x) => {
+            rawlsDAO.getWorkspace(ns, name) flatMap { workspaceResponse =>
+              // this is technically vulnerable to a race condition in which the workspace attributes have changed
+              // between the time we retrieved them and here, where we update them.
```
rushtong, line 79: From what I can grep, a second update to attributes will also result in a call to this actor method. If that's the case, what you'll get is two update actors trying to post to ES at roughly the same time so it's a coin toss as two which one is the last one. I think this will be true of _any_ actor-based implementation.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
```diff
@@ -51,4 +53,16 @@ trait LibraryServiceSupport {
     Document(workspace.workspaceId, attrfields ++ idfields)
   }
 
+  def defaultSchema: String = FileUtils.readAllTextFromResource(LibraryService.schemaLocation)
+
+  def schemaValidate(data: String): Unit = validateJsonSchema(data, defaultSchema)
+  def schemaValidate(data: JsObject): Unit = validateJsonSchema(data.compactPrint, defaultSchema)
+
+  def validateJsonSchema(data: String, schemaStr: String): Unit = {
```
rushtong, line 61: Question: do we want to declare that this throws `ValidationException`? We don't do this at all now so easy answer is no.

rushtong:
👍  LG. Minor comments - nothing to hold this up. I don't have curator permissions so can't test this directly.

PR #271, opened 2016-10-18
GAWB-1115 Use new serialization scheme for Rawls list attributes

helgridly:
This has a sibling PR in Rawls: https://github.com/broadinstitute/rawls/pull/547

A quick summary of these changes:
- Pulled over the entire Rawls attribute hierarchy and serialization scheme (TODO: put this into a library shared by both codebases)
- Changed `attributes` field on workspaces and entities to be `type AttributeMap = Map[AttributeName, Attribute]`
- Changes to TSV import to use new `CreateAttributeEntityReferenceList` entity-update endpoint
  - Authentication
  - Authorization
  - Encryption
  - Audit trails
- Review cycle:
  - LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
@@ -14,6 +14,7 @@ import spray.httpx.unmarshalling._
 import spray.httpx.SprayJsonSupport._
 import spray.json.DefaultJsonProtocol._
 
+import scala.Option
 import scala.concurrent.{ExecutionContext, Future}
 
 /**
@@ -55,13 +56,20 @@ class HttpRawlsDAO( implicit val system: ActorSystem, implicit val executionCont
   override def getAllLibraryPublishedWorkspaces: Future[Seq[RawlsWorkspace]] = {
     val adminToken = HttpGoogleServicesDAO.getAdminUserAccessToken
 
+    def isPublished(rw:RawlsWorkspace) = {
+      rw.attributes.get(LibraryService.publishedFlag) match {
+        case Some(AttributeBoolean(true)) => true
+        case _ => false
+      }
+    }
+
     val allPublishedPipeline = addCredentials(OAuth2BearerToken(adminToken)) ~> sendReceive
     allPublishedPipeline(Get(rawlsAdminWorkspaces)) map {response =>
       response.entity.as[Seq[RawlsWorkspace]] match {
         case Right(srw) =>
           logger.info("admin workspace list got: " + srw.length + " raw workspaces")
           val published = srw.collect {
-            case rw:RawlsWorkspace if rw.attributes.getOrElse(LibraryService.publishedFlag, AttributeBoolean(false)).asInstanceOf[AttributeBoolean].value => rw
+            case rw:RawlsWorkspace if isPublished(rw) => rw
           }
           logger.info("admin workspace list collected: " + published.length + " published workspaces")
           published
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
@@ -61,7 +61,7 @@ class HttpRawlsDAO( implicit val system: ActorSystem, implicit val executionCont
         case Right(srw) =>
           logger.info("admin workspace list got: " + srw.length + " raw workspaces")
           val published = srw.collect {
-            case rw:RawlsWorkspace if rw.attributes.getOrElse(LibraryService.publishedFlag, "false").toBoolean => rw
+            case rw:RawlsWorkspace if rw.attributes.getOrElse(LibraryService.publishedFlag, AttributeBoolean(false)).asInstanceOf[AttributeBoolean].value => rw
           }
           logger.info("admin workspace list collected: " + published.length + " published workspaces")
           published
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/model/Attributes.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/model/Attributes.scala
@@ -38,8 +38,9 @@ object AttributeName {
 }
 
 sealed trait Attribute
-sealed trait AttributeValue extends Attribute
-sealed trait AttributeList[T <: Attribute] extends Attribute { val list: Seq[T] }
+sealed trait AttributeListElementable extends Attribute //terrible name for "this type can legally go in an attribute list"
+sealed trait AttributeValue extends AttributeListElementable
+sealed trait AttributeList[T <: AttributeListElementable] extends Attribute { val list: Seq[T] }
 case object AttributeNull extends AttributeValue
 case class AttributeString(val value: String) extends AttributeValue
 case class AttributeNumber(val value: BigDecimal) extends AttributeValue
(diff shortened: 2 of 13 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
```diff
@@ -61,7 +61,7 @@ class HttpRawlsDAO( implicit val system: ActorSystem, implicit val executionCont
         case Right(srw) =>
           logger.info("admin workspace list got: " + srw.length + " raw workspaces")
           val published = srw.collect {
-            case rw:RawlsWorkspace if rw.attributes.getOrElse(LibraryService.publishedFlag, "false").toBoolean => rw
+            case rw:RawlsWorkspace if rw.attributes.getOrElse(LibraryService.publishedFlag, AttributeBoolean(false)).asInstanceOf[AttributeBoolean].value => rw
```
davidangb: can this now throw a casting error on `asInstanceOf`, where the previous `String.toBoolean` could not?

src/main/scala/org/broadinstitute/dsde/firecloud/model/Attributes.scala
```diff
@@ -0,0 +1,52 @@
...
+      case sName :: Nil => AttributeName.withDefaultNS(sName)
+      case sNamespace :: sName :: Nil => AttributeName(sNamespace, sName)
+      case _ => throw new FireCloudException(s"Attribute string $dName has too many '$delimiter' delimiters")
+    }
+  }
+}
+
+sealed trait Attribute
+sealed trait AttributeValue extends Attribute
+sealed trait AttributeList[T <: Attribute] extends Attribute { val list: Seq[T] }
```
davidangb: I realize you copied this from rawls, but ... should this be: `AttributeList[T <: AttributeValue]`?

Otherwise, we technically support lists-of-lists. Do we do that? Do we need T <: Attribute in order to support AttributeEntityReference, and if so is there another way to do it where we _don't_ support lists-of-lists?

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala
```diff
@@ -9,45 +9,86 @@ import spray.json.DefaultJsonProtocol._
...
+  }
 
-  implicit val impDestination = jsonFormat3(MethodConfigurationId)
-  implicit val impMethodConfigurationCopy = jsonFormat4(MethodConfigurationCopy)
-  implicit val impConfigurationCopyIngest = jsonFormat5(CopyConfigurationIngest)
-  implicit val impMethodConfigurationPublish = jsonFormat3(MethodConfigurationPublish)
-  implicit val impPublishConfigurationIngest = jsonFormat4(PublishConfigurationIngest)
+  def writeAttributeList[T <: Attribute](listType: String, list: Seq[T]): JsValue = {
+    JsObject( Map(LIST_ITEMS_TYPE_KEY -> JsString(listType), LIST_ITEMS_KEY -> JsArray(list.map( writeAttribute ).toSeq:_*)) )
+  }
```
davidangb, line 75: should this verify that `listType` is one of `VALUE_LIST_TYPE` or `REF_LIST_TYPE`?

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala
```diff
@@ -9,45 +9,86 @@ import spray.json.DefaultJsonProtocol._
...
+
+  def writeAttribute(obj: Attribute): JsValue
+  def readAttribute(json: JsValue): Attribute
+}
 
-  implicit val impMethod = jsonFormat8(MethodRepository.Method)
-  implicit val impConfiguration = jsonFormat9(MethodRepository.Configuration)
+//Serializes attribute lists to e.g. [1,2,3]
+//When reading in JSON, assumes that [] is an empty value list, not an empty ref list
+trait PlainArrayAttributeListSerializer extends AttributeListSerializer {
```
davidangb, line 25: `PlainArrayAttributeListSerializer` is never used as far as I can see. Is this the one orchestration would need to talk to cromwell? Maybe we only pull this in once we realize we need it?

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
```diff
@@ -1,59 +1,52 @@
 package org.broadinstitute.dsde.firecloud.service
 
-import org.broadinstitute.dsde.firecloud.model.{AttributeString, Document, RawlsWorkspace}
+import org.broadinstitute.dsde.firecloud.model._
 import org.broadinstitute.dsde.firecloud.model.AttributeUpdateOperations.{AddListMember, AddUpdateAttribute, AttributeUpdateOperation, RemoveAttribute}
-import spray.json.{JsArray, JsObject, JsString, JsValue}
+import org.broadinstitute.dsde.firecloud.model.Attributable.AttributeMap
+import spray.json.{JsArray}
```
davidangb: nit: remove braces

davidangb (commented):
A few inline comments. I have not done any functional testing.

I'd also love to see some unit tests that accept a blob of json, unmarshal to an AttributeMap, and then verify the AttributeMap looks correct.

src/main/scala/org/broadinstitute/dsde/firecloud/model/Attributes.scala
dvoet: we don't support lists of lists

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala
helgridly: Yes, this is the one to talk to Cromwell. It got pulled in with the other one. I'm _slightly_ in favour of leaving it here so the code is line-for-line identical, but not that much in favour if you have stronger objections.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
helgridly: what do you want the behaviour to be here? i can do either:
a) nothing
b) throw an exception if the value isn't a bool
c) auto-convert strings "true" or "false" to bools (be liberal in what you accept), and throw an exception if the string is otherwise or the value is e.g. a number

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala
helgridly: it's only ever called from `writeListType`, where that's always true. and there are going to be other types in future (think: sets), so i'd say no

src/main/scala/org/broadinstitute/dsde/firecloud/model/Attributes.scala
helgridly: > should this be: `AttributeList[T <: AttributeValue]`?

no

> Do we [support lists-of-lists]?

no

> Do we need T <: Attribute in order to support AttributeEntityReference  

yes

> if so is there another way to do it where we don't support lists-of-lists  

i guess we could have an intermediate `sealed trait AttributeTypeThatCanGoInLists`... i'll let @dvoet make a call on whether that's madness

helgridly: he called it not madness. trying now to see if it's free

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
davidangb: This is lame-duck code anyway and will be removed once broadinstitute/rawls#552 lands, so don't put too much effort into it - therefore, I suggest taking the easiest route. But my desired functionality would be that if the Attribute is an AttributeBoolean(true), then true; if the Attribute is anything else (e.g. AttributeBoolean(false), AttributeString, AttributeNumber, etc) then false.

helgridly: OK. I will _not_ convert `AttributeString("true")` to `true`, then, and you'll have to re-publish any workspaces published with the stringy version of the tag.

helgridly:
@davidangb I've done a bunch of work on this, so back to you again.

davidangb:
👍 looks good to me, modulo:
- accepting my work on unit tests and seeing everything pass
- any additional changes on the rawls side that cause any changes here.

PR #278, opened 2016-10-25
GAWB-972: programmatically set/update mappings

skwalker:
  - Authentication
  - Authorization
  - Encryption
  - Audit trails
- Review cycle:
  - LR reviews
- [ ] **TL** sign off
- [x] **LR** sign off


src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOSupport.scala
```diff
@@ -39,4 +42,20 @@ trait ElasticSearchDAOSupport extends LazyLogging {
...
+    val maps = definition.properties flatMap { case (label: String, detail: AttributeDetail) =>
+      detailFromAttribute(label, detail)
+    }
+    ESDatasetProperty(maps).toJson.prettyPrint
+  }
+
+  def detailFromAttribute(label: String, detail: AttributeDetail) = {
+    detail match {
+      // weirdness for when we have an array type: https://www.elastic.co/guide/en/elasticsearch/reference/1.4/mapping-array-type.html
+      case AttributeDetail(t, Some(items)) => assert(t == "array"); Map(label -> Right(ESArray(ESItem(ESDetail(items.`type`)))))
```
davidangb: simplify this to:

```
case AttributeDetail("array", Some(items))
```

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOSupport.scala
```diff
@@ -39,4 +42,20 @@ trait ElasticSearchDAOSupport extends LazyLogging {
     }
   }
 
-}
+  def makeMapping(attributeJson: String): String = {
+    val definition = attributeJson.parseJson.convertTo[AttributeDefinition]
+    val maps = definition.properties flatMap { case (label: String, detail: AttributeDetail) =>
+      detailFromAttribute(label, detail)
+    }
```
davidangb, line 49: could you use `map` instead of `flatMap` here, and then have the `detailFromAttribute` cases return `(label -> value)` instead of `Map(label -> value)`?

In other words, you're currently creating a map of maps, then flattening it out; instead, you could just transform the entries.

src/main/scala/org/broadinstitute/dsde/firecloud/model/ESMapping.scala
```diff
@@ -0,0 +1,12 @@
+package org.broadinstitute.dsde.firecloud.model
+
+case class AttributeDefinition(properties: Map[String, AttributeDetail])
+case class AttributeDetail(`type`: String, items: Option[AttributeDetail]=None)
+
+case class ESMapping(mappings: ESDataset)
+case class ESDataset(dataset: ESDatasetProperty)
+case class ESDatasetProperty(properties: Map[String, Either[ESDetail, ESArray]])
```
davidangb: I'm confused about the need for `ESDataset` here. Could `ESMapping` just contain `properties: Map[String, Either[ESDetail, ESArray]]`?

Further: if you had a common parent for ESDetail and ESArray, something like:

```
sealed trait ESProperty
case class ESArray(properties: ESItem) extends ESProperty
case class ESDetail(`type`: String) extends ESProperty
```

then you could get rid of the use of `Either`, and just have a `Map[String, ESProperty]`.

Finally: we may not need the `ESArray` type at all. See https://www.elastic.co/guide/en/elasticsearch/reference/current/array.html ... I don't think that ES needs to be told that the field is an array.

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala
```diff
@@ -241,6 +241,17 @@ object ModelJsonProtocol {
...
+  implicit val AttributeDetailFormat: RootJsonFormat[AttributeDetail] = rootFormat(lazyFormat(jsonFormat2(AttributeDetail.apply)))
+  implicit val AttributeDefinitionFormat = jsonFormat1(AttributeDefinition.apply)
+
+
+  implicit val ESDetailFormat = jsonFormat1(ESDetail.apply)
+  implicit val ESItemFormat = jsonFormat1(ESItem.apply)
+  implicit val ESArrayFormat = jsonFormat1(ESArray.apply)
+  implicit val ESDatasetPropertyFormat = jsonFormat1(ESDatasetProperty.apply)
+  implicit val ESDatasetFormat = jsonFormat1(ESDataset.apply)
+  implicit val ESMappingFormat = jsonFormat1(ESMapping.apply)
```
davidangb: do you need the `.apply` in these? We don't need them elsewhere.

davidangb (changes requested):
see comments inline!

davidangb (approved):

davidangb:
👍 looks good to me, pending Andrea's review!

davidangb:
This is now obsolete, superseded by #284, and should be closed/abandoned

PR #280, opened 2016-10-25
GAWB-1119: use plain-array deserialization for library metadata update

davidangb:
part of GAWB-1119, but doesn't necessarily fix it all the way.

The library metadata-update API accepts plain json - not the new attribute list format. It does this because:
- we can then take advantage of json schema validation, since metadata is defined in json schema
- we don't need to support entity references

The default deserializer in ModelJsonProtocol uses the new attribute list format. To fix this, I needed to bring a new implicit into scope that uses the plain json format instead.
  - Authentication
  - Authorization
  - Encryption
  - Audit trails
- Review cycle:
  - LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
@@ -5,10 +5,8 @@ import akka.pattern._
 import com.typesafe.scalalogging.slf4j.LazyLogging
 import org.broadinstitute.dsde.firecloud.Application
 import org.broadinstitute.dsde.firecloud.dataaccess.{RawlsDAO, SearchDAO}
-
 import org.broadinstitute.dsde.firecloud.model.Attributable.AttributeMap
-import org.broadinstitute.dsde.firecloud.model.{AttributeName, Document, ErrorReport, RequestCompleteWithErrorReport, UserInfo}
-import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol.impRawlsWorkspace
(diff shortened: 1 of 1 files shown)


helgridly:
I think we should be consistent in our representation of attributes, so I'm 😢 on using the plain array representation in some cases and not others. My initial reaction is "update the json schema" but I don't have the context to know if that's possible/easy.

davidangb:
1) This fixes a runtime error in an existing endpoint, so I'm pushing to merge this PR as-is (with any functional/syntax changes, of course), then have a followup conversation about API signature if needed.

2) I don't think we should be consistent here. I've been trying to keep Library datasets and FireCloud workspaces isolated, and treat the APIs differently. If we had made Library a standalone service, this would be less of a concern, but today ... I'd prefer to treat it as a coincidence that Library dataset metadata is backed by workspace attributes, and therefore not tie the two APIs together. Library dataset metadata uses standard json for updates; workspace attributes use an array of operations, or our own customized json.

( this may be better discussed in person )

skwalker:
Fixes the 400 errors we're getting trying to catalog a dataset so I'm ok with it 👍

helgridly:
> I've been trying to keep Library datasets and FireCloud workspaces isolated, and treat the APIs differently

I think this is probably the philosophical difference underpinning our disagreement. Since it's a user-facing issue I'm going to bow out because such decisions are for POs.

rushtong:
👍 Code is saving and returning the right content. UI still needs to fix the display, but this part is working well.

rushtong:
Got the 👍 from hipchat to merge.

PR #281, opened 2016-10-25
clean up futures

davidangb:
Does this work better for you? Since the test failures were intermittent, I found it hard to verify that this really fixed it, unless I ran tests over and over and over ...

btw if you didn't already know,

```
sbt "test-only *LibraryApiServiceSpec"
```

might be your friend.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
@@ -83,24 +83,13 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
         Future(RequestCompleteWithErrorReport(Forbidden, "must be an owner"))
       } else {
         val operations = updatePublishAttribute(value)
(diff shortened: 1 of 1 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -83,24 +83,13 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
...
+          else
             searchDAO.deleteDocument(newws.workspaceId)
-          }
-          //wsfuture
-          true
+          RequestComplete(newws)
         }
-//        println(response)
-//        Future(RequestComplete(wsfuture))
-        wsfuture map (RequestComplete(_))
```
davidangb, line 103: we discussed this in person, but using `wsfuture map` both here and around the searchDAO calls is unsafe, because you can't know which will return first.

ahaessly:
when i run all the tests, it seems pretty reliable in showing the timing issue. when i use the bad code and just run the test alone, it actually passes.

PR #288, opened 2016-11-01
GAWB-23: Create endpoints for importing and exporting workspace attributes as tsv files

ansingh7115:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -61,4 +71,71 @@ class WorkspaceService(protected val argUserInfo: UserInfo, val rawlsDAO: RawlsD
...
+          case Success(attrs) =>
+              rawlsDAO.getWorkspace(workspaceNamespace, workspaceName) flatMap { workspaceResponse =>
+                  val allOperations = getWorkspaceAttributeCalls(attributePairs)
+                  rawlsDAO.patchWorkspaceAttributes(workspaceNamespace, workspaceName, allOperations) map (RequestComplete(_))
+                }
+        }
+      }
+  }
+
+  private def getWorkspaceAttributeCalls(attributePairs: Seq[(String,String)]): Seq[AttributeUpdateOperation] = {
```
MatthewBemis, line 126: for readability, you can name the tuple fields using `case` like so:

```
    attributePairs.map { case (name, value) =>
      if (value.equals("__DELETE__")) {
        new RemoveAttribute(new AttributeName("default", name))
      }
      else {
        new AddUpdateAttribute(new AttributeName("default", name), new AttributeString(value))
      }
    }
```

the same thing applies to exportWorkspaceAttributes above

src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
```diff
@@ -549,6 +578,17 @@ class WorkspaceApiServiceSpec extends BaseServiceSpec with WorkspaceApiService w
           }
         }
       }
+
+      "when calling POST on the workspaces/*/*/importAttributes path" - {
+        "should 200 OK if it has the correct headers and valid internals" in {
```
MatthewBemis, line 583: maybe also test a failure case or two here? failing the check in `checkFirstRowDistinct` may be a good one

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -61,4 +71,71 @@ class WorkspaceService(protected val argUserInfo: UserInfo, val rawlsDAO: RawlsD
...
+  }
+
+   /*
+  Import attributes on a workspace from a TSV file
+  */
+  private def importWorkspaceAttributeTSV(workspaceNamespace: String, workspaceName: String, tsv: TSVLoadFile): Future[PerRequestMessage] = {
+    checkFirstRowDistinct(tsv) {
+        val attributePairs = colNamesToWorkspaceAttributeNames(tsv.headers).zip(tsv.tsvData.head)
+        Try(scala.util.parsing.json.JSONObject(attributePairs.toMap).toString().parseJson.asJsObject.convertTo[AttributeMap]) match {
+          case Failure(ex:ParsingException) => Future(RequestCompleteWithErrorReport(StatusCodes.BadRequest, "Invalid json supplied", ex))
```
MatthewBemis: if this is an error that the user is going to see, "Invalid json supplied" will not make a whole lot of sense to them because they're uploading a tsv

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -61,4 +71,85 @@ class WorkspaceService(protected val argUserInfo: UserInfo, val rawlsDAO: RawlsD
     }
   }
 
+  def exportWorkspaceAttributes(workspaceNamespace: String, workspaceName: String, filename: String): Future[PerRequestMessage] = {
+    log.info("WE'RE DOING STUFF!!!")
```
MatthewBemis: ❗️❗️

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -61,4 +71,85 @@ class WorkspaceService(protected val argUserInfo: UserInfo, val rawlsDAO: RawlsD
...
+    Try(rawlsDAO.getWorkspace(workspaceNamespace, workspaceName)) match {
+      case Failure(regret) => Future(RequestCompleteWithErrorReport(StatusCodes.BadRequest, regret.getMessage))
+      case Success(workspaceFuture) => workspaceFuture map { workspaceResponse =>
+          val headerString = "workspace:" + (workspaceResponse.workspace.get.attributes map { attribute =>
```
MatthewBemis: you can also do the tuple field naming trick here

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -61,4 +71,85 @@ class WorkspaceService(protected val argUserInfo: UserInfo, val rawlsDAO: RawlsD
...
+    }
+  }
+
+  private def colNamesToWorkspaceAttributeNames(headers: Seq[String]): Seq[String] = {
+    val newHead = headers.head.stripPrefix("workspace:")
+    Seq(newHead) ++ headers.tail
+  }
+
+  private def checkIf2Rows(tsv: TSVLoadFile)(op: => Future[PerRequestMessage]): Future[PerRequestMessage] = {
+    log.info(tsv.tsvData.toString())
```
MatthewBemis: is this log statement meant to be left in?

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
ansingh7115: Crap, I totally forgot to take that out

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -61,4 +71,83 @@ class WorkspaceService(protected val argUserInfo: UserInfo, val rawlsDAO: RawlsD
     }
   }
 
+  def exportWorkspaceAttributes(workspaceNamespace: String, workspaceName: String, filename: String): Future[PerRequestMessage] = {
+    Try(rawlsDAO.getWorkspace(workspaceNamespace, workspaceName)) match {
+      case Failure(regret) => Future(RequestCompleteWithErrorReport(StatusCodes.BadRequest, regret.getMessage))
```
dvoet, line 78: 2 comments: 1) you don't need a full fledged Future for something we already know the value of. In this case Future.successful creates something that looks like a Future but completes immediately. 2) Why is this a bad request? Seems like you are masking the underlying error. Maybe it was a 404? Maybe it was a communications error?

dvoet: why do you even need to handle the error here at all? Is there something special about the error handling that is different than what per request would do?

dvoet: and 1 more: this is probably not doing what you think it is. The Try is not wrapping the actual call to rawls. It is wrapping only the construction of the call. rawlsDAO.getWorkspace(workspaceNamespace, workspaceName) creates a Future, the Try wraps the creation of the Future, not the actual execution. So calling this error a bad request is wrong, it is an internal server error.

TLDR; get rid of the Try

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -61,4 +71,83 @@ class WorkspaceService(protected val argUserInfo: UserInfo, val rawlsDAO: RawlsD
...
+      case Success(workspaceFuture) => workspaceFuture map { workspaceResponse =>
+          val headerString = "workspace:" + (workspaceResponse.workspace.get.attributes map { case (attName, attValue) =>
+                attName.name}).mkString("\t").replaceAll("description\t", "")
```
dvoet: how are namespaces handled? Looks like they are completely dropped. Will that be a problem for library things?

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -61,4 +71,83 @@ class WorkspaceService(protected val argUserInfo: UserInfo, val rawlsDAO: RawlsD
...
+          val valueString = (workspaceResponse.workspace.get.attributes map { attribute =>
+              impAttributeFormat.write(attribute._2).toString().replaceAll("\"","")}).mkString("\t").replaceAll("null\t", "")
```
dvoet: what is attribute._2? replace attribute 1 line up with a case statement that extracts the tuple to make the code more clear

replacing null\t with "", is it right to drop the tab?

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -61,4 +71,83 @@ class WorkspaceService(protected val argUserInfo: UserInfo, val rawlsDAO: RawlsD
...
+          RequestComplete(StatusCodes.OK, headerString + "\n" + valueString)
+        }
+    }
+  }
+
+  def importAttributesFromTSV(workspaceNamespace: String, workspaceName: String, tsvString: String): Future[PerRequestMessage] = {
+    Try(TSVParser.parse(tsvString)) match {
+      case Failure(regret) => Future(RequestCompleteWithErrorReport(StatusCodes.BadRequest, regret.getMessage))
```
dvoet, line 95: Future.successful

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -61,4 +71,83 @@ class WorkspaceService(protected val argUserInfo: UserInfo, val rawlsDAO: RawlsD
...
+
+   /*
+  Import attributes on a workspace from a TSV file
+  */
+  private def importWorkspaceAttributeTSV(workspaceNamespace: String, workspaceName: String, tsv: TSVLoadFile): Future[PerRequestMessage] = {
+    checkIf2Rows(tsv) {
+      checkFirstRowDistinct(tsv) {
+        val attributePairs = colNamesToWorkspaceAttributeNames(tsv.headers).zip(tsv.tsvData.head)
+        Try(scala.util.parsing.json.JSONObject(attributePairs.toMap).toString().parseJson.asJsObject.convertTo[AttributeMap]) match {
+          case Failure(ex: ParsingException) => Future(RequestCompleteWithErrorReport(StatusCodes.BadRequest, "Invalid json supplied", ex))
```
dvoet, line 114: how could this happen? The user is not even giving us json so telling them invalid json does not seem helpful.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiService.scala
```diff
@@ -85,6 +101,26 @@ trait WorkspaceApiService extends HttpService with FireCloudRequestBuilding
                     }
                   }
                 } ~
+                path("exportAttributes") {
```
dvoet, line 103: why are there 2 exportAttributes end points?

dvoet:
need export tests

PR #289, opened 2016-11-01
GAWB-1218: Modify workspace updateAttributes to take the new attribute map

MattPutnam:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off


src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -30,21 +32,34 @@ object WorkspaceService {
...
-class WorkspaceService(protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDAO, val thurloeDAO: ThurloeDAO) extends Actor {
+class WorkspaceService(protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDAO, val thurloeDAO: ThurloeDAO) extends Actor with WorkspaceServiceSupport {
 
   implicit val system = context.system
   import system.dispatcher
   val log = Logging(system, getClass)
 
   implicit val userInfo = argUserInfo
 
+  import WorkspaceService._
```
skwalker, line 43: This seems a little strange to me -- is there a way we could avoid this?

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -20,6 +21,7 @@ import scala.concurrent.{ExecutionContext, Future}
  */
 object WorkspaceService {
   sealed trait WorkspaceServiceMessage
+  case class UpdateWorkspaceAttributes(workspaceNamespace: String, workspaceName: String, newAttributes: AttributeMap) extends WorkspaceServiceMessage
   case class UpdateWorkspaceACL(workspaceNamespace: String, workspaceName: String, aclUpdates: Seq[WorkspaceACLUpdate], originEmail: String) extends WorkspaceServiceMessage
 
   def props(workspaceServiceConstructor: UserInfo => WorkspaceService, userInfo: UserInfo): Props = {
```
skwalker, line 27: should this be updated to be more like Library Service with a constructor method?

src/main/scala/org/broadinstitute/dsde/firecloud/service/AttributeSupport.scala
```diff
@@ -0,0 +1,34 @@
...
+  * Created by putnam on 11/1/16.
+  */
+trait AttributeSupport {
+  /**
+    * given a set of existing attributes and a set of new attributes, calculate the attribute operations
+    * that need to be performed
+    */
+  def generateAttributeOperations(existingAttrs: AttributeMap, newAttrs: AttributeMap, attributeFilter: AttributeName => Boolean): Seq[AttributeUpdateOperation] = {
+    val oldKeys = existingAttrs.keySet.filter(attributeFilter)
+    val newFields = newAttrs.seq.filterKeys(attributeFilter)
```
davidangb: `filterKeys` is unsafe, and we should avoid it.

See https://issues.scala-lang.org/browse/SI-4776 and broadinstitute/rawls#390 ... in that other PR we spent a long time tracking down mysterious errors; let's avoid those here.

One possible solution: rewrite so that both old and new operate on the entire map, instead of using keys for old and map for new? That way, the `attributeFilter` argument can operate against `AttributeMap.filter`.

davidangb (commented):
one important comment inline.

In this PR, `WorkspaceService` mixes in the `WorkspaceServiceSupport` trait, which extends `AttributeSupport`. You could just as easily eliminate `WorkspaceServiceSupport` entirely, and have `WorkspaceService` mix in `AttributeSupport`.

To match that, you could have `LibraryService` mix in both `LibraryServiceSupport` and `AttributeSupport`, and remove the `generateAttributeOperations` wrapper from `LibraryServiceSupport`.

Finally, you're changing the signature of the updateAttributes endpoint:
1) Is there anything else using that? Are there any end users that use that via scripts?
2) Assuming we are okay changing the signature, why use standard json instead of the FireCloud items/itemsType json? We talked about this in person; how did you reach this decision?

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
MattPutnam: The way this works other places is that the import is at the top.  I like importing things right where they're used though, and it feels weirder to import stuff above where it's defined.

MattPutnam:
Maybe instead of changing this endpoint, add another one called "set attributes" or something?

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
dvoet: I don't think you need the import at all since it is a companion class

MattPutnam: You'd think not, but I get a compiler error otherwise.

skwalker:
This looks good to me 👍

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1542,6 +1542,38 @@ paths:
           description: Workspace not found
         500:
           description: Internal Error
+      security:
+        - googleoauth:
+            - openid
+            - email
+            - profile
+            - https://www.googleapis.com/auth/devstorage.full_control
```
davidangb: based on #290, please remove the security field here.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1542,6 +1542,38 @@ paths:
...
+  /api/workspaces/{workspaceNamespace}/{workspaceName}/setAttributes:
+    patch:
+      tags:
+      - Workspaces
+      operationId: setAttributes
+      summary: |
+        Set attributes on a workspace.
+      produces:
+      - application/json
```
davidangb: also remove produces field here, based on #290.

davidangb (commented):
two swagger updates, comments inline.

davidangb:
👍

PR #290, opened 2016-11-02
GAWB-408: swagger updates

davidangb:
part of GAWB-408, but does not resolve the ticket. I'm trying to make multiple PRs as I progress, to keep chance of merge conflict lower and to keep PRs manageable.

In this PR:
* set version number to something closer to the version in build.sbt
* use `$ref` more aggressively to reuse/centralize code
* use global definitions for `security` and `produces` to centralize code

This PR reduces lines of code from 2966 to 2414.

There should be zero functional change here (beyond the version number). Everything should work just as before, but the code is cleaner.

I suggest reviewing individual commits; they're pretty clean.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -26,12 +26,6 @@ paths:
       responses:
         200:
           description: List of workspaces.
-      security:
-        - googleoauth:
-            - openid
-            - email
-            - profile
-            - https://www.googleapis.com/auth/devstorage.full_control
     post:
       tags:
         - Workspaces
@@ -57,12 +51,6 @@ paths:
           description: Workspace by that name already exists
         500:
           description: Internal Server Error
-      security:
-        - googleoauth:
-            - openid
-            - email
-            - profile
-            - https://www.googleapis.com/auth/devstorage.full_control
   /api/methods:
     get:
       tags:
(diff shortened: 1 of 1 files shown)


ahaessly:
at line 1236, part of the produces list was left in. I think this one should have the produces with both text and json (since it had both before the edit).

in the commented out paths the security fields are still there. it is probably easy to remove them so when/if they are used again this code won't be duplicated (starting at line 1431 through 1578)
I think there could be some parameter reuse for these methods as well, but probably not as important right now.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1763,20 +1227,12 @@ paths:
...
-            - https://www.googleapis.com/auth/devstorage.full_control
   /register/profile:
     get:
       tags:
         - Profile
       operationId: getAll
       summary: Returns a list of all keys and values stored in the user profile service for the currently logged-in user.
-      produces:
-        - application/json
         - text/plain
```
ahaessly, line 1238: here is where i think it should have the produces with both types.

davidangb:
Good catch on the broken `produces:` key for /register/profile - fixed.

I intentionally left the `security` keys in place for the commented-out paths at 1431 through 1578 - each one of them is slightly different than the global setting. One uses cloud-billing instead of devstorage.full_control; the others leave out devstorage.full_control.

As mentioned earlier, I plan to verify what _should_ be in each setting later in the story; for now, I'm trying to just refactor, so I haven't asked if the security keys in those endpoints are correct or not.

Back to you, @ahaessly

PR #292, opened 2016-11-07
Support logout via new auth flow

dmohs:
Support for GAWB 317: https://github.com/broadinstitute/firecloud-ui/pull/548

See that PR for notes.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off


src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -56,6 +38,9 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
   import system.dispatcher
   val log = Logging(system, getClass)
 
+  // TODO(dmohs): This should be passed like other actors.
```
dvoet: do this or enter a jira

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -106,119 +91,14 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
       }
 
     case GetNIHStatus(userInfo: UserInfo) =>
-      val profileResponse = getProfile(userInfo, false)
+      val profileResponse = temporaryThurloeDao.getProfile(userInfo, updateExpiration = false)
       profileResponse pipeTo context.parent
```
dvoet: the new pattern would be to pipeTo sender, not parent but I don't know if that should be part of this

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -291,6 +171,7 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
                                       "Profile saved, but unexpected error registering user", e) }
   }
 
+  // TODO(dmohs): DRY
```
dvoet: is this still valid?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
```diff
@@ -36,4 +36,122 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
...
+              Future(RequestCompleteWithErrorReport(InternalServerError,
+                "Could not unmarshal profile response: " + err.toString, Seq()))
+          }
+        case x =>
+          // request for profile returned non-200
+          Future(RequestCompleteWithErrorReport(x, response.toString, Seq()))
+      }
+    } recover {
+      // unexpected error retrieving profile
+      case e: Throwable => RequestCompleteWithErrorReport(InternalServerError, e.getMessage)
```
dvoet: shouldn't per request catch this? do we need this recover?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
```diff
@@ -36,4 +36,122 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
...
+
+                  // if the user's link has expired, the user must re-link.
+                  // NB: we use a separate val here in case we need to change the logic. For instance, we could
+                  // change the logic later to be "link has expired OR user hasn't logged in within 24 hours"
+                  val loginRequired = (howSoonExpire >= 0)
+
+                  getNIHStatusResponse(pipeline, loginRequired, profile, linkExpireSeconds)
+
+                case _ =>
+                  Future(RequestComplete(NotFound, "Linked NIH username not found"))
```
dvoet: Futures like this (a couple more below) should be Future.successful

davidangb:
unit tests are failing - please fix. Also, you mention notes about this change in the corresponding UI PR, but I don't see any notes there ... what should I be looking at?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
```diff
@@ -36,4 +36,122 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
...
   }
 
+  override def getProfile(userInfo: UserInfo, updateExpiration: Boolean): Future[PerRequestMessage] = {
+    val pipeline = addFireCloudCredentials ~> addCredentials(userInfo.accessToken) ~> sendReceive
+    val profileReq = Get(UserService.remoteGetAllURL.format(userInfo.getUniqueId))
+
+    pipeline(profileReq) flatMap { response: HttpResponse =>
+      response.status match {
+        case OK =>
+          val profileEither = response.entity.as[ProfileWrapper]
```
dvoet: should user `~> unmarshal[ProfileWrapper]` in the pipeline then don't need to match on status or `case Right` below

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
```diff
@@ -36,4 +36,122 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
...
+          isDbgapAuthorized = Some(authorized),
+          lastLinkTime = Some(profile.lastLinkTime.getOrElse(0L)),
+          linkExpireTime = Some(linkExpireSeconds),
+          descriptionSinceLastLink = Some(DateUtils.prettySince(profile.lastLinkTime.getOrElse(0L))),
+          descriptionUntilExpires = Some(DateUtils.prettySince(linkExpireSeconds))
+        )
+      )
+    } recover {
+      // unexpected error retrieving dbgap group status
+      case e: Throwable => RequestCompleteWithErrorReport(InternalServerError, e.getMessage)
```
dvoet: should be handled by per request, recover code not needed

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
```diff
@@ -36,4 +36,122 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
     }
   }
 
+  override def getProfile(userInfo: UserInfo, updateExpiration: Boolean): Future[PerRequestMessage] = {
```
dvoet: DAO functions should not be returning PerRequestMessages. That should be in the service layer. The function name implies that this function should be returning the profile. It actually looks like it returns a NIHStatus within a PerRequestMessage. Further, this code does a bunch more than just getting the profile (updating link expiration, etc.) which is all business logic. DAOs should be strictly data access logic. I would put the complicated code in this function in the service layer.

src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
```diff
@@ -26,189 +37,120 @@ object OAuthService {
...
-                authentication, we are extra careful.
-             */
-            case e: Exception =>
-              log.error("problem during OAuth redirect", e)
-              requestContext.complete(StatusCodes.Unauthorized, e.getMessage)
-          }
-        }
+
+  override def receive = {
+    case HandleOauthCode(code: String, redirectUri: String) => {
```
dvoet: 1 line: `handleOauthCode(code, redirectUri) pipeTo sender`

src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
```diff
@@ -26,189 +37,120 @@ object OAuthService {
...
+  def props(oauthService: () => OAuthService): Props = {
+    Props(oauthService())
+  }
+
+  def constructor(app: Application)()(implicit executionContext: ExecutionContext) =
+    new OAuthService(app.thurloeDAO)
 }
 
-trait OAuthService extends HttpService with PerRequestCreator with FireCloudDirectives with FireCloudRequestBuilding {
+class OAuthService(val thurloeDao: ThurloeDAO)(implicit protected val executionContext: ExecutionContext) extends Actor
```
dvoet: a good model to follow is to have a trait `OAuthService` which has all the good stuff in it and a concrete class `OAuthServiceActor extends Actor with OAuthService` which has just the receive function. This way the trait can be tested in isolation of the actor system.

src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
```diff
@@ -26,189 +37,120 @@ object OAuthService {
...
+    request.setRedirectUri(redirectUri)
+    try {
+      val response = request.execute
+      val idToken = response.parseIdToken()
+      val accessToken = response.getAccessToken
+      val refreshToken = Option(response.getRefreshToken)
+      updateNihStatus(idToken.getPayload.getSubject, accessToken)
+      refreshToken match {
+        case Some(x) => completeWithRefreshToken(accessToken, x)
+        case None => Future(RequestComplete(StatusCodes.NoContent))
```
dvoet, line 76: Future.successful

src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
```diff
@@ -26,189 +37,120 @@ object OAuthService {
...
     }
+  }
 
-
-  private def initiateAuth(state: String, approvalPrompt: String, requestContext: RequestContext) =  {
-    val gcsAuthUrl = HttpGoogleServicesDAO.getGoogleRedirectURI(state, approvalPrompt=approvalPrompt)
-    requestContext.redirect(gcsAuthUrl, StatusCodes.TemporaryRedirect)
+  private def updateNihStatus(subjectId: String, accessToken: String) = {
+    val userInfo = UserInfo("", OAuth2BearerToken(accessToken), -1, subjectId)
+    thurloeDao.getProfile(userInfo, updateExpiration = true)
```
dvoet: the fact that **get**Profile is the result of **update**NihStatus is concerning (see other comments on biz logic vs data access logic)

src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
```diff
@@ -26,189 +37,120 @@ object OAuthService {
...
+      httpTransport, jsonFactory, clientSecrets, authScopes
+    ).build
+    val request = flow.newTokenRequest(code)
+    request.setRedirectUri(redirectUri)
+    try {
+      val response = request.execute
+      val idToken = response.parseIdToken()
+      val accessToken = response.getAccessToken
+      val refreshToken = Option(response.getRefreshToken)
+      updateNihStatus(idToken.getPayload.getSubject, accessToken)
```
dvoet: where does the result of this go?

src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
dvoet: I see a bunch of spray client code in this actor I would expect that to be in DAO classes.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/OauthApiService.scala
```diff
@@ -0,0 +1,84 @@
...
+          val params = requestContext.request.entity.data.asString.parseJson.asJsObject
+          params.fields.get("code") match {
+            case None => completeWithMissingKey(requestContext, "code")
+            case Some(jsvCode) =>
+              try {
+                val code = jsvCode.convertTo[String]
+                params.fields.get("redirectUri") match {
+                  case None => completeWithMissingKey(requestContext, "redirectUri")
+                  case Some(jsvRedirectUri) =>
+                    try {
```
dvoet: do we need this inner try as well?

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/OauthApiService.scala
```diff
@@ -0,0 +1,84 @@
...
+  def completeWithBadValue(rc: RequestContext, key: String, message: String) = rc.complete(
+    StatusCodes.BadRequest,
+    Map("error" -> Map("summary" -> s"Invalid value for '$key'", "detail" -> message))
+  )
+
+  val oauthRoutes: Route =
+    path("handle-oauth-code") {
+      post { requestContext =>
+        try {
+          val params = requestContext.request.entity.data.asString.parseJson.asJsObject
```
dvoet, line 38: why do we do this instead of modeling with a case class and using the usual json formatting?

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/OauthApiService.scala
dvoet: in fact it is the 3rd level

dvoet (commented):
I am not sure how far you want to take this refactoring but I would not say it is yet done. It is certainly a step in the right direction though. I don't see any show stopper issues.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
dmohs: https://broadinstitute.atlassian.net/browse/GAWB-1286

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
dmohs: Yes, it's duplicated by `sendNotifications` in ThurloeDAO. Probably beyond the scope of this PR?

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/OauthApiService.scala
dmohs: I attempted this with the next path (`handle-oauth-code-v2`). Despite the benefit of concision, I can't get the usability I want. First, it rejects the request if you don't specify a Content-Type header, which makes this less approachable for curl users but has no affect on our ability to parse the message. Second, when I specify an invalid value, it only tells me that there was an invalid value—it doesn't tell me which field.

I'd prefer to write less code, but not at the expense of user experience. Help would be appreciated.

src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
```diff
@@ -497,7 +497,8 @@ class WorkspaceApiServiceSpec extends BaseServiceSpec with WorkspaceApiService w
           }
         }
 
-        "should 200 OK if it has the full set of required attribute headers, plus optionals" in {
+        // TODO(dmohs): Flaky.
```
dmohs: Note: I intend to un-ignore these as part of this change as soon as I figure out why they are flaky.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
dvoet: if it is still valid then enter a JIRA and remove the TODO, it is useless to anyone but you - and my guess is that it will be useless to you too in about a month ;)

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -56,6 +38,10 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
   import system.dispatcher
   val log = Logging(system, getClass)
 
+  // TODO(dmohs): This should be passed like other actors.
+  val temporaryThurloeDao = new HttpThurloeDAO
+  val temporaryRawlsDao = new HttpRawlsDAO
+
   override def receive: Receive = {
```
dvoet, line 45: we can keep going down the refactor rabbit hole so just let me know if you want to stop.

a) each case in the receive function should be in the form `case Message(args) => callFunction(args) pipeTo sender`
b) notice I pipeTo sender, not context.parent

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -106,117 +92,24 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
       }
```
dvoet: I know you did not write it but 

```
syncWhiteListResult onComplete {
  case _ => profileResponse pipeTo parent
}
```

is terrible. This should use a for comprehension or a flatMap and the overall result should be piped, not a pipe within the onComplete.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
```diff
@@ -34,6 +34,16 @@ class HttpRawlsDAO( implicit val system: ActorSystem, implicit val executionCont
     }
   }
 
+  override def isDbGapAuthorized(accessToken: OAuth2BearerToken): Future[Boolean] = {
+    val pipeline = addCredentials(accessToken) ~> sendReceive
+    pipeline(Get(RawlsDAO.pathToUrl(RawlsDAO.groupPath(FireCloudConfig.Nih.rawlsGroupName)))) map {
+      response => response.status match {
+        case x if x == OK => true
```
dvoet: I think you can do just
`case OK => true`

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
```diff
@@ -87,5 +97,25 @@ class HttpRawlsDAO( implicit val system: ActorSystem, implicit val executionCont
   private def getWorkspaceUrl(ns: String, name: String) = FireCloudConfig.Rawls.authUrl + FireCloudConfig.Rawls.workspacesPath + s"/%s/%s".format(ns, name)
   private def patchWorkspaceAclUrl(ns: String, name: String) = rawlsWorkspaceACLUrl.format(ns, name)
 
+  override def getRefreshTokenStatus(userInfo: UserInfo): Future[Option[DateTime]] = {
+    val pipeline = addCredentials(userInfo.accessToken) ~> sendReceive
+    pipeline(Get(RawlsDAO.pathToUrl(RawlsDAO.refreshTokenDatePath))) map { response =>
+      response.status match {
+        case OK =>
+          Some(DateTime.parse(unmarshal[RawlsTokenDate].apply(response).refreshTokenUpdatedDate))
```
dvoet: use Option instead of Some directly so you don't run the risk of getting a Some(null)

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
```diff
@@ -36,4 +34,64 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
     }
   }
 
+  override def getProfile(userInfo: UserInfo): Future[Option[Profile]] = {
+    val pipeline = addFireCloudCredentials ~> addCredentials(userInfo.accessToken) ~> sendReceive
+    pipeline(Get(UserService.remoteGetAllURL.format(userInfo.getUniqueId))) map { response =>
+      response.status match {
+        case StatusCodes.OK => Some(Profile(unmarshal[ProfileWrapper].apply(response)))
```
dvoet, line 41: again with the Option instead of Some

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
```diff
@@ -36,4 +34,64 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
...
+        case _ => throwBadResponse(response)
+      }
+    }
+  }
+
+  def getIsDbGapAuthorized(accessToken: OAuth2BearerToken): Future[Boolean] = {
+    val pipeline = addFireCloudCredentials ~> addCredentials(accessToken) ~> sendReceive
+    pipeline(Get(UserService.groupUrl(FireCloudConfig.Nih.rawlsGroupName))) map { response =>
+      response.status match {
+        case x if x == OK => true
```
dvoet: case OK

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/RawlsDAO.scala
```diff
@@ -4,12 +4,23 @@ import com.typesafe.scalalogging.slf4j.LazyLogging
...
 /**
   * Created by davidan on 9/23/16.
   */
+
+object RawlsDAO {
+  lazy val refreshTokenPath = FireCloudConfig.Rawls.authPrefix + "/user/refreshToken"
+  lazy val refreshTokenDatePath = FireCloudConfig.Rawls.authPrefix + "/user/refreshTokenDate"
+
+  def pathToUrl(path: String) = FireCloudConfig.Rawls.baseUrl + path
+  def groupPath(group: String): String = FireCloudConfig.Rawls.authPrefix + s"/user/group/$group"
```
dvoet: I am not a big fan of this pattern. It leads to code like `RawlsDAO.pathToUrl(RawlsDAO.groupPath("group"))` which seems to complex to me. When do you ever want RawlsDAO.groupPath or any other path on its own? We should just have Url defs/vals e.g. `def groupUrl(group: String) = FireCloudConfig.Rawls.baseUrl + FireCloudConfig.Rawls.authPrefix + s"/user/group/$group"`

dmohs: Yes I think I see and agree.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -101,122 +85,27 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
       } recover { case e: Throwable => RequestCompleteWithErrorReport(InternalServerError,
         "Unexpected error updating NIH link", e) }
       // Complete syncWhitelist and ignore as neither success nor failure are useful to the client
-      syncWhiteListResult onComplete {
-        case _ => profileResponse pipeTo parent
-      }
+      syncWhiteListResult pipeTo sender
```
dvoet: I think you changed the behavior here. The old behavior was do syncWhiteListResult then whatever the response pipe back profileResponse. I think the right way to do this is convert syncWhiteListResult from `Future[x]` to `Future[Try[x]]` (see FutureSupport.toFutureTry in rawls). Then `syncWhiteListResultTry flatMap { _ => profileResponse }` pipeTo sender

dvoet:
one comment then 👍

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
davidangb: +1

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -56,10 +38,13 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
   import system.dispatcher
   val log = Logging(system, getClass)
 
+  // GAWB-1286
```
davidangb, line 41: not code - but the referenced jira ticket needs some explanation. I don't understand what's happening here, or why these are temporary. Is this to avoid updating all the other callers of ProfileClient?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
```diff
@@ -34,6 +34,16 @@ class HttpRawlsDAO( implicit val system: ActorSystem, implicit val executionCont
     }
   }
 
+  override def isDbGapAuthorized(accessToken: OAuth2BearerToken): Future[Boolean] = {
+    val pipeline = addCredentials(accessToken) ~> sendReceive
+    pipeline(Get(RawlsDAO.groupUrl(FireCloudConfig.Nih.rawlsGroupName))) map {
```
davidangb: for centralization reasons, consider having isDbGapAuthorized accept the full UserInfo object, and use `userAuthedRequest` instead of writing out the pipeline yourself.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
```diff
@@ -87,5 +97,25 @@ class HttpRawlsDAO( implicit val system: ActorSystem, implicit val executionCont
   private def getWorkspaceUrl(ns: String, name: String) = FireCloudConfig.Rawls.authUrl + FireCloudConfig.Rawls.workspacesPath + s"/%s/%s".format(ns, name)
   private def patchWorkspaceAclUrl(ns: String, name: String) = rawlsWorkspaceACLUrl.format(ns, name)
 
+  override def getRefreshTokenStatus(userInfo: UserInfo): Future[Option[DateTime]] = {
+    val pipeline = addCredentials(userInfo.accessToken) ~> sendReceive
+    pipeline(Get(RawlsDAO.refreshTokenDateUrl)) map { response =>
```
davidangb: same here. avoid boilerplate and use `userAuthedRequest`. If RestJsonClient doesn't have the methods you need, consider adding them there for reuse elsewhere.

EDIT: there's a lot of places in this PR that should use `userAuthedRequest` or `requestToObject`, potentially with some enhancements.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
```diff
@@ -87,5 +97,25 @@ class HttpRawlsDAO( implicit val system: ActorSystem, implicit val executionCont
...
 
+  override def saveRefreshToken(accessToken: String, refreshToken: String): Future[Unit] = {
+    val pipeline = addCredentials(OAuth2BearerToken(accessToken)) ~> sendReceive
+    pipeline(Put(RawlsDAO.refreshTokenUrl, RawlsToken(refreshToken))) map
+      { _ => () }
+  }
+
+  private def throwBadResponse(response: HttpResponse) = {
+    throw new FireCloudExceptionWithErrorReport(ErrorReport("Rawls", response))
+  }
```
davidangb: this is only used in one place. Is it worth breaking out into a separate method? I'm fine either way.

I do think the error should be "FireCloud" and not "Rawls", though, since it's user-facing.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
```diff
@@ -36,4 +34,64 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
     }
   }
 
+  override def getProfile(userInfo: UserInfo): Future[Option[Profile]] = {
+    val pipeline = addFireCloudCredentials ~> addCredentials(userInfo.accessToken) ~> sendReceive
+    pipeline(Get(UserService.remoteGetAllURL.format(userInfo.getUniqueId))) map { response =>
+      response.status match {
+        case StatusCodes.OK => Option(Profile(unmarshal[ProfileWrapper].apply(response)))
```
davidangb: `Some`, not `Option`

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
```diff
@@ -36,4 +34,64 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
...
+
+  def getIsDbGapAuthorized(accessToken: OAuth2BearerToken): Future[Boolean] = {
+    val pipeline = addFireCloudCredentials ~> addCredentials(accessToken) ~> sendReceive
+    pipeline(Get(UserService.groupUrl(FireCloudConfig.Nih.rawlsGroupName))) map { response =>
+      response.status match {
+        case OK => true
+        case _ => false
+      }
+    }
+  }
```
davidangb: why is this in the Thurloe DAO? This is contacting Rawls. And ... the same method exists in the Rawls DAO.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
```diff
@@ -36,4 +34,64 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
...
+
+  override def maybeUpdateNihLinkExpiration(userInfo: UserInfo, profile: Profile): Future[Unit] = {
+    val pipeline = addFireCloudCredentials ~> addCredentials(userInfo.accessToken) ~>
+      sendReceive
```
davidangb: micro-optimization: move the `val pipeline` inside the match. Otherwise, you're creating an object that never gets used.

src/main/scala/org/broadinstitute/dsde/firecloud/model/Profile.scala
```diff
@@ -105,6 +107,7 @@ object Profile {
         case Some(time) => Some(time.toLong)
         case _ => None
       },
+      // TODO(dmohs): Now we always get this from the authoritative source (Google).
```
davidangb: GAWB-83

src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
```diff
@@ -1,214 +1,103 @@
...
+    new OAuthService(app.rawlsDAO, app.thurloeDAO)
 }
 
-trait OAuthService extends HttpService with PerRequestCreator with FireCloudDirectives with FireCloudRequestBuilding {
+class OAuthService(val rawlsDao: RawlsDAO, val thurloeDao: ThurloeDAO)
+  (implicit protected val executionContext: ExecutionContext) extends Actor
+  with LazyLogging {
 
-  private implicit val executionContext = actorRefFactory.dispatcher
   lazy val log = LoggerFactory.getLogger(getClass)
```
davidangb: if you're using `LazyLogging`, get rid of this logger.

src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
```diff
@@ -1,214 +1,103 @@
...
+      httpTransport, jsonFactory, clientSecrets, authScopes
+    ).build
+    val request = flow.newTokenRequest(code)
+    request.setRedirectUri(redirectUri)
+    try {
+      val response = request.execute
+      val idToken = response.parseIdToken()
+      val accessToken = response.getAccessToken
+      val refreshToken = Option(response.getRefreshToken)
+      val userInfo = UserInfo("", OAuth2BearerToken(accessToken), -1, idToken.getPayload.getSubject)
```
davidangb: not critical, but I think it would be better to create a a companion object in UserInfo.scala:

```
object UserInfo {
  def apply(accessToken: OAuth2BearerToken, id: String) = new UserInfo("", accessToken, -1, id)
}
```

and let this line be just:

```
val userInfo = UserInfo(OAuth2BearerToken(accessToken), idToken.getPayload.getSubject)
```

src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
```diff
@@ -1,214 +1,103 @@
...
+        ageDaysCount match {
+          case x if x < 90 =>
+            log.debug(s"User's refresh token is $x days old; all good!")
+            RequestComplete(StatusCodes.NoContent)
+          case x =>
+            log.info(s"User's refresh token is $x days old; requesting a new one.")
+            RequestComplete(StatusCodes.OK, Map("requiresRefresh" -> true))
+        }
+      case None =>
+        RequestComplete(StatusCodes.OK, Map("requiresRefresh" -> true))
```
davidangb, line 97: I find these responses confusing. It's impossible to get `{"requiresRefresh":false}`; you get a 200 OK if you _do_ need a refresh. I think this should either return `requiresRefresh: true | false`, or it should return 200 if all set, and some other error code (204?, 205?) if a a refresh is needed.

src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
```diff
@@ -1,214 +1,103 @@
...
+        _.map { thurloeDao.maybeUpdateNihLinkExpiration(userInfo, _) }
+      }
+      refreshToken match {
+        case Some(x) =>
+          rawlsDao.saveRefreshToken(accessToken, x) map { _ => RequestComplete(StatusCodes.NoContent)}
+        case None => Future(RequestComplete(StatusCodes.NoContent))
       }
+    } catch {
+      case e: TokenResponseException =>
+        Future(RequestComplete(HttpResponse(StatusCodes.BadRequest, e.getContent)))
```
davidangb, line 80: consider moving a bunch of this code (back) into the Google DAO - ideally, that class would (eventually) encapsulate all the connectivity to Google.

src/main/scala/org/broadinstitute/dsde/firecloud/service/UserService.scala
```diff
@@ -133,9 +132,8 @@ trait UserService extends HttpService with PerRequestCreator with FireCloudReque
               }
               tryParseScopes match {
                 case Success(scopes) =>
-                  // user does not have appropriate scopes.  Ask user to enable them via OAuth
-                  val redirectURI = HttpGoogleServicesDAO.getGoogleRedirectURI(state, "auto", Option(scopes.requiredScopes ++ HttpGoogleServicesDAO.userLoginScopes))
-                  requestContext.complete(Forbidden, ErrorReport(Forbidden, BillingAccountRedirect(redirectURI).toJson.toString))
+                  // user does not have appropriate scopes.
+                  requestContext.complete(Forbidden)
```
davidangb: doesn't this change end user functionality? (caveat: I haven't read through the UI PR yet, maybe you're handling it there).

src/main/scala/org/broadinstitute/dsde/firecloud/service/UserService.scala
```diff
@@ -145,7 +143,7 @@ trait UserService extends HttpService with PerRequestCreator with FireCloudReque
         }
       } ~
       path("profile" / "refreshTokenDate") {
-        passthrough(OAuthService.remoteTokenDateUrl, HttpMethods.GET)
+        passthrough(RawlsDAO.refreshTokenDateUrl, HttpMethods.GET)
```
davidangb: is this endpoint used anymore, now that you've added /api/refresh-token-status ?

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/OauthApiService.scala
```diff
@@ -0,0 +1,97 @@
...
+  implicit val impHandleOathCodeParams = jsonFormat2(HandleOauthCodeParams)
+
+  val oauthRoutes: Route =
+    path("handle-oauth-code") {
+      post { requestContext =>
+        try {
+          val params = requestContext.request.entity.data.asString.parseJson.asJsObject
+          params.fields.get("code") match {
+            case None => completeWithMissingKey(requestContext, "code")
+            case Some(jsvCode) =>
```
davidangb: if you use `Some(jsvCode:JsString)`, you're both being extra-safe, and have a head start on converting it later.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/OauthApiService.scala
davidangb: I don't care about supporting curl users for login routes, so let's not jump through hoops there.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/OauthApiService.scala
davidangb: yes - let's try to make this whole thing a bit more terse. How about something like:

```
          val fields = Try(requestContext.request.entity.data.asString.parseJson.asJsObject.fields)
            .toOption.getOrElse(throw(DeserializationException(...)))
          val jsvCode = fields.get("code")
          val jsvRedirectUri = fields.get("redirectUri")
          (jsvCode, jsvRedirectUri) match {
            case (Some(code:JsString), Some(redirectUri:JsString)) =>
              perRequest(requestContext,
                OAuthService.props(oauthServiceConstructor),
                OAuthService.HandleOauthCode(code.value, redirectUri.value)
              )
            case _ => completeWithSomethingLike(s"'code' and 'redirectUri' required; found: {code: %s, redirectUri: %s}".format(jsvCode, jsvRedirectUri))
          }
```

(note that this is pseudocode, won't compile)

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/OauthApiService.scala
```diff
@@ -0,0 +1,97 @@
...
+      path("handle-oauth-code-v2") {
+        post {
+          entity(as[HandleOauthCodeParams]) { params => requestContext =>
+            perRequest(requestContext,
+              OAuthService.props(oauthServiceConstructor),
+              OAuthService.HandleOauthCode(params.code, params.redirectUri)
+            )
+          }
+        }
+      } ~
```
davidangb: are both `handle-oauth-code` and `handle-oauth-code-v2` used? Remove whatever is not in use.

davidangb (commented):
first pass review done, back to you. I have reviewed this for code/architecture, not 100% for security. Once I look at the UI PR, I'll understand the whole thing and can think about security too.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
dmohs: The comment just above this line says, "Complete syncWhitelist and ignore as neither success nor failure are useful to the client." I also tested linking on dev. Maybe this shouldn't be piped anywhere?

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
dmohs: All I can really say with confidence is that ProfileClient doesn't look enough like OauthService (for example) to receive app.thurloeDao and whatnot. Suggestions welcome.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
dmohs: Doug's concern was that you could get Some(null) in that case, assuming I understood and implemented this correctly.

src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala
dmohs: This PR is wearing me out. I'll keep pushing this to be better as I touch it.

src/main/scala/org/broadinstitute/dsde/firecloud/service/UserService.scala
dmohs: I would guess that you now know that I have handled this case in the UI.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
dmohs: This also works:
```
syncWhiteListResult map(Success(_)) recover { case t => Failure(t) } flatMap { _ => profileResponse } pipeTo sender
```
but I'm inclined to think (based on the comment) that this line should just be deleted.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
davidangb: *I* understand - I'm saying that were someone to pick up GAWB-1286, that ticket is not particularly understandable, and there's nothing in code comments to explain either.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
davidangb: If nothing else, this PR changes what's in the response. Previously the response was `profileResponse`, now it is `syncWhiteListResult`.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
davidangb: That's fair - if `unmarshal[ProfileWrapper].apply(response)` can return null, makes complete sense. I thought it would throw an error if it encountered a problem.

src/main/scala/org/broadinstitute/dsde/firecloud/service/UserService.scala
davidangb: Previously, this endpoint would return:
* 403 if the user is not activated (standard behavior for all protected endpoints)
* 403 with a redirect containing the missing oauth scope if the user is activated, but hasn't oauthed to the right scope

For users who hit the API directly, this is a functional change that needs to be approved/documented in jira or elsewhere, for QA if nothing else. Your UI change technically looks vulnerable to a weird corner case, but is probably safe because users can't reach that page without being activated anyway.

What is your reasoning behind this change? The billing scope isn't actually required? The BillingAccountRedirect object isn't necessary in the response, and API users should look for an empty 403 to mean a scope problem? Something else?

If this change stays as-is, then the BillingAccountRedirect class and its jsonformat should be deleted. If the billing scope isn't actually required, the `tryParseScopes` section should be deleted.

davidangb (commented):
comments inline ...

src/main/scala/org/broadinstitute/dsde/firecloud/service/UserService.scala
```diff
@@ -117,35 +117,25 @@ trait UserService extends HttpService with PerRequestCreator with FireCloudReque
...
-            case Success(response) => requestContext.complete(response.status, response.entity)
-            case Failure(regrets) => respondWithErrorReport(regrets, requestContext)
+          pipeline(Get(UserService.billingAccountsUrl)) map { response =>
+            response.status match {
+              // The proxy would prevent a user from getting here if their account wasn't enabled,
+              // so a 403 from Rawls is effectively guaranteed to mean that the user hasn't enabled
+              // the billing accounts scope.
+              case StatusCodes.Forbidden =>
+                requestContext.complete(
+                  StatusCodes.Forbidden, Map("billingPermissionsRequired" -> true))
```
davidangb: I thought we discussed (via Slack) sending the error message from rawls - which contains the name of the missing scope - back to the user? Without that, API users have no way of knowing which scope is missing ... and if the back-end logic changes to require new/different scopes, we have no way of communicating that.

This PR has gone on for a long time and I guess I won't hold it up any longer on this point, but I'd prefer to pass the scope details to the user, instead of obfuscating them.

davidangb:
one comment inline, your call on what to do about it, otherwise ok. I haven't tested the latest changes functionally - counting on you to do that.

src/main/scala/org/broadinstitute/dsde/firecloud/service/UserService.scala
dvoet: I agree with @davidangb, the design was that rawls passes back the required scopes and then whoever is responsible should ask for those scopes. I don't know why that was dropped. Looking to the comment above, "effectively guaranteed," I don't believe you.

dvoet: oops, didn't see the later commit, never mind

davidangb:
👍 approve!

PR #298, opened 2016-11-09
Use JAVA_OPTS like in Rawls

dmohs:
No JIRA. This bit me today, so I brought it over.


dvoet:
@dmohs is this still relevant

PR #300, opened 2016-11-10
Ah gawb 973: Library: simple text search

ahaessly:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [x] **LR** sign off


src/main/resources/swagger/api-docs.yaml
```diff
@@ -1425,6 +1425,49 @@ paths:
           description: Method Configuration failed to resolve input expressions with the supplied Entity
         500:
           description: Internal Error
+  '/api/libraries':
```
skwalker: we're trying to alphabetize the swagger so this should be further up (line 307ish)

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1425,6 +1425,49 @@ paths:
...
+    get:
+      tags:
+        - Library
+      operationId: FindDocuments
+      summary: |
+        lists published library datasets
+      produces:
```
skwalker: also don't need produces anymore! (check the readme in src/main/resources/swagger

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
```diff
@@ -93,4 +99,20 @@ class ElasticSearchDAO(servers:Seq[Authority], indexName: String) extends Search
       case e: Exception => logger.warn(s"ES index '%s' could not be recreated and may be in an unstable state.".format(indexName))
     }
   }
+
+  override def findDocuments(term: String, from: Int = 0, size: Int = 10) : LibrarySearchResponse = {
```
skwalker: Might be nicer if findDocuments took in term as an Option(String) and then you could just match the None case to use findAll vs make the queryStr from the String

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
```diff
@@ -1,21 +1,28 @@
...
 import spray.http.Uri.Authority
+import spray.json._
+
+import scala.concurrent.Future
 
 class ElasticSearchDAO(servers:Seq[Authority], indexName: String) extends SearchDAO with ElasticSearchDAOSupport {
 
   private val client: TransportClient = buildClient(servers)
   private final val datatype = "dataset"
+  private final val findAll = "{ \"query\": { \"match_all\" : {}}}"
```
skwalker: I think it would be nicer if we could make these case classes and then use the basic json formatter class to turn them into strings

src/main/resources/swagger/api-docs.yaml
davidangb: and you don't need the quotes.

src/main/resources/swagger/api-docs.yaml
davidangb: well ... if this endpoint truly returns both text/plain or application/json, then it does need to be overridden here. If it only returns json, leave it out.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -2345,6 +2388,19 @@ definitions:
...
+    properties:
+      searchTerm:
+        type: string
+        description: The term to search for
+      from:
+        type: integer
+        description: Where in the results list to start (used for pagination)
+      size:
+        type: integer
+        description: How many results to return
```
davidangb, line 2675: `SearchTerm` should also be alphabetized amongst the definitions, and have a newline before it.

Also, specify a `default` for from and size.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
```diff
@@ -1,21 +1,28 @@
 package org.broadinstitute.dsde.firecloud.dataaccess
 
 import org.broadinstitute.dsde.firecloud.model.Document
+import  org.broadinstitute.dsde.firecloud.model.LibrarySearchResponse
```
davidangb: nit: to be Scala-ish, combine this line and the previous one as:

```
import org.broadinstitute.dsde.firecloud.model.{Document, LibrarySearchResponse}
```

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
davidangb: +1

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/SearchDAO.scala
```diff
@@ -14,5 +14,5 @@ trait SearchDAO extends LazyLogging {
   def bulkIndex(docs: Seq[Document]): Unit
   def indexDocument(doc: Document): Unit
   def deleteDocument(id: String): Unit
-
+  def findDocuments(term: String, from: Int = 0, size: Int = 10): LibrarySearchResponse
```
davidangb: This signature is going to have to change as soon as we support facets - we'll need more criteria than a String. I think we'll need to create some kind of `SearchCriteria` object, holding the search term, from/size for pagination, sort criteria, and (soon) facet selections, eventually things like sort. I'm fine leaving this PR as-is, but if you feel like tackling that refactor, all good.

EDIT: oh wait, you have a `LibrarySearchParams` model class already! Why not have this method accept `LibrarySearchParams` instead? Then you could also handle the default values for `from` and `size` in the model, not here; you could address Sarah's comment about accepting `Option[String]` for the search term in the model too.

src/main/scala/org/broadinstitute/dsde/firecloud/model/Library.scala
```diff
@@ -0,0 +1,18 @@
+package org.broadinstitute.dsde.firecloud.model
```
davidangb: combine `Library.scala` and `Search.scala`? I have no opinion what it should be named, but the model classes feel like they belong together.

src/main/scala/org/broadinstitute/dsde/firecloud/model/Library.scala
```diff
@@ -0,0 +1,18 @@
...
+  */
+
+  case class LibrarySearchParams(
+    searchTerm: String,
+    from: Int = 0,
+    size: Int = 10)
+
+  case class LibrarySearchResponse(
+    total: Int,
+    results: Array[JsObject])
```
davidangb: let's brainstorm this. My gut tells me we should use something more strongly typed than JsObject here, for consumption by Scala code. This class might need a constructor (read: `apply` method) that accepts a JsObject, and/or accepts a SearchHit object.

The parallel to the `Document` model class is interesting. `Document` is intended to be created by Scala and sent to ES, so it makes sense for it to hold a JsObject, because ES expects a JsObject. `LibrarySearchResponse` is intended to be read _from_ ES and consumed by Scala, so it makes sense for it to hold a more strongly-typed object.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/LibraryApiService.scala
```diff
@@ -79,8 +81,28 @@ trait LibraryApiService extends HttpService with FireCloudRequestBuilding
                 LibraryService.IndexAll)
             }
           }
+        } ~
+        pathPrefix("libraries") {
```
davidangb: this should probably be `pathEnd` or `pathEndOrSingleSlash` (see http://spray.io/documentation/1.2.4/spray-routing/predefined-directives-alphabetically/), to be precise about path-matching.

Please ignore (or fix!) all the other routes that do this wrong :(

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/LibraryApiService.scala
```diff
@@ -79,8 +81,28 @@ trait LibraryApiService extends HttpService with FireCloudRequestBuilding
...
+              }
+            }
+          } ~
+          get {
+            respondWithJSON {
+              parameters('from ? 0, 'size ? 10) { (from, size ) =>
+                requestContext =>
+                  perRequest(requestContext,
+                    LibraryService.props(libraryServiceConstructor, userInfo),
+                    LibraryService.FindDocuments("", from, size))
```
davidangb: yeah, I'd definitely pass `None` here instead of `""`.

src/test/scala/org/broadinstitute/dsde/firecloud/service/LibraryApiServiceSpec.scala
```diff
@@ -96,11 +99,38 @@ class LibraryApiServiceSpec extends BaseServiceSpec with LibraryApiService {
...
             this.searchDAO.asInstanceOf[MockSearchDAO].deleteDocumentInvoked = false
           }
         }
       }
     }
+
+    "when retrieving datasets" - {
+      "GET on " + librariesPath - {
+        "should retrieve all datasets" in {
+          this.searchDAO.asInstanceOf[MockSearchDAO].indexDocumentInvoked = false
```
davidangb: findDocumentsInvoked, not indexDocumentInvoked

src/test/scala/org/broadinstitute/dsde/firecloud/service/LibraryApiServiceSpec.scala
```diff
@@ -96,11 +99,38 @@ class LibraryApiServiceSpec extends BaseServiceSpec with LibraryApiService {
...
+          new RequestBuilder(HttpMethods.GET)(librariesPath) ~> dummyUserIdHeaders("1234") ~> sealRoute(libraryRoutes) ~> check {
+            status should equal(OK)
+            assert(this.searchDAO.asInstanceOf[MockSearchDAO].findDocumentsInvoked, "indexDocument should have been invoked")
```
davidangb: findDocumentsInvoked, not indexDocument (in the string message)

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1425,6 +1425,49 @@ paths:
...
+        - in: query
+          description: which result to start at (for pagination)
+          name: from
+          type: integer
+          required: false
+        - in: query
+          description: how many results to return
+          name: size
+          type: integer
+          required: false
```
davidangb: for `from` and `size` parameters, specify a `default`: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameterObject

src/main/scala/org/broadinstitute/dsde/firecloud/model/Library.scala
davidangb: Also, I think the search response should echo the `LibrarySearchParams` it used back to the caller. This makes handling the response super easy, either from the UI or via script - the caller can read everything it needs from the response packet.

davidangb (commented):
Back to you, comments inline!

src/main/scala/org/broadinstitute/dsde/firecloud/model/Library.scala
ahaessly: Do you think they should both be put with ESMapping?

ahaessly:
issues above should be addressed

src/main/scala/org/broadinstitute/dsde/firecloud/model/ESMapping.scala
```diff
@@ -1,7 +1,81 @@
...
+  def apply(id: String, jsonStr: String) = new Document(id, jsonStr.parseJson.asJsObject)
+}
+
+
+// classes to convert from json body and to json response
+case class LibrarySearchParams(
+                                searchTerm: Option[String],
+                                from: Option[Int],
+                                size: Option[Int]) {
+}
```
davidangb: what about:

```
case class LibrarySearchParams(
  searchTerm: Option[String],
  from: Int = 0,
  size: Int = 10) {
}
```

this would remove the need for `LibrarySearchConstants`, and consumers of this class could simply use `from` and `size` instead of `from.getOrElse(...)` and `size.getOrElse(...)`.

nitpick: indentation. Your code is using cljs indentation ... just use a couple of spaces instead of aligning to the open-paren.

src/main/scala/org/broadinstitute/dsde/firecloud/model/ESMapping.scala
```diff
@@ -1,7 +1,81 @@
...
+    implicit val impAttributeFormat: AttributeFormat = new AttributeFormat with PlainArrayAttributeListSerializer
+    new Document(id, valMap.toJson.asJsObject)
+  }
+  def apply(id: String, jsonStr: String) = new Document(id, jsonStr.parseJson.asJsObject)
+}
+
+
+// classes to convert from json body and to json response
+case class LibrarySearchParams(
+                                searchTerm: Option[String],
```
skwalker: nit: you can leave searchTerm on the previous line and then make sure the from and size to fields line up with it.
(same thing with LibrarySearchResponse below)

src/main/scala/org/broadinstitute/dsde/firecloud/model/ESMapping.scala
```diff
@@ -1,7 +1,81 @@
...
+                                  from: Int,
+                                  size: Int,
+                                  total: Int,
+                                  results: Array[JsValue]) {
+  def this (params: LibrarySearchParams, total: Int, results: Array[JsValue]) =
+    this(params.searchTerm, params.from.getOrElse(LibrarySearchConstants.defaultFrom), params.size.getOrElse(LibrarySearchConstants.defaultSize), total, results)
+}
+
+
+// classes to create the following 2 query formats for ES
```
skwalker: forgot to delete?

src/main/scala/org/broadinstitute/dsde/firecloud/model/ESMapping.scala
```diff
@@ -1,7 +1,81 @@
...
+
+case class LibrarySearchResponse(
+                                  searchTerm: Option[String],
+                                  from: Int,
+                                  size: Int,
+                                  total: Int,
+                                  results: Array[JsValue]) {
+  def this (params: LibrarySearchParams, total: Int, results: Array[JsValue]) =
+    this(params.searchTerm, params.from.getOrElse(LibrarySearchConstants.defaultFrom), params.size.getOrElse(LibrarySearchConstants.defaultSize), total, results)
+}
```
davidangb: how about:

```
case class LibrarySearchResponse(
  request: LibrarySearchParams,
  total: Int,
  results: Array[JsValue]) {
}
```

?
Is there any reason to flatten it out? By reusing the `LibrarySearchParams` object, the response object will gracefully adapt as we add new criteria to the params.

Also, I'd still really like to see this use something more strongly typed than `JsValue` - could this be an `AttributeMap`? I haven't run the code yet, so Im not sure exactly what the value looks like.

Finally - in Scala-land, use `Seq` to represent a generic list-of-things, not `Array`, if possible.

src/main/scala/org/broadinstitute/dsde/firecloud/model/ESMapping.scala
```diff
@@ -1,7 +1,81 @@
...
+import spray.json.{JsObject, JsValue}
+import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
+
 case class AttributeDefinition(properties: Map[String, AttributeDetail])
 case class AttributeDetail(`type`: String, items: Option[AttributeDetail]=None)
 
 case class ESDatasetProperty(properties: Map[String, ESDetail])
 case class ESDetail(`type`: String)
+
+
```
skwalker: Maybe rename this file since it's doing much more than elastic search mappings now

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -117,4 +119,9 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
     }
   }
 
+  def findDocuments(criteria: LibrarySearchParams) : Future[PerRequestMessage] = {
```
skwalker: nit: there's an extra space between the close parenthesis and the colon

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/LibraryApiService.scala
```diff
@@ -79,8 +80,30 @@ trait LibraryApiService extends HttpService with FireCloudRequestBuilding
...
+            post {
+              respondWithJSON {
+                entity(as[LibrarySearchParams]) { params => requestContext =>
+                  perRequest(requestContext,
+                    LibraryService.props(libraryServiceConstructor, userInfo),
+                    LibraryService.FindDocuments(params))
+                }
+              }
+            }
+//            get {
```
skwalker: can remove now

skwalker (commented):
This looks good! I'm wondering if we would want this endpoint to be /api/library/search instead of /api/libraries? Just since it is one library with many datasets in it?

src/main/scala/org/broadinstitute/dsde/firecloud/model/ESMapping.scala
```diff
@@ -1,7 +1,81 @@
...
+object ESWildcardObj {
+  def apply (term: String) :ESWildcard = ESWildcard(ESWildcardSearchTerm("*"+term+"*"))
+}
+
+case class ESQuery(query: Either[ESMatchAll,ESWildcard]) {
+}
+
+
+case class ESWildcardSearchTerm(_all: String) {
+}
```
davidangb: I liked where you were going with the QueryMap. How about something like this for the set of classes:

```
case class ESQuery(query: QueryMap)

sealed trait QueryMap

case object ESMatchAll extends QueryMap {
  val match_all = Map.empty[String,String]
}

case class ESWildcard(wildcard: ESWildcardSearchTerm) extends QueryMap {
  def apply (term: String) = ESWildcard(ESWildcardSearchTerm(s"*%s*".format(term)))
}
case class ESWildcardSearchTerm(_all: String)
```

couple points:
- unless you have a reason to create a new one each time, ESMatchAll can be a singleton. However, I don't know if spray-json knows how to de/serialize objects - you might need to keep this as a class.
- if your `apply` creates an `ESWildcard`, I think the apply should be in `ESWildcard`
- I've consolidated so that `ESWildcardObj` is no longer necessary
- at the syntax level, you can/should leave the `{}` off classes that have no body

Lastly, I don't think we should add the wildcards before/after - we should search on whatever the user entered, not munge it. Let's talk in person about that.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
```diff
@@ -93,4 +97,22 @@ class ElasticSearchDAO(servers:Seq[Authority], indexName: String) extends Search
...
+
+  override def findDocuments(criteria: LibrarySearchParams) : LibrarySearchResponse = {
+    val searchStr =
+      criteria.searchTerm match {
+        case None => ESQuery(Left(new ESMatchAll)).toJson.prettyPrint
+        case Some(searchTerm:String) => ESQuery(Right(ESWildcardObj(searchTerm))).toJson.prettyPrint
+      }
+    val searchReq = client.prepareSearch(indexName).setQuery(searchStr)
+    searchReq.setFrom(criteria.from.getOrElse(LibrarySearchConstants.defaultFrom))
+    searchReq.setSize(criteria.size.getOrElse(LibrarySearchConstants.defaultSize))
```
davidangb: nitpick: for syntax cleanliness, could this could be:

```
    val searchReq = client.prepareSearch(indexName).setQuery(searchStr)
        .setFrom(criteria.from.getOrElse(LibrarySearchConstants.defaultFrom))
        .setSize(criteria.size.getOrElse(LibrarySearchConstants.defaultSize))
```

this is pure style - so if you have preference in the other direction, Im ok.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
```diff
@@ -93,4 +97,22 @@ class ElasticSearchDAO(servers:Seq[Authority], indexName: String) extends Search
       case e: Exception => logger.warn(s"ES index '%s' could not be recreated and may be in an unstable state.".format(indexName))
     }
   }
+
+  override def findDocuments(criteria: LibrarySearchParams) : LibrarySearchResponse = {
+    val searchStr =
+      criteria.searchTerm match {
+        case None => ESQuery(Left(new ESMatchAll)).toJson.prettyPrint
+        case Some(searchTerm:String) => ESQuery(Right(ESWildcardObj(searchTerm))).toJson.prettyPrint
+      }
```
davidangb: as a tiny optimization, use `.compactPrint` instead of `.prettyPrint`, so we're sending slightly fewer bytes over the wire - this is only between services, right?

src/main/scala/org/broadinstitute/dsde/firecloud/model/ESMapping.scala
```diff
@@ -1,7 +1,81 @@
 package org.broadinstitute.dsde.firecloud.model
```
davidangb: I love the consolidation, but now the name of this file is misleading - it's much more than just the mapping.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/LibraryApiService.scala
```diff
@@ -79,8 +80,30 @@ trait LibraryApiService extends HttpService with FireCloudRequestBuilding
...
+//              respondWithJSON {
+//                parameters('from ? 0, 'size ? 10) { (from, size) =>
+//                  requestContext =>
+//                    perRequest(requestContext,
+//                      LibraryService.props(libraryServiceConstructor, userInfo),
+//                      LibraryService.FindDocuments(None, from, size))
+//                }
+//              }
+//            }
```
davidangb: delete instead of commenting out

davidangb (commented):
back to you, @ahaessly , comments inline!

ahaessly:
ok, i think i finally got all of your comments incorporated. i had some git issues so i had to force push this branch to origin. if you have trouble, let me know

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala
```diff
@@ -104,6 +104,22 @@ object ModelJsonProtocol {
     }
   }
 
+  implicit object impQueryMap extends JsonFormat[QueryMap] {
+    override def write(inputmap: QueryMap): JsValue = inputmap match {
+      case _ : ESMatchAll => inputmap.asInstanceOf[ESMatchAll].toJson
+      case _ : ESWildcard => inputmap.asInstanceOf[ESWildcard].toJson
+    }
```
davidangb, line 144:
```
    override def write(inputmap: QueryMap): JsValue = inputmap match {
      case matchall : ESMatchAll => matchall.toJson
      case wildcard : ESWildcard => wildcard.toJson
      case _ => throw new SerializationException("unexpected QueryMap type")
    }
```

if you use variable names when matching the type, the variables will automatically be the right type, and you don't have to manually cast.

davidangb (commented):
one scala syntax comment inline.

I also saw that you created a serializer for the QueryMap (yay!) ... and then I got all frustrated that my pet peeve of LibrarySearchParams didn't have one.

See #306 - would that work for you?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
```diff
@@ -101,7 +101,7 @@ class ElasticSearchDAO(servers:Seq[Authority], indexName: String) extends Search
   override def findDocuments(criteria: LibrarySearchParams) : LibrarySearchResponse = {
     val searchStr =
     criteria.searchTerm match {
-      case None => ESQuery(new ESMatchAll).toJson.compactPrint
+      case None | Some("") => ESQuery(new ESMatchAll).toJson.compactPrint
```
davidangb, line 103: are you concerned about whitespace-only strings, e.g. what if the user enters "   "?

You could do:

```
case None | Some(x) if x.trim == ""
```

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
```diff
@@ -102,11 +102,12 @@ class ElasticSearchDAO(servers:Seq[Authority], indexName: String) extends Search
     val searchStr =
     criteria.searchTerm match {
       case None | Some("") => ESQuery(new ESMatchAll).toJson.compactPrint
-      case Some(searchTerm:String) => ESQuery(new ESWildcard(searchTerm)).toJson.compactPrint
+      case Some(searchTerm:String) => ESQuery(new ESWildcard(searchTerm.toLowerCase)).toJson.compactPrint
```
davidangb, line 104: is the lowercase necessary? I didn't think ES did case-sensitive searches (and if it does, let's change that).

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala
```diff
@@ -104,6 +104,59 @@ object ModelJsonProtocol {
...
+        case None => None
+        case _ => throw DeserializationException("unexpected json type for " + SIZE)
+      }
+
+      (from, size) match {
+        case (None, None) => LibrarySearchParams(term)
+        case (Some(f), None) => LibrarySearchParams(term, f)
+        case (None, Some(s)) => LibrarySearchParams(term, size=s)
+        case (Some(f), Some(s)) => LibrarySearchParams(term, f, s)
+      }
```
davidangb: could you replace the entire `(from, size) match` with `LibrarySearchParams(term, from, size)`, and then inside of ModelJsonProtocol, have:

```
case class LibrarySearchParams(
  searchTerm: Option[String],
  from: Int = 0,
  size: Int = 10)

object LibrarySearchParams {
  def apply(searchTerm: Option[String], from: Option[Int], size: Option[Int]) = {
    new LibrarySearchParams(searchTerm, from.getOrElse(0), size.getOrElse(10))
  }
}
```

this does mean you have the defaults (0,10) in two places, but they're within a few lines of each other in the class and its companion. Disclaimer: I haven't actually tried this.

davidangb (approved):
one comment inline to be a little more scala-ish, but I'm okay even without that change. Thanks for sticking with this! 👍

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
ahaessly: the standard analyzer uses by default a token filter 
"lowercase". we can define a custom analyzer if we want case sensitive matching

davidangb:
👍 for latest commit.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
```diff
@@ -93,4 +97,22 @@ class ElasticSearchDAO(servers:Seq[Authority], indexName: String) extends Search
       case e: Exception => logger.warn(s"ES index '%s' could not be recreated and may be in an unstable state.".format(indexName))
     }
   }
+
+  override def findDocuments(criteria: LibrarySearchParams) : LibrarySearchResponse = {
+    val searchStr =
+    criteria.searchTerm match {
+      case None | Some("") => ESQuery(new ESMatchAll).toJson.compactPrint
```
skwalker, line 103: is the compactPrint necessary?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
```diff
@@ -93,4 +97,22 @@ class ElasticSearchDAO(servers:Seq[Authority], indexName: String) extends Search
...
+      case Some(searchTerm:String) => ESQuery(new ESWildcard(searchTerm.toLowerCase)).toJson.compactPrint
```
skwalker, line 104: nit: spacing

src/main/scala/org/broadinstitute/dsde/firecloud/model/ElasticSearch.scala
```diff
@@ -0,0 +1,74 @@
...
+trait QueryMap
+
+case class ESQuery(query: QueryMap)
+
+case class ESMatchAll(match_all: Map[String,String]) extends QueryMap {
+  def this() = this(Map.empty[String,String])
+}
+
+case class ESWildcard(wildcard: ESWildcardSearchTerm) extends QueryMap {
+  def this (term: String) = this(ESWildcardSearchTerm(term))
```
skwalker: nit: spacing

skwalker (approved):
looks good! I have a couple of nitpick-y comments but they really don't matter either way

PR #303, opened 2016-11-14
GAWB-23: Import and Export Workspace Attributes

ansingh7115:
3 new endpoints added for importing and exporting workspace attributes as tsvs

UserInfo case class was refactored so that it extends trait WithAccessToken. A new case class called AccessToken was created, which also extends WithAccessToken trait. This AccessToken is different from UserInfo in that it only has an access token, no email, id or expiration.

This new generic WithAccessToken replaces UserInfo in the WorkspaceService. It should be used more broadly, but that refactor will be in a different PR, as it will get a little too messy. 

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -76,4 +97,58 @@ class WorkspaceService(protected val argUserInfo: UserInfo, val rawlsDAO: RawlsD
...
+      val valueString = (attributes map { case (attName, attValue) => impPlainAttributeFormat.write(attValue) }).mkString("\t")
+      RequestCompleteWithHeaders((StatusCodes.OK, headerString + "\n" + valueString),
+        HttpHeaders.`Content-Disposition`.apply("attachment", Map("filename" -> filename)),
+        HttpHeaders.`Content-Type`(`text/plain`))
+    }
+  }
+
+  def importAttributesFromTSV(workspaceNamespace: String, workspaceName: String, tsvString: String): Future[PerRequestMessage] = {
+    Try(TSVParser.parse(tsvString)) match {
+      case Failure(regret) => Future.successful(RequestCompleteWithErrorReport(StatusCodes.BadRequest, regret.getMessage))
```
helgridly: this try/failure case is equivalent to `withTSVFile` in EntityClient. can you make them both use the same code?

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -76,4 +97,58 @@ class WorkspaceService(protected val argUserInfo: UserInfo, val rawlsDAO: RawlsD
...
+          case "workspace" =>
+            importWorkspaceAttributeTSV(workspaceNamespace, workspaceName, tsv)
+          case _ =>
+            Future.successful(RequestCompleteWithErrorReport(StatusCodes.BadRequest, "Invalid first column header should start with \"workspace\""))
+        }
+    }
+  }
+
+  private def importWorkspaceAttributeTSV(workspaceNamespace: String, workspaceName: String, tsv: TSVLoadFile): Future[PerRequestMessage] = {
+    if (tsv.tsvData.length != 1) {
```
helgridly: feedback from before: make this a utility function `checkTSVRequiredRows` or something

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -76,4 +97,58 @@ class WorkspaceService(protected val argUserInfo: UserInfo, val rawlsDAO: RawlsD
...
+      Future(RequestCompleteWithErrorReport(StatusCodes.BadRequest, "Your file does not have the correct number of rows. There should be 2."))
+    } else {
+      val attributeNames = Seq(tsv.headers.head.stripPrefix("workspace:")) ++ tsv.headers.tail
+      val distinctAttributes = attributeNames.distinct
+      if (attributeNames.size != distinctAttributes.size) {
+        Future(RequestCompleteWithErrorReport(StatusCodes.BadRequest, "Duplicated attribute keys are not allowed"))
```
helgridly: don't we handle this in entity tsvs too?

helgridly: (thinking about code duplication here)

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -76,4 +97,58 @@ class WorkspaceService(protected val argUserInfo: UserInfo, val rawlsDAO: RawlsD
...
+      } else {
+        rawlsDAO.getWorkspace(workspaceNamespace, workspaceName) flatMap { workspaceResponse =>
+          Try(getWorkspaceAttributeCalls(attributeNames.zip(tsv.tsvData.head))) match {
+            case Failure(regret) => Future.successful(RequestCompleteWithErrorReport(StatusCodes.BadRequest,
+              "One or more of your values are not in the correct format"))
```
helgridly, line 120: this too.

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -76,4 +97,58 @@ class WorkspaceService(protected val argUserInfo: UserInfo, val rawlsDAO: RawlsD
...
+            case Success(attributeCalls) => rawlsDAO.patchWorkspaceAttributes(workspaceNamespace, workspaceName, attributeCalls) map (RequestComplete(_))
+          }
+        }
+      }
+    }
+  }
+
+  private def getWorkspaceAttributeCalls(attributePairs: Seq[(String,String)]): Seq[AttributeUpdateOperation] = {
+    attributePairs.map { case (name, value) =>
+      if (value.equals("__DELETE__"))
```
helgridly: This definitely exists for entities. We should make more of an effort to share this code rather than duplicate it.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupport.scala
```diff
@@ -0,0 +1,113 @@
...
+    val entitiesToUpdate = tsv.tsvData.map(_.headOption.get)
+    val distinctEntities = entitiesToUpdate.distinct
+    if ( entitiesToUpdate.size != distinctEntities.size ) {
+      Future( RequestCompleteWithErrorReport(BadRequest,
+        "Duplicated entities are not allowed in TSV: " + entitiesToUpdate.diff(distinctEntities).distinct.mkString(", ")) )
+    } else {
+      op
+    }
+  }
+
```
ansingh7115, line 62: So, upon further investigation, getWorkspaceAttributeCalls (used by WorkspaceService) and setAttributesOnEntity (used by EntityClient) can't really share code (both below). This is because WorkspaceService uses rawlsDAO.patchWorkspaceAttributes(...) which takes a Seq[AttributeUpdateOperation] while EntityClient uses batchCalltoRawls which passes the updates on as Seq[EntityUpdateDefinition]

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -84,21 +70,6 @@ class EntityClient (requestContext: RequestContext) extends Actor with FireCloud
...
-          " TSV: " + entitiesToUpdate.diff(distinctEntities).distinct.mkString(", ")) )
-    } else {
-      op
-    }
-  }
-
-  /**
    * Bail with a 400 Bad Request if the tsv is trying to set members on a collection type.
    * Otherwise, carry on. */
   private def checkNoCollectionMemberAttribute( tsv: TSVLoadFile, memberTypeOpt: Option[String] )(op: => Future[PerRequestMessage]): Future[PerRequestMessage] = {
```
ansingh7115: I'm unsure if I should move functions like this (and validateMembershipTSV) to TSVFileSupport - they do stuff with TSVLoadFiles but they are only used in EntityClient and not on WorkspaceService.

helgridly: My gut says yes - consolidating TSV operations in TSVFileSupport seems like the right move to me.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupport.scala
helgridly: Under the hood these both use the same data structures: compare `upsertAttrOperation` (line 5 below) with the `write` function in AttributeUpdateOperations.scala - they both construct the same JSON:

```
{
    "op" : "AddUpdateAttribute",
    "attributeName" : "my_attribute",
    "attributeValue" : 42
}
```

So the inside of the first block of `setAttributesOnEntity` builds something the same shape as `getWorkspaceAttributeCalls`, and with a bit of fiddling I think you could get some reuse out of it.

That said: for the sake of getting this PR out the door, I'm willing to come back to this refactor later, as long as we don't make a habit of doing so.

PR #309, opened 2016-11-17
Export selected columns

dvoet:
https://broadinstitute.atlassian.net/browse/GAWB-1232
Added an optional parameter to tsv download end point to specify attribute names, in order, to be in the resulting tsv. Note that this does not trickle down to rawls so the full data set still goes over the wire between orch and rawls.... meh. The first commit deals with just about all of this.

Also refactored the areas I touched to match new orch structure. The refactor also fixed https://broadinstitute.atlassian.net/browse/GAWB-1305 because the new code does not custom parse json responses and I think the regression came with the attribute list format refactor.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
```diff
@@ -11,102 +10,98 @@ import spray.json.JsValue
...
+
+    val headers: IndexedSeq[String] = entityHeader +: requestedHeadersSansId.getOrElse(defaultHeaders(entityType, filteredEntities, headerRenamingMap))
+    val rows: IndexedSeq[IndexedSeq[String]] = filteredEntities.filter { _.entityType == entityType }
       .map { entity =>
         makeRow(entity, headers, headerRenamingMap)
       }.toIndexedSeq
     exportToString(headers, rows)
   }
 
+  def defaultHeaders(entityType: String, filteredEntities: Seq[RawlsEntity], headerRenamingMap: Map[String, String]) = {
```
breilly2, line 62: Nice extracted function and simplification. I'll have to remember to keep `collect` in mind.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
```diff
@@ -11,102 +10,98 @@ import spray.json.JsValue
...
+    * @return new list of RawlsEntity
     */
-  private def filterAttributeFromEntities(entities: List[EntityWithType], attributeName: String): List[EntityWithType] = {
+  private def filterAttributeFromEntities(entities: Seq[RawlsEntity], attributeName: String): Seq[RawlsEntity] = {
     entities map {
       entity =>
-        val attributes = entity.attributes.getOrElse(Map.empty) filterNot {
-          attribute => attribute._1.equals(attributeName)
+        val attributes = entity.attributes filterNot {
+          case (thisAttributeName, _) => thisAttributeName == AttributeName.withDefaultNS(attributeName)
```
breilly2, line 86: Another nice trick I'll have to keep in mind.

breilly2:
👍 Good test coverage added for GAWB-1232.
I'm not certain that I see how GAWB-1305 was fixed. Does it need separate testing or is the problem surfaced in the existing tests now that they're using the Attribute\* case classes/objects (or maybe something completely different)?

breilly2:
I also appreciate the refactoring. Took a bit of comparison to understand what changed, but I think the pattern that you're pushing forward makes sense, and it's what I would expect after working in Rawls.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
```diff
@@ -11,102 +10,98 @@ import spray.json.JsValue
...
-    val rows: List[IndexedSeq[String]] = entities.filter { _.entityType == entityType }.flatMap {
+    val rows: Seq[IndexedSeq[String]] = entities.filter { _.entityType == entityType }.flatMap {
       entity =>
-        entity.attributes.getOrElse(Map.empty).filter {
+        entity.attributes.filter {
           // To make the membership file, we need the array of elements that correspond to the set type.
           // All other top-level properties are not necessary and are only used for the data load file.
-          attribute => attribute._1.equals(collectionMemberType) }.flatMap {
-          m =>
-            Try(m._2.convertTo[List[JsValue]]) match {
```
dvoet, line 23: @breilly2 GAWB-1305 was fixed here. The old code attempted to treat the attribute value as a list and if it couldn't just return IndexedSeq.empty (which is why the file was empty). The new way doesn't touch json so the problem is no longer possible.

PR #312, opened 2016-11-21
GAWB-1174 estimate storage cost

breilly2:
The first commit implements the functionality and includes a test that depends on the current price list. The remaining commits include clean-up and introducing a `MockGoogleServicesDAO` so that the test won't break when storage prices change.

This in only the orchestration layer. I have not made UI changes yet.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/RawlsDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/RawlsDAO.scala
@@ -29,6 +29,7 @@ trait RawlsDAO extends LazyLogging {
   lazy val rawlsCuratorUrl = FireCloudConfig.Rawls.authUrl + "/user/role/curator"
   lazy val rawlsAdminWorkspaces = FireCloudConfig.Rawls.authUrl + "/admin/workspaces?attributeName=library:published&valueBoolean=true"
   lazy val rawlsWorkspaceACLUrl = FireCloudConfig.Rawls.workspacesUrl + "/%s/%s/acl"
+  lazy val rawlsBucketUsageUrl = FireCloudConfig.Rawls.workspacesUrl + "/%s/%s/bucketUsage"
   def rawlsEntitiesOfTypeUrl(workspaceNamespace: String, workspaceName: String, entityType: String) = FireCloudConfig.Rawls.workspacesUrl + s"/$workspaceNamespace/$workspaceName/entities/$entityType"
 
   def isAdmin(userInfo: UserInfo): Future[Boolean]
@@ -36,6 +37,8 @@ trait RawlsDAO extends LazyLogging {
   def isDbGapAuthorized(userInfo: UserInfo): Future[Boolean]
 
   def isLibraryCurator(userInfo: UserInfo): Future[Boolean]
+
+  def getBucketUsage(ns: String, name: String)(implicit userInfo: UserInfo): Future[RawlsBucketUsageResponse]
 
   def getWorkspace(ns: String, name: String)(implicit userToken: WithAccessToken): Future[RawlsWorkspaceResponse]
 
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/RawlsDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/RawlsDAO.scala
@@ -38,7 +38,7 @@ trait RawlsDAO extends LazyLogging {
 
   def isLibraryCurator(userInfo: UserInfo): Future[Boolean]
 
-  def getBucketUsage(ns: String, name: String)(implicit userInfo: UserInfo): Future[RawlsBucketUsageResponse]
+  def getBucketUsage(ns: String, name: String)(implicit userInfo: WithAccessToken): Future[RawlsBucketUsageResponse]
 
   def getWorkspace(ns: String, name: String)(implicit userToken: WithAccessToken): Future[RawlsWorkspaceResponse]
 
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/RawlsDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/RawlsDAO.scala
@@ -29,7 +29,7 @@ trait RawlsDAO extends LazyLogging {
   lazy val rawlsCuratorUrl = FireCloudConfig.Rawls.authUrl + "/user/role/curator"
   lazy val rawlsAdminWorkspaces = FireCloudConfig.Rawls.authUrl + "/admin/workspaces?attributeName=library:published&valueBoolean=true"
   lazy val rawlsWorkspaceACLUrl = FireCloudConfig.Rawls.workspacesUrl + "/%s/%s/acl"
-  lazy val rawlsBucketUsageUrl = FireCloudConfig.Rawls.workspacesUrl + "/%s/%s/bucketUsage"
+  def rawlsBucketUsageUrl(workspaceNamespace: String, workspaceName: String) = FireCloudConfig.Rawls.workspacesUrl + s"/$workspaceNamespace/$workspaceName/bucketUsage"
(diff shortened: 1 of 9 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -70,6 +73,19 @@ class WorkspaceService(protected val argUserToken: WithAccessToken, val rawlsDAO
...
 
+  def getStorageCostEstimate(workspaceNamespace: String, workspaceName: String) = {
+    Future.sequence(Seq(
+      rawlsDAO.getBucketUsage(workspaceNamespace, workspaceName),
+      googleServicesDAO.fetchPriceList
+    )) map {
+      case Seq(usage: RawlsBucketUsageResponse, priceList: GooglePriceList) =>
+        val rate = priceList.prices.cpBigstoreStorage.us
+        val estimate: BigDecimal = BigDecimal(usage.usageInBytes) / 1000000000 * rate
+        RequestComplete(WorkspaceStorageCostEstimate(f"$$$estimate%.2f"))
```
breilly2, line 81: Choosing to fully format the dollar value server-side to avoid having to do any rounding/formatting client-side. Also, the price list currently only gives us US dollars, so there are no other formatting options.

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -39,10 +40,10 @@ object WorkspaceService {
   }
 
   def constructor(app: Application)(userToken: WithAccessToken)(implicit executionContext: ExecutionContext) =
-    new WorkspaceService(userToken, app.rawlsDAO, app.thurloeDAO)
+    new WorkspaceService(userToken, app.rawlsDAO, app.thurloeDAO, app.googleServicesDAO)
 }
 
-class WorkspaceService(protected val argUserToken: WithAccessToken, val rawlsDAO: RawlsDAO, val thurloeDAO: ThurloeDAO) extends Actor with AttributeSupport with TSVFileSupport {
+class WorkspaceService(protected val argUserToken: WithAccessToken, val rawlsDAO: RawlsDAO, val thurloeDAO: ThurloeDAO, val googleServicesDAO: GoogleServicesDAO) extends Actor with AttributeSupport with TSVFileSupport {
```
breilly2, line 46: This is the only place where I made changes to inject a mock `GoogleServicesDAO` for tests. Other places that currently use `HttpGoogleServicesDAO` directly are traits which would require deeper refactoring to allow for injection of a mock. I'd like to leave that until it is actually needed. Let me know if you feel strongly that the refactoring should be done now.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -44,7 +44,7 @@ object GooglePriceListJsonProtocol extends DefaultJsonProtocol {
 }
 import org.broadinstitute.dsde.firecloud.dataaccess.GooglePriceListJsonProtocol._
 
-object HttpGoogleServicesDAO extends FireCloudRequestBuilding {
+object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuilding {
```
dvoet, line 61: this probably should not be an object but that could be for another day

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/RawlsDAO.scala
```diff
@@ -17,12 +17,15 @@ trait RawlsDAO extends LazyLogging {
   lazy val rawlsCuratorUrl = FireCloudConfig.Rawls.authUrl + "/user/role/curator"
   lazy val rawlsAdminWorkspaces = FireCloudConfig.Rawls.authUrl + "/admin/workspaces?attributeName=library:published&valueBoolean=true"
   lazy val rawlsWorkspaceACLUrl = FireCloudConfig.Rawls.workspacesUrl + "/%s/%s/acl"
+  lazy val rawlsBucketUsageUrl = FireCloudConfig.Rawls.workspacesUrl + "/%s/%s/bucketUsage"
```
dvoet: I prefer the style of the following next line of code. Use string interpolation instead of a format string with a later call to .format. It combines the string definition and the function into 1 line, removes a string that is otherwise useless and maybe confusing, and allows us to not have a getter function.

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -70,6 +73,19 @@ class WorkspaceService(protected val argUserToken: WithAccessToken, val rawlsDAO
 
   }
 
+  def getStorageCostEstimate(workspaceNamespace: String, workspaceName: String) = {
+    Future.sequence(Seq(
```
dvoet: instead of Future.sequence I think you want zip, ```rawlsDAO.getBucketUsage(workspaceNamespace, workspaceName).zip(googleServicesDAO.fetchPriceList)```

I am a little concerned about type erasure in the code as it stands.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiService.scala
```diff
@@ -159,6 +159,13 @@ trait WorkspaceApiService extends HttpService with FireCloudRequestBuilding
             } ~
             path("unlock") {
               passthrough(workspacePath + "/unlock", HttpMethods.PUT)
+            } ~
+            path("storageCostEstimate") {
+              get { requestContext =>
+                perRequest(requestContext,
+                  WorkspaceService.props(workspaceServiceConstructor, userInfo),
+                  WorkspaceService.GetStorageCostEstimate(workspaceNamespace, workspaceName))
+              }
```
dvoet, line 201: can we expose a pass through that is the storage in bytes as well? Seems useful from a general api perspective. Or can we include the size in the cost estimate response?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/RawlsDAO.scala
breilly2: I like that better, too. I just followed the wrong pattern because it was the more prevalent one. I'll look into updating them all to avoid the same mistake in the future.

breilly2: The only other one here that uses `format` is `rawlsWorkspaceACLUrl`. However, it's not currently covered by any tests. They'd need to be integration tests which firecloud-orchestration currently does not have and I am not up to setting that up right now. I'll fix mine, but unfortunately I'm going to have to leave `rawlsWorkspaceACLUrl` as is.

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
breilly2: Yes, that makes sense. I didn't like having to explicitly specify the types. I guess maybe `Future.sequence` works better when all the futures are of the same type?

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiService.scala
breilly2: I'm a little hesitant to add something we don't currently need (more to test and maintain), but the passthrough is simple enough so I will add that.

breilly2:
Updated with changes based on all feedback and rebased onto develop as of this morning.

MatthewBemis:
:+1:

dvoet (approved):
👍

PR #313, opened 2016-11-21
GAWB-1126: duos autocomplete, orch part

skwalker:
local config PR: https://github.com/broadinstitute/firecloud-develop/pull/205
ui PR: https://github.com/broadinstitute/firecloud-ui/pull/577

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off


src/main/resources/library/attribute-definitions.json
```diff
@@ -4,14 +4,16 @@
   "title": "Library attribute definitions, v1",
   "description": "Constraints, facet definitions, and display definitions for FireCloud Library",
   "type": "object",
-  "required": ["library:datasetName", "library:datasetDescription", "library:datasetCustodian",
-               "library:datasetDepositor", "library:datasetOwner", "library:institute", "library:indication", "library:numSubjects", "library:projectName", "library:datatype",
+  "required": ["library:datasetName", "library:datasetDescription", "library:datasetCustodian", "library:datasetDepositor",
+               "library:datasetOwner", "library:institute", "library:numSubjects", "library:projectName", "library:datatype",
```
davidangb: is indication no longer required?

src/main/resources/library/attribute-definitions.json
```diff
@@ -4,14 +4,16 @@
...
-                    "library:dataUseORSPConsentGroupNumber", "library:dataUseBSPSampleCollectionID", "library:studyDesign", "library:cellType", "library:coverage",
-                    "library:ethnicity", "library:primaryDiseaseSite", "library:broadInternalResearchProjectID", "library:broadInternalResearchProjectName",
+  "propertyOrder": ["library:numSubjects", "library:datatype", "library:dataUseRestriction", "library:diseaseOntology",
+                    "library:indication", "library:datasetName", "library:datasetDescription", "library:datasetCustodian",
+                    "library:datasetDepositor", "library:datasetOwner", "library:institute",  "library:projectName",
+                    "library:reference", "library:dataFileFormats", "library:technology", "library:profilingProtocol",
+                    "library:dataUseORSPConsentGroupNumber", "library:dataUseBSPSampleCollectionID", "library:studyDesign",
+                    "library:cellType", "library:coverage", "library:ethnicity", "library:primaryDiseaseSite",
+                    "library:broadInternalResearchProjectID", "library:broadInternalResearchProjectName",
```
davidangb: could you explain why the order changed?

src/main/resources/library/attribute-definitions.json
```diff
@@ -43,9 +45,18 @@
...
     },
+    "library:diseaseOntologyDOID": {
+      "type": "string",
+      "hidden": true
+    },
+    "library:diseaseOntology": {
+      "type": "string",
+      "title": "Cohort Phenotypic Inclusion Criteria (Disease Ontology)",
+      "typeahead": "ontology"
+    },
```
davidangb, line 54: not necessary for this PR, but: at some future point we may need to indicate *which* ontology gets used for typeahead.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/LibraryApiService.scala
```diff
@@ -20,10 +20,17 @@ trait LibraryApiService extends HttpService with FireCloudRequestBuilding
   private implicit val ec: ExecutionContext = actorRefFactory.dispatcher
 
   lazy val rawlsCuratorUrl = FireCloudConfig.Rawls.authUrl + "/user/role/curator"
+  val duosAutocompleteUrl = FireCloudConfig.Duos.baseUrl + "/autocomplete?types=disease&q=%s"
 
   val libraryServiceConstructor: UserInfo => LibraryService
 
   val libraryRoutes: Route =
+    path("duos-autocomplete" / Segment) { (searchTerm) =>
```
davidangb: nit: I'd make this `"duos" / "autocomplete"` ... that way, we can add more endpoints under `library/duos/` in a nice consistent way.

davidangb (commented):
a few comments inline, most importantly changes to the schema that aren't documented as requirements anywhere.

src/main/resources/library/attribute-definitions.json
skwalker: I was trying to get the order to work so that the diseaseOntology would be directly followed by indication in both edit and view mode, which Moran said she wanted. Then I realized that the order here doesn't really work because of the primary properties, which supersedes this -- only in the view mode. So I tried to change it so that the order would match in edit and view mode, because I thought that would make more sense.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/LibraryApiService.scala
skwalker: ahh, good call

src/main/resources/library/attribute-definitions.json
skwalker: Need to talk to Moran about this, I think I might have gotten confused / shouldn't have made this change.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/LibraryApiService.scala
```diff
@@ -20,10 +20,17 @@ trait LibraryApiService extends HttpService with FireCloudRequestBuilding
...
 
   val libraryServiceConstructor: UserInfo => LibraryService
 
   val libraryRoutes: Route =
+    path("duos" / "autocomplete" / Segment) { (searchTerm) =>
+      get { requestContext =>
+        val extReq = Get(duosAutocompleteUrl.format(searchTerm))
+        externalHttpPerRequest(requestContext, extReq)
+      }
+    } ~
```
davidangb, line 33: needs swagger!

src/main/resources/library/attribute-definitions.json
davidangb: understood. Add this info to the jira ticket (or ask Moran to do it) - otherwise an eagle-eyed QA might reject it.

src/main/resources/library/attribute-definitions.json
```diff
@@ -43,9 +46,18 @@
...
```
davidangb, line 54: My one request towards generic-ness: rename these attrs to "fooId" and "fooLabel" ... so, `library:diseaseOntologyId` and `library:diseaseOntologyLabel`. While we don't need to make our code totally generic, it would be a pain to have to update attributes already saved onto workspaces.

By using a naming convention of "fooId" and "fooLabel", we could write code in the future that programmatically generates the names. So, if we added a new ontology-driven property for, say, the research area, we could know to expect `library:researchAreaId" and `library:researchAreaLabel`.

Unfortunately, renaming these here will require code changes in the UI.

davidangb (commented):
comments inline!

src/main/resources/library/attribute-definitions.json
ahaessly: what is the value of typeahead used for? (just thinking about when will do autocomplete for ES)

skwalker: the value of typeahead is used in the UI so that we deal with it properly, with the select2 typeahead

src/main/resources/library/attribute-definitions.json
skwalker: Totally meant to do this in response to your UI comments yesterday, but then I forgot..

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1852,6 +1852,20 @@ paths:
...
+      - Library
+      operationId: duosAutocomplete
+      summary: |
+        Searches DUOS for a disease ontology.
+      responses:
+        200:
+          description: Matching disease ontologies.
+        500:
+          description: Internal Server Error
+      security: []
```
ahaessly, line 1873: is security not required because it's just a passthrough to duos which is open?

skwalker: yep, exactly

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1852,6 +1852,20 @@ paths:
...
       security: []
 
+  /duos/autocomplete/{queryTerm}:
+    get:
+      tags:
+      - Library
+      operationId: duosAutocomplete
+      summary: |
+        Searches DUOS for a disease ontology.
+      responses:
```
davidangb, line 1868: this needs:
```
      parameters:
        - name: queryTerm
          in: path
          description: The query term (word fragment) which the service should try and complete.
          required: true
          type: string
```
otherwise the swagger is nonfunctional - no way to enter the term.

davidangb:
👍

skwalker:
OK, talked to Moran and got things into the order she wants so this PR is ready to be submitted once people take a final look and make sure I didn't do anything stupid.

davidangb:
:+1: again for latest changes

PR #315, opened 2016-11-22
Remove logging noise in tests

dmohs:
With this change, the only log noise in tests is this:

```
[MockServer thread for port: 8990] i.n.util.internal.ThreadLocalRandom - Failed to generate a seed from SecureRandom within 3 seconds. Not enough entrophy?
```

I have filed a separate issue to get that fixed:

https://github.com/jamesdbloom/mockserver/issues/299

### DIFF
--- a/src/test/resources/reference.conf
+++ b/src/test/resources/reference.conf
@@ -1,7 +1,7 @@
 include "src/main/resources/reference.conf"
 
 akka {
-  loglevel = "OFF"
+  loglevel = "ERROR"
 }
 
 auth {
--- a/src/test/resources/reference.conf
+++ b/src/test/resources/reference.conf
@@ -1,5 +1,9 @@
 include "src/main/resources/reference.conf"
 
+akka {
+  loglevel = "OFF"
+}
+
 auth {
   googleClientId = "dummy"
   googleSecretsJson = """{"web":{"auth_provider_x509_cert_url":"","auth_uri":"","client_id":"","client_secret":"","javascript_origins":[],"redirect_uris":[],"token_uri":""}}"""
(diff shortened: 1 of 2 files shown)


davidangb:
This is super-small, but I'd still prefer to use `loglevel = "ERROR"` instead, so we see true errors.

There *is* one error currently happening in tests, which is fallout from #297 - that other PR specifies "/dev/null" for a pem file, which ends up causing a null pointer elsewhere. We should resolve that error the right way.

thanks for doing this.

dmohs:
If I provide a valid pem file, I get this:
```
[ERROR] [11/22/2016 18:43:26.814] [org-broadinstitute-dsde-firecloud-service-NIHSyncEndpointsSpec-akka.actor.default-dispatcher-3] [akka://org-broadinstitute-dsde-firecloud-service-NIHSyncEndpointsSpec/user/perRequest$default$41810386095434746/perRequest$default$41810386096650592] 401 Unauthorized
com.google.api.client.auth.oauth2.TokenResponseException: 401 Unauthorized
```
I tried to see if I could just turn off logging for this one test, but I can't see a way to do that. Since I expect tests to trigger errors as part of what we test, I feel like `OFF` may be the correct option for the general case.

dvoet:
👍  whatever

davidangb:
I feel pretty strongly we should not be suppressing valid errors. There's only one test that throws an error - and after investigating, it's clear the test itself is invalid and needs work. See #317 (assigned to you, @dmohs ) - I've used ERROR loglevel and disabled the crappy test. End result is the same - noise is reduced in the test suite.

dmohs:
Confirmed that running with "ERROR" level and the ignored test gives nice clean output (except for the "entrophy" warning).

PR #318, opened 2016-11-28
GAWB-1130: Updating attribute definitions to support new wizard

MattPutnam:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [x] **LR** sign off


src/main/resources/library/attribute-definitions.json
```diff
@@ -4,146 +4,281 @@
...
-               "library:datasetDepositor", "library:datasetOwner", "library:institute", "library:indication", "library:numSubjects", "library:projectName", "library:datatype",
-               "library:dataUseRestriction", "library:studyDesign", "library:cellType"],
-  "propertyOrder": ["library:datasetName", "library:datasetDescription", "library:datasetCustodian",
-                    "library:datasetDepositor", "library:datasetOwner", "library:institute", "library:indication", "library:numSubjects", "library:projectName", "library:datatype",
-                    "library:reference", "library:dataFileFormats", "library:technology", "library:profilingProtocol", "library:dataUseRestriction",
-                    "library:dataUseORSPConsentGroupNumber", "library:dataUseBSPSampleCollectionID", "library:studyDesign", "library:cellType", "library:coverage",
-                    "library:ethnicity", "library:primaryDiseaseSite", "library:broadInternalResearchProjectID", "library:broadInternalResearchProjectName",
-                    "library:broadInternalCohortName", "library:broadInternalSeqProjectNumbers"],
```
rushtong: Thank you for pulling these out individually. Small changes are more github-review-friendly with the less dense format.

rushtong:
👍

davidangb:
Look at `/src/test/resources/test-attribute-definitions.json`. We have separate test and runtime definitions in the thought that PO could make minor changes to the runtime definitions without disturbing test framework. That may be a bad assumption.

But either way ... with a refactoring this major, we need to update the test file and see what, if anything, breaks, and deal with it. We could take this opportunity to have the tests just use the runtime file, too.

davidangb:
after looking at this more in-depth ... removing the top-level `required` section breaks validation - in that it allows invalid documents to pass. #324 changes the unit tests to use the runtime schema, and shows that a bunch of tests fail.

rushtong:
The `AttributeName` case class in `org.broadinstitute.dsde.firecloud.model.Attributable` assumes `default:` and `library:...` prefixes. Not sure how well that will play with this renaming.

src/test/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSpec.scala
```diff
@@ -314,65 +314,38 @@ class LibraryServiceSpec extends FreeSpec with LibraryServiceSupport with Attrib
...
-      }
-      "fails with two missing keys" in {
-        val testSchema = FileUtils.readAllTextFromResource("test-attribute-definitions.json")
-        val defaultData = testLibraryMetadata.parseJson.asJsObject
-        val sampleData = defaultData.copy(defaultData.fields-"library:datasetName"-"library:datasetOwner").compactPrint
-        val ex = intercept[ValidationException] {
-          validateJsonSchema(sampleData, testSchema)
-        }
-        assertResult(2){ex.getViolationCount}
-      }
```
MattPutnam: See https://broadinstitute.atlassian.net/browse/GAWB-1389

rushtong:
`LibraryServiceSpec` is using the prefixed properties for testing.

src/test/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSpec.scala
rushtong, line 343: Not your code, but all of the schema tests in this class should be pulled out - they have nothing to do with `LibraryService`. Likewise with a bunch of attribute tests.

rushtong:
Ignore my comments about `library:...` - I was thrown off by the disconnect between the removing it from the schema and what actually exists as workspace attributes prefixed with `library:`. That prefix should still exist and needs to be handled.

rushtong:
👍

davidangb:
Hmmm ... I still don't like moving all the `required` logic into the wizard section. I do understand there is conditional logic in there that would need work (and could be done later), but there's still a lot of attributes that are straight-up simply required.

The `attribute-definitions.json` is a json schema definition, and the json-schema standard defines where `required` should live (http://json-schema.org/latest/json-schema-validation.html#rfc.section.5.15). Your branch still validates incoming data (https://github.com/broadinstitute/firecloud-orchestration/blob/mp-gawb-1130/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala#L68), but at this point it's limited to validating data types, enums, and minimums for numeric fields - we've dropped all the required stuff.

Moving the `required` definition means we can't take advantage of third-party libraries, and makes us non-standard.

src/test/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSpec.scala
```diff
@@ -315,25 +316,24 @@ class LibraryServiceSpec extends FreeSpec with LibraryServiceSupport with Attrib
...
       "fails with one missing key" in {
-        val testSchema = FileUtils.readAllTextFromResource("test-attribute-definitions.json")
+        val testSchema = FileUtils.readAllTextFromResource("library/attribute-definitions.json")
         val defaultData = testLibraryMetadata.parseJson.asJsObject
         val sampleData = defaultData.copy(defaultData.fields-"library:datasetName").compactPrint
         val ex = intercept[ValidationException] {
           validateJsonSchema(sampleData, testSchema)
         }
         assertResult(1){ex.getViolationCount}
-        assert(ex.getMessage.contains("library:datasetName"))
```
MattPutnam, line 333: This assert doesn't work any more.  Because of the more complicated validation, these always fail with something like "only 1 subschema matches out of 2".

src/main/resources/library/attribute-definitions.json
```diff
@@ -5,145 +5,292 @@
...
       "items": { "type": "string" },
-      "title": "Seq Project Numbers",
-      "inputHint": "Broad internal IDs"
+      "consentCode": "DS"
+    },
+    "library:NCU": {
+      "title": "Future commercial use is prohibited",
+      "type": "string",
+      "enum": ["Yes", "No"],
+      "consentCode": "NCU"
```
davidangb, line 154: why use a string, with Yes/No enum, instead of a boolean?

src/main/resources/library/attribute-definitions.json
```diff
@@ -5,145 +5,292 @@
...
+      "library:broadInternalSeqProjectNumbers"],
+    "consentCodes": ["GRU", "HMB", "DS", "NCU", "NPU", "NDMS", "NAGR", "NCTRL", "RS-G", "RS-PD", "RS-POP"]
+  },
+  "wizard":
+  [
+    {
+      "title": "Define Data Set attributes",
+      "questions":
+      [
+        {"property": "library:datasetName", "required": true},
```
davidangb, line 236: the `required` clauses are now duplicated between the wizard and the top-level required statement - could that be avoided?

src/test/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSpec.scala
davidangb: could you change the assert to look for the expected error message, instead of removing it? (in multiple places) ... see #329.

davidangb (commented):
see comments/questions inline.

I didn't like losing the specificity on some of the unit tests, so I figured out what was going on with them - see #329.

src/main/resources/library/attribute-definitions.json
rushtong: I'm confused by why we're defining property attributes in two places. Wouldn't it be more clear if `required`, etc., were part of the top-level properties? Then the list of questions could just be references to the top level properties instead.

src/main/resources/library/attribute-definitions.json
```diff
@@ -5,145 +5,292 @@
...
+        {"property": "library:reference", "inputHint": "e.g. hg19, GRC38;  To which genome build the data was aligned, if relevant"},
+        {"property": "library:dataFileFormats", "inputHint": "e.g. VCF, BAM; Tag all relevant"},
+        {"property": "library:technology", "inputHint": "e.g. Illumina, 10X"},
+        {"property": "library:profilingProtocol"},
+        {"property": "library:dataUseRestriction", "required": true},
+        {"property": "library:dataUseBSPSampleCollectionID"},
+        {"property": "library:studyDesign", "inputHint": "e.g Case/Control, Trio, Tumor/normal, cases only - somatic,  cases only - germline, controls", "required": true},
+        {"property": "library:cellType", "required": true},
+        {"property": "library:coverage"},
+        {"property": "library:ethnicity", "inputHint": "e.g. Caucasians, African-americans, Latino, East asians, South Asians, Finnish, Non-Finnish Europeans; check all relevant"},
```
rushtong: Should `African-americans` be `African-Americans` ?

rushtong: Similar question about `East asians, South Asians`

src/main/resources/library/attribute-definitions.json
MattPutnam: I was thinking (perhaps too) generically--it's already not as simple as saying that a prop is required or not required.  I'm envisioning a scenario where the same question is required in one circumstance but not another.

src/main/resources/library/attribute-definitions.json
MattPutnam: Yeah, I need to update the properties with the latest from Moran.

src/main/resources/library/attribute-definitions.json
MattPutnam: I wanted to control whether it was "Yes/No" or "True/False" or whatever else.  There's already a "Yes/No/Unspecified".

davidangb: Hm. I'd think if all boolean properties are either "Yes/No" or "True/False", then it's an issue for UI rendering - the underlying value is a boolean, the UI decides how to display it. This becomes more important when we consider future use - at some point we'll be performing backend logic on these values (`if (NCU) then foo else bar`) ... that logic will have to parse strings to understand their values - and they'll have to be the same string values as were saved as of a certain version of the schema.

"Yes/No/Unspecified" could be boolean true/false/null.

I won't hold this PR for this, but think we'll need to deal with it sooner rather than later.

PR #320, opened 2016-11-29
Ah gawb 976 - Library: term facets; facet definitions for 4 

ahaessly:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -345,6 +345,22 @@ paths:
         200:
           description: Matching datasets
 
+  /api/library/search/aggregations:
+    post:
+      tags:
+        - Library
+      operationId: GetAggregations
+      summary: Returns the aggregation information for your term
+      parameters:
+        - name: fields
+          description: attribute to get aggregations for
+          required: true
+          schema:
+            $ref: '#/definitions/AggregationRef'
+          in: body
+      responses:
+        200:
+          description: aggregations
   /api/library/user/role/curator:
     get:
       responses:
@@ -2087,6 +2103,20 @@ parameters:
 ## DEFINITIONS
 ##########################################################################################
 definitions:
+
+  AggregationRef:
+    properties:
+      fields:
+        type: array
+        items:
+          type: string
+        description: The fields to get aggregations for
+      maxResults:
+        type: integer
+        default: 10
+        description: How many results to return, will be applied to all fields
+    required:
+      - fields
 
   AttributeUpdateOperation:
     properties:
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -345,22 +345,6 @@ paths:
         200:
           description: Matching datasets
 
-  /api/library/search/aggregations:
-    post:
-      tags:
-        - Library
-      operationId: GetAggregations
-      summary: Returns the aggregation information for your term
-      parameters:
-        - name: fields
-          description: attribute to get aggregations for
-          required: true
-          schema:
-            $ref: '#/definitions/AggregationRef'
-          in: body
-      responses:
-        200:
-          description: aggregations
   /api/library/user/role/curator:
     get:
       responses:
@@ -2104,20 +2088,6 @@ parameters:
 ##########################################################################################
 definitions:
 
-  AggregationRef:
-    properties:
-      fields:
-        type: array
-        items:
-          type: string
-        description: The fields to get aggregations for
-      maxResults:
-        type: integer
-        default: 10
-        description: How many results to return, will be applied to all fields
-    required:
-      - fields
-
   AttributeUpdateOperation:
     properties:
       op:
@@ -2750,12 +2720,20 @@ definitions:
 
   SearchTermRef:
     properties:
-      searchTerm:
+      searchString:
         type: string
         description: The term to search for
-      fieldTerms:
+      searchFields:
         type: object
-        description: Map[String, Array[String]]
+        description: Map[String, Array[String]] Each entry contains the column name (i.e. "library:indication") and a list of terms that will be or'ed in the query (["cancer", "sleep apnea"])
+      fieldAggregations:
+        type: array
+        items:
+          type: string
+        description: The list of fields for which you would like to retrieve aggregations
+      maxAggregations:
+        type: integer
+        description: Set if you want more than 10 results, will be applied to all field aggregations
       from:
         type: integer
         default: 0
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -2753,6 +2753,9 @@ definitions:
       searchTerm:
         type: string
         description: The term to search for
+      fieldTerms:
+        type: object
+        description: Map[String, Array[String]]
       from:
         type: integer
         default: 0
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
@@ -11,11 +11,13 @@ import org.elasticsearch.action.delete.{DeleteRequest, DeleteRequestBuilder, Del
 import org.elasticsearch.action.index.{IndexRequest, IndexRequestBuilder, IndexResponse}
 import org.elasticsearch.action.search.{SearchRequest, SearchRequestBuilder, SearchResponse}
 import org.elasticsearch.client.transport.TransportClient
+import org.elasticsearch.search.aggregations.{AggregationBuilders}
 import org.parboiled.common.FileUtils
 import spray.http.Uri.Authority
 import spray.json._
 import spray.json.DefaultJsonProtocol._
-
+import org.elasticsearch.search.aggregations.bucket.terms.Terms
+import scala.collection.JavaConverters._
 
 class ElasticSearchDAO(servers:Seq[Authority], indexName: String) extends SearchDAO with ElasticSearchDAOSupport {
 
@@ -113,5 +115,27 @@ class ElasticSearchDAO(servers:Seq[Authority], indexName: String) extends Search
       hit.getSourceAsString.parseJson
     }
     new LibrarySearchResponse(criteria, searchResults.getHits.totalHits().toInt, sourceDocuments)
+  }
+
+  def getAggregations(fields: Seq[String], numBuckets: Option[Int]) : Seq[LibraryAggregationResponse] = {
+    val searchReq = client.prepareSearch(indexName)
+    searchReq.setSize(0)
+    fields map { field: String =>
+      val terms = AggregationBuilders.terms(field)
+      if (None != numBuckets) {
+        terms.size(numBuckets.get)
+      }
+      searchReq.addAggregation(terms.field(field + ".raw"))
+    }
+    val aggResults = executeESRequest[SearchRequest, SearchResponse, SearchRequestBuilder] (searchReq)
+
+    fields map { field: String =>
+      val terms: Terms = aggResults.getAggregations().get(field)
+      LibraryAggregationResponse(terms.getName(),
+        AggregationFieldResults(terms.getSumOfOtherDocCounts.toInt,
+          terms.getBuckets.asScala.toSeq map { bucket: Terms.Bucket =>
+            AggregationTermResult(bucket.getKey.toString, bucket.getDocCount.toInt)
+          }))
+    }
   }
 }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
@@ -9,14 +9,11 @@ import org.elasticsearch.action.admin.indices.exists.indices.{IndicesExistsReque
 import org.elasticsearch.action.bulk.{BulkRequest, BulkRequestBuilder, BulkResponse}
 import org.elasticsearch.action.delete.{DeleteRequest, DeleteRequestBuilder, DeleteResponse}
 import org.elasticsearch.action.index.{IndexRequest, IndexRequestBuilder, IndexResponse}
-import org.elasticsearch.action.search.{SearchRequest, SearchRequestBuilder, SearchResponse}
 import org.elasticsearch.client.transport.TransportClient
 import org.parboiled.common.FileUtils
 import spray.http.Uri.Authority
 import spray.json._
 import spray.json.DefaultJsonProtocol._
-import org.elasticsearch.search.aggregations.bucket.terms.Terms
-import scala.collection.JavaConverters._
 
 class ElasticSearchDAO(servers:Seq[Authority], indexName: String) extends SearchDAO with ElasticSearchDAOSupport with ElasticSearchDAOQuerySupport {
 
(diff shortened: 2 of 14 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -2720,9 +2720,20 @@ definitions:
 
   SearchTermRef:
     properties:
-      searchTerm:
+      searchString:
         type: string
         description: The term to search for
+      searchFields:
+        type: object
+        description: Map[String, Array[String]] Each entry contains the column name (i.e. "library:indication") and a list of terms that will be or'ed in the query (["cancer", "sleep apnea"])
```
davidangb, line 2728: is this the facet value selections? If so, I'd prefer a name other than "search".

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
```diff
@@ -0,0 +1,54 @@
...
+import org.elasticsearch.search.aggregations.AggregationBuilders
+
+/**
+  * Created by ahaessly on 11/28/16.
+  */
+trait ElasticSearchDAOQuerySupport {
+
+  def createESMatch(data: Map[String, Seq[String]]): Seq[QueryMap] = {
+    (data map {
+      case (k: String, v: Vector[String]) => ESMatch(Map(k -> v.mkString(" ")))
```
davidangb, line 24: this accepts `Seq[String]` as an argument, why check for `Vector[String]` here instead of `Seq[String]`?

Is `createESMatch` used anywhere?

src/main/resources/swagger/api-docs.yaml
```diff
@@ -2720,9 +2720,20 @@ definitions:
...
+      fieldAggregations:
+        type: array
+        items:
+          type: string
+        description: The list of fields for which you would like to retrieve aggregations
+      maxAggregations:
+        type: integer
+        description: Set if you want more than 10 results, will be applied to all field aggregations
```
davidangb, line 2736: nit: add newline before the next endpoint

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
```diff
@@ -0,0 +1,54 @@
...
+      case (Some(searchTerm: String), _) => createListOfMusts(criteria.searchFields) :+ new ESMatch(searchTerm.toLowerCase)
+    }
+
+    ESQuery(ESConstantScore(ESFilter(ESBool(ESMust(qmseq))))).toJson.compactPrint
+  }
+
+  def addAggregations(searchReq: SearchRequestBuilder, criteria: LibrarySearchParams) = {
+    criteria.fieldAggregations map { field: String =>
+      val terms = AggregationBuilders.terms(field)
+      if (None != criteria.maxAggregations) {
```
davidangb: syntax nit: use `isDefined` instead of checking for equality against `None`.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
```diff
@@ -0,0 +1,54 @@
...
+        terms.size(criteria.maxAggregations.get)
```
davidangb: you're handling `size` here. Is it worth adding `order` as a param and handling sorting of facet values here too? I'm fine if the only options we support for ordering are `_count` (desc) and `_term` (asc)

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOSupport.scala
```diff
@@ -52,8 +52,10 @@ trait ElasticSearchDAOSupport extends LazyLogging {
 
   def detailFromAttribute(label: String, detail: AttributeDetail) = {
     detail match {
-      case AttributeDetail("array", Some(items)) => label -> ESDetail(items.`type`)
-      case AttributeDetail(t, None) => label -> ESDetail(detail.`type`)
+      case AttributeDetail("array", Some(items), Some(true)) => label -> new ESAggregatableType(items.`type`)
+      case AttributeDetail("array", Some(items), _) => label -> ESType(items.`type`)
+      case AttributeDetail(t, None, Some(true)) => label -> new ESAggregatableType(detail.`type`)
+      case AttributeDetail(t, None, None) => label -> ESType(detail.`type`)
```
davidangb, line 58: for the last two cases, can you just use `t` instead of `detail.type`?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
```diff
@@ -98,20 +99,25 @@ class ElasticSearchDAO(servers:Seq[Authority], indexName: String) extends Search
...
+    addAggregations(searchReq, criteria)
     val searchResults = executeESRequest[SearchRequest, SearchResponse, SearchRequestBuilder] (searchReq)
 
     val sourceDocuments = searchResults.getHits.getHits.toList map { hit =>
       hit.getSourceAsString.parseJson
     }
-    new LibrarySearchResponse(criteria, searchResults.getHits.totalHits().toInt, sourceDocuments)
+    val aggResults = criteria.fieldAggregations map { field: String =>
+      val terms: Terms = searchResults.getAggregations().get(field)
+      LibraryAggregationResponse(terms.getName(),
```
davidangb: syntax nit: IntelliJ really wants to use the Scala syntax for this, without the empty parens:
```
val terms: Terms = aggResults.getAggregations.get(field)
LibraryAggregationResponse(terms.getName,
```
and I hate seeing IntelliJ complain!

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
```diff
@@ -98,20 +99,25 @@ class ElasticSearchDAO(servers:Seq[Authority], indexName: String) extends Search
...
+        AggregationFieldResults(terms.getSumOfOtherDocCounts.toInt,
+          terms.getBuckets.asScala.toSeq map { bucket: Terms.Bucket =>
```
davidangb: syntax nit: `toSeq` is redundant; just remove it

src/main/scala/org/broadinstitute/dsde/firecloud/model/ElasticSearch.scala
```diff
@@ -9,11 +9,19 @@ import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
...
+trait ESPropertyFields
+case class ESDatasetProperty(properties: Map[String, ESPropertyFields])
+case class ESType(`type`: String) extends ESPropertyFields
+case class ESAggregatableType(`type`: String, fields:ESRaw) extends ESPropertyFields {
+  def this(str:String) = this(str, new ESRaw(str))
+}
+case class ESRaw(raw: ESAggregateProperties) {
+  def this(str: String) = this(ESAggregateProperties(str, "not_analyzed"))
+}
+case class ESAggregateProperties(`type`: String, index:String)
```
davidangb, line 24: these all look good. For discussion purpose: why did you choose `this()` constructors, instead of creating a companion object with `apply` constructors, e.g.:
```
case class ESRaw(raw: ESAggregateProperties)
object ESRaw {
  def apply(str: String) = ESRaw(ESAggregateProperties(str, "not_analyzed"))
}
```
The `apply` approach is probably more Scala-esque, and it eliminates the need for `new` everywhere; you can say `ESRaw(str)` instead of `new ESRaw(str)`.

src/main/scala/org/broadinstitute/dsde/firecloud/model/ElasticSearch.scala
```diff
@@ -37,38 +45,76 @@ object Document {
...
+  // when the search term should search all columns/attributes
+  def this(value: String) = this(Map("_all" -> value))
 }
 
-case class ESWildcardSearchTerm(_all: String)
+case class ESTerm(term: Map[String, String]) extends QueryMap
+
+case class ESMatchAll(match_all: Map[String,String]) extends QueryMap {
+  def this() = this(Map.empty[String,String])
+}
```
davidangb, line 118: these look good, but I haven't matched them up with the queries we actually send to ES. I'd love to - outside of the PR process - pick a handful of query use cases and go over the query formation (i.e. json) with you to understand what it's really doing.

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala
```diff
@@ -104,22 +104,31 @@ object ModelJsonProtocol {
...
-      case Some(term) => JsObject(Map(SEARCH_TERM -> JsString(term), FROM -> JsNumber(params.from), SIZE -> JsNumber(params.size)))
+    override def write(params: LibrarySearchParams): JsValue = {
+      val results = scala.collection.mutable.HashMap(SEARCH_FIELDS -> params.searchFields.toJson, FIELD_AGGREGATIONS -> params.fieldAggregations.toJson, FROM -> params.from.toJson, SIZE -> params.size.toJson)
+      if (params.searchString != None) {
+        results += (SEARCH_STRING -> JsString(params.searchString.get))
+      }
+      if (params.maxAggregations != None) {
+        results += (MAX_AGGREGATIONS -> JsNumber(params.maxAggregations.get))
+      }
+      JsObject(results.toMap)
```
davidangb, line 122: as a Scala rule, stay away from mutable objects. This is a pretty simple and safe case, but my Scala-sense gets all twingy here. I might rewrite this as:
```
    override def write(params: LibrarySearchParams): JsValue = {
      val fields:Seq[Option[(String, JsValue)]] = Seq(
          Some(SEARCH_FIELDS -> params.searchFields.toJson),
          Some(FIELD_AGGREGATIONS -> params.fieldAggregations.toJson),
          Some(FROM -> params.from.toJson),
          Some(SIZE -> params.size.toJson),
          params.searchString map {SEARCH_STRING -> JsString(_)},
          params.maxAggregations map {MAX_AGGREGATIONS -> JsNumber(_)}
        )

      JsObject( fields.filter(_.isDefined).map{_.get}.toMap )
    }

```
you could even combine those two statements into one, if you feel like it.

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala
```diff
@@ -131,21 +140,71 @@ object ModelJsonProtocol {
         case None => None
         case _ => throw DeserializationException("unexpected json type for " + SIZE)
       }
-      LibrarySearchParams(term, from, size)
+      val fields: Map[String, Seq[String]] = data.getOrElse(SEARCH_FIELDS, None) match {
+        case some:JsObject => some.convertTo[Map[String,Seq[String]]]
+        case None => Map.empty
+      }
```
davidangb, line 146: this needs the `case _` ... you could theoretically have a `JsArray` instead of a `JsObject`.

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala
```diff
@@ -131,21 +140,71 @@ object ModelJsonProtocol {
...
+      case must : ESMust => must.toJson
+      case should : ESShould => should.toJson
+      case _ => throw new SerializationException("unexpected QueryMap type")
+    }
+
+    override def read(json: JsValue): ESLogicType = {
+      json.asJsObject.fields.keys.head match {
+        case "must" => impESMust.read(json)
+        case "should" => impESShould.read(json)
+        case _ => throw DeserializationException("unexpected json type")
```
davidangb, line 172: need a check for size != 1 on fields?

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala
```diff
@@ -131,21 +140,71 @@ object ModelJsonProtocol {
...
+      }
+      val maxAggs = data.getOrElse(MAX_AGGREGATIONS, None) match {
+        case JsNumber(x) => Some(x.intValue)
+        case None => None
+        case _ => throw DeserializationException("unexpected json type")
+      }
+      LibrarySearchParams(term, fields, aggs, maxAggs, from, size)
+    }
+  }
```
davidangb, line 159: this is a lot of boilerplate, but I'm struggling for a better way to do it. I'll keep thinking on it, but there may not be a super elegant way.

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala
```diff
@@ -131,21 +140,71 @@ object ModelJsonProtocol {
...
+    override def write(input: ESPropertyFields): JsValue = input match {
+      case estype : ESType => estype.toJson
+      case esaggtype : ESAggregatableType => esaggtype.toJson
+      case _ => throw new SerializationException("unexpected ESProperty type")
+    }
+
+    override def read(json: JsValue): ESPropertyFields = {
+      json.asJsObject.fields.keys.size match {
+        case 1 => ESTypeFormat.read(json)
+        case 2 => ESAggregatableTypeFormat.read(json)
```
davidangb, line 207: this is ok for now, but checking the size doesn't feel extensible if we add more types.

davidangb (commented):
comments inline, focused on syntax and json-stuff!

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala
davidangb: I don't know if this is actually better, but here's an option:
```
    override def read(json: JsValue): LibrarySearchParams = {
      val data = json.asJsObject.fields
      val term = data.getOrElse(SEARCH_STRING, None) match {
        case JsString(str) if str.trim == "" => None
        case JsString(str) => Some(str.trim)
        case None => None
        case _ => throw DeserializationException("unexpected json type for " + SEARCH_STRING)
      }

      val fields = convertingReader[Map[String, Seq[String]]](SEARCH_FIELDS, data, Map.empty)
      val aggs = convertingReader[Seq[String]](FIELD_AGGREGATIONS, data, Seq.empty)

      val from: Option[Int] = intReader(FROM, data)
      val size: Option[Int] = intReader(SIZE, data)
      val maxAggs = intReader(MAX_AGGREGATIONS, data)

      LibrarySearchParams(term, fields, aggs, maxAggs, from, size)
    }

    private def intReader(fieldName:String, data:Map[String,JsValue]):Option[Int] = {
      data.get(fieldName) match {
        case Some(x:JsNumber) => Some(x.value.toInt)
        case None => None
        case _ => throw DeserializationException(s"unexpected json type for $fieldName")
      }
    }
    private def convertingReader[T](fieldName:String, data:Map[String,JsValue], default:T):T = {
      data.getOrElse(fieldName, None) match {
        case j:JsValue => Try(j.convertTo[T]).toOption.getOrElse(
          throw DeserializationException(s"unexpected json type for $fieldName")
        )
        case None => default
      }
    }
  }
```

(I used `data.get` in `intReader` and `data.getOrElse` in `convertingReader` just for fun - they should probably be consistent)

In this approach, you could move `intReader` and `convertingReader` somewhere else to make the code readable.

EDIT: this doesn't even compile. Bah.

src/main/resources/swagger/api-docs.yaml
ahaessly: searchString is the free text search
searchFields will contain the facet selections - do you prefer a different name? matchFields? matchTerms?

src/main/resources/swagger/api-docs.yaml
ahaessly: this is all part of the same endpoint...

src/main/scala/org/broadinstitute/dsde/firecloud/model/ElasticSearch.scala
ahaessly: There is actually a reason - not sure if you have run across this before. If I create a companion object for a type that needs a json format, the standard formater no longer references the class. instead it references the object. (at least this is what i think is happening). I can change the name of the object to solve this problem, but then this is no longer a companion object... but maybe that is still preferable.

this is the code addition just to demonstrate the error:
object ESRaw{}
the error happens whether this is declared before or after the class
here is the compiler error:
app_1    | [error] /app/src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala:351: type mismatch;
app_1    | [error]  found   : org.broadinstitute.dsde.firecloud.model.ESRaw.type
app_1    | [error]  required: ? => ?
app_1    | [error]  Note: implicit value ESRawFormat is not applicable here because it comes after the application point and it lacks an explicit result type
app_1    | [error]   implicit val ESRawFormat = jsonFormat1(ESRaw)

src/main/scala/org/broadinstitute/dsde/firecloud/model/ElasticSearch.scala
ahaessly: no problem - would it be useful to add some debug level logging statements to print this out?

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala
ahaessly: do you prefer this?
    override def read(json: JsValue): ESPropertyFields = {
      if (json.asJsObject.fields.contains("fields")) {
        ESAggregatableTypeFormat.read(json)
      } else {
        ESTypeFormat.read(json)
      }
    }

PR #323, opened 2016-12-02
gawb 1366: change wildcard search to match search

ahaessly:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/model/ElasticSearch.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/model/ElasticSearch.scala
@@ -54,10 +54,12 @@ case class LibrarySearchResponse(
 }
 
 
-// classes to create the ES queries in json format
-// {"query":{"match_all":{}}}"
-// {"query":{"match":{"_all":"<str>"}}}"
-trait QueryMap
+/** classes to create the ES queries in json format
+  * {"query":{"match_all":{}}}"
+  * {"query":{"match":{"_all":"some-search-string"}}}"
+  */
+
+sealed trait QueryMap
 
 case class ESQuery(query: QueryMap)
 
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/model/ElasticSearch.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/model/ElasticSearch.scala
@@ -56,19 +56,18 @@ case class LibrarySearchResponse(
 
 // classes to create the ES queries in json format
 // {"query":{"match_all":{}}}"
-// {"query":{"wildcard":{"_all":"%s"}}}"
+// {"query":{"match":{"_all":"<str>"}}}"
 trait QueryMap
 
 case class ESQuery(query: QueryMap)
 
+case class ESMatch(`match`: Map[String, String]) extends QueryMap {
+  // when the search term should search all columns/attributes
+  def this(value: String) = this(Map("_all" -> value))
+}
+
 case class ESMatchAll(match_all: Map[String,String]) extends QueryMap {
   def this() = this(Map.empty[String,String])
 }
-
-case class ESWildcard(wildcard: ESWildcardSearchTerm) extends QueryMap {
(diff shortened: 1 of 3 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala
```diff
@@ -138,14 +138,14 @@ object ModelJsonProtocol {
   implicit object impQueryMap extends JsonFormat[QueryMap] {
     override def write(inputmap: QueryMap): JsValue = inputmap match {
       case matchall : ESMatchAll => matchall.toJson
-      case wildcard : ESWildcard => wildcard.toJson
+      case wildcard : ESMatch => wildcard.toJson
```
davidangb: nit: change `case widcard` to a more appropriate variable name? (but not "match" because that's a Scala word)

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala
```diff
@@ -138,14 +138,14 @@ object ModelJsonProtocol {
...
       case _ => throw new SerializationException("unexpected QueryMap type")
     }
 
     override def read(json: JsValue): QueryMap = {
       json.asJsObject.fields.keys.head match {
         case "match_all" => impESMatchAll.read(json)
-        case "wildcard" => impESWildcard.read(json)
+        case "wildcard" => impESMatch.read(json)
```
davidangb: should this be `case "match_all"` or `case "match"`?

src/main/scala/org/broadinstitute/dsde/firecloud/model/ElasticSearch.scala
```diff
@@ -56,19 +56,18 @@ case class LibrarySearchResponse(
 
 // classes to create the ES queries in json format
 // {"query":{"match_all":{}}}"
-// {"query":{"wildcard":{"_all":"%s"}}}"
+// {"query":{"match":{"_all":"<str>"}}}"
```
rushtong: Do we need these comments or should this be in javadoc format?

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala
ahaessly: wow - good catch. i guess this means we don't really convert back from json to QueryMap

rushtong: I think this should be `case "match"` if I'm reading the `QueryMap` trait right.

src/main/scala/org/broadinstitute/dsde/firecloud/model/ElasticSearch.scala
```diff
@@ -56,19 +56,18 @@ case class LibrarySearchResponse(
...
 trait QueryMap
```
rushtong: Do you want to make this a sealed trait? I think if you do that, then when you use it in case statements, the compiler (and IDE) can give you warnings about whether you're handling all of them or not.

src/main/scala/org/broadinstitute/dsde/firecloud/model/ElasticSearch.scala
ahaessly: it's not so useful in this version, but once we incorporate facets, it's helpful for me to be able to look at the example to understand the query generated instead of trying to re-work it mentally from the classes. i'm fine changing it to a different comment format (/**)

src/main/scala/org/broadinstitute/dsde/firecloud/model/ElasticSearch.scala
ahaessly: sounds good

davidangb:
👍

rushtong:
👍

PR #325, opened 2016-12-02
WorkspaceApiServiceSpec refactoring (no JIRA, though related to work for GAWB-1174)

breilly2:
This is just the refactoring that I had to do before adding a reliable test for a bucketUsage passthrough. Instead of a global mock setup, each test configures the specific mock behavior that it needs. The changes exposed a number of situations where there was mock behavior that looked like it was needed but actually wasn't. I believe these changes bring the existing tests to a state where they pass for the right reasons.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

… the test that it is needed for.

Improved WorkspaceApiServiceSpec to test what should be rather than what should not be.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/model/WorkspaceACL.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/model/WorkspaceACL.scala
@@ -75,5 +75,4 @@ object WorkspaceACLJsonSupport {
   implicit val WorkspaceACLFormat = jsonFormat1(WorkspaceACL)
 
   implicit val WorkspaceACLUpdateFormat = jsonFormat2(WorkspaceACLUpdate)
-  implicit val ListOfWorkspaceACLUpdatesFormat = listFormat[WorkspaceACLUpdate]
 }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/model/WorkspaceACL.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/model/WorkspaceACL.scala
@@ -75,4 +75,5 @@ object WorkspaceACLJsonSupport {
   implicit val WorkspaceACLFormat = jsonFormat1(WorkspaceACL)
 
   implicit val WorkspaceACLUpdateFormat = jsonFormat2(WorkspaceACLUpdate)
+  implicit val ListOfWorkspaceACLUpdatesFormat = listFormat[WorkspaceACLUpdate]
 }
--- a/src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
@@ -15,6 +15,7 @@ import spray.http._
 import spray.http.StatusCodes._
 import spray.httpx.SprayJsonSupport._
 import spray.json._
+import spray.json.DefaultJsonProtocol._
 
 class WorkspaceApiServiceSpec extends BaseServiceSpec with WorkspaceApiService with BeforeAndAfterEach {
 
--- a/src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
@@ -2,17 +2,16 @@ package org.broadinstitute.dsde.firecloud.service
 
 import org.broadinstitute.dsde.firecloud.FireCloudConfig
 import org.broadinstitute.dsde.firecloud.mock.MockUtils._
-import org.broadinstitute.dsde.firecloud.mock.{MockUtils, MockTSVFormData}
+import org.broadinstitute.dsde.firecloud.mock.{MockTSVFormData, MockUtils}
 import org.broadinstitute.dsde.firecloud.model._
+import org.broadinstitute.dsde.firecloud.model.WorkspaceACLJsonSupport._
 import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
 import org.broadinstitute.dsde.firecloud.webservice.WorkspaceApiService
-
 import org.mockserver.integration.ClientAndServer
 import org.mockserver.integration.ClientAndServer._
 import org.mockserver.model.HttpRequest._
 import org.scalatest.BeforeAndAfterEach
-import spray.http.{MediaTypes, HttpEntity, HttpMethods}
-
+import spray.http._
 import spray.http.StatusCodes._
 import spray.httpx.SprayJsonSupport._
 import spray.json._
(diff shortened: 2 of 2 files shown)


src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
```diff
@@ -27,159 +26,57 @@ class WorkspaceApiServiceSpec extends BaseServiceSpec with WorkspaceApiService w
...
+        .withHeaders(MockUtils.header).withStatusCode(status.intValue))
+  }
 
-    workspaceServer
-      .when(
-        request()
-          .withMethod("POST")
-          .withPath(s"${workspaceBasePath}/${workspace.namespace.get}/${workspace.name.get}/entities/batchUpsert")
-          .withHeader(authHeader))
+  def stubRawlsServiceWithError(method: HttpMethod, path: String, status: StatusCode) = {
```
breilly2, line 64: I don't think this is actually called. I meant to remove it before committing.

src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
```diff
@@ -27,159 +26,57 @@ class WorkspaceApiServiceSpec extends BaseServiceSpec with WorkspaceApiService w
...
-          .withMethod("POST")
-          .withPath(s"${workspaceBasePath}/${workspace.namespace.get}/${workspace.name.get}/entities/batchUpdate")
-          .withHeader(authHeader))
-      .respond(
-        org.mockserver.model.HttpResponse.response()
-          .withHeaders(MockUtils.header)
-          .withStatusCode(NoContent.intValue)
-          .withBody(rawlsErrorReport(NoContent).toJson.compactPrint)
-      )
+  override def beforeEach(): Unit = {
```
breilly2, line 75: This is `beforeEach` now instead of `beforeAll` so that stub behaviors don't bleed from one test to the next.

src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
```diff
@@ -288,397 +185,419 @@ class WorkspaceApiServiceSpec extends BaseServiceSpec with WorkspaceApiService w
   "WorkspaceService Passthrough Tests" - {
 
     "Passthrough tests on the /workspaces path" - {
-      "MethodNotAllowed error is not returned for HTTP GET and POST methods" in {
```
breilly2, line 291: POST for this isn't actually a passthrough, so the test has been moved to a new section below.

src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
```diff
@@ -288,397 +185,419 @@ class WorkspaceApiServiceSpec extends BaseServiceSpec with WorkspaceApiService w
...
+        Patch(updateAttributesPath, "[]") ~> dummyUserIdHeaders("1234") ~> sealRoute(workspaceRoutes) ~> check {
+          status should equal(OK)
         }
       }
     }
 
 
     "Passthrough tests on the /workspaces/%s/%s/acl path" - {
-      "MethodNotAllowed error is not returned for HTTP GET and PATCH methods" in {
-        List(HttpMethods.GET, HttpMethods.PATCH) map {
```
breilly2, line 337: PATCH is not actually a passthrough, so the test has been moved to a new section below.

src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
```diff
@@ -288,397 +185,419 @@ class WorkspaceApiServiceSpec extends BaseServiceSpec with WorkspaceApiService w
...
-            new RequestBuilder(method)(aclPath) ~> dummyUserIdHeaders("1234") ~> sealRoute(workspaceRoutes) ~> check {
-              status shouldNot equal(MethodNotAllowed)
-            }
-        }
-      }
-    }
-
-
-    "Passthrough tests on the /workspaces/%s/%s/clone path" - {
-      "MethodNotAllowed error is not returned for POST method" in {
```
breilly2, line 348: POST on /clone is not actually a passthrough, so the test has been moved to a new section below.

breilly2:
Github has made this more difficult to read than it should be. If you have trouble, I recommend checking out the branch and doing a local compare ignoring whitespace.

src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
rushtong: Please double check this with Travis by re-running the build a few times. I believe we made this change to deal with spurious travis tests specific to this class. The only reliable fix we could find was to use `beforeAll`.

breilly2: @rushtong Thanks for the suggestion. Is that just a matter of clicking "Restart build" a few times on the build result in Travis or is there a more preferred way?

rushtong: Exactly. The errors were very spurious - the only way to reproduce was to restart some variable number of times.

dvoet:
👍

src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
breilly2: I restarted the Travis job a couple dozen times throughout the day yesterday and didn't see any failures. Since there was some known flakiness (GAWB-1297) and I have no evidence that this makes that problem any worse, I think it's worth keeping this clean-up.

rushtong: 👍

src/main/scala/org/broadinstitute/dsde/firecloud/model/WorkspaceACL.scala
```diff
@@ -75,4 +75,5 @@ object WorkspaceACLJsonSupport {
   implicit val WorkspaceACLFormat = jsonFormat1(WorkspaceACL)
 
   implicit val WorkspaceACLUpdateFormat = jsonFormat2(WorkspaceACLUpdate)
+  implicit val ListOfWorkspaceACLUpdatesFormat = listFormat[WorkspaceACLUpdate]
```
MatthewBemis: why was this needed?

breilly2: The test for `PATCH` on `/workspaces/_/_/acl` was previously set up as a passthrough test even though the implementation is not a passthrough. When I split the test out and didn't provide a request body, it returned 400 Bad Request instead of 200 OK. Therefore, I had to provide a `List` of `WorkspaceACLUpdate` which there had not yet been a need to explicitly marshal. Without this line, the test does not compile.

This is the technique I had used in a previous project to configure marshalling of a list of something. However, I don't see any other use of `listFormat` in orchestration or rawls. Is there a different preferred way to accomplish this here?

MatthewBemis: if you import `import spray.json.DefaultJsonProtocol._` in WorkspaceApiServiceSpec instead of this, it will work. this has all of the implicits for handling collections (and more). my understanding is that if the application knows how to unmarshal a `List` and it knows how to unmarshal a `WorkspaceACLUpdate`, then it knows how to unmarshal a `List[WorkspaceACLUpdate]`

the reason I bring this up is that this diverges from the Rawls implementation of `WorkspaceACL` (which is where this entire file was pulled from). eventually this code won't exist in orchestration and will be part of a dependency so it can't be different

breilly2: Good call. Thanks. Fixed.

MatthewBemis:
:+1:

PR #330, opened 2016-12-08
LMSVN stuff for orchestration

rushtong:
See  https://broadinstitute.atlassian.net/browse/GAWB-1352
Companion UI PR: https://github.com/broadinstitute/firecloud-ui/pull/600

Add a new property that is not meant to be sent to ES nor is meant to be user-editable. 

---
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/main/resources/library/attribute-definitions.json
+++ b/src/main/resources/library/attribute-definitions.json
@@ -207,8 +207,7 @@
       "description": "Library Metadata Schema Version Number",
       "type": "integer",
       "default": 1,
-      "indexable": false,
-      "editable": false
+      "indexable": false
     }
   },
   "consentCodes": {
--- a/src/main/resources/library/attribute-definitions.json
+++ b/src/main/resources/library/attribute-definitions.json
@@ -201,6 +201,14 @@
     },
     "library:orsp": {
       "type": "string"
+    },
+    "library:lmsvn": {
+      "title": "Library Metadata Schema Version Number",
+      "description": "Library Metadata Schema Version Number",
+      "type": "integer",
+      "default": 1,
+      "indexable": false,
+      "editable": false
     }
   },
   "consentCodes": {
--- a/src/main/resources/library/attribute-definitions.json
+++ b/src/main/resources/library/attribute-definitions.json
@@ -232,7 +232,7 @@
       "library:dataUseBSPSampleCollectionID", "library:studyDesign", "library:cellType", "library:coverage",
       "library:ethnicity", "library:primaryDiseaseSite", "library:broadInternalResearchProjectID",
       "library:broadInternalResearchProjectName", "library:broadInternalCohortName",
-      "library:broadInternalSeqProjectNumbers"],
+      "library:broadInternalSeqProjectNumbers", "library:lmsvn"],
     "consentCodes": ["GRU", "HMB", "DS", "NCU", "NPU", "NDMS", "NAGR", "NCTRL", "RS-G", "RS-PD", "RS-POP"]
   },
   "wizard":
@@ -265,7 +265,8 @@
         {"property": "library:broadInternalResearchProjectID"},
         {"property": "library:broadInternalResearchProjectName"},
         {"property": "library:broadInternalCohortName"},
-        {"property": "library:broadInternalSeqProjectNumbers", "inputHint": "Broad internal IDs"}
+        {"property": "library:broadInternalSeqProjectNumbers", "inputHint": "Broad internal IDs"},
+        {"property": "library:lmsvn", "editable": false}
       ]
     },
     {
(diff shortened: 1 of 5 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOSupport.scala
```diff
@@ -44,18 +44,21 @@ trait ElasticSearchDAOSupport extends LazyLogging {
...
     }
-    ESDatasetProperty(maps).toJson.prettyPrint
+    ESDatasetProperty(attributeDetailMap).toJson.prettyPrint
   }
 
-  def detailFromAttribute(label: String, detail: AttributeDetail) = {
+  def detailFromAttribute(label: String, detail: AttributeDetail): (String, ESPropertyFields) = {
+    val itemType = detail match {
+      case x if x.`type` == "array" && x.items.isDefined => x.items.get.`type`
+      case x => x.`type`
```
rushtong: this needs to be safer, something like `case _ => detail.type`

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala
```diff
@@ -5,7 +5,7 @@ import spray.http.StatusCode
 import spray.http.StatusCodes.BadRequest
 import org.broadinstitute.dsde.firecloud.model.MethodRepository.{AgoraPermission, FireCloudPermission}
 import spray.json._
-import spray.json.DefaultJsonProtocol._
+import spray.json.DefaultJsonProtocol.{jsonFormat4, _}
```
rushtong: Unnecessary

src/main/resources/library/attribute-definitions.json
```diff
@@ -257,7 +265,8 @@
         {"property": "library:broadInternalResearchProjectID"},
         {"property": "library:broadInternalResearchProjectName"},
         {"property": "library:broadInternalCohortName"},
-        {"property": "library:broadInternalSeqProjectNumbers", "inputHint": "Broad internal IDs"}
+        {"property": "library:broadInternalSeqProjectNumbers", "inputHint": "Broad internal IDs"},
+        {"property": "library:lmsvn", "editable": false}
```
skwalker, line 268: does editable need to be here and above (line 211)? I feel like it probably only needs to be in one place?

src/main/resources/library/attribute-definitions.json
```diff
@@ -201,6 +201,14 @@
     },
     "library:orsp": {
       "type": "string"
+    },
+    "library:lmsvn": {
+      "title": "Library Metadata Schema Version Number",
+      "description": "Library Metadata Schema Version Number",
+      "type": "integer",
+      "default": 1,
```
skwalker, line 209: Should we be starting at 0? (just looking at 1417)

src/main/scala/org/broadinstitute/dsde/firecloud/model/ElasticSearch.scala
```diff
@@ -9,7 +9,12 @@ import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
 
 case class AttributeDefinition(properties: Map[String, AttributeDetail])
 
-case class AttributeDetail(`type`: String, items: Option[AttributeDetail] = None, aggregate: Option[Boolean] = None)
+case class AttributeDetail(
+  `type`: String,
+  items: Option[AttributeDetail] = None,
+  aggregate: Option[Boolean] = None,
+  indexable: Option[Boolean] = Some(true)
```
davidangb: I'd prefer that `aggregate` and `indexable` be consistent: either both default to None, or both default to Some.

davidangb (commented):
one minor comment inline, otherwise 👍

src/main/scala/org/broadinstitute/dsde/firecloud/model/ElasticSearch.scala
rushtong: 👍

skwalker (approved):
works well, 2 tiny inline comments

src/main/resources/library/attribute-definitions.json
rushtong: Good catch ... it was originally only in one place, but with Matt's refactor, I needed to add it here and didn't remove it from above. I'm ambivalent about having these kinds of property attributes spread out at all - I'd really like these to be one thing so we don't have duplication.

src/main/resources/library/attribute-definitions.json
rushtong: GAWB-1417 describes going from v1 to v2, so I think this being v1 is correct.

skwalker: oh oops I thought I saw 0 someone but never mind!

src/main/resources/library/attribute-definitions.json
skwalker: yeah, I don't really like the duplication that's there now either, but thanks for this fix! PR looks good to me 👍

davidangb: could probably add this in to GAWB-1389.

davidangb (approved):
👍 based on latest changes!

src/main/resources/library/attribute-definitions.json
rushtong: Added to GAWB-1389, thanks David!

PR #333, opened 2016-12-14
Pass object metadata errors through instead of parsing them

MatthewBemis:
Currently 403s and 404s getting object metadata will be masked as 500 errors because we will incorrectly and unsuccessfully try to coerce Google's error response into the ObjectMetadata case class.

Related to GAWB-1144

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
@@ -6,9 +6,9 @@ import com.google.api.client.googleapis.auth.oauth2.GoogleCredential
 import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport
 import com.google.api.client.json.jackson2.JacksonFactory
 import com.google.api.services.storage.{Storage, StorageScopes}
-import org.broadinstitute.dsde.firecloud.FireCloudConfig
+import org.broadinstitute.dsde.firecloud.{FireCloudExceptionWithErrorReport, FireCloudConfig}
 import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol.impGoogleObjectMetadata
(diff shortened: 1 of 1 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -152,7 +152,10 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
   def getObjectMetadata(bucketName: String, objectKey: String, authToken: String)
                     (implicit actorRefFactory: ActorRefFactory, executionContext: ExecutionContext): Future[ObjectMetadata] = {
     val request = Get( getObjectResourceUrl(bucketName, objectKey) ) ~> addCredentials(OAuth2BearerToken(authToken)) ~> sendReceive
-    request map (_.entity.asString.parseJson.convertTo[ObjectMetadata])
+    request map { response =>
+      if(response.status.isSuccess) response.entity.asString.parseJson.convertTo[ObjectMetadata]
```
breilly2: It's a shame you can't just put an `unmarshal` in the response pipeline, but I don't see another way in this case due to needing to do something different with failures.

breilly2 (approved):

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
dvoet: I think unmarshal does the right thing. It does not do the json conversion if there is an error. Spray should throw and exception for an unsuccessful response code. To convert that to a different exception use recover on the Future.

dvoet (approved):
👍

PR #335, opened 2016-12-15
Facet render hints for aggregate types.

rushtong:
This is required for https://broadinstitute.atlassian.net/browse/GAWB-976

The facet rendering on the library page requires more information than is currently available for the aggregate types. For example, both phenotype and project are string types, but require different forms of rendering in the UI. This change allows the UI to make the right display decision. 

Even though `library:numSubjects` is not currently an aggregate field, the UI does call for it at some point so this PR tries to anticipate that requirement.

---

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/main/resources/library/attribute-definitions.json
+++ b/src/main/resources/library/attribute-definitions.json
@@ -51,30 +51,35 @@
     "library:indication": {
       "title": "Cohort Phenotype/Indication",
       "type": "string",
-      "aggregate": true
+      "aggregate": true,
+      "renderHint": "text"
     },
     "library:numSubjects": {
       "title": "No. of Subjects",
       "type": "integer",
       "minimum": 0,
-      "default": 0
+      "default": 0,
+      "renderHint": "slider"
     },
     "library:projectName": {
       "title": "Project Name",
       "type": "string",
-      "aggregate": true
+      "aggregate": true,
+      "renderHint": "checkbox"
     },
     "library:dataCategory": {
       "title": "Data Category",
       "type": "array",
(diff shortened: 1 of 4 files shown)


skwalker:
👍 looks fine to me

davidangb:
Some thoughts:
1. I'm fine with this as-is.
2. We've been using `renderHint` to control display on the cataloging (edit) screen. This PR overrides `renderHint` to control display of facets. Was this intentional? I assume so ... just wanted to make sure.
3. I suspect in the future we may have more configs for facets. In that case, it would probably make sense to go for a structure like:
```
    "library:projectName": {
      "title": "Project Name",
      "type": "string",
      "aggregate": {
        "renderHint": "checkbox",
        "autocomplete": false,
        "order": { "_count" : "asc" }
      }
    },
```
but we don't have to do that yet.

rushtong:
2. Technically, this doesn't override them in the wizard because the wizard
isn't using these properties - it's using a separate set of properties
defined locally for just the wizard. The only reason I was using
`renderHint` was to be consistent.

3. I like your suggestion. Let me look into how much of a change it is in
orchestration.

davidangb:
agreed on bullet point 2 - and by "override" I really meant, "you're using the same name as", not that you're overwriting the other renderHints. All set!

davidangb:
👍

davidangb (approved):

PR #338, opened 2016-12-20
GAWB-1434: grab the appropriate 'foreign key' columns when downloading entity tsv

davidangb:
I'm not familiar with the requirements for column renaming in TSV downloads (or import), so perhaps this is the wrong fix ... but it seems to fix the issue as reported in GAWB-1434.

When we construct the download, we were searching for the wrong column name for "foreign key" references - e.g. the participant attribute of a sample. We were looking for the renamed column ("participant" -> "participant_id"), but that's not what exists in the entity.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
@@ -100,7 +100,7 @@ object TSVFormatter {
     headerRenamingMap:Map[String, String]): IndexedSeq[String] = {
     val rowMap: Map[Int, String] =  entity.attributes map {
       case (attributeName, attribute) =>
-        val columnPosition = headerValues.indexOf(headerRenamingMap.getOrElse(attributeName.name, attributeName.name))
+        val columnPosition = headerValues.indexOf(attributeName.name)
         val cellValue = AttributeStringifier(attribute)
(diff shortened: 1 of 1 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
```diff
@@ -100,7 +100,7 @@ object TSVFormatter {
     headerRenamingMap:Map[String, String]): IndexedSeq[String] = {
     val rowMap: Map[Int, String] =  entity.attributes map {
       case (attributeName, attribute) =>
-        val columnPosition = headerValues.indexOf(headerRenamingMap.getOrElse(attributeName.name, attributeName.name))
+        val columnPosition = headerValues.indexOf(attributeName.name)
```
dvoet, line 103: I am not sure this is right. If I am reading this right it is saying the header value (for a sample entity) has an entry for participant and we are not finding it because we are looking for participant_id. This code change makes it look for participant which allows us to find the data BUT the header should be participant_id. So this changes the wrong side of the equation. headerValues should have participant_id instead of participant.

dvoet:
I am going to change it to remove headerRenamingMap entirely.

PR #340, opened 2017-01-05
Remove tech debt from ProfileClient

dmohs:
https://broadinstitute.atlassian.net/browse/GAWB-1286

I'm not sure this brings "ProfileClient/ProfileClientActor into line with desired architecture" but it does get rid of the temporary DAO stuff (the title of this ticket). I believe this is enough refactoring for now.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off


src/main/scala/org/broadinstitute/dsde/firecloud/model/Profile.scala
```diff
@@ -131,6 +132,8 @@ case class NIHStatus(
 
 object NIHStatus {
 
+  implicit val impNihStatus = jsonFormat7(NIHStatus.apply)
```
dvoet: why move this?

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
```diff
@@ -0,0 +1,51 @@
...
+        profile.linkedNihUsername match {
+          case Some(_) =>
+            rawlsDao.isDbGapAuthorized(userInfo) map { x =>
+              RequestComplete(NIHStatus(profile, Some(x)))
+            }
+          case None =>
+            Future.successful(RequestComplete(StatusCodes.NotFound))
+        }
+      case None => Future.successful(RequestComplete(StatusCodes.NotFound))
+    } pipeTo sender
```
dvoet: I don't think this pipeTo is right here

dvoet (changes requested):

src/main/scala/org/broadinstitute/dsde/firecloud/model/Profile.scala
dmohs: Sweet I was hoping you'd ask this! So, my thought is that you should be able to use NIHStatus in just about any way you want (including JSON) simply by importing the object. You should not have to also import ModelJsonProtocol just because you decided you wanted to serialize the object.

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
dmohs: I removed this and it seems to work, but since this is completely redundant (right?) I'm not sure why it worked before.

dvoet (approved):

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockThurloeDAO.scala
```diff
@@ -12,12 +12,20 @@ import scala.util.{Success, Try}
  *
  */
 class MockThurloeDAO extends ThurloeDAO {
+  var nextGetProfileResponse:Option[Profile] = None
+  val testProfile = Profile("Rich", "Hickey", "CTO", None, "Datomic", "Data Storage",
+    "NYC", "NY", "USA", "David Mohs", "Not for no profit")
+
+  def reset() = {
```
breilly2, line 20: Instead of a reset method, I'd rather see a new instance of `MockThurloeDAO` for each test. Using a reset method makes it too easy to add state to the mock and forget to add code to the reset method which could cause unintended cross-talk between tests.

src/test/scala/org/broadinstitute/dsde/firecloud/service/BaseServiceSpec.scala
```diff
@@ -3,14 +3,20 @@ package org.broadinstitute.dsde.firecloud.service
 import org.broadinstitute.dsde.firecloud.Application
 import org.broadinstitute.dsde.firecloud.dataaccess._
 import org.broadinstitute.dsde.firecloud.mock.MockGoogleServicesDAO
+import org.scalatest.BeforeAndAfter
 
-class BaseServiceSpec extends ServiceSpec {
+class BaseServiceSpec extends ServiceSpec with BeforeAndAfter {
```
breilly2, line 8: It's not intuitive to me that the base spec includes `BeforeAndAfter` while the individual tests (i.e. `NIHServiceSpec`) actually override `before()`. Could `BaseServiceSpec` override `before()`? This is where `reset()` is defined and it theoretically should be done for all tests. (Of course, this only applies if `reset()` stays.)

src/test/scala/org/broadinstitute/dsde/firecloud/service/BaseServiceSpec.scala
```diff
@@ -3,14 +3,20 @@ package org.broadinstitute.dsde.firecloud.service
...
-  val rawlsDAO:RawlsDAO = new MockRawlsDAO
-  val searchDAO:SearchDAO = new MockSearchDAO
-  val thurloeDAO:ThurloeDAO = new MockThurloeDAO
-  val googleServicesDAO:GoogleServicesDAO = new MockGoogleServicesDAO
-  val app:Application = new Application(agoraDAO, rawlsDAO, searchDAO, thurloeDAO, googleServicesDAO)
+  val agoraDao:MockAgoraDAO = new MockAgoraDAO
+  val googleServicesDao:MockGoogleServicesDAO = new MockGoogleServicesDAO
+  val rawlsDao:MockRawlsDAO = new MockRawlsDAO
+  val searchDao:MockSearchDAO = new MockSearchDAO
+  val thurloeDao:MockThurloeDAO = new MockThurloeDAO
```
breilly2, line 14: Good change. Tests are absolutely allowed to be aware that mocks are being used (though now the explicit types aren't even needed).

src/test/scala/org/broadinstitute/dsde/firecloud/service/NIHServiceSpec.scala
```diff
@@ -1,106 +1,66 @@
...
-          .withHeaders(MockUtils.header).withStatusCode(OK.intValue)
-          .withBody("true")
-      )
-
-  }
-
-  override def afterAll(): Unit = {
-    profileServer.stop()
-    workspaceServer.stop()
+  before {
```
breilly2: I feel like this should be in `BaseServiceSpec`. (I misspoke above saying it's an override. `BeforeAndAfterEach.beforeEach()` is mean to be used as an override... not sure why the two are different.)

src/main/scala/org/broadinstitute/dsde/firecloud/model/Profile.scala
breilly2: Counter argument: could it cause any harm to bring the implicit `JsonFormat` into scope in code that doesn't need it? It's probably fine as long as this is the only way we need to marshal/unmarshal the object. If we did need to vary marshalling, that would be a reason to keep the formatter closer to the code that uses it.

src/test/scala/org/broadinstitute/dsde/firecloud/service/NIHServiceSpec.scala
```diff
@@ -1,106 +1,66 @@
...
+        ))
+        Get(targetUri) ~> dummyUserIdHeaders(uniqueId) ~> sealRoute(nihRoutes) ~> check {
           status should equal(OK)
           val nihStatus = responseAs[NIHStatus]
           nihStatus.loginRequired shouldBe(true)
         }
       }
     }
 
+    // This is done after the previous test to ensure reset() is working.
```
breilly2: This adds to my dislike of `reset()`.

src/test/scala/org/broadinstitute/dsde/firecloud/service/NIHServiceSpec.scala
```diff
@@ -1,106 +1,66 @@
...
-import org.mockserver.model.HttpRequest._
-
+import org.broadinstitute.dsde.firecloud.webservice.NihApiService
 import spray.http.StatusCodes._
 import spray.httpx.SprayJsonSupport._
-import spray.json._
 
 
-class NIHServiceSpec extends ServiceSpec with NIHService {
+class NIHServiceSpec extends BaseServiceSpec with NihApiService {
```
breilly2: Overall, these changes are good. It's worth noting that (as discussed in person with Mohs) we do lose a little test coverage of deserializing responses from Thurloe. That should still be (at least mostly) covered by web tests, though, so all we currently lose is some test granularity. What we should have is a HttpThurloeDAOSpec that tests deserialization.

breilly2 (approved):

src/test/scala/org/broadinstitute/dsde/firecloud/service/BaseServiceSpec.scala
dmohs: This is so, so, so, so, so much better.

breilly2 (approved):

PR #344, opened 2017-01-09
GAWB-1468: search suggestions/autocomplete

davidangb:
important: you must perform a reindex-all operation before this will work!

This PR creates a new /api/library/suggest endpoint that accepts the same request criteria as /api/library/search (but ignores anything about aggregations), and returns autocomplete suggestions for the text search box.

For autocomplete suggestions, I'm using the technique described at:
https://www.elastic.co/guide/en/elasticsearch/guide/current/_index_time_search_as_you_type.html
https://qbox.io/blog/multi-field-partial-word-autocomplete-in-elasticsearch-using-ngrams
instead of using ES's suggesters. I'm doing this for all the additional control - including (and in my mind most important) the ability to have the suggestions be contextual. In short, this is the only technique that ensures suggestions will actually return results if you have already filtered using facets. In a small corpus such as ours, that's important.

Skeleton of the implementation:
1) for each ES mapping, copy its value into a new "_suggest" field. That field now contains all data on the record.
2) set up custom ES analyzers and filters on the _suggest field. At index time, we index all the substrings present in each word. I like doing this at index time because our index is small and it keeps runtime fast.
3) at search time, search the _suggest field and match on any of each word's substrings

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
@@ -100,4 +100,8 @@ class ElasticSearchDAO(servers: Seq[Authority], indexName: String) extends Searc
   override def findDocuments(criteria: LibrarySearchParams): Future[LibrarySearchResponse] = {
     findDocumentsWithAggregateInfo(client, indexName, criteria)
   }
+
+  override def suggest(criteria: LibrarySearchParams): Future[LibrarySearchResponse] = {
+    findDocumentsWithAggregateInfo(client, indexName, criteria)
+  }
 }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
@@ -43,27 +43,13 @@ class ElasticSearchDAO(servers: Seq[Authority], indexName: String) extends Searc
   override def createIndex = {
     val mapping = makeMapping(FileUtils.readAllTextFromResource(LibraryService.schemaLocation))
     executeESRequest[CreateIndexRequest, CreateIndexResponse, CreateIndexRequestBuilder](
-      client.admin.indices.prepareCreate(indexName).setSettings(analysisSettings).addMapping(datatype, mapping)
+      client.admin.indices.prepareCreate(indexName)
+        .setSettings(analysisSettings)
+        .addMapping(datatype, mapping)
+      // TODO: set to one shard? https://www.elastic.co/guide/en/elasticsearch/guide/current/relevance-is-broken.html
     )
-    // TODO: set to one shard? https://www.elastic.co/guide/en/elasticsearch/guide/current/relevance-is-broken.html
-
-    /* TODO: eventually, we will make an autocomplete query that looks something like this:
-       {
-          "_source": false,
-            "size" : 8,
-          "query": {
-              "match": {
-                  "_suggest": "sam"
-              }
-          },
-          "highlight": {
-            "fields" : {
-              "_suggest" : {"fragment_size" : 75}
-            }
-          }
-        }
-     */
   }
+
   // will throw an error if index does not exist
   override def deleteIndex = {
     executeESRequest[DeleteIndexRequest, DeleteIndexResponse, DeleteIndexRequestBuilder](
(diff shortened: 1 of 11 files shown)


rushtong:
I see duplicated results with ```{ "searchString": "cancer" }```
```
{
  "searchParams": {
    "size": 10,
    "searchString": "cancer",
    "fieldAggregations": [],
    "filters": {},
    "from": 0
  },
  "total": 5,
  "results": [
    "<em>cancer</em>",
    "<em>Cancer</em>",
    "pediatric <em>cancer</em>",
    "<em>cancer</em> project",
    "This cohort is part of The <em>Cancer</em> Genome Atlas project (https",
    "://<em>cancergenome.nih.gov</em>/abouttcga/overview).  This cohort includes raw data and analysis",
    " of <em>cancer</em> patients samples by genomic DNA copy number arrays, DNA",
    " publication list at: https://<em>cancergenome.nih.gov</em>/publications; Data"
  ],
  "aggregations": []
}
```
Is that intentional?

Second question: is this an ES disconnect to have a total of "5" but have "8" actual results? Or is that due to filtering results in scala?

davidangb:
On the duplicates, are you talking about "cancer" and "Cancer"? Yeah, Scala doesn't see those as duplicates because of the case difference ... hm. I suppose we should filter those out.

On the total of 5 vs. 8 ... it makes sense to me why this happens but I totally grant that it's a bad API. Under the covers, ES is really finding 5 *datasets* that have matching suggestions, but a few of those datasets have multiple suggestion strings. In this API we suppress the actual dataset results and just collate the suggestions.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
```diff
@@ -143,5 +153,32 @@ trait ElasticSearchDAOQuerySupport extends ElasticSearchDAOSupport {
...
+        if (hit.getHighlightFields.containsKey(fieldSuggest)) {
+          hit.getHighlightFields.get(fieldSuggest).fragments map {t => JsString(t.toString)}
+        } else {
+          None
+        }
+      }).distinct
+
+      LibrarySearchResponse(
+        criteria,
+        searchResult.getHits.totalHits().toInt,
```
rushtong: Since the total results from ES are not necessarily representative of the results, maybe this can instead just be a count of suggestions.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
```diff
@@ -42,9 +43,13 @@ class ElasticSearchDAO(servers: Seq[Authority], indexName: String) extends Searc
   override def createIndex = {
     val mapping = makeMapping(FileUtils.readAllTextFromResource(LibraryService.schemaLocation))
     executeESRequest[CreateIndexRequest, CreateIndexResponse, CreateIndexRequestBuilder](
-      client.admin.indices.prepareCreate(indexName).addMapping(datatype, mapping)
+      client.admin.indices.prepareCreate(indexName)
+        .setSettings(analysisSettings)
+        .addMapping(datatype, mapping)
+      // TODO: set to one shard? https://www.elastic.co/guide/en/elasticsearch/guide/current/relevance-is-broken.html
```
rushtong, line 49: This is fascinating reading! Can we push this TODO into a story?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
```diff
@@ -143,5 +153,32 @@ trait ElasticSearchDAOQuerySupport extends ElasticSearchDAOSupport {
...
+    searchFuture map {searchResult =>
+
+      // autocomplete query can return duplicate suggestions. De-dupe them here.
+      val suggestions:List[JsString] = (searchResult.getHits.getHits.toList flatMap { hit =>
+        if (hit.getHighlightFields.containsKey(fieldSuggest)) {
+          hit.getHighlightFields.get(fieldSuggest).fragments map {t => JsString(t.toString)}
+        } else {
+          None
+        }
+      }).distinct
```
rushtong, line 175: As mentioned in another comment, this is case sensitive and should likely be case-insensitive to clarify the number of results.

rushtong (approved):
👍 with minor suggestions in line. Looks good, works as needed.

davidangb:
I created GAWB-1493 to track the relevance/shards issue, and have two small code changes to address your comments.

rushtong:
👍 Looks good!

PR #350, opened 2017-01-20
Return 404 instead of 401 when path not found

dmohs:


MatthewBemis (approved):
Change looks good to me and I've also made the mental note to be more careful about the order that things fail in when working with spray routes 😧

:+1:

PR #362, opened 2017-01-31
Ah gawb 1396: Library: autocomplete during metadata intake 

ahaessly:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/resources/library/attribute-definitions.json
+++ b/src/main/resources/library/attribute-definitions.json
@@ -21,7 +21,8 @@
     },
     "library:datasetVersion": {
       "title": "Dataset Version",
-      "type": "string"
+      "type": "string",
+      "typeahead": "populate"
     },
     "library:datasetDescription": {
       "title": "Cohort Description",
@@ -30,27 +31,32 @@
     "library:datasetCustodian": {
       "title": "Dataset Custodian",
       "type": "string",
-      "inputHint": "e.g. Project Manager"
+      "inputHint": "e.g. Project Manager",
+      "typeahead": "populate"
     },
     "library:datasetDepositor": {
       "title": "Dataset Depositor",
       "type": "string",
-      "inputHint": "e.g. Project Manager"
+      "inputHint": "e.g. Project Manager",
+      "typeahead": "populate"
     },
     "library:contactEmail": {
       "title": "Contact Email",
       "type": "string",
-      "renderHint": {"datatype": "email"}
+      "renderHint": {"datatype": "email"},
+      "typeahead": "populate"
     },
     "library:datasetOwner": {
       "title": "Dataset Owner",
       "type": "string",
-      "inputHint": "e.g. Prinicipal Investigator"
+      "inputHint": "e.g. Prinicipal Investigator",
+      "typeahead": "populate"
     },
     "library:institute": {
       "title": "Research Institute",
       "type": "array",
-      "items": { "type": "string" }
+      "items": { "type": "string" },
+      "typeahead": "populate"
     },
     "library:indication": {
       "title": "Cohort Phenotype/Indication",
@@ -86,13 +92,15 @@
       "inputHint": "e.g. TCGA, TopMed, ExAC ; tag all relevant associated projects",
       "aggregate": {
         "renderHint": "checkbox"
-      }
+      },
+      "typeahead": "populate"
     },
     "library:dataCategory": {
       "title": "Data Category",
       "type": "array",
       "items": { "type": "string" },
-      "inputHint": "e.g. Simple Nucleotide Variation (SNVs), Raw Sequencing data, Copy Number Variation, Structural Variation, Expression Quantification, Transcriptome Assembly, DNA-Methylation, Clinical phenotypes, Biospecimen metadata"
+      "inputHint": "e.g. Simple Nucleotide Variation (SNVs), Raw Sequencing data, Copy Number Variation, Structural Variation, Expression Quantification, Transcriptome Assembly, DNA-Methylation, Clinical phenotypes, Biospecimen metadata",
+      "typeahead": "populate"
     },
     "library:datatype": {
       "title": "Experimental Strategy",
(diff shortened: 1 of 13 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
```diff
@@ -110,6 +110,10 @@ class ElasticSearchDAO(servers: Seq[Authority], indexName: String) extends Searc
     autocompleteSuggestions(client, indexName, criteria, groups)
   }
 
+  override def fieldSuggest(field: String, text: String): Future[Seq[String]] = {
```
rushtong: For my own edification - I was curious about the override here. Strictly, it's not necessary, but good practice to explicitly state it to prevent accidental override errors.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
```diff
@@ -174,6 +175,22 @@ trait ElasticSearchDAOQuerySupport extends ElasticSearchDAOSupport {
     response
   }
 
+  def populateSuggestions(client: TransportClient, indexname: String, field: String, text: String) : Future[Seq[String]] = {
```
rushtong: Super nitpick - camel case `indexname`

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
```diff
@@ -174,6 +175,22 @@ trait ElasticSearchDAOQuerySupport extends ElasticSearchDAOSupport {
...
+    suggestion.text(text)
+    suggestion.field(field + ".suggest")
+    val suggestQuery = client.prepareSearch(indexname).addSuggestion(suggestion)
+    logger.debug(s"populate suggestions query: $suggestQuery.toJson")
+    val results = Future[SearchResponse](executeESRequest[SearchRequest, SearchResponse, SearchRequestBuilder](suggestQuery))
+
+    results map { suggestResult =>
+      val compSugg: CompletionSuggestion = suggestResult.getSuggest.getSuggestion(suggestionName)
+      val options : Seq[CompletionSuggestion.Entry.Option] = compSugg.getEntries.get(0).getOptions.asScala
+      options map { option: CompletionSuggestion.Entry.Option => option.getText().string() }
```
rushtong: This also works and doesn't give me weak warnings in intellij: `option.getText.string()`

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOSupport.scala
```diff
@@ -55,15 +55,20 @@ trait ElasticSearchDAOSupport extends LazyLogging {
...
+  def createType(label: String, detail: AttributeDetail): (String, ESPropertyFields) = {
     val itemType = detail match {
       case x if x.`type` == "array" && x.items.isDefined => x.items.get.`type`
       case _ => detail.`type`
     }
-    detail match {
-      case x if x.aggregate.isDefined => label -> ESAggregatableType(itemType)
-      case _ => label -> ESType(itemType)
+    val searchSuggest = itemType == "string"
+    val createSuggest = searchSuggest && !(noCreateSuggest.contains(label))
```
rushtong: Unnecessary parens around `noCreateSuggest.contains(label)`

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOSupport.scala
```diff
@@ -55,15 +55,20 @@ trait ElasticSearchDAOSupport extends LazyLogging {
...
+    val isaggregate = detail match {
```
rushtong: Super nit again - camel case `isaggregate`

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockSearchDAO.scala
```diff
@@ -42,4 +43,8 @@ class MockSearchDAO extends SearchDAO {
     Future(LibrarySearchResponse(librarySearchParams, 0, Seq[JsValue](), Seq[LibraryAggregationResponse]()))
   }
 
+  override def fieldSuggest(field: String, text: String): Future[Seq[String]] = {
```
rushtong: I don't see any tests that use this - do we need one?

src/main/scala/org/broadinstitute/dsde/firecloud/model/ElasticSearch.scala
```diff
@@ -33,27 +33,40 @@ trait ESPropertyFields {
...
-case class ESType(`type`: String, copy_to: String) extends ESPropertyFields
+case class ESType(`type`: String, fields: Option[Map[String,ESInnerField]], copy_to: Option[String] = None ) extends ESPropertyFields
 object ESType extends ESPropertyFields {
-  def apply(`type`: String):ESType = ESType(`type`, ElasticSearch.fieldSuggest)
+  def apply(`type`: String, hasPopulateSuggest: Boolean, hasSearchSuggest: Boolean, isAggregatable: Boolean):ESType =  {
+    val fields: Option[Map[String,ESInnerField]] = (hasPopulateSuggest, isAggregatable) match {
+      case (true, true) => Option(Map("suggest" -> completionField, "raw" -> rawField(`type`)))
+      case (true, false) => Option(Map("suggest" -> completionField))
+      case (false, true) => Option(Map("raw" -> rawField(`type`)))
+      case (false, false) => None
```
rushtong: My only concern about this block is maintainability. If we have one more boolean in the match condition, this explodes to 9 cases. Not something to hold up the PR but something to keep in mind.

rushtong:
Functional testing on a non-suggestable field gives me back a 500 error that might be better handled. For example, orch knows which fields are and aren't, so it should send back a Bad Request in that case.

```
{
  "statusCode": 500,
  "source": "FireCloud",
  "timestamp": 1485966309924,
  "causes": [
    {
      "source": "FireCloud",
      "message": "all shards failed",
      "causes": [
        {
          "source": "FireCloud",
          "message": "Field [indication.suggest] is not a completion suggest field",
          "causes": [],...
```

rushtong:
👍 with minor comments and one suggestion for better error handling.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
ahaessly: i just assumed it was required since i saw it on the other methods declared in SearchDAO but defined in this file

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
ahaessly: ooh, and even better
`option.getText.string`

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOSupport.scala
ahaessly: i wrote this code before i added the property to the attribute file. i'll use the setting in the file now

src/main/scala/org/broadinstitute/dsde/firecloud/model/ElasticSearch.scala
ahaessly: yes, we wanted to reduce the multiple case classes to one. i'm definitely open to suggestions for how to improve.

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockSearchDAO.scala
ahaessly: this is tested by the last test in LibraryAPIServiceSpec, the call to populate suggest invokes the mock dao

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
davidangb: could you rename this method to avoid confusion with `ElasticSearch.fieldSuggest` (in model)?

src/main/resources/swagger/api-docs.yaml
```diff
@@ -493,6 +493,30 @@ paths:
...
+          name: text
+          type: string
+          description: the initial text to complete
+      responses:
+        200:
+          description: Field specific suggestions
+        400:
+          description: Bad Request, specified field that was not configured for suggestions
+          schema:
+            $ref: '#/definitions/ErrorReport'
```
davidangb, line 519: in terms of API design, I'd prefer to see the input text as a query param, e.g. `/api/library/populate/suggest/{field}?q={text}`

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
```diff
@@ -174,6 +175,22 @@ trait ElasticSearchDAOQuerySupport extends ElasticSearchDAOSupport {
...
+    suggestion.field(field + ".suggest")
+    val suggestQuery = client.prepareSearch(indexName).addSuggestion(suggestion)
+    logger.debug(s"populate suggestions query: $suggestQuery.toJson")
+    val results = Future[SearchResponse](executeESRequest[SearchRequest, SearchResponse, SearchRequestBuilder](suggestQuery))
+
+    results map { suggestResult =>
+      val compSugg: CompletionSuggestion = suggestResult.getSuggest.getSuggestion(suggestionName)
+      val options : Seq[CompletionSuggestion.Entry.Option] = compSugg.getEntries.get(0).getOptions.asScala
+      options map { option: CompletionSuggestion.Entry.Option => option.getText.string }
+    }
```
davidangb, line 191: may be easier in person ... but why would we get back these two different structures from the same request?

src/main/scala/org/broadinstitute/dsde/firecloud/model/ElasticSearch.scala
```diff
@@ -33,27 +34,40 @@ trait ESPropertyFields {
...
+    }
+    val copyto = hasSearchSuggest match {
+      case true => Option(ElasticSearch.fieldSuggest)
+      case false => None
+    }
```
davidangb: this could just be:
```
if (hasSearchSuggest) { ...
``
instead of a `match` clause.

File: src/main/scala/org/broadinstitute/dsde/firecloud/model/ElasticSearch.scala (hunk 7 above)
Comment (davidangb) [reply src/main/scala/org/broadinstitute/dsde/firecloud/model/ElasticSearch.scala]: without refactoring the case classes, you could refactor this  `match` block. How about something like this:
```
    val innerFields = Map.empty[String,ESInnerField] ++
      (if (isAggregatable) Map("raw" -> rawField(`type`)) else Nil) ++
      (if (hasPopulateSuggest) Map("suggest" -> completionField) else Nil)

    val fields = if (innerFields.isEmpty) None else Some(innerFields)
```

you might even be able to get away with the final `isEmpty` check, if ES does the right thing if we send the empty map.

warning: I did not test this code!

File: src/main/resources/library/attribute-definitions.json [hunk 11]
```diff
@@ -165,7 +184,8 @@
     },
     "library:primaryDiseaseSite": {
       "title": "Primary Disease Site",
-      "type": "string"
+      "type": "string",
+      "typeahead": "populate"
```
davidangb, line 188: is it worth assuming `typeahead: populate` as the default, and only adding a `typeahead` key to those String properties we do NOT want typeahead on?

davidangb (changes requested):
comments inline. Unit tests are failing, too - those will need updates.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
ahaessly: not sure which requests you are talking about. this query didn't actually have a search part, so there were no hits. it was just asking for suggestions, which comes back in a different suggest section.

davidangb: ignore me! I totally misread this code - my bad.

davidangb (approved):
looking good!

src/main/resources/library/attribute-definitions.json
ahaessly: i'm assuming we might have different types of typeahead in the future. so i think it would be better to have this be stating what kinds of type ahead apply instead of a negative

PR #364, opened 2017-02-01
HOTFIX: Check that users are on the whitelist before linking them

MatthewBemis:
Note that I haven't added any new unit tests for this. We will need to revamp accordingly but that can be done in a follow-up ticket since the priority is to get this fix out ASAP.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
@@ -3,7 +3,7 @@ package org.broadinstitute.dsde.firecloud.core
 import akka.actor.{Actor, ActorRefFactory, Props}
 import akka.event.Logging
 import akka.pattern.pipe
-import org.broadinstitute.dsde.firecloud.FireCloudConfig
+import org.broadinstitute.dsde.firecloud.{FireCloudExceptionWithErrorReport, FireCloudConfig}
 import org.broadinstitute.dsde.firecloud.core.ProfileClient._
 import org.broadinstitute.dsde.firecloud.dataaccess.{HttpGoogleServicesDAO, HttpRawlsDAO, HttpThurloeDAO}
 import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
@@ -67,7 +67,7 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
       profileResponse pipeTo sender
 
     case UpdateNIHLinkAndSyncSelf(userInfo: UserInfo, nihLink: NIHLink) =>
-      val syncWhiteListResult = syncWhitelist(Some(userInfo.getUniqueId))
+      val syncWhiteListResult = syncWhitelist(Some((userInfo.getUniqueId, nihLink.linkedNihUsername)))
       val pipeline = authHeaders(requestContext) ~> sendReceive
       val profilePropertyMap = nihLink.propertyValueMap
       val propertyUpdates = updateUserProperties(pipeline, userInfo, profilePropertyMap)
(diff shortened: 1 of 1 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -208,7 +209,10 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
           RawlsGroupMemberList(userSubjectIds = Some(subjectIds))
         }}
       }
-      case Some(id) => Future.successful(RawlsGroupMemberList(userSubjectIds = Some(Seq(id))))
+      case Some((fcUser, nihUser)) =>
+        if(whitelist contains nihUser) Future.successful(RawlsGroupMemberList(userSubjectIds = Some(Seq(fcUser))))
```
MatthewBemis, line 213: this was the key part. before it was returning the user regardless of them being on the whitelist or not. now we actually check their NIH name against the whitelist.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -208,7 +209,10 @@ class ProfileClientActor(requestContext: RequestContext) extends Actor with Fire
...
+        else Future.failed(throw new FireCloudExceptionWithErrorReport(ErrorReport(StatusCodes.Forbidden,
```
MatthewBemis, line 214: I'm not sure if this bubbles up in the way that we'd like but it seems to _work_ the way we want it to. Basically, if the user isn't a part of the whitelist then we don't want to do *anything* that follows. If we don't stop the operation here then we'll end up updating their thurloe profile to say that their TCGA access is "pending" when in reality we should just immediately fail and not update anything in thurloe

dvoet: what happens if we just return an empty RawlsGroupMemberList instead?

MatthewBemis: that technically wouldn't grant them access to the TCGA data because it wouldn't put them in the dbGap group (but it would still call UPDATE on the group, with the empty payload)

the issue is that in `UpdateNIHLinkAndSyncSelf`, we can't really tell the difference between when we've updated the group with an empty `RawlsGroupMemberList` and when we've updated the group with a populated `RawlsGroupMemberList`

as a result, the thurloe entries `lastLinkTime`, `linkExpireTime`, and `linkedNihUsername` would all still get updated- and this would trigger the UI to say "You have successfully linked, your access will be granted shortly"

so my choice (mainly in the interest of time) was to go nuclear and throw an exception instead of rewriting UpdateNIHLinkAndSyncSelf. the exception allows us to fail quickly and not store any inaccurate junk in thurloe

dvoet: technically, the junk is not inaccurate. The inaccuracy is the statement "your access will be granted shortly"

under what conditions does it say no access?

dmohs (approved):
In my judgement, the overall logic in this file is not easy to follow (not your fault). I think we need to refactor it or else we risk introducing additional bugs in the future.

However, I am confident that this change fixes the immediate problem of giving access to non-whitelist members. I'm comfortable pushing this through if re-enabling TCGA access is indeed a high priority.

dvoet (approved):
agreed that this is good enough to fix the issue and that this needs a good refactor.

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
MatthewBemis: The UI logic is confusing to follow but it looks like these are the conditions:

If your link time has expired, then it will say "Expired"
If you have the 3 valid entries in thurloe but aren't a part of the rawls dbGap group, it will say "Your access will be granted shortly"
If you have the 3 thurloe entries + dbGap group member it will say "Authorized"
For any other condition, it will say "Not Authorized", but I've never actually seen this case happen


You're right that the junk is not inaccurate- because they are technically linked to NIH. It just seems like we'd need a new state for "You're successfully linked to NIH, but you don't have the proper permissions to view TCGA data"

PR #370, opened 2017-02-08
Ah gawb 1321: Re-publishing upon cataloging an already-published dataset

ahaessly:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
@@ -83,8 +83,7 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
               // between the time we retrieved them and here, where we update them.
               val allOperations = generateAttributeOperations(workspaceResponse.workspace.attributes, userAttrs,
                 k => k.namespace == AttributeName.libraryNamespace && k.name != LibraryService.publishedFlag.name)
-              val republish = workspaceResponse.workspace.attributes.contains(publishedFlag) &&
-                workspaceResponse.workspace.attributes.get(publishedFlag).fold(false)(_.asInstanceOf[AttributeBoolean].value)
+              val republish = workspaceResponse.workspace.attributes.get(publishedFlag).fold(false)(_.asInstanceOf[AttributeBoolean].value)
               rawlsDAO.patchWorkspaceAttributes(ns, name, allOperations) map { newws =>
                 if (republish) {
                   // we do not need to delete before republish
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
@@ -83,7 +83,15 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
               // between the time we retrieved them and here, where we update them.
               val allOperations = generateAttributeOperations(workspaceResponse.workspace.attributes, userAttrs,
                 k => k.namespace == AttributeName.libraryNamespace && k.name != LibraryService.publishedFlag.name)
-              rawlsDAO.patchWorkspaceAttributes(ns, name, allOperations) map (RequestComplete(_))
+              val republish = workspaceResponse.workspace.attributes.contains(publishedFlag) &&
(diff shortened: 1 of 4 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -83,7 +83,15 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
               // between the time we retrieved them and here, where we update them.
               val allOperations = generateAttributeOperations(workspaceResponse.workspace.attributes, userAttrs,
                 k => k.namespace == AttributeName.libraryNamespace && k.name != LibraryService.publishedFlag.name)
-              rawlsDAO.patchWorkspaceAttributes(ns, name, allOperations) map (RequestComplete(_))
+              val republish = workspaceResponse.workspace.attributes.contains(publishedFlag) &&
+                workspaceResponse.workspace.attributes.get(publishedFlag).fold(false)(_.asInstanceOf[AttributeBoolean].value)
```
davidangb: the `contains` part of this clause is redundant - the `attributes.get(publishedFlag) ...` handles the case where the attribute is missing and properly folds it to false.

optional feedback: in general, the codebase uses `getOrElse` instead of `fold`. I don't think one it's necessarily better, but it's more consistent. That said, I think this fold syntax is nice and terse and I'm good with it.

This line will have a problem if the library:published attribute is something other than a boolean. Do we care?

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockRawlsDAO.scala
```diff
@@ -54,14 +54,40 @@ class MockRawlsDAO  extends RawlsDAO {
...
+      AttributeName("default", "d") -> AttributeString("escape quo\"te"),
+      AttributeName("default", "e") -> AttributeString("v1"),
+      AttributeName("default", "f") -> AttributeValueList(Seq(
+        AttributeString("v6"),
+        AttributeNumber(999),
+        AttributeBoolean(true)
+      ))),
+    "",
+    Map("" -> Map("" -> "")),
+    Some(Map("" -> ""))
```
davidangb: use `Map.empty` or `None` instead of a map with the empty string.

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockRawlsDAO.scala
```diff
@@ -54,14 +54,40 @@ class MockRawlsDAO  extends RawlsDAO {
...
+    "attributes",
+    "att",
+    Option(false),
+    "ansingh",
+    "date",
+    Some("date"),
+    Map(AttributeName("default", "a") -> AttributeBoolean(true),
+      AttributeName("default", "b") -> AttributeNumber(1.23),
+      AttributeName("default", "c") -> AttributeString(""),
+      AttributeName("library", "published") -> AttributeBoolean(true),
```
davidangb, line 68: it's worth having at least another `library:` attribute, to ensure that calling code finds the right one.

davidangb (commented):
some comments inline. Looks like the unit test is failing, too!

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
ahaessly: my problem with get or else here, was that i couldn't just use a false as the or else because the get would return an AttributeBoolean. so my or else would have had to construct a false AttributeBoolean which seemed clunky. that's why i chose the fold, because i could execute the function on the AttributeBoolean directly if it was there and skip it if not.

ahaessly: and i like it even better that it takes care of the case where it's missing!

ahaessly: we should be the only ones creating the publish flag.... and when someone sets metadata via swagger, it validates against the attribute file... i think we're ok

skwalker (approved):
looks good to me

davidangb (approved):
👍 lgtm!

PR #379, opened 2017-02-17
GAWB-1260 Use rawlsModel library for Rawls classes

helgridly:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off


src/main/scala/org/broadinstitute/dsde/firecloud/model/Workspace.scala
```diff
@@ -1,108 +1,39 @@
...
-  attributes: AttributeMap,
-  bucketName: String,
-  accessLevels: Map[String, Map[String, String]],
-  realm: Option[Map[String, String]],
-  isProtected: Boolean) {
-  def this(rw: RawlsWorkspace) =
-    this(rw.workspaceId, rw.namespace, rw.name, rw.isLocked, rw.createdBy, rw.createdDate,
-      rw.lastModified, rw.attributes, rw.bucketName, rw.accessLevels, rw.realm,
-      rw.realm.flatMap(_.get("realmName").map(_ == FireCloudConfig.Nih.rawlsGroupName)).getOrElse(false))
+object WorkspaceCreate {
```
breilly2, line 14: This generally looks better. (Can this just be called `Workspace` now?) I do worry a little about the realm parts... not because I see a problem with these changes but because I know how fragile this code was previously.

@MatthewBemis, can you have a look specifically at the changes here around determining `isProtected`? In light of this, I'm very glad for the approach that we took with testing today!

breilly2 (commented):
There is a crazy amount of good in here! Deletion of duplicated code, collapsing of redundant case classes. If this truly doesn't break anything, then bravo! ⭐️ 

As mentioned in a comment, I'd like to have @MatthewBemis take a look at one particular part. Then, as far as I'm concerned, 👍

src/main/scala/org/broadinstitute/dsde/firecloud/model/Workspace.scala
helgridly: There's already a `Workspace` class in rawlsModel. This is different because it turns the "protected" bool into an NIH realm.

MatthewBemis: Changes look good to me :+1:

PR #381, opened 2017-02-22
GAWB-1618, GAWB-1477: Implement new NIH link expiration logic, eliminate ProfileClient

MatthewBemis:
This PR does a bunch of things. Sorry to whoever is reviewing it.

-Implemented new NIH link expiration logic
-Remove outdated tests that test the old logic
-Refactored `ProfileClient.scala` by moving the remaining pieces to `NihService.scala`. Rewrote all functions to use rawls and thurloe DAOs instead of making direct http requests
-Built up `MockThurloeDAO` to allow for us to test all of the whitelist details
-Built up parts of `MockRawlsDAO` to allow for us to test that the dbGapAuthorizedUsers group was getting updated when linking and when whitelist syncing
-Pulled in `withDefaultApiServices` from rawls to use in `NihApiService`, which addresses GAWB-1477
-Pulled the guts of `NihService` out into a trait to allow for easier testing
-Improved error handling in HttpThurloeDAO with `wrapExceptions`

Jyotin is aware of these changes and the impact that it has on FISMA docs.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


src/main/resources/swagger/api-docs.yaml
```diff
@@ -3240,7 +3240,6 @@ definitions:
 
   NIHStatus:
     required:
-      - loginRequired
```
MatthewBemis, line 3306: so this was actually never used. maybe it was used at one point? the new logic shouldn't need it.

when you link, the only fields we set in thurloe are `linkedNihUsername` and `linkExpireTime`. `linkExpireTime` will be now + 30 days. 

when the whitelist syncs every 6 hours, it will filter out and remove all linked accounts that were linked > 30 days ago.

the UI will display the "you must re-link within 24 hours" banner once you reach the 29th day.

if there is no `linkExpireTime` set, you will be removed upon the next whitelist sync. this should never happen but there are safegaurds in place to make it work in the case where it does

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
```diff
@@ -97,17 +79,13 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
...
-      case (Some(lastLink), Some(expire)) =>
+    profile.linkExpireTime match {
+      case Some(expire) =>
         // User in good standing; return the expire time unchanged
         expire
       case _ =>
-        // Either last-link or expire is missing. Reset the user's expiration.
-        DateUtils.nowPlus24Hours
+        // Expiration date is missing. Expire their access, forcing them to re-link
+        DateUtils.nowMinus1Hour
```
MatthewBemis: i think that this is the appropriate safeguard now but want to double check

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
```diff
@@ -97,17 +79,13 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
 object HttpThurloeDAO {
   // Not private to allow this to be called by tests.
   def calculateExpireTime(profile:Profile): Long = {
```
dvoet: this should not be called calculate anymore

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
dvoet: isn't this whole function now `profile.linkExpireTime.getOrElse(DateUtils.nowMinus1Hour)`? Might as well use 0 instead of DateUtils.nowMinus1Hour.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
dvoet: I actually think this function should be inlined and the HttpThurloeDAO object deleted.

dvoet (changes requested):
I would really like to see some tests that expiration actually happens.

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
```diff
@@ -61,7 +61,6 @@ class FireCloudServiceActor extends HttpServiceActor with FireCloudDirectives
     statusService.routes ~ nihRoutes ~ billingService.routes
 
   val userService = new UserService with ActorRefFactoryContext
-  val nihSyncService = new NIHSyncService with ActorRefFactoryContext
```
MatthewBemis, line 65: this is now inside of NihService

src/main/scala/org/broadinstitute/dsde/firecloud/core/ProfileClient.scala
```diff
@@ -1,216 +0,0 @@
...
-  }
-
-  /**
-    * Rawls can come back with a conflict (500 wrapping a 409) if the user exists either in
-    * Rawls or LDAP when registering. That is not a real error so we can consider that a success.
-    *
-    * @param postResponse The HttpResponse
-    * @return A conflict or not
-    */
-  def isConflict(postResponse: HttpResponse): Boolean = {
```
MatthewBemis, line 110: I didn't re-implement this because it wasn't being used anymore

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
```diff
@@ -47,67 +43,42 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
...
+
   override def saveProfile(userInfo: UserInfo, profile: BasicProfile): Future[Unit] = {
-    val pipeline = addFireCloudCredentials ~> addCredentials(userInfo.accessToken) ~> sendReceive
     val profilePropertyMap = profile.propertyValueMap ++ Map("email" -> userInfo.userEmail)
     Future.sequence(profilePropertyMap map({
       case (key, value) => saveKeyValue(userInfo, key, value)
     })) map({ _.forall({ _ => true }) })
   }
 
-  override def maybeUpdateNihLinkExpiration(userInfo: UserInfo, profile: Profile): Future[Unit] = {
```
MatthewBemis: no longer needed with new expiration logic

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
```diff
@@ -47,67 +43,42 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
...
-  }
-
   private def throwBadResponse(response: HttpResponse) = {
     throw new FireCloudExceptionWithErrorReport(FCErrorReport("Thurloe", response))
   }
 }
-
-object HttpThurloeDAO {
-  // Not private to allow this to be called by tests.
-  def calculateExpireTime(profile:Profile): Long = {
```
MatthewBemis: no longer needed with new expiration logic

src/main/scala/org/broadinstitute/dsde/firecloud/service/NIHSyncService.scala
```diff
@@ -1,21 +0,0 @@
...
-import org.slf4j.LoggerFactory
-import spray.routing._
-
-trait NIHSyncService extends HttpService with PerRequestCreator with FireCloudDirectives with StandardUserInfoDirectives {
-
-  private implicit val executionContext = actorRefFactory.dispatcher
-  lazy val log = LoggerFactory.getLogger(getClass)
-
-  val routes: Route =
-      path("sync_whitelist") {
```
MatthewBemis, line 15: this exists in NihApiService as its own route

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
```diff
@@ -1,84 +1,185 @@
...
-  private def getStatus(userInfo: UserInfo): Future[PerRequestMessage] = {
+  private def getNihStatus(userInfo: UserInfo): Future[PerRequestMessage] = {
     thurloeDao.getProfile(userInfo) flatMap {
       case Some(profile) =>
         profile.linkedNihUsername match {
           case Some(_) =>
-            rawlsDao.isDbGapAuthorized(userInfo) map { x =>
-              RequestComplete(NihStatus(profile, Some(x)))
+            rawlsDao.isDbGapAuthorized(userInfo) map { isDbGapAuthorized =>
+              RequestComplete(NihStatus(profile.copy(linkExpireTime = Option(profile.linkExpireTime.getOrElse(0L))), Some(isDbGapAuthorized)))
```
MatthewBemis, line 81: if the linkExpireTime is missing from their profile, we return their expire time as the beginning of time so they'll be prompted to re-link in the UI. they will also be removed from the dbGap group once the whitelist syncs again, but that's separate logic

src/test/scala/org/broadinstitute/dsde/firecloud/mock/MockGoogleServicesDAO.scala
```diff
@@ -14,7 +14,7 @@ import scala.concurrent.{ExecutionContext, Future}
 
 class MockGoogleServicesDAO extends GoogleServicesDAO {
   override def getAdminUserAccessToken: String = ""
-  override def getBucketObjectAsInputStream(bucketName: String, objectKey: String): InputStream = null
+  override def getBucketObjectAsInputStream(bucketName: String, objectKey: String): InputStream = new ByteArrayInputStream("firecloud-dev".getBytes("UTF-8"))
```
MatthewBemis, line 17: this is the mock whitelist which contains only the user `firecloud-dev`

src/test/scala/org/broadinstitute/dsde/firecloud/service/ApiServiceSpec.scala
```diff
@@ -0,0 +1,40 @@
...
+import spray.routing.HttpService
+import spray.testkit.ScalatestRouteTest
+
+import scala.concurrent.duration._
+
+/**
+  * Created by mbemis on 3/1/17.
+  */
+
+// common trait to be inherited by API service tests
```
MatthewBemis, line 18: i've only done this for NihApiServiceSpec.

this pattern was stolen from rawls and should nicely address GAWB-1477

src/test/scala/org/broadinstitute/dsde/firecloud/service/NihApiServiceSpec.scala
```diff
@@ -0,0 +1,93 @@
...
+import org.broadinstitute.dsde.firecloud.model.JWTWrapper
+import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
+import org.broadinstitute.dsde.firecloud.utils.DateUtils
+import spray.http.StatusCodes._
+
+import scala.concurrent.ExecutionContext
+
+class NihApiServiceSpec extends ApiServiceSpec {
+  val dbGapAuthorizedUsersGroupName = FireCloudConfig.Nih.rawlsGroupName
+  val validJwtOnWhitelist = JWTWrapper("eyJhbGciOiJIUzI1NiJ9.ZmlyZWNsb3VkLWRldg.NPXbSpTmAOUvJ1HX85TauAARnlMKfqBsPjumCC7zE7s")
```
MatthewBemis, line 15: this jwt decodes to `firecloud-dev`, which is on the test whitelist in the mock google dao

src/test/scala/org/broadinstitute/dsde/firecloud/service/NihApiServiceSpec.scala
```diff
@@ -0,0 +1,93 @@
...
+  val validJwtNotOnWhitelist = JWTWrapper("eyJhbGciOiJIUzI1NiJ9.bm90LWEtcmVhbC11c2Vy.L6JKlzZ9zp2mvDAbhGKRJIbW99x4cN-UjWJNZuSdnbc")
```
MatthewBemis, line 16: this jwt decodes to a username that is definitely not on the test whitelist in the mock google dao

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
```diff
@@ -101,6 +101,18 @@ class HttpRawlsDAO( implicit val system: ActorSystem, implicit val executionCont
     }
   }
 
+  override def addMemberToGroup(groupName: String, memberList: RawlsGroupMemberList): Future[Boolean] = {
```
dvoet: put admin in the name of this and the next fuction

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
```diff
@@ -47,67 +43,42 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
...
+      Try(response.status match {
         case StatusCodes.OK => ()
         case _ => throwBadResponse(response)
-      }
+      })
     }
   }
 
+  override def updateProfile(userInfo: UserInfo, propertyValueMap: Map[String, String]): Future[Iterable[Try[Unit]]] = {
+    Future.sequence(propertyValueMap.map { case (key, value) => saveKeyValue(userInfo, key, value)})
```
dvoet: do we really have to do these one by one? Is there no batch way to set a bunch of keys in thurloe? If not, open a ticket for that please.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
```diff
@@ -47,67 +43,42 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
...
-          Some(FireCloudKeyValue(Some(key), Some(value)))
-        ))
-    } map { response =>
-      response.status match {
+  override def getAllUserValuesForKey(key: String): Future[Map[String, String]] = {
+    val queryUri = Uri(UserService.remoteGetQueryURL).withQuery(Map("key"->key))
+
+    adminAuthedRequest(Get(queryUri), false, true) map { response =>
+      if(response.status != OK) throwBadResponse(response)
+      else response.entity.as[Seq[ThurloeKeyValue]] match {
```
dvoet: use `unmarshal[Seq[ThurloeKeyValue]]` in the pipeline instead of an if statement to check the response then the match.

dvoet: All this internal error handling is messy. It looks like the only reason we have throwBadResponse is to wrap it up in an error report that has thurloe as a source. The latest version of rawls-model has a trait called ErrorReportable to help with that. I would use that, create a function called wrapExceptions that encloses all the public functions that call apis in this class and wraps raw exceptions in error report exceptions with the right source.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
dvoet: I would also call this saveKeyValues

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
```diff
@@ -1,84 +1,185 @@
...
+  def updateNihLinkAndSyncSelf(userInfo: UserInfo, nihLink: NIHLink) = {
+    val syncWhiteListResult = syncWhitelist(Some((userInfo.getUniqueId, nihLink.linkedNihUsername)))
+    val profilePropertyMap = nihLink.propertyValueMap
+    val propertyUpdates = thurloeDao.updateProfile(userInfo, profilePropertyMap)
+    val profileResponse = propertyUpdates.flatMap { responses =>
+      val allSucceeded = responses.forall(_.isSuccess)
+
+      allSucceeded match {
+        case true => Future.successful(RequestComplete(OK))
+        case false => Future.successful(RequestComplete(InternalServerError))
```
dvoet: use RequestCompleteWithErrorReport and give a message

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
```diff
@@ -1,84 +1,185 @@
...
-      case None => Future.successful(RequestComplete(StatusCodes.NotFound))
+      case None => Future.successful(RequestComplete(NotFound))
     }
   }
+
+  def updateNihLinkAndSyncSelf(userInfo: UserInfo, nihLink: NIHLink) = {
+    val syncWhiteListResult = syncWhitelist(Some((userInfo.getUniqueId, nihLink.linkedNihUsername)))
+    val profilePropertyMap = nihLink.propertyValueMap
+    val propertyUpdates = thurloeDao.updateProfile(userInfo, profilePropertyMap)
+    val profileResponse = propertyUpdates.flatMap { responses =>
```
dvoet: since all outcomes are Future.successful, use map instead of flatMap

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
```diff
@@ -1,84 +1,185 @@
...
+      val allSucceeded = responses.forall(_.isSuccess)
+
+      allSucceeded match {
+        case true => Future.successful(RequestComplete(OK))
+        case false => Future.successful(RequestComplete(InternalServerError))
+      }
+    } recover { case e: Throwable => RequestCompleteWithErrorReport(InternalServerError,
```
dvoet: I don't think this recover can happen, remove it

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
```diff
@@ -1,84 +1,185 @@
...
+    }
+
+    syncWhiteListResult map (Success(_)) recover { case t => Failure(t) } flatMap {
+      case Success(_) => profileResponse
+      case Failure(t) => Future(RequestCompleteWithErrorReport(InternalServerError,
+        "Error updating NIH link", t))
+    }
+  }
+
+  def syncWhitelist(subjectId: Option[(String, String)] = None): Future[PerRequestMessage] = Try {
```
dvoet: The Try does not do anything, remove it

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
```diff
@@ -1,84 +1,185 @@
...
+
+      allSucceeded match {
+        case true => Future.successful(RequestComplete(OK))
+        case false => Future.successful(RequestComplete(InternalServerError))
+      }
+    } recover { case e: Throwable => RequestCompleteWithErrorReport(InternalServerError,
+      "Unexpected error updating NIH link", e)
+    }
+
+    syncWhiteListResult map (Success(_)) recover { case t => Failure(t) } flatMap {
```
dvoet: I can't follow this error handling logic... seems to me whatever is going on here is handled by the default logic in Futures. Change this to a for comprehension (with better var names):
```
    val z = for {
      x <- syncWhiteListResult
      y <- profileResponse
    } yield {
      y
    }
    z.recover { case t => RequestCompleteWithErrorReport(InternalServerError, "Error updating NIH link", t) }
```

or the 1 liner
```
    syncWhiteListResult flatMap { _ => profileResponse } recover { case t => RequestCompleteWithErrorReport(InternalServerError, "Error updating NIH link", t) }
```

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
dvoet: I think this function should be split in 2, one for a single user and 1 for all users. Looking at the logic it looks like to separate code paths just bundled into 1 function. Splitting it will make intensions clear.

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
```diff
@@ -1,84 +1,185 @@
...
+      else RequestCompleteWithErrorReport(InternalServerError, "Error synchronizing NIH whitelist")
+    }
+  }.get recover {
+    // intentionally quash errors so as not to leak sensitive data
+    case _ => RequestCompleteWithErrorReport(InternalServerError, "Error synchronizing NIH whitelist")
+  }
+
+}
+
+object NIHWhitelistUtils extends FireCloudRequestBuilding {
```
dvoet: is there a reason this is a separate object?

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
```diff
@@ -1,84 +1,185 @@
...
 
   override def receive = {
-    case GetStatus(userInfo: UserInfo) => getStatus(userInfo) pipeTo sender
+    case GetNihStatus(userInfo: UserInfo) => getNihStatus(userInfo) pipeTo sender
+    case UpdateNIHLinkAndSyncSelf(userInfo: UserInfo, nihLink: NIHLink) => updateNihLinkAndSyncSelf(userInfo: UserInfo, nihLink: NIHLink) pipeTo sender
+    case SyncWhitelist => syncWhitelist() pipeTo sender
   }
 
-  private def getStatus(userInfo: UserInfo): Future[PerRequestMessage] = {
+  private def getNihStatus(userInfo: UserInfo): Future[PerRequestMessage] = {
```
dvoet: split this class into an an Actor class that has only the receive function and a trait that has all the goods. This makes testing easier.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/RestJsonClient.scala
```diff
@@ -16,22 +18,32 @@ import scala.concurrent.{ExecutionContext, Future}
...
+trait RestJsonClient extends FireCloudRequestBuilding {
   implicit val system: ActorSystem
   implicit val executionContext: ExecutionContext
 
-  def userAuthedRequest(req: HttpRequest, compressed: Boolean = false)(implicit userInfo: WithAccessToken): Future[HttpResponse] = {
-    val pipeline = if (compressed) {
-      addCredentials(userInfo.accessToken) ~> addHeader (`Accept-Encoding`(gzip)) ~> sendReceive ~> decode(Gzip)
-    } else {
-      addCredentials(userInfo.accessToken) ~> sendReceive
+  def userAuthedRequest(req: HttpRequest, compressed: Boolean = false, useFireCloudHeader: Boolean = false)(implicit userInfo: WithAccessToken): Future[HttpResponse] = {
```
dvoet: userAuthedRequest and adminAuthedRequest are identical except for the function to add creds, make an internal function that takes the addCreds function as a param.
```
  def userAuthedRequest(req: HttpRequest, compressed: Boolean = false, useFireCloudHeader: Boolean = false)(implicit userInfo: WithAccessToken): Future[HttpResponse] = authedRequest(addCredentials(userInfo.accessToken))(req, compressed, useFireCloudHeader)
  def adminAuthedRequest(req: HttpRequest, compressed: Boolean = false, useFireCloudHeader: Boolean = false): Future[HttpResponse] = authedRequest(addAdminCredentials)(req, compressed, useFireCloudHeader)

  private def authedRequest(addCreds: RequestTransformer)(req: HttpRequest, compressed: Boolean = false, useFireCloudHeader: Boolean = false)(implicit userInfo: WithAccessToken): Future[HttpResponse] = {
    val pipeline = (compressed, useFireCloudHeader) match {
      case (true, true) => addCreds ~> addFireCloudCredentials ~> addHeader (`Accept-Encoding`(gzip)) ~> sendReceive ~> decode(Gzip)
      case (true, false) => addCreds ~> addHeader (`Accept-Encoding`(gzip)) ~> sendReceive ~> decode(Gzip)
      case (false, true) => addCreds ~> addFireCloudCredentials ~> addHeader (`Accept-Encoding`(gzip)) ~> sendReceive ~> decode(Gzip)
      case _ => addCreds ~> sendReceive
    }
    pipeline(req)
  }
```

src/test/scala/org/broadinstitute/dsde/firecloud/service/NihApiServiceSpec.scala
```diff
@@ -0,0 +1,93 @@
...
+  /* Test scenario:
+     1 user that is linked but their TCGA access has expired. they should be removed
+     1 user that is linked but they have no expiration date stored in their profile. they should be removed
+     1 user that is linked and has active TCGA access. they should remain in the dbGapAuthorizedUsers group
+   */
+  it should "return NoContent and properly sync the whitelist for users of different link statuses" in withDefaultApiServices { services =>
+    Post("/sync_whitelist") ~> sealRoute(services.syncRoute) ~> check {
+      status should equal(NoContent)
+    }
+    services.rawlsDao.groups(dbGapAuthorizedUsersGroupName) should contain theSameElementsAs Set(services.thurloeDao.linkedUserId)
```
dvoet: is there reason to distrust `theSameElementsAs`? Given that it is a singleton set you could just as well test equality

dvoet (dismissed):

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
MatthewBemis: yes, thurloe only takes one kv pair at a time. ticket filed here: https://broadinstitute.atlassian.net/browse/GAWB-1734

src/main/scala/org/broadinstitute/dsde/firecloud/model/Profile.scala
```diff
@@ -61,7 +61,6 @@ case class Profile (
     pi: String,
     nonProfitStatus: String,
     linkedNihUsername: Option[String] = None,
-    lastLinkTime: Option[Long] = None,
```
breilly2, line 57: I'm guessing that this isn't a "user-provided profile field" but wanted to ask in case this warrants incrementing `currentVersion` above.

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockThurloeDAO.scala
```diff
@@ -12,27 +15,125 @@ import scala.util.{Success, Try}
...
+        FireCloudKeyValue(Some("title"), Some("Mr.")),
+        FireCloudKeyValue(Some("contactEmail"), Some("bobby.testerson@example.com")),
+        FireCloudKeyValue(Some("institute"), Some("Broad")),
+        FireCloudKeyValue(Some("institutionalProgram"), Some("Something")),
+        FireCloudKeyValue(Some("programLocationCity"), Some("Cambridge")),
+        FireCloudKeyValue(Some("programLocationState"), Some("MA")),
+        FireCloudKeyValue(Some("programLocationCountry"), Some("USA")),
+        FireCloudKeyValue(Some("pi"), Some("Abby Testerson")),
+        FireCloudKeyValue(Some("nonProfitStatus"), Some("true")),
+        FireCloudKeyValue(Some("linkedNihUsername"), Some("firecloud-dev")),
```
breilly2: I like that you have covered a lot of different test scenarios!

It looks to me like the only `linkedNihUsername` and `linkExpireTime` are the significant variables for the tests. Can the rest be extracted into a set that is cloned (or something) to emphasize the test variables? Better yet to have the tests do the cloning and setting `linkedNihUsername` and `linkExpireTime` for the specific test conditions.

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
```diff
@@ -1,84 +1,174 @@
...
 import spray.httpx.SprayJsonSupport._
 import spray.json.DefaultJsonProtocol._
 
 import scala.concurrent.{ExecutionContext, Future}
+import scala.io.Source
+import scala.util.{Failure, Success, Try}
 
 
 case class NihStatus(
-  loginRequired: Boolean,
```
breilly2, line 20: Should also remove `loginRequired` and `descriptionSinceLastLink` from api-docs.yaml. Also, I'm assuming we don't have to worry about any external (non-FireCloud) clients when making API changes to this service, right?

src/test/scala/org/broadinstitute/dsde/firecloud/service/NihApiServiceSpec.scala
```diff
@@ -0,0 +1,96 @@
...
+    Post("/nih/callback", validJwtOnWhitelist) ~> dummyUserIdHeaders(services.thurloeDao.normalUserId) ~> sealRoute(services.nihRoutes) ~> check {
+      println(response)
+      status should equal(OK)
+    }
+    assert(services.rawlsDao.groups(dbGapAuthorizedUsersGroupName).contains(services.thurloeDao.normalUserId))
+  }
+
+  it should "return InternalServerError when NIH linking with a valid JWT and the user is NOT on the whitelist" in withDefaultApiServices { services =>
+    Post("/nih/callback", validJwtNotOnWhitelist) ~> dummyUserIdHeaders(services.thurloeDao.normalUserId) ~> sealRoute(services.nihRoutes) ~> check {
+      status should equal(InternalServerError)
```
breilly2, line 55: Why `InternalServerError`? Does it maybe come from of a whitelist sync failure or something like that? If not, maybe there's a more appropriate status to use.

If this is just asserting the existing behavior that would require code changes in UI to change, I think it's worth a clean-up ticket in the backlog.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/RestJsonClient.scala
```diff
@@ -16,22 +18,25 @@ import scala.concurrent.{ExecutionContext, Future}
...
-    } else {
-      addCredentials(userInfo.accessToken) ~> sendReceive
+  def userAuthedRequest(req: HttpRequest, compressed: Boolean = false, useFireCloudHeader: Boolean = false)(implicit userInfo: WithAccessToken): Future[HttpResponse] = authedRequest(addCredentials(userInfo.accessToken))(req, compressed, useFireCloudHeader)
+  def adminAuthedRequest(req: HttpRequest, compressed: Boolean = false, useFireCloudHeader: Boolean = false): Future[HttpResponse] = authedRequest(addAdminCredentials)(req, compressed, useFireCloudHeader)
+
+  private def authedRequest(addCreds: RequestTransformer)(req: HttpRequest, compressed: Boolean = false, useFireCloudHeader: Boolean = false): Future[HttpResponse] = {
+    val pipeline = (compressed, useFireCloudHeader) match {
+      case (true, true) => addCreds ~> addFireCloudCredentials ~> addHeader (`Accept-Encoding`(gzip)) ~> sendReceive ~> decode(Gzip)
+      case (true, false) => addCreds ~> addHeader (`Accept-Encoding`(gzip)) ~> sendReceive ~> decode(Gzip)
+      case (false, true) => addCreds ~> addFireCloudCredentials ~> addHeader (`Accept-Encoding`(gzip)) ~> sendReceive ~> decode(Gzip)
```
breilly2: `compressed` is `false`, so this should not be using gzip?

breilly2 (changes requested):
Only minor things to change, specifically swagger API and the gzip thing (If I'm reading it correctly). The rest of my comments are only suggestions and things to make sure they don't need to be changed.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/RestJsonClient.scala
MatthewBemis: you're right- nice catch

src/main/scala/org/broadinstitute/dsde/firecloud/model/Profile.scala
MatthewBemis: yes that's right. `currentVersion` is used to prompt users to fill out a new version of the profile when logging in so no change needed here

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
MatthewBemis: this _is_ a public facing API but it's really unlikely that anyone is consuming this outside of firecloud UI

src/test/scala/org/broadinstitute/dsde/firecloud/service/NihApiServiceSpec.scala
MatthewBemis: the internalservererror was done as part of the quick hotfix to prevent non-whitelisted NIH users from getting access to controlled TCGA data. we do have a ticket to improve the error messaging and everything related to this situation

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockThurloeDAO.scala
```diff
@@ -20,79 +20,37 @@ class MockThurloeDAO extends ThurloeDAO {
...
-        FireCloudKeyValue(Some("title"), Some("Mr.")),
-        FireCloudKeyValue(Some("contactEmail"), Some("bobby.testerson@example.com")),
-        FireCloudKeyValue(Some("institute"), Some("Broad")),
-        FireCloudKeyValue(Some("institutionalProgram"), Some("Something")),
-        FireCloudKeyValue(Some("programLocationCity"), Some("Cambridge")),
-        FireCloudKeyValue(Some("programLocationState"), Some("MA")),
-        FireCloudKeyValue(Some("programLocationCountry"), Some("USA")),
-        FireCloudKeyValue(Some("pi"), Some("Abby Testerson")),
-        FireCloudKeyValue(Some("nonProfitStatus"), Some("true")),
+      linkedUserId -> baseProfile.++(Set(
```
breilly2: Nice! I think this is much easier to read and understand the intent.

breilly2 (approved):
👍

PR #382, opened 2017-02-23
Ah gawb 1510: return groups for discoverability

ahaessly:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -466,7 +466,10 @@ paths:
       responses:
         200:
           description: Names of the allowed groups
-
+          schema:
+            type: array
+            items:
+              type: string
   /api/library/search:
     post:
       tags:
@@ -560,11 +563,13 @@ paths:
           in: body
       responses:
         200:
-          description: Success
+          description: The groups that have been assigned to be able to discover the dataset
           schema:
-            $ref: '#/definitions/Workspace'
+            type: array
+            items:
+              type: string
         500:
-          description: Rawls Internal Error
+          description: Error
           schema:
             $ref: '#/definitions/ErrorReport'
 
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -456,6 +456,17 @@ paths:
           schema:
             $ref: '#/definitions/LibraryBulkIndexResponse'
 
+  /api/library/groups:
+    get:
+      tags:
+        - Library
+      operationId: groups
+      summary: |
+        returns groups that can be used for discoverable datasets
+      responses:
+        200:
+          description: Names of the allowed groups
+
   /api/library/search:
     post:
       tags:
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -544,6 +544,30 @@ paths:
       summary: Query to see if you're a library curator
       operationId: isCurator
 
+  /api/library/{workspaceNamespace}/{workspaceName}/discoverableBy:
+    put:
+      tags:
+        - Library
+      operationId: updateDiscoverableGroups
+      summary: |
+        set the groups that can discover this library dataset
+      parameters:
+        - $ref: '#/parameters/workspaceNamespaceParam'
+        - $ref: '#/parameters/workspaceNameParam'
+        - name: discoverableGroups
+          description: List of group names, or empty list for no restrictions
+          required: true
+          in: body
+      responses:
+        200:
+          description: Success
+          schema:
+            $ref: '#/definitions/Workspace'
+        500:
+          description: Rawls Internal Error
+          schema:
+            $ref: '#/definitions/ErrorReport'
+
   /api/library/{workspaceNamespace}/{workspaceName}/metadata:
     put:
       tags:
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -547,7 +547,7 @@ paths:
       summary: Query to see if you're a library curator
       operationId: isCurator
 
-  /api/library/{workspaceNamespace}/{workspaceName}/discoverableBy:
+  /api/library/{workspaceNamespace}/{workspaceName}/discoverableGroups:
     put:
       tags:
         - Library
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
@@ -20,9 +20,8 @@ import spray.json.JsonParser.ParsingException
 import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol.{impLibraryBulkIndexResponse, impLibrarySearchResponse}
 import org.broadinstitute.dsde.rawls.model.WorkspaceJsonSupport.{AttributeNameFormat, WorkspaceFormat}
 import org.broadinstitute.dsde.rawls.model.Attributable.AttributeMap
-import org.broadinstitute.dsde.rawls.model.AttributeUpdateOperations.{AddListMember, AttributeUpdateOperation, CreateAttributeValueList, RemoveAttribute}
+import org.broadinstitute.dsde.rawls.model.AttributeUpdateOperations.{AddListMember, AttributeUpdateOperation, RemoveAttribute}
 
-import scala.collection.JavaConverters._
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success, Try}
 
@@ -33,7 +32,7 @@ object LibraryService {
 
   sealed trait LibraryServiceMessage
   case class UpdateAttributes(ns: String, name: String, attrsJsonString: String) extends LibraryServiceMessage
-  case class UpdateDiscoverableBy(ns: String, name: String, newGroups: Seq[String]) extends LibraryServiceMessage
+  case class UpdateDiscoverableByGroups(ns: String, name: String, newGroups: Seq[String]) extends LibraryServiceMessage
   case class SetPublishAttribute(ns: String, name: String, value: Boolean) extends LibraryServiceMessage
   case object IndexAll extends LibraryServiceMessage
   case class FindDocuments(criteria: LibrarySearchParams) extends LibraryServiceMessage
@@ -58,7 +57,7 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
 
   override def receive = {
     case UpdateAttributes(ns: String, name: String, attrsJsonString: String) => updateAttributes(ns, name, attrsJsonString) pipeTo sender
-    case UpdateDiscoverableBy(ns: String, name: String, newGroups: Seq[String]) => updateDiscoverableBy(ns, name, newGroups) pipeTo sender
+    case UpdateDiscoverableByGroups(ns: String, name: String, newGroups: Seq[String]) => updateDiscoverableByGroups(ns, name, newGroups) pipeTo sender
     case SetPublishAttribute(ns: String, name: String, value: Boolean) => setWorkspaceIsPublished(ns, name, value) pipeTo sender
     case IndexAll => asAdmin {indexAll} pipeTo sender
     case FindDocuments(criteria: LibrarySearchParams) => findDocuments(criteria) pipeTo sender
@@ -81,8 +80,8 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
   /*
    * This attribute is allowed to be modified by on owner or someone with canShare for the workspace
    */
-  def updateDiscoverableBy(ns: String, name: String, newGroups: Seq[String]): Future[PerRequestMessage] = {
-    if (FireCloudConfig.ElasticSearch.discoverGroupNames.containsAll(newGroups.asJavaCollection)) {
+  def updateDiscoverableByGroups(ns: String, name: String, newGroups: Seq[String]): Future[PerRequestMessage] = {
+    if (newGroups.forall { g => FireCloudConfig.ElasticSearch.discoverGroupNames.contains(g) }) {
       rawlsDAO.getWorkspace(ns, name) map { workspaceResponse =>
         if (workspaceResponse.accessLevel >= WorkspaceAccessLevels.Owner) {
           // this is technically vulnerable to a race condition in which the workspace attributes have changed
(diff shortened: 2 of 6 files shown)


src/test/scala/org/broadinstitute/dsde/firecloud/service/LibraryApiServiceSpec.scala
```diff
@@ -204,6 +205,15 @@ class LibraryApiServiceSpec extends BaseServiceSpec with LibraryApiService {
           }
         }
       }
+      "GET on " + libraryGroupsPath - {
+        "should return the all broad users group" in {
+          new RequestBuilder(HttpMethods.GET)(libraryGroupsPath) ~> dummyUserIdHeaders("1234") ~> sealRoute(libraryRoutes) ~> check {
+            status should equal(OK)
+            val respdata = response.entity.asString
+            assert(respdata.contains("all_broad_users"))
```
rushtong: Instead of hard-coding this, it might be better to read the values from the config. This way, if we add more, the test doesn't have to change. Something like:
```
FireCloudConfig.ElasticSearch.discoverGroupNames.foreach { group => assert(respdata.contains(group)) }
```

rushtong (approved):
👍 LG, WG. One minor comment inline.

src/test/scala/org/broadinstitute/dsde/firecloud/service/LibraryApiServiceSpec.scala
davidangb: even better, instead of checking `contains`, turn the expected and actual values into `Set`s, and check for equality between the sets.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -456,6 +456,17 @@ paths:
...
+    get:
+      tags:
+        - Library
+      operationId: groups
+      summary: |
+        returns groups that can be used for discoverable datasets
+      responses:
+        200:
+          description: Names of the allowed groups
+
```
davidangb: for clarity:
```
      responses:
        200:
          description: Names of the allowed groups
          schema:
            type: array
            items:
              type: string
```

davidangb (commented):
couple minor comments inline, otherwise looking good!

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -54,14 +57,51 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
...
+    case PopulateSuggest(field: String, text: String) => populateSuggest(field: String, text: String) pipeTo sender
   }
 
+  def hasAccessOrCurator(workspaceResponse: WorkspaceResponse, neededLevel: WorkspaceAccessLevels.WorkspaceAccessLevel): Boolean = {
+    val wsaccess = workspaceResponse.accessLevel >= neededLevel
+    if (!wsaccess)
+      tryIsCurator(userInfo) map { value => return value }
+    wsaccess
+  }
+
```
ahaessly: this function doesn't work because i can't return the value from inside the map (i.e. the future).
i would like some sort of convenience function because we do these types of checks in several places below... any suggestions on what to do?

src/main/resources/swagger/api-docs.yaml
```diff
@@ -533,6 +547,32 @@ paths:
       summary: Query to see if you're a library curator
       operationId: isCurator
 
+  /api/library/{workspaceNamespace}/{workspaceName}/discoverableBy:
```
rushtong: nitpick on endpoint name - I think it should be more noun-like for convention. `discoverableGroups` seems most appropriate since you're using that term already.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -533,6 +547,32 @@ paths:
...
+    put:
```
rushtong: Do we need a get? Not sure if that comes back as part of workspace or not.

rushtong: Also ... why isn't there a `POST` for this? Initially, one would be `POST`ing a new group and subsequent calls would be making a `PUT` request.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -54,14 +57,52 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
...
+    if (FireCloudConfig.ElasticSearch.discoverGroupNames.containsAll(newGroups.asJavaCollection)) {
+      rawlsDAO.getWorkspace(ns, name) map { workspaceResponse =>
+        if (workspaceResponse.accessLevel >= WorkspaceAccessLevels.Owner) {
+          // this is technically vulnerable to a race condition in which the workspace attributes have changed
+          // between the time we retrieved them and here, where we update them.
+          val remove = Seq(RemoveAttribute(discoverableWSAttribute))
+          val operations = newGroups map (group => AddListMember(discoverableWSAttribute, new AttributeString(group)))
+          RequestComplete(patchWorkspace(ns, name, remove ++ operations, isPublished(workspaceResponse)))
+        } else {
+          RequestCompleteWithErrorReport(Forbidden, "must be an owner to set discoverable By")
```
rushtong: nitpick: `Must be an owner to set the groups this workspace is discoverable by.` seems better. My inner grammar police is screaming not to end a sentence with a preposition, but apparently this is considered OK now.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -54,14 +57,52 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
...
+
+  def isPublished(workspaceResponse: WorkspaceResponse): Boolean = {
+    workspaceResponse.workspace.attributes.get(publishedFlag).fold(false)(_.asInstanceOf[AttributeBoolean].value)
+  }
+
+  /*
+   * This attribute is allowed to be modified by on owner or someone with canShare for the workspace
+   */
+  def updateDiscoverableBy(ns: String, name: String, newGroups: Seq[String]): Future[PerRequestMessage] = {
+    if (FireCloudConfig.ElasticSearch.discoverGroupNames.containsAll(newGroups.asJavaCollection)) {
```
rushtong: Could be rewritten without the java collection coercion:
```
if (newGroups.forall { g => FireCloudConfig.ElasticSearch.discoverGroupNames.contains(g) })
```

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -54,14 +57,52 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
...
+          // between the time we retrieved them and here, where we update them.
+          val remove = Seq(RemoveAttribute(discoverableWSAttribute))
+          val operations = newGroups map (group => AddListMember(discoverableWSAttribute, new AttributeString(group)))
+          RequestComplete(patchWorkspace(ns, name, remove ++ operations, isPublished(workspaceResponse)))
+        } else {
+          RequestCompleteWithErrorReport(Forbidden, "must be an owner to set discoverable By")
+        }
+      }
+    } else {
+      Future(RequestCompleteWithErrorReport(BadRequest, s"groups must be subset of allowable groups: %s".format(FireCloudConfig.ElasticSearch.discoverGroupNames)))
```
rushtong: When group names becomes more than one name, you'll want this to be better formatted:
```
FireCloudConfig.ElasticSearch.discoverGroupNames.toArray.mkString(", ")
```

rushtong (approved):
👍  LG, WG. Minor comments inline - feel free to address as you wish.

src/main/resources/swagger/api-docs.yaml
ahaessly: this is really just setting a library value on an existing workspace. it's broken out into a separate call just for the API because it has different access control checks than the other library fields and these checks are handled when setting the value from the UI.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -54,14 +56,52 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
...
+      Future.successful(wsaccess)
+  }
+
+  def isPublished(workspaceResponse: WorkspaceResponse): Boolean = {
+    workspaceResponse.workspace.attributes.get(publishedFlag).fold(false)(_.asInstanceOf[AttributeBoolean].value)
+  }
+
+  /*
+   * This attribute is allowed to be modified by on owner or someone with canShare for the workspace
+   */
```
davidangb: this comment doesn't match the implementation below.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -81,38 +121,65 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
...
-                k => k.namespace == AttributeName.libraryNamespace && k.name != LibraryService.publishedFlag.name)
-              val republish = workspaceResponse.workspace.attributes.get(publishedFlag).fold(false)(_.asInstanceOf[AttributeBoolean].value)
-              rawlsDAO.patchWorkspaceAttributes(ns, name, allOperations) map { newws =>
-                if (republish) {
-                  // we do not need to delete before republish
-                  publishDocument(newws)
-                }
-                RequestComplete(newws)
+            rawlsDAO.getWorkspace(ns, name) map  { workspaceResponse =>
+              if (workspaceResponse.accessLevel >= WorkspaceAccessLevels.Write) {
```
davidangb: you shouldn't need this check; Rawls handles it. Is there a reason to enforce it here? Are you getting a head start on handling catalog permission?

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -81,38 +121,65 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
...
+            RequestComplete(ws)
+          }
+        } else {
+          // verify owner on workspace
+          if (workspaceResponse.accessLevel < WorkspaceAccessLevels.Write)
+            Future(RequestCompleteWithErrorReport(Forbidden, "must be a curator or have at least write privileges"))
           else
-            removeDocument(ws)
-          RequestComplete(ws)
+            Future(RequestCompleteWithErrorReport(Forbidden, "must be an owner or a curator"))
```
davidangb: fyi - check out GAWB-1683. That ticket is *NOT* in scope for this PR - let's avoid scope creep and keep PRs small - mentioning it just so you're aware.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -81,38 +121,65 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
...
+            if (value)
+              publishDocument(ws)
+            else
+              removeDocument(ws)
+            RequestComplete(ws)
+          }
+        } else {
+          // verify owner on workspace
+          if (workspaceResponse.accessLevel < WorkspaceAccessLevels.Write)
+            Future(RequestCompleteWithErrorReport(Forbidden, "must be a curator or have at least write privileges"))
```
davidangb: error message is now wrong re: curator - and in the case where the workspace isn't yet published, user would need owner privs.

src/test/scala/org/broadinstitute/dsde/firecloud/service/LibraryApiServiceSpec.scala
```diff
@@ -102,6 +105,13 @@ class LibraryApiServiceSpec extends BaseServiceSpec with LibraryApiService {
           }
         }
       }
+      "POST as writer on " + publishedPath() - {
+        "should be Forbidden" in {
+          new RequestBuilder(HttpMethods.POST)(publishedPath("unpublishedwriter")) ~> dummyUserIdHeaders("1234") ~> sealRoute(libraryRoutes) ~> check {
+            status should equal(Forbidden)
+          }
+        }
+      }
```
davidangb, line 114: add a test for POST as writer on an already-published workspace, which should succeed.

src/test/scala/org/broadinstitute/dsde/firecloud/service/LibraryApiServiceSpec.scala
```diff
@@ -204,6 +224,15 @@ class LibraryApiServiceSpec extends BaseServiceSpec with LibraryApiService {
           }
         }
       }
+      "GET on " + libraryGroupsPath - {
+        "should return the all broad users group" in {
+          new RequestBuilder(HttpMethods.GET)(libraryGroupsPath) ~> dummyUserIdHeaders("1234") ~> sealRoute(libraryRoutes) ~> check {
+            status should equal(OK)
+            val respdata = response.entity.asString.parseJson.convertTo[Seq[String]]
+            assert(respdata.forall { g => FireCloudConfig.ElasticSearch.discoverGroupNames.contains(g) })
```
davidangb: instead of testing `forall`, turn them into `Set`s and compare equality between the sets.

davidangb (commented):
A few comments inline, I haven't tested!

src/test/scala/org/broadinstitute/dsde/firecloud/service/LibraryApiServiceSpec.scala
rushtong: The `forall` was my suggestion. I think what you need is to check that the set of `FireCloudConfig.ElasticSearch.discoverGroupNames` contains all of the elements in `respdata`. Simple set equality will fail as soon as we have more than one discoverable group and the user is only adding one of them to the workspace.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
ahaessly: i did have extra checks that rawls didn't do... but i'll fix since i'm not sure what the solution will be next

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
ahaessly: refactored this code. the removeDocument is for unpublishing not addressing 1683, but admit code was confusing. hopefully my changes are clearer, even if duplicating some code. let me know if you have a better way :)

src/test/scala/org/broadinstitute/dsde/firecloud/service/LibraryApiServiceSpec.scala
ahaessly: greg, your response was for a different part of the code, and i made the change there. this is the test and these groups should always have the same members

rushtong: Thanks Andrea!

ahaessly:
i don't think i force pushed, but i don't see my responses anymore. let me know if i haven't answered a question you would like me to

davidangb (approved):
👍 looks good!

PR #386, opened 2017-02-28
GAWB-1644: copy ontology parent-node information into Library

davidangb:
Warning: PO has not seen this yet; it is possible functionality will need to change.

When indexing datasets into Library's ElasticSearch, if the dataset is tagged with an ontology node id for the "indication" field, query consent-ontology for that node's parents. Then, save the parent information in Library's ES.

By having the parent information present in ES, we enable ontology-aware search.

To test:
* catalog a dataset. While cataloging, make a note of the indication's node id - for instance, `http://purl.obolibrary.org/obo/DOID_1485` for cystic fibrosis.
* publish the dataset.
* look up the node id using consent-ontology's search API: https://consent-ontology.dsde-dev.broadinstitute.org/swagger/#!/Search/get_search, and make note of the parents' labels. For instance, monogenic disease is a parent of cystic fibrosis.
* go to Library in the UI and enter a parent label into the main search box, and see that your dataset is returned in search results.

Notes:
* I have explicitly chosen to NOT make parent-node definitions or synonyms searchable. This is a precision/recall tradeoff - I believe making these searchable will decrease precision and increase recall by too much.
* I have explicitly chosen to NOT make parent-node labels available in autocomplete suggestions. This is in order to not clutter the suggestions.
* I am not utilizing the `order` value of the parent to assist with relevancy. This was a stretch goal; I'll handle that in a future story.

There's a ton of commits in this PR, sorry!

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpOntologyDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpOntologyDAO.scala
@@ -1,0 +1,22 @@
+package org.broadinstitute.dsde.firecloud.dataaccess
+
+import akka.actor.ActorSystem
+import org.broadinstitute.dsde.firecloud.model.Ontology.SearchResponse
+import org.broadinstitute.dsde.firecloud.utils.RestJsonClient
+import spray.client.pipelining._
+import spray.http.StatusCodes._
+import spray.http.{HttpResponse, OAuth2BearerToken}
+import spray.httpx.SprayJsonSupport._
+import spray.httpx.unmarshalling._
+import spray.json.DefaultJsonProtocol._
+
+import scala.concurrent.{ExecutionContext, Future}
+
+
+class HttpOntologyDAO(implicit val system: ActorSystem, implicit val executionContext: ExecutionContext)
+  extends OntologyDAO with RestJsonClient {
+
+  override def search(term: String): Future[List[SearchResponse]] = {
+    requestToObject[List[SearchResponse]](Get(ontologySearchUrl))
+  }
+}
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpOntologyDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpOntologyDAO.scala
@@ -2,7 +2,6 @@ package org.broadinstitute.dsde.firecloud.dataaccess
 
 import akka.actor.ActorSystem
 import org.broadinstitute.dsde.firecloud.model.Ontology.TermResource
-import org.broadinstitute.dsde.firecloud.model.UserInfo
 import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
 import org.broadinstitute.dsde.firecloud.utils.RestJsonClient
 import spray.client.pipelining._
@@ -15,7 +14,7 @@ import scala.concurrent.{ExecutionContext, Future}
 class HttpOntologyDAO(implicit val system: ActorSystem, implicit val executionContext: ExecutionContext)
   extends OntologyDAO with RestJsonClient {
 
-  override def search(term: String)(implicit ec: ExecutionContext, userInfo: UserInfo): Future[List[TermResource]] = {
-    requestToObject[List[TermResource]](Get(ontologySearchUrl))
+  override def search(term: String)(implicit ec: ExecutionContext): Future[List[TermResource]] = {
+    unAuthedRequestToObject[List[TermResource]](Get(ontologySearchUrl))
   }
 }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpOntologyDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpOntologyDAO.scala
@@ -15,7 +15,7 @@ import scala.concurrent.{ExecutionContext, Future}
 class HttpOntologyDAO(implicit val system: ActorSystem, implicit val executionContext: ExecutionContext)
   extends OntologyDAO with RestJsonClient {
 
-  override def search(term: String)(implicit userInfo: UserInfo): Future[List[TermResource]] = {
+  override def search(term: String)(implicit ec: ExecutionContext, userInfo: UserInfo): Future[List[TermResource]] = {
     requestToObject[List[TermResource]](Get(ontologySearchUrl))
   }
 }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpOntologyDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpOntologyDAO.scala
@@ -1,7 +1,9 @@
 package org.broadinstitute.dsde.firecloud.dataaccess
 
 import akka.actor.ActorSystem
-import org.broadinstitute.dsde.firecloud.model.Ontology.SearchResponse
+import org.broadinstitute.dsde.firecloud.model.Ontology.TermResource
+import org.broadinstitute.dsde.firecloud.model.WithAccessToken
+import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
 import org.broadinstitute.dsde.firecloud.utils.RestJsonClient
 import spray.client.pipelining._
 import spray.http.StatusCodes._
@@ -16,7 +18,7 @@ import scala.concurrent.{ExecutionContext, Future}
 class HttpOntologyDAO(implicit val system: ActorSystem, implicit val executionContext: ExecutionContext)
   extends OntologyDAO with RestJsonClient {
 
-  override def search(term: String): Future[List[SearchResponse]] = {
-    requestToObject[List[SearchResponse]](Get(ontologySearchUrl))
+  override def search(term: String)(implicit userToken: WithAccessToken): Future[List[TermResource]] = {
+    requestToObject[List[TermResource]](Get(ontologySearchUrl))
   }
 }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpOntologyDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpOntologyDAO.scala
@@ -2,14 +2,11 @@ package org.broadinstitute.dsde.firecloud.dataaccess
 
 import akka.actor.ActorSystem
 import org.broadinstitute.dsde.firecloud.model.Ontology.TermResource
-import org.broadinstitute.dsde.firecloud.model.WithAccessToken
+import org.broadinstitute.dsde.firecloud.model.UserInfo
 import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
 import org.broadinstitute.dsde.firecloud.utils.RestJsonClient
 import spray.client.pipelining._
-import spray.http.StatusCodes._
-import spray.http.{HttpResponse, OAuth2BearerToken}
 import spray.httpx.SprayJsonSupport._
-import spray.httpx.unmarshalling._
 import spray.json.DefaultJsonProtocol._
 
 import scala.concurrent.{ExecutionContext, Future}
@@ -18,7 +15,7 @@ import scala.concurrent.{ExecutionContext, Future}
 class HttpOntologyDAO(implicit val system: ActorSystem, implicit val executionContext: ExecutionContext)
   extends OntologyDAO with RestJsonClient {
 
-  override def search(term: String)(implicit userToken: WithAccessToken): Future[List[TermResource]] = {
+  override def search(term: String)(implicit userInfo: UserInfo): Future[List[TermResource]] = {
     requestToObject[List[TermResource]](Get(ontologySearchUrl))
   }
 }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpOntologyDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpOntologyDAO.scala
@@ -7,6 +7,7 @@ import org.broadinstitute.dsde.firecloud.utils.RestJsonClient
 import spray.client.pipelining._
 import spray.http.Uri
 import spray.httpx.SprayJsonSupport._
+import spray.httpx.unmarshalling._
 import spray.json.DefaultJsonProtocol._
 
 import scala.concurrent.{ExecutionContext, Future}
@@ -15,7 +16,15 @@ import scala.concurrent.{ExecutionContext, Future}
 class HttpOntologyDAO(implicit val system: ActorSystem, implicit val executionContext: ExecutionContext)
   extends OntologyDAO with RestJsonClient {
 
-  override def search(term: String)(implicit ec: ExecutionContext): Future[List[TermResource]] = {
-    unAuthedRequestToObject[List[TermResource]](Get(Uri(ontologySearchUrl).withQuery(("id", term))))
+  override def search(term: String): Future[Option[List[TermResource]]] = {
+    unAuthedRequest(Get(Uri(ontologySearchUrl).withQuery(("id", term)))) map { response =>
+      if (response.status.isSuccess)
+        response.entity.as[List[TermResource]] match {
+          case Right(obj) => Some(obj)
+          case Left(error) => None
+        }
+      else
+        None
+    }
   }
 }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpOntologyDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpOntologyDAO.scala
@@ -5,6 +5,7 @@ import org.broadinstitute.dsde.firecloud.model.Ontology.TermResource
 import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
 import org.broadinstitute.dsde.firecloud.utils.RestJsonClient
 import spray.client.pipelining._
+import spray.http.Uri
 import spray.httpx.SprayJsonSupport._
 import spray.json.DefaultJsonProtocol._
 
@@ -15,6 +16,6 @@ class HttpOntologyDAO(implicit val system: ActorSystem, implicit val executionCo
   extends OntologyDAO with RestJsonClient {
 
   override def search(term: String)(implicit ec: ExecutionContext): Future[List[TermResource]] = {
-    unAuthedRequestToObject[List[TermResource]](Get(ontologySearchUrl))
+    unAuthedRequestToObject[List[TermResource]](Get(Uri(ontologySearchUrl).withQuery(("id", term))))
   }
 }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpOntologyDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpOntologyDAO.scala
@@ -1,6 +1,7 @@
 package org.broadinstitute.dsde.firecloud.dataaccess
 
 import akka.actor.ActorSystem
+import com.typesafe.scalalogging.slf4j.LazyLogging
 import org.broadinstitute.dsde.firecloud.model.Ontology.TermResource
 import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
 import org.broadinstitute.dsde.firecloud.utils.RestJsonClient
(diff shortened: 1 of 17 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpOntologyDAO.scala
```diff
@@ -0,0 +1,21 @@
...
+
+import scala.concurrent.{ExecutionContext, Future}
+
+
+class HttpOntologyDAO(implicit val system: ActorSystem, implicit val executionContext: ExecutionContext)
+  extends OntologyDAO with RestJsonClient {
+
+  override def search(term: String)(implicit ec: ExecutionContext, userInfo: UserInfo): Future[List[TermResource]] = {
+    requestToObject[List[TermResource]](Get(ontologySearchUrl))
+  }
```
davidangb: you technically don't need the `UserInfo` object here, because consent-ontology is open and doesn't require authentication. Removing `UserInfo` would simplify some other code in the unit tests.

HOWEVER, `requestToObject` does require `UserInfo`. In order to remove `UserInfo` you'd have to make a new function similar to `requestToObject` that doesn't work with authentication.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/OntologyDAO.scala
```diff
@@ -0,0 +1,15 @@
...
+import org.broadinstitute.dsde.firecloud.model.UserInfo
+
+import scala.concurrent.{ExecutionContext, Future}
+
+
+trait OntologyDAO {
+
+  lazy val ontologySearchUrl = FireCloudConfig.Duos.baseUrl + "/search"
+
+  def search(term: String)(implicit ec: ExecutionContext, userInfo: UserInfo): Future[List[TermResource]]
```
davidangb: the `term` argument is never appended to the url!

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -119,27 +124,30 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
   }
 
   def publishDocument(ws: Workspace): Unit = {
-    searchDAO.indexDocument(indexableDocument(ws))
+    indexableDocument(ws, ontologyDAO) onSuccess {
+      case document => searchDAO.indexDocument(document)
+    }
```
davidangb: prefer `map` over `onSuccess` for futures, something like:
```
indexableDocument(ws, ontologyDAO) map { doc:Document => searchDAO.indexDocument(doc) }
```
or more concisely:
```
indexableDocument(ws, ontologyDAO) map { searchDAO.indexDocument }
```
( I haven't checked to see if that compiles or works!)

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -20,6 +20,7 @@ import spray.json.JsonParser.ParsingException
 import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol.{impLibraryBulkIndexResponse, impLibrarySearchResponse}
 import org.broadinstitute.dsde.rawls.model.WorkspaceJsonSupport.{AttributeNameFormat, WorkspaceFormat}
 import org.broadinstitute.dsde.rawls.model.Attributable.AttributeMap
+import spray.http.StatusCode
```
davidangb: import not needed

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
```diff
@@ -39,7 +38,15 @@ trait LibraryServiceSupport {
...
       AttributeName.withDefaultNS("workspaceId") -> AttributeString(workspace.workspaceId)
     )
-    Document(workspace.workspaceId, attrfields ++ idfields)
+    val fields = attrfields ++ idfields
+
+    workspace.attributes.get(AttributeName.withLibraryNS("diseaseOntologyID")) match {
+      case Some(id: AttributeString) =>
+        ontologyDAO.search(id.value) map { resources =>
+          Document(workspace.workspaceId, fields + (AttributeName.withDefaultNS("parents") -> AttributeValueRawJson(resources.toString())))
+        }
```
davidangb: I *think* that in
```
ontologyDAO.search(id.value) map { resources =>
```
the value of `resources` will be the target node and the parents, not just the parent information. You'll need to do a little massaging. It might be as easy as `resources.parents.toString`.

src/test/scala/org/broadinstitute/dsde/firecloud/service/BaseServiceSpec.scala
```diff
@@ -12,9 +12,10 @@ class BaseServiceSpec extends ServiceSpec with BeforeAndAfter {
   val rawlsDao:MockRawlsDAO = new MockRawlsDAO
   val searchDao:MockSearchDAO = new MockSearchDAO
   val thurloeDao:MockThurloeDAO = new MockThurloeDAO
+  val ontologyDao:MockOntologyDAO = new MockOntologyDAO
 
   val app:Application =
-    new Application(agoraDao, rawlsDao, searchDao, thurloeDao, googleServicesDao)
+    new Application(agoraDao, rawlsDao, searchDao, thurloeDao, googleServicesDao, ontologyDao)
```
davidangb: Tests are failing because you changed the signature of `LibraryServiceSupport.indexableDocument`, but didn't change the calling code in the tests.  In `LibraryServiceSpec`, change the class declaration from:
```
import org.scalatest.FreeSpec
...
class LibraryServiceSpec extends FreeSpec
```
to
```
import org.scalatest.FreeSpecLike
...
class LibraryServiceSpec extends BaseServiceSpec with FreeSpecLike 
```
This will bring `ontologyDao` into scope, and you can change the six instances of
```
indexableDocument(w)
```
to
```
indexableDocument(w, ontologyDao)
```

That'll get you past the first error, but then you'll have problems with a missing `UserInfo` ... see my first comment. I'm surprised we don't already have a mock `UserInfo` in our test suite anywhere, but I can't find it (at least right now) ... you can hack this by adding:
```
  implicit val userInfo:UserInfo = UserInfo("fakeToken", "fakeSubjectId")
```
into `BaseServiceSpec`.

davidangb (commented):
Overall, I think this is the right approach, but I have a bunch of comments inline!

src/main/scala/org/broadinstitute/dsde/firecloud/model/Ontology.scala
```diff
@@ -0,0 +1,23 @@
...
+object Ontology {
+
+  case class TermResource(
+    id: String,
+    ontology: String,
+    usable: Boolean,
+    label: String,
+    definition: String,
+    synonyms: List[String],
+    parents: List[TermParent]
```
rushtong: synonyms and parents are both optional - there are lots of cases where they don't exist. Might make sense to use Options.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
rushtong: +1, pretty sure David's right.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -119,27 +124,30 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
...
       else {
-        val toIndex:Seq[Document] = workspaces.map {workspace => indexableDocument(workspace)}
-        val bi = searchDAO.bulkIndex(toIndex)
-        val statusCode = if (bi.hasFailures) {
-          InternalServerError
-        } else {
-          OK
+        val toIndex: Future[Seq[Document]] = Future.sequence(workspaces map { indexableDocument(_, ontologyDAO) })
+        toIndex map { documents =>
+          val indexed = searchDAO.bulkIndex(documents)
```
rushtong: nitpick - the variable name looks like a boolean but it's a full `LibraryBulkIndexResponse`. One could also iterate through the failures and dig out the individual errors, but that seems like overkill since the whole `indexed` is being returned as the error body.

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockOntologyDAO.scala
```diff
@@ -0,0 +1,48 @@
...
+        synonyms=null
+      ),
+      TermParent(
+        id="http://purl.obolibrary.org/obo/DOID_4",
+        order=4,
+        label="disease",
+        definition="A disease is a disposition (i) to undergo pathological processes that (ii) exists in an organism because of one or more disorders in that organism.",
+        synonyms=null
+      )
+    )))
```
rushtong: See https://github.com/broadinstitute/firecloud-orchestration/blob/develop/src/test/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSpec.scala#L125 for another approach to populating a `TermResource` object. Depending on the situation, that can be more concise than creating full objects.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -119,27 +123,28 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
...
-        RequestComplete(NoContent)
+        Future(RequestComplete(NoContent))
       else {
-        val toIndex:Seq[Document] = workspaces.map {workspace => indexableDocument(workspace)}
-        val bi = searchDAO.bulkIndex(toIndex)
-        val statusCode = if (bi.hasFailures) {
-          InternalServerError
-        } else {
-          OK
+        val toIndex: Future[Seq[Document]] = Future.sequence(workspaces map { indexableDocument(_, ontologyDAO) })
```
davidangb, line 184: in practice, do you run into any issues with this? We currently have ~80 published workspaces in dev ... which means we fire off 80 requests to consent-ontology as fast as possible. Any problems?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpOntologyDAO.scala
```diff
@@ -0,0 +1,21 @@
...
+import spray.json.DefaultJsonProtocol._
+
+import scala.concurrent.{ExecutionContext, Future}
+
+
+class HttpOntologyDAO(implicit val system: ActorSystem, implicit val executionContext: ExecutionContext)
+  extends OntologyDAO with RestJsonClient {
+
+  override def search(term: String)(implicit ec: ExecutionContext): Future[List[TermResource]] = {
+    unAuthedRequestToObject[List[TermResource]](Get(Uri(ontologySearchUrl).withQuery(("id", term))))
```
davidangb: this will throw a runtime error if it doesn't find the id, if I understand the code correctly.

This means that if even one of the published datasets has an unknown id, we'll fail to reindex-all, and we'll throw errors when publishing that dataset. I think we want error-trapping somewhere ... probably inside `indexableDocument`.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
```diff
@@ -39,7 +41,19 @@ trait LibraryServiceSupport {
       AttributeName.withDefaultNS("namespace") -> AttributeString(workspace.namespace),
       AttributeName.withDefaultNS("workspaceId") -> AttributeString(workspace.workspaceId)
     )
-    Document(workspace.workspaceId, attrfields ++ idfields)
+    val fields = attrfields ++ idfields
+
+    workspace.attributes.get(AttributeName.withLibraryNS("diseaseOntologyID")) match {
+      case Some(id: AttributeString) =>
+        ontologyDAO.search(id.value) map { (response: Seq[TermResource]) =>
```
davidangb: see one of my previous comments - probably want error-trapping here, so that if `id.value` is not found, we'll continue to index the document, just without any parent information.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/RestJsonClient.scala
```diff
@@ -21,6 +21,15 @@ trait RestJsonClient {
...
 
+  def unAuthedRequest(req: HttpRequest, compressed: Boolean = false): Future[HttpResponse] = {
+    val pipeline = if (compressed) {
+      addHeader(`Accept-Encoding`(gzip)) ~> sendReceive ~> decode(Gzip)
+    } else {
+      sendReceive
+    }
+    pipeline(req)
+  }
+
```
davidangb: look for reuse here, something like:
```
  private def constructPipeline(compressed: Boolean): SendReceive = {
    if (compressed) {
      addHeader(`Accept-Encoding`(gzip)) ~> sendReceive ~> decode(Gzip)
    } else {
      sendReceive
    }
  }

  def unAuthedRequest(req: HttpRequest, compressed: Boolean = false): Future[HttpResponse] = {
    val pipeline = constructPipeline(compressed)
    pipeline(req)
  }

  def userAuthedRequest(req: HttpRequest, compressed: Boolean = false)(implicit userInfo: WithAccessToken): Future[HttpResponse] = {
    val pipeline = addCredentials(userInfo.accessToken) ~> constructPipeline(compressed)
    pipeline(req)
  }
```

src/main/scala/org/broadinstitute/dsde/firecloud/utils/RestJsonClient.scala
```diff
@@ -42,4 +51,17 @@ trait RestJsonClient {
...
+      response.status match {
+        case s if s.isSuccess =>
+          response.entity.as[T] match {
+            case Right(obj) => obj
+            case Left(error) => throw new FireCloudExceptionWithErrorReport(FCErrorReport(response))
+          }
+        case f => throw new FireCloudExceptionWithErrorReport(FCErrorReport(response))
+      }
+    }
+  }
```
davidangb: also here, maybe something like:
```
  def requestToObject[T](req: HttpRequest, compressed: Boolean = false)(implicit userInfo: WithAccessToken, unmarshaller: Unmarshaller[T]): Future[T] = {
    doRequest( true, req, compressed )
  }

  def unAuthedRequestToObject[T](req: HttpRequest, compressed: Boolean = false)(implicit unmarshaller: Unmarshaller[T]): Future[T] = {
    doRequest( false, req, compressed )
  }

  private def doRequest[T](auth: Boolean, req: HttpRequest, compressed: Boolean = false)(implicit unmarshaller: Unmarshaller[T]): Future[T] = {
    val resp = if (auth)
      userAuthedRequest( req, compressed )
    else
      unAuthedRequest(req, compressed)
    
    resp map { response =>
      response.status match {
        case s if s.isSuccess =>
          response.entity.as[T] match {
            case Right(obj) => obj
            case Left(error) => throw new FireCloudExceptionWithErrorReport(FCErrorReport(response))
          }
        case f => throw new FireCloudExceptionWithErrorReport(FCErrorReport(response))
      }
    }
  }

```
( might want to refactor the two public methods to `authedRequestToObject` and `unAuthedRequestToObject` for clarity )

I haven't actually tested this, or the previous comment, but you get the idea!

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
```diff
@@ -39,7 +41,19 @@ trait LibraryServiceSupport {
...
+          val parents = response.get(0).parents
+          parents match {
+            case Some(list) => Document(workspace.workspaceId, fields + (AttributeName.withDefaultNS("parents") -> AttributeValueRawJson(list.toJson.compactPrint)))
+            case _ => Document(workspace.workspaceId, fields)
+          }
```
davidangb: I'm curious what `AttributeValueRawJson` generates.

src/test/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSpec.scala
```diff
@@ -340,7 +344,24 @@ class LibraryServiceSpec extends FreeSpec with LibraryServiceSupport with Attrib
...
+        ))
+        val expected = Document(testUUID.toString, Map(
+          AttributeName.withLibraryNS("diseaseOntologyID") -> AttributeString("DOID_9220"),
+          AttributeName.withDefaultNS("name") -> AttributeString(testWorkspace.name),
+          AttributeName.withDefaultNS("namespace") -> AttributeString(testWorkspace.namespace),
+          AttributeName.withDefaultNS("workspaceId") -> AttributeString(testWorkspace.workspaceId),
+          AttributeName.withDefaultNS("parents") -> AttributeValueRawJson(ontologyDao.resources.get(0).parents.toJson.compactPrint)
+        ))
+        assertResult(expected) {
+          Await.result(indexableDocument(w, ontologyDao), dur)
```
davidangb: Since we're holding each other to good testing practices, other unit tests I'd like to see:
* case where the diseaseOntologyID returns a node without any parent information
* case where the diseaseOntologyID is not found

davidangb (commented):
some comments inline about refactoring.

The big question is: given this code, what is the data structure that actually gets indexed in ElasticSearch? This will help us understand what kind of mappings we do/don't need.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
davidangb: Answering my own question: we haven't seen any problems with this.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpOntologyDAO.scala
```diff
@@ -0,0 +1,30 @@
...
+
+import scala.concurrent.{ExecutionContext, Future}
+
+
+class HttpOntologyDAO(implicit val system: ActorSystem, implicit val executionContext: ExecutionContext)
+  extends OntologyDAO with RestJsonClient {
+
+  override def search(term: String): Future[Option[List[TermResource]]] = {
+    unAuthedRequest(Get(Uri(ontologySearchUrl).withQuery(("id", term)))) map { response =>
+      if (response.status.isSuccess)
```
rushtong: Not really sure you need to handle the success case - you have either something you can unmarshal, or None. Untested, but I think you can get away with just:
```
        response.entity.as[List[TermResource]] match {
          case Right(obj) => Some(obj)
          case Left(error) => None
        }
```

src/main/scala/org/broadinstitute/dsde/firecloud/model/ElasticSearch.scala
```diff
@@ -68,6 +71,8 @@ object ESType extends ESPropertyFields {
 
 }
 
+case class ESNestedType(properties:Map[String,ESInnerField], `type`:String="nested") extends ESPropertyFields
```
rushtong, line 74: In the current case of parent label->order mapping, there is virtually no chance of having duplicate keys. This might not be true when re-used. Probably not worth worrying about for this PR.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
rushtong: And to add to that answer, the ontology lookups to ES have been pretty zippy through all of the testing I've done. If we run into issues here, I'd be worried.

rushtong (approved):
👍 LG, and worked for me with a sleep apnea example. Minor comments but nothing to hold this up.

src/main/scala/org/broadinstitute/dsde/firecloud/model/ElasticSearch.scala
davidangb: The `ESNestedType` is only used to create a mapping in ES - it's not used to hold values coming from the ontology. So I don't think this comment is applicable?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpOntologyDAO.scala
davidangb: good call, will change.

src/main/scala/org/broadinstitute/dsde/firecloud/model/ElasticSearch.scala
rushtong: Ah, yes, that makes sense. No worries then.

rushtong (approved):
👍 LG!

src/main/scala/org/broadinstitute/dsde/firecloud/model/ElasticSearch.scala
```diff
@@ -11,6 +11,9 @@ object ElasticSearch {
   final val fieldAll = "_all"
   final val fieldSuggest = "_suggest"
   final val fieldDiscoverableByGroups = "_discoverableByGroups"
+  final val fieldOntologyParents = "parents"
+  final val fieldOntologyParentsOrder = "order"
+  final val fieldOntologyParentsLabel = "label"
```
rushtong, line 16: Curious about why we're not using underscores like we are with `_all`, `_suggest`, etc.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
```diff
@@ -39,7 +42,30 @@ trait LibraryServiceSupport {
...
+          } else {
+            fields
+          }
+          Document(workspace.workspaceId, parentFields)
+        }
+      case _ => Future(Document(workspace.workspaceId, fields))
+    }
+  }
+
+  // wraps the ontologyDAO call, handles Nones/nulls, and returns a [Future[Seq].
```
rushtong, line 61: nit: `[Future[Seq]` should be `[Future[Seq]]`

PR #388, opened 2017-03-01
GAWB-1260 Fix isProtected on workspace list

breilly2:
I was a little quick to approve PR #379. There were some problems that affected the isProtected flag on workspace list. These changes bring back some case classes that had been previously deleted. They are used to augment the response from Rawls by interpreting the dbGap realm to mean that the workspace is "protected" (it's not a pure passthrough). I've also updated the tests to make sure that this works correctly.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/model/Workspace.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/model/Workspace.scala
@@ -47,6 +47,13 @@ case class UIWorkspaceResponse(
     this(Option(wlr.accessLevel.toString), None, Option(new UIWorkspace(wlr.workspace)), Option(wlr.workspaceSubmissionStats), Option(wlr.owners.toList))
 }
 
+/** A Firecloud UI focused result object that performs extra translation on the result from Rawls, specifically
+  * interpreting the NIH realm as a binary "protected" flag.
+  *
+  * Note: Depending on the direction that firecloud-orchestration takes in the future, we may keep this here or move
+  * this logic into firecloud-ui as part of https://broadinstitute.atlassian.net/browse/GAWB-1674. See discussion in
+  * https://github.com/broadinstitute/firecloud-orchestration/pull/388.
+  */
 case class UIWorkspace(
   workspaceId: String,
   namespace: String,
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/model/Workspace.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/model/Workspace.scala
@@ -2,6 +2,7 @@ package org.broadinstitute.dsde.firecloud.model
 
 import org.broadinstitute.dsde.firecloud.FireCloudConfig
 import org.broadinstitute.dsde.rawls.model.Attributable.AttributeMap
+import org.broadinstitute.dsde.rawls.model.WorkspaceAccessLevels.WorkspaceAccessLevel
 import org.broadinstitute.dsde.rawls.model._
 
 case class WorkspaceCreate(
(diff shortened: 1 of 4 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/model/Workspace.scala
```diff
@@ -34,6 +35,37 @@ object WorkspaceCreate {
 
 }
 
+case class UIWorkspaceResponse(
```
breilly2, line 38: Some of the fields here now use case classes from RawlsModel.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiService.scala
```diff
@@ -52,12 +72,14 @@ trait WorkspaceApiService extends HttpService with FireCloudRequestBuilding
...
-            passthrough(rawlsWorkspacesRoot, HttpMethods.GET)
+            mapHttpResponse(transformListWorkspaceRequest) {
+              passthrough(rawlsWorkspacesRoot, HttpMethods.GET)
+            }
           } ~
           post {
             requireUserInfo() { _ =>
               entity(as[WorkspaceCreate]) { createRequest => requestContext =>
-                val extReq = Post(FireCloudConfig.Rawls.workspacesUrl, createRequest)
+                val extReq = Post(FireCloudConfig.Rawls.workspacesUrl, WorkspaceCreate.toWorkspaceRequest(createRequest))
```
breilly2, line 82: This had to come back explicitly (also for clone) because there's no reason for the compiler to go looking for an implicit type conversion.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiService.scala
```diff
@@ -52,12 +72,14 @@ trait WorkspaceApiService extends HttpService with FireCloudRequestBuilding
       pathPrefix("workspaces") {
         pathEnd {
           requireUserInfo() { _ =>
-            passthrough(rawlsWorkspacesRoot, HttpMethods.GET)
+            mapHttpResponse(transformListWorkspaceRequest) {
```
breilly2, line 75: This (and the singular GET) are not pure passthrough methods. They interpret the realm from Rawls into the isProtected property.

src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
```diff
@@ -281,7 +281,7 @@ class WorkspaceApiServiceSpec extends BaseServiceSpec with WorkspaceApiService w
           //generally this is not how we want to treat the response
           //it should already be returned as JSON but for some strange reason it's being returned as text/plain
           //here we take the plain text and force it to be json so we can get the test to work
-          assert(WorkspaceCreate.isProtected(entity.asString.parseJson.convertTo[WorkspaceResponse].workspace))
+          assert(entity.asString.parseJson.convertTo[UIWorkspaceResponse].workspace.get.isProtected)
```
breilly2, line 323: This is testing GET, not post, so it needs to convert to UIWorkspaceResponse and not use WorkspaceCreate.

src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
```diff
@@ -412,10 +422,31 @@ class WorkspaceApiServiceSpec extends BaseServiceSpec with WorkspaceApiService w
...
+      val response = Workspace("namespace", "name", realm, "foo", "bar", DateTime.now(), DateTime.now(), "bob", Map(), Map(), Map())
+      stubRawlsService(HttpMethods.POST, clonePath, Created, Some(response.toJson.compactPrint))
+
+      val isProtected: Option[Boolean] = Option(true)
+      val orchestrationRequest: WorkspaceCreate = WorkspaceCreate("namespace", "name", Map(), isProtected)
+      val rawlsRequest = WorkspaceRequest("namespace", "name", realm, Map(AttributeName("library", "published") -> AttributeBoolean(false)))
+      Post(clonePath, orchestrationRequest) ~> dummyUserIdHeaders("1234") ~> sealRoute(workspaceRoutes) ~> check {
+        rawlsServer.verify(request().withPath(clonePath).withMethod("POST").withBody(rawlsRequest.toJson.prettyPrint))
+        status should equal(Created)
+        responseAs[Workspace] should equal(response)
```
rushtong: From slack discussion: `rawlsRequest` should start off with `AttributeName("library", "published") -> AttributeBoolean(true)`, and then that will be transformed inside the clone endpoint. At that point, the rawls stub should see `AttributeName("library", "published") -> AttributeBoolean(false)` instead.

Additionally, one could start off with an empty attribute map. The transform will populate the false value so that when Rawls sees the create request, it will use that value instead of whatever exists in the source workspace.

helgridly:
So from what I can tell, what you've done here is restored the UIWorkspace classes in order to translate "has NIH realm" to "is protected". I think my intention was to have the UI handle that instead, but I forgot to follow up on that.

My gut tells me that since Realms is going to be a more general concept, Orchestration should pass it through and let the UI deal with it. If the FireCloud (cancer-specific) UI wants to interpret that as "protected access if the realm is NIH" then it can do so. I think the answer depends on whether you see Orchestration as an NIH/cancer-y facade, or a passthrough/firewall service.

breilly2:
That is correct. I guess it does depend on your point of view. In the absence of UI changes, I went with changing orchestration. We have some stories in our backlog to support multiple realms in the UI at which point the NIH realm and "protected" concept stops being a special case (or at least less of one).

I can add some comments in the code and relevant jira ticket to make sure we address this at some point. For now, I like having this here because of the test coverage it allows for with current techniques.

BTW, I think this is the first time I've heard anyone talk about Firecloud as being cancer-specific.

helgridly:
I'm okay with this reasoning. Let's tag the other TLs and see if there are any strong objections, though I doubt there will be. @davidangb @dvoet

MatthewBemis (approved):
I'm also cool with this in a restorative sense. Ideally I don't think that orch should care about specific realms at all- it would be the UI that decides to treat dbGap protected workspaces differently. But we can deal with that discussion when we start to build out the UI for realms. This gets us back to where we were before so :+1:

breilly2:
I added some code comments, including notes about possibly moving the special case to firecloud-ui. Also added another test specifically for Greg's case of cloning a published workspace.

breilly2:
One more commit. I had to rebase to get @rushtong's latest changes because it affects a test that I added.

PR #391, opened 2017-03-03
GAWB-1419: return if you have access to the workspace for a dataset with library results

skwalker:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
@@ -144,32 +144,12 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
     }
   }
 
-  def updateAccess(docs: LibrarySearchResponse, workspaces: Seq[WorkspaceListResponse]) = {
-    val ids = workspaces collect {
-      case workspaceResponse: WorkspaceListResponse if (workspaceResponse.accessLevel != WorkspaceAccessLevels.NoAccess) =>
-        workspaceResponse.workspace.workspaceId
-    }
-
-    val updatedResults = docs.results.map { document =>
-      val docId = document.asJsObject.fields.get("workspaceId")
-      val newJson = docId match {
-        case Some(id) =>  val workspaceId = id.toString.replace("\"", ""); document.asJsObject.fields + ("workspaceAccess" -> JsBoolean(ids.contains(workspaceId)))
-        case None => document.asJsObject.fields + ("workspaceAccess" -> JsBoolean(true)) // is this the right thing to do?
-      }
-      JsObject(newJson)
-    }
-
-    docs.copy(results = updatedResults)
-  }
-
   def findDocuments(criteria: LibrarySearchParams): Future[PerRequestMessage] = {
     getEffectiveDiscoverGroups(rawlsDAO) map { userGroups =>
-      // calling them first makes the futures execute in parallel?
-      // http://buransky.com/scala/scala-for-comprehension-with-concurrently-running-futures/
       val docsFuture = searchDAO.findDocuments(criteria, userGroups)
       val idsFuture = rawlsDAO.getWorkspaces
       val searchResults = for {
-        docs <- docsFuture //
+        docs <- docsFuture
         ids <- idsFuture
       } yield updateAccess(docs, ids)
       (RequestComplete(searchResults))
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
@@ -144,27 +144,28 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
     }
   }
 
-  def findDocuments(criteria: LibrarySearchParams): Future[PerRequestMessage] = {
-    // get workspace ids user has access to, then add that to this
-    getEffectiveDiscoverGroups(rawlsDAO) map {userGroups =>
-      // flatmap get workspace ids or use for ting with a yield
-      searchDAO.findDocuments(criteria, userGroups)} map { documents =>
-      // update documents
-      val docsWithAccess = documents.map { docs =>
-        rawlsDAO.getWorkspacesIds map { accessibleWorkspaceIds =>
-          val accessChecked = docs.results.map { document =>
-            val docId = document.asJsObject.fields.get("workspaceId")
-            val newJson = docId match {
-              case Some(id) =>  val workspaceId = id.toString.replace("\"", ""); document.asJsObject.fields + ("access" -> JsBoolean(accessibleWorkspaceIds.contains(workspaceId)))
-              case None => document.asJsObject.fields + ("access" -> JsBoolean(true)) // ??
-            }
-            JsObject(newJson)
-          }
-          log.info("all workspaces: " + accessChecked)
-          accessChecked
-        }
+  def updateAccess(docs: LibrarySearchResponse, ids: Seq[String]) = {
+    docs.results.map { document =>
+      val docId = document.asJsObject.fields.get("workspaceId")
+      val newJson = docId match {
+        case Some(id) =>  val workspaceId = id.toString.replace("\"", ""); document.asJsObject.fields + ("access" -> JsBoolean(ids.contains(workspaceId))) // two lines?
+        case None => document.asJsObject.fields + ("access" -> JsBoolean(true)) // ??
       }
-      (RequestComplete(docsWithAccess))
+      JsObject(newJson)
+    }
+  }
+
+  def findDocuments(criteria: LibrarySearchParams): Future[PerRequestMessage] = {
+    getEffectiveDiscoverGroups(rawlsDAO) map { userGroups =>
+      // calling them first makes the futures execute in parallel?
+      // http://buransky.com/scala/scala-for-comprehension-with-concurrently-running-futures/
+      val docsFuture = searchDAO.findDocuments(criteria, userGroups)
+      val idsFuture = rawlsDAO.getWorkspacesIds
+      val searchResults = for {
+        docs <- docsFuture //
+        ids <- idsFuture
+      } yield updateAccess(docs, ids)
+      (RequestComplete(searchResults))
     }
   }
 
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
@@ -144,12 +144,17 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
     }
   }
 
-  def updateAccess(docs: LibrarySearchResponse, ids: Seq[String]) = {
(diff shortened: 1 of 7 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -144,9 +144,36 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
     }
   }
 
+  def updateAccess(docs: LibrarySearchResponse, workspaces: Seq[WorkspaceListResponse]) = {
```
skwalker: should I write a test for this or something?

skwalker:
the one other thing I haven't checked is if project owners now can click datasets because I think this ticket came out as a bug where you don't have bucket access but do have workspace access as a project owner (not sure how to test but I _think_ it should work now)

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -144,9 +144,36 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
...
+    val ids = workspaces collect {
+      case workspaceResponse: WorkspaceListResponse if (workspaceResponse.accessLevel != WorkspaceAccessLevels.NoAccess) =>
+        workspaceResponse.workspace.workspaceId
+    }
+
+    val updatedResults = docs.results.map { document =>
+      val docId = document.asJsObject.fields.get("workspaceId")
+      val newJson = docId match {
+        case Some(id) =>  val workspaceId = id.toString.replace("\"", ""); document.asJsObject.fields + ("workspaceAccess" -> JsBoolean(ids.contains(workspaceId)))
```
davidangb: 1) make sure that `workspaceId` is a `JsString`
2) use `JsString.value` instead of `JsValue.toString`; the `toString` quite annoyingly is the thing that adds the quotes around the value.
```
        case Some(id:JsString) =>  val workspaceId = id.value; document.asJsObject.fields + ("workspaceAccess" -> JsBoolean(ids.contains(workspaceId)))
```

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -144,9 +144,36 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
...
+        case None => document.asJsObject.fields + ("workspaceAccess" -> JsBoolean(true)) // is this the right thing to do?
```
davidangb: in the UI, I think the right thing to do is to send the user to the workspace and let Rawls decide what to do.

In the API, I think the right thing to do is to have three states: `true` if the user has access, `false` if the user does not, and empty/null/missing if we can't determine access. So how about:
```
case _ => document
```
which would leave the `workspaceAccess` field out of the response entirely? This would of course require corresponding logic in the UI to handle the missing value.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -144,9 +144,36 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
     }
   }
 
+  def updateAccess(docs: LibrarySearchResponse, workspaces: Seq[WorkspaceListResponse]) = {
+    val ids = workspaces collect {
+      case workspaceResponse: WorkspaceListResponse if (workspaceResponse.accessLevel != WorkspaceAccessLevels.NoAccess) =>
+        workspaceResponse.workspace.workspaceId
+    }
```
davidangb: open for discussion: is it better to have `workspaceAccess: true|false` as a Boolean in the response, or to copy the value of `workspaceResponse.accessLevel` into the response?

In the latter, we're returning much more granular info in case we want to use it in the UI - and it's more consistent with what the workspace listing returns.

Also, consider moving the `updateAccess` function into `LibraryServiceSupport` - rule of thumb here is that `LibraryService` contains the top-level methods (typically those called from the API routes), and `LibraryServiceSupport` contains utilities to support those top-level methods.

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockRawlsDAO.scala
```diff
@@ -99,6 +99,15 @@ class MockRawlsDAO  extends RawlsDAO {
...
 
+  override def getWorkspaces(implicit userInfo: WithAccessToken): Future[Seq[WorkspaceListResponse]] = {
+    Future(Seq(WorkspaceListResponse(WorkspaceAccessLevels.ProjectOwner, newWorkspace, WorkspaceSubmissionStats(None, None, runningSubmissionsCount = 0), List.empty),
+      WorkspaceListResponse(WorkspaceAccessLevels.Read, newWorkspace, WorkspaceSubmissionStats(None, None, runningSubmissionsCount = 0), List.empty),
+      WorkspaceListResponse(WorkspaceAccessLevels.Owner, rawlsWorkspaceWithAttributes, WorkspaceSubmissionStats(None, None, runningSubmissionsCount = 0), List.empty),
+      WorkspaceListResponse(WorkspaceAccessLevels.Owner, publishedRawlsWorkspaceWithAttributes, WorkspaceSubmissionStats(None, None, runningSubmissionsCount = 0), List.empty),
+      WorkspaceListResponse(WorkspaceAccessLevels.Owner, newWorkspace, WorkspaceSubmissionStats(None, None, runningSubmissionsCount = 0), List.empty)))
+  }
+
+
```
davidangb, line 110: could you write unit tests that call either `LibraryService.findDocuments` or `LibraryServiceSupport. updatedResults` and verify correctness?

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -144,9 +144,36 @@ class LibraryService (protected val argUserInfo: UserInfo, val rawlsDAO: RawlsDA
...
+
+    docs.copy(results = updatedResults)
+  }
+
   def findDocuments(criteria: LibrarySearchParams): Future[PerRequestMessage] = {
-    getEffectiveDiscoverGroups(rawlsDAO) map {userGroups =>
-      searchDAO.findDocuments(criteria, userGroups)} map (RequestComplete(_))
+    getEffectiveDiscoverGroups(rawlsDAO) map { userGroups =>
+      // calling them first makes the futures execute in parallel?
+      // http://buransky.com/scala/scala-for-comprehension-with-concurrently-running-futures/
```
davidangb: nit: update comments. Is there still a question? :)

davidangb (changes requested):
first pass review; comments inline!

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
skwalker: ahh that's beautiful, thank you

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
skwalker: just wanted to make sure it was correct!

skwalker:
re: returning true|false vs the actual accessLevel -- why would we need that information in the UI? I think we should just return true/false because otherwise, the UI would just have to do the check for accessLevel != "No Access" itself. Maybe we could just change it to return the accessLevel if we do need to know the accessLevel in the UI in the future.

davidangb (approved):
👍 as-is.

For the sake of discussion, on the topic of returning the access level (owner/reader/writer/etc) vs. a boolean: either way will achieve the requirements of the story. But looking at this from a high level, I'm shooting for our APIs to be flexible and contain a lot of detail, in case users script against them, and our UI offers a lot of guidance and rails. From that perspective, I'd want the API to return the access level, and have the UI interpret it.

But again, either way is fine, we can change it in the future if we find a use case.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
```diff
@@ -63,15 +63,13 @@ trait LibraryServiceSupport {
...
-      case workspaceResponse: WorkspaceListResponse if (workspaceResponse.accessLevel != WorkspaceAccessLevels.NoAccess) =>
-        workspaceResponse.workspace.workspaceId
-    }
-
     val updatedResults = docs.results.map { document =>
       val docId = document.asJsObject.fields.get("workspaceId")
       val newJson = docId match {
-        case Some(id: JsString) => document.asJsObject.fields + ("workspaceAccess" -> JsBoolean(ids.contains(id.value)))
+        case Some(id: JsString) => workspaces.find((workspaceResponse: WorkspaceListResponse) => workspaceResponse.workspace.workspaceId == id.value) match {
+          case Some(workspaceResponse) => document.asJsObject.fields + ("workspaceAccess" -> JsString(workspaceResponse.accessLevel.toString))
```
skwalker: should this not be a JsString?

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
```diff
@@ -62,4 +62,20 @@ trait LibraryServiceSupport {
...
+          case _ => document.asJsObject.fields + ("workspaceAccess" -> JsString(WorkspaceAccessLevels.NoAccess.toString))  // or just document.asJsObject.fields ??
+        }
+        case _ => document.asJsObject.fields
+      }
+      JsObject(newJson)
+    }
+
+    docs.copy(results = updatedResults)
+  }
```
davidangb, line 84: two comments:
1) at line 71, if you can't find the workspace id in the rawls response, I think you did the right thing and set it to NoAccess.
2) at line 69, `workspaces.find` (i.e. `Seq.find`) is, I believe, an O(n) operation. In the most common use case where you have 20 results from Library, you're executing O(n) 20 times. In this situation, I would create an intermediate `Map` and do your lookups into it, something like:

```
    val workspaceAccess = (workspaces map {w:WorkspaceListResponse => w.workspace.workspaceId -> w.accessLevel}).toMap

    val updatedResults = docs.results.map { document =>
      val docId = document.asJsObject.fields.get("workspaceId")
      val newJson = docId match {
        case Some(id: JsString) => workspaceAccess.get(id.value) match {
          case Some(level) => document.asJsObject.fields + ("workspaceAccess" -> JsString(level.toString))
          case _ => document.asJsObject.fields + ("workspaceAccess" -> JsString(WorkspaceAccessLevels.NoAccess.toString))
        }
        case _ => document.asJsObject.fields
      }
      JsObject(newJson)
    }
```

davidangb (commented):
one optimization inline, otherwise looking good!

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
skwalker: yep, using a map makes way more sense, just made that change

davidangb (approved):
👍

rushtong (approved):
👍 Looks good, works good!

PR #396, opened 2017-03-10
GAWB-1665: ontology parent node search

davidangb:
runtime code changes are relatively small; there's a lot of lines of code for tests.

Big problem I was running into with ontology-parent search was quality of search results.

Changes in this PR are:
1) When pulling ontology parents from consent-ontology, don't fail if the definition is missing/null. Technically not part of this jira ticket, but found while working on it. 
2) Enable (English) stop words in the default analyzer. This improved quality because searches for e.g. "disease of mental health" no longer match on the word "of".
3) Don't include ontology parents in the global "_all" meta-field. Instead, query ES for (_all OR parents.label). For the parents.label part of that clause, be a bit stricter - 3<75% - for the number of terms to require. This OR search improves quality because it won't span multiple search terms across parents and another field. Previously, a search for "infectious disease" could match "infectious" in any attribute, coupled with the "disease" root node which is a parent on every dataset. Now, we have to match "infectious disease" inside any parent OR any other attribute, but not across.
4) Make parents available in search suggestions. I was originally against this, but after trying it out both ways, I like having the suggestions available.

Individual commits in this PR are pretty clean, if you want to review commit-by-commit.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/OntologySearchSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/OntologySearchSpec.scala
@@ -1,0 +1,106 @@
+package org.broadinstitute.dsde.firecloud.integrationtest
+
+import com.typesafe.scalalogging.slf4j.LazyLogging
+import org.broadinstitute.dsde.firecloud.integrationtest.ESIntegrationSupport._
+import org.scalatest.{BeforeAndAfterAll, FreeSpec, Matchers}
+
+class OntologySearchSpec extends FreeSpec with Matchers with BeforeAndAfterAll with LazyLogging with SearchResultValidation {
+
+  override def beforeAll = {
+    // use re-create here, since instantiating the DAO will create it in the first place
+    searchDAO.recreateIndex()
+    // make sure we specify refresh=true here; otherwise, the documents may not be available in the index by the
+    // time the tests start, leading to test failures.
+    logger.info("indexing fixtures ...")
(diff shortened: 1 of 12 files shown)


src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/OntologySearchSpec.scala
```diff
@@ -0,0 +1,106 @@
...
+  /*
+    OntologySearchTextFixtures has five datasets, with these ontology nodes:
+
+    CSA_9220  : central sleep apnea < sleep apnea < sleep disorder < disease of mental health < disease
+    E_4325    : Ebola hemorrhagic fever < viral infectious disease < disease by infectious agent < disease
+    L_1240    : leukemia < hematologic cancer < immune system cancer < organ system cancer < cancer < disease of cellular proliferation < disease
+    HC_2531   : hematologic cancer < immune system cancer < organ system cancer < cancer < disease of cellular proliferation < disease
+    D_4       : Disease
+    None      : no doid
+   */
```
rushtong: I would suggest one more case to test which is for a node that has multiple parents at the same distance (order) away. "http://purl.obolibrary.org/obo/DOID_0050696" is a good one for that.

rushtong (approved):
👍 Love it! Searching works pretty much as I expected via the API no matter what I threw at it.

rushtong (approved):
👍 LGWG

PR #397, opened 2017-03-13
GAWB-1765, GAWB-1735: passthrough tests

davidangb:
GAWB-1765: orch passthrough directive mangles query parameters
GAWB-1735: Add optional namespace parameter to method repo API call
plus a bunch of tests

This PR started because I wanted to write unit tests for the passthrough directives. By writing those tests, I found some bugs in the directives.

I fixed a problem where query parameters would not pass through correctly: `path?param` would get encoded and pass through as `path%3Fparam` .. which broke query parameters. This issue blocks GAWB-1735 and is mentioned in GAWB-1303.

I also fixed a problem where the `passthroughAllPaths` directive did not support a path like `foo/bar` as the source. This isn't hugely important because it doesn't block anything - and I'd like to eliminate `passthroughAllPaths` altogether (See GAWB-1303).

see individual commits to isolate the bugfixes.

There's a bunch more unit tests we could write around the passthrough directive - testing the compression flag, POST/PUT payloads, etc - but I think this is a good start.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -750,12 +750,11 @@ paths:
         - $ref: '#/parameters/methodNamespaceParam'
         - $ref: '#/parameters/methodNameParam'
         - $ref: '#/parameters/methodSnapshotId'
-        # GAWB-1303
-        # - name: onlyPayload
-        #   in: query
-        #   description: Boolean to return only the payload of the method.
-        #   required: false
-        #   type: string
+        - name: onlyPayload
+          in: query
+          description: Boolean to return only the payload of the method.
+          required: false
+          type: string
       responses:
         200:
           description: A single method.
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -208,54 +208,55 @@ paths:
       parameters:
         - name: namespace
           in: query
-          description: Namespace of method.
+          description: Namespace of configuration.
           required: false
           type: string
         - name: name
           in: query
-          description: Name of method.
+          description: Name of configuration.
           required: false
           type: string
         - name: snapshotId
           in: query
-          description: Snapshot ID of method.
+          description: Snapshot ID of configuration.
           required: false
           type: number
         - name: synopsis
           in: query
-          description: Synopsis of method.
+          description: Synopsis of configuration.
           required: false
           type: string
         - name: documentation
           in: query
-          description: Documentation of method.
+          description: Documentation of configuration.
           required: false
           type: string
         - name: owner
           in: query
-          description: Owner of method.
+          description: Owner of configuration.
           required: false
           type: string
         - name: payload
           in: query
-          description: Payload of method in WDL.
+          description: Payload of configuration in WDL.
           required: false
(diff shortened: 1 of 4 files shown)


rushtong:
Ignore this comment. 

~Suggestion: can we have a test for html encoded segments similar to the hex-encoded test? Something like this to ensure we're not double encoding or un-encoding things:~
```
      "path with a uri-encoded segment" - {
        "should hit the path exactly" in {
          validateUri(
            "/one/%20space/three",
            "/one/%20space/three"
          )
        }
      }
```

rushtong:
👍

src/main/resources/swagger/api-docs.yaml
```diff
@@ -202,61 +202,60 @@ paths:
...
+          in: query
+          description: Payload of method in WDL.
+          required: false
+          type: string
+        - name: entityType
+          in: query
+          description: Type of Method - Task or Workflow
+          required: false
+          type: string
+        - name: excludedFields
```
rushtong: This doesn't seem to work correctly - not sure if the issue is what orchestration is sending to agora or if agora just isn't honoring this query param.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -202,61 +202,60 @@ paths:
...
-#          description: Fields to include in JSON response
-#          required: false
-#          type: string
+      description: |
+        The configurations endpoint returns all configurations the requester
+        has permissions to read that fit the filter criteria.
+      parameters:
+        - name: namespace
+          in: query
+          description: Namespace of method.
```
rushtong: Here and throughout these parameters, this is for a configuration, not a method.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -648,58 +647,57 @@ paths:
...
+          in: query
+          description: Payload of method in WDL.
+          required: false
+          type: string
+        - name: entityType
+          in: query
+          description: Type of Method - Task or Workflow
+          required: false
+          type: string
+        - name: excludedFields
```
rushtong: Similar issue to configurations - the response doesn't honor this query param.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -648,58 +647,57 @@ paths:
...
+          in: query
+          description: Fields to exclude in JSON response.
+          required: false
+          type: string
+        - name: includedFields
```
rushtong: I also can't get this to work. Tried using `payload` which is returned when you get the individual method or config, but no dice. The query looks correctly formed so this is either the param orchestration is sending or a bug in agora that doesn't honor it.

rushtong:
Confirmed in agora swagger that `excludedFields` and `includedFields` are not honored - orchestration is doing the right thing.

davidangb:
yup, I also found that included/excluded fields don't work in agora. I've re-hidden those params in orch swagger.

rushtong (approved):
👍 LG!

PR #399, opened 2017-03-13
Passthrough for searching on ORSP IDs

rushtong:
Addresses https://broadinstitute.atlassian.net/browse/GAWB-1759

* Passthrough for retrieving a consent from an ORSP ID.
* New models for valid consent and error cases.

Commit history is a mess, I don't recommend looking at it.

Since this touches authentication and authorization, we have a separate story to deal with FISMA changes. See https://broadinstitute.atlassian.net/browse/GAWB-1718 for addressing that aspect of this work.

Requires DUOS change so FC users (who have accounts in DUOS) can make OAuth-ed API calls:
* ~https://github.com/broadinstitute/consent/pull/233~ completed
* https://github.com/broadinstitute/firecloud-develop/pull/351/files

Requires config changes.
* Local: https://github.com/broadinstitute/firecloud-develop/pull/338
* Dev: https://github.com/broadinstitute/firecloud-develop/pull/339

## Testing Notes:
Once https://github.com/broadinstitute/firecloud-develop/pull/351 lands, you can test this branch against dev. Otherwise, spin up a local consent instance and point to `local.broadinstitute.org` in the orchestration config for consent.
* Update configs using https://github.com/broadinstitute/firecloud-develop/pull/338
* Spin up local orchestration
* Sign into swagger as a user that exists in both FC and DUOS.
* Query on some ORSP IDs. Good ones include:
  * OD-310 (should be not found)
  * OD-320 (should be valid)
  * OD-320 (should be not approved) 
  * Any random string
* Also log in as a user in FC, but not in DUOS. Any query should get a 401.

---

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


src/test/scala/org/broadinstitute/dsde/firecloud/service/LibraryApiServiceSpec.scala
```diff
@@ -237,5 +237,34 @@ class LibraryApiServiceSpec extends BaseServiceSpec with LibraryApiService with
...
+          }
+        }
+        "should return a Not Found error on unknown missing" in {
+          val orspId = randomStringFromCharList(10, ('a' to 'z') ++ ('A' to 'Z') ++ ('0' to '9'))
+          Get(duosConsentOrspIdPath + orspId) ~> dummyUserIdHeaders("1234") ~> sealRoute(libraryRoutes) ~> check {
+            status should equal(NotFound)
+          }
+        }
+      }
+    }
```
rushtong, line 268: These tests are filling the logs with exceptions - I need to figure out a way to clean that up.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/LibraryApiService.scala
```diff
@@ -47,6 +47,13 @@ trait LibraryApiService extends HttpService with FireCloudRequestBuilding
     } ~
     pathPrefix("api") {
       requireUserInfo() { userInfo =>
+        path("duos" / "consent" / "orsp" / Segment) { (orspId) =>
+          get { requestContext =>
+            perRequest(requestContext,
+              LibraryService.props(libraryServiceConstructor, userInfo),
+              LibraryService.SearchOrspId(orspId))
```
rushtong, line 54: @davidangb - I wonder if by making this an actor call that goes through a DAO, I'm making this more complex than it needs to be. I am thinking of making this a passthrough which would likely simplify the code a lot and also allow me to get rid of the exceptions in the tests by using MockServer.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpOntologyDAO.scala
```diff
@@ -50,4 +56,30 @@ class HttpOntologyDAO(implicit val system: ActorSystem, implicit val executionCo
...
+              throw new FireCloudExceptionWithErrorReport(ErrorReport(x, consentError.message))
+            case Left(err) =>
+              throw new FireCloudExceptionWithErrorReport(ErrorReport(x, response.toString))
+          }
+        case _ =>
+          logger.error(response.toString)
+          throw new FireCloudExceptionWithErrorReport(FCErrorReport(response))
+      }
+    }
+  }
```
davidangb, line 83: either:
* the orspIdSearch functionality should move to a new ConsentDAO or OrspDAO, or:
* this class (and its parents/siblings) should rename from "Ontology" to something like "DUOS"

src/main/scala/org/broadinstitute/dsde/firecloud/model/Ontology.scala
```diff
@@ -1,5 +1,28 @@
...
+    dulName: Option[String] = None,
+    translatedUseRestriction: Option[String] = None
+  )
+
+  case class ConsentError(
+    message: String,
+    code: Int
+  )
+
+}
```
davidangb, line 25: also here, the file Ontology.scala should probably be renamed, or move the DUOS object to its own file.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -224,4 +226,17 @@ class LibraryService (protected val argUserInfo: UserInfo,
...
+
+  def searchOrspId(orspId: String): Future[PerRequestMessage] = {
+    // We need this implicit to unmarshal the consent in the response
+    import ModelJsonProtocol.impDuosConsent
+    ontologyDAO.orspIdSearch(userInfo, orspId) map { RequestComplete(_) } recoverWith {
+      case e: FireCloudExceptionWithErrorReport =>
+        val status = e.errorReport.statusCode.getOrElse(NotFound)
+        Future(RequestCompleteWithErrorReport(status, e.errorReport.message))
+      case e: FireCloudException =>
+        Future(RequestCompleteWithErrorReport(InternalServerError, s"error searching for ORSP ID '%s'".format(orspId)))
```
davidangb, line 238: under what conditions would `ontologyDAO.orspIdSearch` throw `FireCloudException`? If this is just a catchall case, I'd prefer `case e: Exception`.

src/test/scala/org/broadinstitute/dsde/firecloud/service/LibraryApiServiceSpec.scala
```diff
@@ -237,5 +237,34 @@ class LibraryApiServiceSpec extends BaseServiceSpec with LibraryApiService with
         }
       }
     }
+
+    "when searching for ORSP IDs" - {
+      "GET on " + duosConsentOrspIdPath - {
+        "should return a valid consent for '12345'" in {
+          Get(duosConsentOrspIdPath + "12345") ~> dummyUserIdHeaders("1234") ~> sealRoute(libraryRoutes) ~> check {
+            status should equal(OK)
+            val consent = response.entity.asString.parseJson.convertTo[Consent]
```
davidangb, line 246: interesting: `response.entity.asString.parseJson.convertTo[Consent]` works, but the simpler `responseAs[Consent]` fails to find the right implicit. (just babbling, not requesting any change)

src/test/scala/org/broadinstitute/dsde/firecloud/service/LibraryApiServiceSpec.scala
davidangb: I think http://stackoverflow.com/a/19441215 has your answer. `MockOntologyDAO.orspIdSearch` throws exceptions outside of a Future. If you modify the mock to be structured like the runtime dao, the exceptions are handled by the caller's `recover*`, something like this:

```
  override def orspIdSearch(userInfo: UserInfo, orspId: String): Future[Option[Consent]] = {
    Future.successful(true) map { x =>
      orspId match {
        case x if x.equals("unapproved") =>
          throw new FireCloudExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, "Unapproved"))
        case x if x.equals("missing") =>
          throw new FireCloudExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, "Not Found"))
        case x if x.equals("12345") =>
          Some(Consent(consentId = "consent-id-12345", name = "12345", translatedUseRestriction = Some("Translation")))
        case _ => None
      }
    }
  }
```
(there's probably a cleaner way to write this, instead of `Future.successful(true) map ...`

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -224,4 +226,17 @@ class LibraryService (protected val argUserInfo: UserInfo,
       case e: FireCloudException => Future(RequestCompleteWithErrorReport(BadRequest, s"suggestions not available for field %s".format(field)))
     }
   }
+
+  def searchOrspId(orspId: String): Future[PerRequestMessage] = {
+    // We need this implicit to unmarshal the consent in the response
+    import ModelJsonProtocol.impDuosConsent
+    ontologyDAO.orspIdSearch(userInfo, orspId) map { RequestComplete(_) } recoverWith {
```
davidangb, line 233: instead of using `recoverWith` and having your exception cases return `Future(RequestComplete...)`, you could use `recover` and have your exception cases return `RequestComplete...`

(yes I am also guilty of using recoverWith when I should use recover.

davidangb (commented):
comments inline!

skwalker (commented):
Authorization doesn't seem to be working properly; if you're a user in FC, but not in DUOS, you can still search consent and get results (not getting 401s). Issue seems to be more in consent than in this code, since even in consent I'm getting results when I should get permissions error.

rushtong:
@skwalker - yes, authorization isn't working the way I thought it was working - my mistake. Endpoints marked with `@PermitAll` never hit the `authorize` methods of the various authentication filters, so all OAuthed users will succeed. I need to make a bunch of changes so will be updating and revising the PR (and stories) to account for this reality.

rushtong:
closing in favor of https://github.com/broadinstitute/firecloud-orchestration/pull/407

PR #401, opened 2017-03-15
GAWB-1740: Passthroughs for Groups APIs

MatthewBemis:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -109,7 +109,7 @@ paths:
             - email
             - profile
 
-  /api/billing/{projectId}/{role}/{userEmail}:
+  /api/billing/{projectId}/{role}/{email}:
     put:
       responses:
         200:
@@ -139,8 +139,8 @@ paths:
           type: string
           enum: ["user", "owner"]
         - in: path
-          description: User Email
-          name: userEmail
+          description: email of user or group to add
+          name: email
           required: true
           type: string
       tags:
@@ -181,8 +181,8 @@ paths:
           type: string
           enum: ["user", "owner"]
         - in: path
-          description: User Email
-          name: userEmail
+          description: email of user or group to add
+          name: email
           required: true
           type: string
       tags:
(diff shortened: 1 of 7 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/FireCloudDirectives.scala
```diff
@@ -49,10 +49,8 @@ trait FireCloudDirectives extends spray.routing.Directives with PerRequestCreato
     val path = encodeUri(unencodedPath)
     // POST, PUT, PATCH
     if (inMethod.isEntityAccepted) {
-      method(inMethod) {
-        respondWithJSON { requestContext =>
```
MatthewBemis: I made this change because in some cases rawls and other services return plaintext. Since we force this header onto the response of all passthroughs, when we actually want to pass plaintext through, Swagger-UI will show "no response from server" because the content type doesn't match what's being served.

I'm not sure if this has other implications but I at least wanted to bring it up for discussion

src/main/scala/org/broadinstitute/dsde/firecloud/service/UserService.scala
```diff
@@ -126,33 +129,56 @@ trait UserService extends HttpService with PerRequestCreator with FireCloudReque
...
+                } ~
+                delete {
+                  passthrough(UserService.rawlsGroupMemberUrl(groupName, role, email), HttpMethods.DELETE)
+                }
+              }
+          }
       } ~
-      pathPrefix("profile") {
-        // GET /profile - get all keys for current user
+      pathPrefix("register") {
```
MatthewBemis: note to self: i broke this path

src/main/resources/swagger/api-docs.yaml
```diff
@@ -412,6 +412,206 @@ paths:
...
+            - openid
+            - email
+            - profile
+
+  /api/groups/{groupName}:
+    get:
+      responses:
+        200:
+          description: Successfully loaded group
+        403:
```
MatthewBemis: actually I need to go through and make sure that these status codes are all correct

dvoet (approved):

PR #402, opened 2017-03-15
library reindex-all safety

davidangb:
Creating this as a PR against develop. Our current alpha snapshot is at the head of develop (d326817dc214afd06b16daf3d3289cf500b2eaee) as I write this - so this PR should be representative of what we'd hotfix.

Changes in this PR:
1. Don't delete/reset the ES index unless we successfully generated the documents to index.
2. Add error-trapping (via `recoverWith` to `LibraryServiceSupport.lookupParentNodes`
3. Optimize our queries to ontology - only query for unique nodes. Previously we would re-query for a node we already looked up. Now we build a cache.
4. Add a bunch of logging statements to help debug.
5. This is the real fix: only make one query to ontology at a time, and block waiting for it to return.

The blocking I/O is bad!!!! So I've marked it as deprecated. But I think it'll get us working again while we look at a longer-term fix.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
@@ -1,5 +1,6 @@
 package org.broadinstitute.dsde.firecloud.service
 
+import com.typesafe.scalalogging.slf4j.LazyLogging
 import org.broadinstitute.dsde.firecloud.FireCloudConfig
 import org.broadinstitute.dsde.firecloud.dataaccess.{OntologyDAO, RawlsDAO}
 import org.broadinstitute.dsde.firecloud.model.Ontology.TermParent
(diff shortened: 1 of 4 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
```diff
@@ -64,7 +95,12 @@ trait LibraryServiceSupport {
     ontologyDAO.search(leafId) map {
       case Some(terms) if terms.nonEmpty =>
         terms.head.parents.getOrElse(Seq.empty)
-      case None => Seq.empty
+      case None => Seq.empty[TermParent]
+    } recoverWith {
+      case ex:Exception => {
+        logger.warn(s"exception getting term parents from ontology: ${ex.getMessage}")
```
rushtong: The recover handles any exception from `ontologyDAO.search(leafId)` which is not the just the parents, but the full ontology term. Should this be `logger.warn(s"exception getting term from ontology: ${ex.getMessage}")` ?

rushtong (approved):
👍  Minor logging comment, otherwise looks good, works good. Indexing was pretty quick - pretty sure there's little impact from the `Await` calls for now. Definitely an issue for the future though.

ahaessly (approved):
looks good.

PR #404, opened 2017-03-16
GAWB-1405

skwalker:
rawls passthroughs for https://github.com/broadinstitute/rawls/pull/649

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -2228,6 +2228,7 @@ paths:
         operationId: getTags
         summary: |
           Get all tags used in FireCloud (for autocomplete)
+        description: list of all tags used in FireCloud and their associated frequencies
         parameters:
           - in: query
             description: Query string
@@ -2236,10 +2237,10 @@ paths:
             type: string
         responses:
           200:
-            description: list of all tags used in FireCloud and their associated frequencies
-            type: array
-            items:
-              $ref: '#/definitions/WorkspaceTag'
+            schema:
+              type: array
+              items:
+                $ref: '#/definitions/WorkspaceTag'
           500:
             description: Rawls Internal Error
 
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -2221,7 +2221,7 @@ paths:
         500:
           description: Internal Error
 
-  /api/workspaces/tags/autocomplete/{query}:
+  /api/workspaces/tags:
       get:
         tags:
          - Workspaces
@@ -2229,14 +2229,17 @@ paths:
         summary: |
           Get all tags used in FireCloud (for autocomplete)
         parameters:
-          - name: query
+          - in: query
             description: Query string
-            required: true
+            name: q
+            required: false
             type: string
-            in: path
         responses:
           200:
-            description: List of all tags used in FireCloud
+            description: list of all tags used in FireCloud and their associated frequencies
+            type: array
+            items:
+              $ref: '#/definitions/WorkspaceTag'
           500:
             description: Rawls Internal Error
 
@@ -3932,3 +3935,16 @@ definitions:
       isProtected:
         type: boolean
         description: Whether or not the data should be restricted to dbGaP-enabled users.  Optional, default false.
+
+  WorkspaceTag:
+    required:
+      - tag
+      - count
+    type: object
+    properties:
+      tag:
+        type: string
+        description: tag string
+      count:
+        type: number
+        description: number of usages of the tag across Firecloud
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -2221,6 +2221,25 @@ paths:
         500:
           description: Internal Error
 
+  /api/workspaces/tags/autocomplete/{query}:
+      get:
+        tags:
+         - Workspaces
+        operationId: getTags
+        summary: |
+          Get all tags used in FireCloud (for autocomplete)
+        parameters:
+          - name: query
+            description: Query string
+            required: true
+            type: string
+            in: path
+        responses:
+          200:
+            description: List of all tags used in FireCloud
+          500:
+            description: Rawls Internal Error
+
   /cookie-authed/download/b/{bucket}/o/{object}:
     get:
       tags:
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiService.scala
@@ -85,6 +85,13 @@ trait WorkspaceApiService extends HttpService with FireCloudRequestBuilding
             }
           }
         } ~
+        pathPrefix("tags") {
+          pathEnd {
+            requireUserInfo() { _ =>
+              passthrough(rawlsWorkspacesRoot + "/tags", HttpMethods.GET)
+            }
+          }
+        } ~
         pathPrefix(Segment / Segment) { (workspaceNamespace, workspaceName) =>
           val workspacePath = rawlsWorkspacesRoot + "/%s/%s".format(workspaceNamespace, workspaceName)
           pathEnd {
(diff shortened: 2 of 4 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -2181,6 +2181,25 @@ paths:
...
+          Get all tags used in FireCloud (for autocomplete)
+        parameters:
+          - name: query
+            description: Query string
+            required: true
+            type: string
+            in: path
+        responses:
+          200:
+            description: List of all tags used in FireCloud
```
rushtong: Should this look like the rawls return ... i.e. an array of strings?

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiService.scala
```diff
@@ -85,6 +85,15 @@ trait WorkspaceApiService extends HttpService with FireCloudRequestBuilding
             }
           }
         } ~
+        pathPrefix("tags") {
+          pathPrefix("autocomplete" / Segment) { (query) =>
+            pathEnd {
+              requireUserInfo() { _ =>
+                passthrough(rawlsWorkspacesRoot + "/tags/autocomplete/%s".format(query), HttpMethods.GET)
```
rushtong: In practice, I'm seeing 500 errors with illegal characters in a tag string:
```
root spray.http.IllegalUriException: Illegal URI reference, unexpected character ' ' at position 79: 
root https://local.broadinstitute.org:20443/api/workspaces/tags/autocomplete/testing 12345
root                                                                                ^
```

rushtong (changes requested):
Need to handle invalid characters in the pass-through query string.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiService.scala
davidangb: easiest way to address this is probably to create a `Uri`, let that object handle all the escaping, and pass through to `Uri.toString`.

davidangb (commented):
See my comments on the associated rawls PR - depending on what happens in rawls, code may have to change here as well.

rushtong (approved):

src/main/resources/swagger/api-docs.yaml
```diff
@@ -2181,6 +2181,28 @@ paths:
...
+            description: Query string
+            name: q
+            required: false
+            type: string
+        responses:
+          200:
+            description: list of all tags used in FireCloud and their associated frequencies
+            type: array
+            items:
+              $ref: '#/definitions/WorkspaceTag'
```
davidangb: syntax here is not quite right. I think this should be:
```
          200:
            description: list of all tags used in FireCloud and their associated frequencies
            schema:
              type: array
              items:
                $ref: '#/definitions/WorkspaceTag'
```
The swagger plugin for IntelliJ finds these errors!

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiService.scala
```diff
@@ -85,6 +85,20 @@ trait WorkspaceApiService extends HttpService with FireCloudRequestBuilding
             }
           }
         } ~
+        pathPrefix("tags") {
+          pathEnd {
+            requireUserInfo() { _ =>
```
davidangb, line 90: nit: if you don't need the output of `requireUserInfo()`, you can just omit the `_ =>`.

src/test/scala/org/broadinstitute/dsde/firecloud/mock/MockWorkspaceServer.scala
```diff
@@ -158,6 +159,19 @@ object MockWorkspaceServer {
           .withBody(MockUtils.rawlsErrorReport(NotFound).toJson.compactPrint)
       )
 
+    MockWorkspaceServer.workspaceServer
+      .when(
+        request()
+          .withMethod("GET")
+          .withPath(s"${workspaceBasePath}/tags/autocomplete/abc")
```
davidangb: this path should now just be /tags. Actually ... if you're using the `stubRawlsService` util in the test, I don't think this is even needed.

davidangb (commented):
looking good, minor cleanup inline.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiService.scala
skwalker: when I have it just as `requireUserInfo() { ` I get an error 
```found   : spray.routing.RequestContext => Unit
[error]  required: org.broadinstitute.dsde.firecloud.model.UserInfo => (spray.routing.RequestContext => Unit)
[error]               parameter('q.?) { queryString =>```

Comment (davidangb) [reply src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiService.scala:90]: huh. okay! my bad.

**Review by davidangb — APPROVED** [review]
👍 !

File: src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala [hunk 6]
```diff
@@ -439,7 +450,17 @@ class WorkspaceApiServiceSpec extends BaseServiceSpec with WorkspaceApiService w
           }
         }
       }
+
+    // how to add queries?
+    "Passthrough tests on the workspaces/tags path" - {
+      "OK status is returned for GET" in {
+        stubRawlsService(HttpMethods.GET, tagAutocompletePath, OK)
+        Get(Uri(tagAutocompletePath).withQuery(("q", "tag"))) ~> dummyUserIdHeaders("1234") ~> sealRoute(workspaceRoutes) ~> check {
```
rushtong: A note on this test. You're doing two things here, stubbing out a rawls call to `tagAutocompletePath` and then making an API call to Orchestration on the same path. This works if and only if orchestration and rawls share the same endpoint path. That isn't always the case and I've seen bugs where the two are assumed to be the same, but aren't. 

I think I would prefer that you use the orchestration path url explicitly here to avoid that confusion. You did that above in the `MethodNotAllowed` test.

src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
```diff
@@ -439,7 +450,17 @@ class WorkspaceApiServiceSpec extends BaseServiceSpec with WorkspaceApiService w
           }
         }
       }
+
+    // how to add queries?
```
rushtong: See https://github.com/broadinstitute/firecloud-orchestration/blob/develop/src/test/scala/org/broadinstitute/dsde/firecloud/service/LibraryApiServiceSpec.scala#L72 for how I dealt with stubbing out different query responses. With those stubbed out, you can make correlated GETs and ensure that you're getting back what you want.

rushtong (changes requested):
Looking good - comments inline!

src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
```diff
@@ -451,12 +456,15 @@ class WorkspaceApiServiceSpec extends BaseServiceSpec with WorkspaceApiService w
         }
       }
 
-    // how to add queries?
     "Passthrough tests on the workspaces/tags path" - {
       "OK status is returned for GET" in {
-        stubRawlsService(HttpMethods.GET, tagAutocompletePath, OK)
-        Get(Uri(tagAutocompletePath).withQuery(("q", "tag"))) ~> dummyUserIdHeaders("1234") ~> sealRoute(workspaceRoutes) ~> check {
+        val tagJsonString = "{ \"tag\": \"tagtest\", \"count\": 3 }"
```
davidangb: Scala-wise, I think it's easier to use the triple-quote:
```
val tagJsonString = """{"tag": "tagtest", "count": 3 }"""
```

rushtong (approved):
👍

davidangb (approved):
👍 again.

PR #410, opened 2017-03-22
GAWB-1693: bring autocomplete suggestions more in line with main text search

davidangb:
following Andrea's suggestion, I made the `autocomplete` analyzer look more like the `default` analyzer. This makes our suggestions more consistent with text search.

I also:
* changed autosuggest queries to be phrase search instead of regular search. This makes results for e.g. "brca o" more relevant; previously it would find anything with both "brca" and "o" anywhere in the text; now they need to be part of a phrase, next to each other.
* reduced our highlighter fragment size from 75 to 50. This sets the limit on the suggestion sizes. Entering "coho" will now suggest "This cohort is part of The Cancer Genome Atlas" instead of "This cohort is part of The Cancer Genome Atlas project (https" ... just a bit terser.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/AutoSuggestSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/AutoSuggestSpec.scala
@@ -1,6 +1,7 @@
 package org.broadinstitute.dsde.firecloud.integrationtest
 
 import com.typesafe.scalalogging.slf4j.LazyLogging
+import org.broadinstitute.dsde.firecloud.dataaccess.ElasticSearchDAOQuerySupport
 import org.broadinstitute.dsde.firecloud.integrationtest.ESIntegrationSupport._
 import org.scalatest.{BeforeAndAfterAll, FreeSpec, Matchers}
 import spray.json.JsString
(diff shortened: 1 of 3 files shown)


src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/AutoSuggestSpec.scala
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  val dur = Duration(2, MINUTES)
+  private def suggestionsFor(txt:String) = {
+    val criteria = emptyCriteria.copy(searchString = Some(txt))
+    Await.result(searchDAO.suggestionsFromAll(criteria, Seq.empty[String]), dur)
+  }
+
+  final val HL_START = "<strong class='es-highlight'>"
+  final val HL_END = "</strong>"
```
rushtong: Minor DRY fix - these two vals could be defined in `ElasticSearchDAOQuerySupport` and then re-used.

rushtong (approved):
Love the test. Also, good opportunity to read up on these ES features.

davidangb:
@rushtong back to you for re-review. Since you reviewed last, I have addressed a use case of suggestions for "brca o".
* reverted the min ngram change; min ngram is back to 1. Otherwise, we wouldn't consider the "o" in "brca o"
* changed suggestion query to use phrase search. Previously, we found anything with both "brca" and "o" anywhere in the text; now they need to be next to each other.
* added a test for "brca o".
* removed the test for "tc", because I dropped the ngram size.
* addressed your feedback about the HL_START and HL_END variables - good catch!!!

rushtong (approved):
👍 Can't find anything to comment on ... looks great!

PR #413, opened 2017-03-24
Ah gawb 1709: orch changes to check permissions

ahaessly:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -75,15 +75,13 @@ class LibraryService (protected val argUserInfo: UserInfo,
...
-          // this is technically vulnerable to a race condition in which the workspace attributes have changed
-          // between the time we retrieved them and here, where we update them.
-          val remove = Seq(RemoveAttribute(discoverableWSAttribute))
-          val operations = newGroups map (group => AddListMember(discoverableWSAttribute, new AttributeString(group)))
-          RequestComplete(patchWorkspace(ns, name, remove ++ operations, isPublished(workspaceResponse)))
-        } else {
-          RequestCompleteWithErrorReport(Forbidden, "must be an owner to set discoverable groups")
+      rawlsDAO.getWorkspace(ns, name) flatMap { workspaceResponse =>
+        val remove = Seq(RemoveAttribute(discoverableWSAttribute))
+        val operations = newGroups map (group => AddListMember(discoverableWSAttribute, new AttributeString(group)))
```
rushtong: Nit: Not sure if your IDE shows this, but you don't need the `new` keyword here.

rushtong (commented):
First pass code review looks good. Still need to run this with rawls for testing.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -53,7 +53,7 @@ class LibraryService (protected val argUserInfo: UserInfo,
                       val searchDAO: SearchDAO,
                       val ontologyDAO: OntologyDAO)
                      (implicit protected val executionContext: ExecutionContext) extends Actor
-  with LibraryServiceSupport with AttributeSupport with PermissionsSupport with SprayJsonSupport with LazyLogging {
+  with LibraryServiceSupport with AttributeSupport with LibraryPermissionsSupport with PermissionsSupport with SprayJsonSupport with LazyLogging {
```
rushtong: Since `LibraryPermissionsSupport` extends `PermissionsSupport`, you get both with just `LibraryPermissionsSupport`

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -114,23 +112,27 @@ class LibraryService (protected val argUserInfo: UserInfo,
...
             }
           }
         }
     }
   }
 
   /*
-   * Update workspace in rawls. If the dataset is currently published, republish with changes
+   * Uses admin credentials to update the workspace in rawls. Will republish if it is currently in the published state.
+   * You must do your own permissions check before using this method
```
rushtong: Tracing through this comment - it is a bit confusing. Why can't we enforce them here? I think you can make it private to reduce the chance of improper usage.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryPermissionsSupport.scala
```diff
@@ -0,0 +1,63 @@
...
+import org.broadinstitute.dsde.firecloud.utils.PermissionsSupport
+import org.broadinstitute.dsde.rawls.model.{ErrorReport, WorkspaceAccessLevels, WorkspaceResponse}
+import spray.http.StatusCodes
+
+import scala.concurrent.Future
+
+/**
+ * Created by ahaessly on 3/21/17.
+ */
+trait LibraryPermissionsSupport extends PermissionsSupport {
```
rushtong: Some tests for the public methods would be good.

rushtong (approved):
👍  Looks good - minor questions/comments inline

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
ahaessly: good catch

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -114,23 +112,27 @@ class LibraryService (protected val argUserInfo: UserInfo,
...
-              // this is technically vulnerable to a race condition in which the workspace attributes have changed
-              // between the time we retrieved them and here, where we update them.
+            rawlsDAO.getWorkspace(ns, name) flatMap { workspaceResponse =>
               val allOperations = generateAttributeOperations(workspaceResponse.workspace.attributes, userAttrs,
                 k => k.namespace == AttributeName.libraryNamespace && k.name != LibraryService.publishedFlag.name)
-              RequestComplete(patchWorkspace(ns, name, allOperations, isPublished(workspaceResponse)))
+              // this is technically vulnerable to a race condition in which the workspace attributes have changed
+              // between the time we retrieved them and here, where we update them.
+              if (userAttrs.contains(discoverableWSAttribute))
+                withDiscoverabilityModifyPermissions(workspaceResponse){adminPatchWorkspaceAndRepublish(ns, name, allOperations, isPublished(workspaceResponse)) map (RequestComplete(_))}
```
davidangb: ooh - this means that someone with (grant OR (read+ AND catalog)) can change arbitrary workspace attributes, as long as the request also contains the discoverableWSAttribute.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -140,21 +142,23 @@ class LibraryService (protected val argUserInfo: UserInfo,
...
-      if (workspaceResponse.accessLevel >= requiredLevel) {
-        val operations = updatePublishAttribute(value)
-        rawlsDAO.patchWorkspaceAttributes(ns, name, operations) map { ws =>
-          if (value)
-            publishDocument(ws)
-          else
-            removeDocument(ws)
-          RequestComplete(ws)
+      if (pub == value)
+        Future(RequestCompleteWithErrorReport(NotModified, s"No changes needed"))
```
davidangb: I don't like using 304 here ... 304 is not an error and has implications for client caching. How about returning `Future(RequestComplete(NoContent))` here?

src/main/scala/org/broadinstitute/dsde/firecloud/utils/PermissionsSupport.scala
```diff
@@ -15,6 +15,7 @@ import scala.concurrent.{ExecutionContext, Future}
  */
 trait PermissionsSupport {
   protected val rawlsDAO: RawlsDAO
+//  implicit val userInfo: UserInfo
```
davidangb: delete instead of commenting out

src/test/scala/org/broadinstitute/dsde/firecloud/service/LibraryApiServiceSpec.scala
```diff
@@ -174,21 +174,32 @@ class LibraryApiServiceSpec extends BaseServiceSpec with LibraryApiService with
...
+        new RequestBuilder(HttpMethods.PUT)(setMetadataPath("publishedwriter"), content) ~> dummyUserIdHeaders("1234") ~> sealRoute(libraryRoutes) ~> check {
           status should equal(OK)
+          assert(this.searchDao.indexDocumentInvoked, "indexDocument should have been invoked")
+        }
+      }
+      "should be allowed when reader and catalog" in {
+        val content = HttpEntity(ContentTypes.`application/json`, testLibraryMetadata)
+        new RequestBuilder(HttpMethods.PUT)(setMetadataPath("publishedreadercatalog"), content) ~> dummyUserIdHeaders("1234") ~> sealRoute(libraryRoutes) ~> check {
+          status should equal(OK)
+          assert(this.searchDao.indexDocumentInvoked, "indexDocument should have been invoked")
```
davidangb: as with my previous comment, this is wrong. The `userAttributeOne`, `userAttributeTwo`, and `description` in the `testLibraryMetadata` should NOT be updated if you only have reader.

src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
```diff
@@ -221,6 +221,12 @@ class WorkspaceApiServiceSpec extends BaseServiceSpec with WorkspaceApiService w
           }
         }
       }
+      "Forbidden error is returned for HTTP POST with library attributes in output" in {
+        val methodConfigs = MethodConfiguration("namespace", "name", "root", Map.empty, Map.empty, Map("value" -> AttributeString("this.library:param")), MethodRepoMethod("methodnamespace", "methodname", 1))
+        Post(methodconfigsPath, methodConfigs) ~> dummyUserIdHeaders("1234") ~> sealRoute(workspaceRoutes) ~> check {
+          status should equal(Forbidden)
+        }
+      }
```
davidangb, line 236: This test is repeated verbatim at line 481 - remove one.

davidangb (changes requested):
comments inline - I think the business logic around updating workspace attributes is incorrect in one case.

I would love to limit our use of admin credentials orch->rawls to the bare minimum of where we need them. Would it be too complicated to only use `adminPatchWorkspaceAndRepublish` in the case where the user only has READ, but use the previous/standard approach of relying on the user's creds if the user has WRITE+?

src/test/scala/org/broadinstitute/dsde/firecloud/service/LibraryApiServiceSpec.scala
ahaessly: the updateAttributes method removes any non-library attributes before updating. so only those that can be modified by catalog + read will be modified.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -114,23 +112,43 @@ class LibraryService (protected val argUserInfo: UserInfo,
...
+                    withModifyPermissions(workspaceResponse){
+                      internalPatchWorkspaceAndRepublish(workspaceResponse.accessLevel, ns, name, allOperations, isPublished(workspaceResponse)) map (RequestComplete(_))
+                    }
+                  }
+              }
+              else {
+                withModifyPermissions(workspaceResponse){
+                  internalPatchWorkspaceAndRepublish(workspaceResponse.accessLevel, ns, name, allOperations, isPublished(workspaceResponse)) map (RequestComplete(_))
+                }
+              }
```
ahaessly: i'm trying to modify this code to look something like this... is this better? 

```
              (if (userAttrs.contains(discoverableWSAttribute)) {
                if (userAttrs.size == 1)
                  withDiscoverabilityModifyPermissions _
                else {
                  // make one call just to check perms... if this succeeds, we'll make the other call outside the block
                  withDiscoverabilityModifyPermissions(workspaceResponse){Future.successful(RequestComplete(""))}
                  withModifyPermissions _
                }
              }
              else {
                withModifyPermissions _
              }) (workspaceResponse) {
                internalPatchWorkspaceAndRepublish(workspaceResponse.accessLevel, ns, name, allOperations, isPublished(workspaceResponse)) map (RequestComplete(_))
              }
```

dvoet:
Furthering my comments on the associated rawls PR, I don't think we should use the admin account for workspace access at all. Access control on workspaces should remain completely and totally within rawls.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
davidangb: What if you rewrote it like this?
```
              val doUpdate:Boolean = userAttrs match {
                case x if x.contains(discoverableWSAttribute) && x.size == 1 => canModifyDiscoverability(workspaceResponse)
                case x if x.contains(discoverableWSAttribute) => canModifyDiscoverability(workspaceResponse) && canModify(workspaceResponse)
                case _ => canModify(workspaceResponse)
              }

              if (doUpdate)
                internalPatchWorkspaceAndRepublish(workspaceResponse.accessLevel, ns, name, allOperations, isPublished(workspaceResponse)) map (RequestComplete(_))
              else
                throw FireCloudException("there was a problem") // not sure what to do here
```

trick is using the can* permission methods instead of the with* methods ...

davidangb:
👍 for code/syntax (I did not test), with one refactoring comment inline. Need to address the higher level architecture concerns.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -113,24 +105,34 @@ class LibraryService (protected val argUserInfo: UserInfo,
...
     }
   }
 
   /*
-   * Update workspace in rawls. If the dataset is currently published, republish with changes
+   * Will republish if it is currently in the published state.
    */
-  def patchWorkspace(ns: String, name: String, allOperations: Seq[AttributeUpdateOperation], isPublished: Boolean) : Future[Workspace] = {
-    rawlsDAO.patchWorkspaceAttributes(ns, name, allOperations) map { newws =>
+  def internalPatchWorkspaceAndRepublish(ns: String, name: String, allOperations: Seq[AttributeUpdateOperation], isPublished: Boolean): Future[Workspace] = {
```
rushtong: This is only used in `LibraryService`, and given the `internal` part of the name, should this be private?

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
```diff
@@ -145,4 +146,22 @@ trait LibraryServiceSupport extends LazyLogging {
     docs.copy(results = updatedResults)
   }
 
+  def isDiscoverableDifferent(workspaceResponse: WorkspaceResponse, userAttrs: AttributeMap): Boolean = {
```
rushtong: Could use a comment describing the logic here.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiService.scala
```diff
@@ -110,8 +110,21 @@ trait WorkspaceApiService extends HttpService with FireCloudRequestBuilding
...
+              }
+            } ~
+            post {
+              requireUserInfo() { _ =>
+                entity(as[MethodConfiguration]) { methodConfig =>
+                  requestContext =>
+                    if (methodConfig.outputs.collect { case param if param._2.value.startsWith("this.library:") => param }.isEmpty)
+                      passthrough(workspacePath + "/methodconfigs", HttpMethods.GET, HttpMethods.POST)
+                    else
+                      requestContext.complete(StatusCodes.Forbidden, "Can not modify library attributes")
```
rushtong: Wording nit: We're saying here that the user cannot have a method write back attributes to the workspace that are in the library namespace. How about `Methods and configurations cannot create or modify library attributes.` ?

rushtong (approved):
👍  Pending other review comments. Minor comments overall.

src/main/scala/org/broadinstitute/dsde/firecloud/model/Workspace.scala
```diff
@@ -29,7 +29,7 @@ object WorkspaceCreate {
     new WorkspaceCreate(
       namespace = wc.namespace,
       name = wc.name,
-      attributes = wc.attributes + (AttributeName("library","published") -> AttributeBoolean(false)),
+      attributes = wc.attributes + (AttributeName("library","published") -> AttributeBoolean(false)) + (AttributeName("library","discoverableByGroups") -> AttributeValueEmptyList),
```
rushtong, line 32: Probably not for this PR - just thinking out loud here. We're at the point where there is a set of library attributes that are non-clone-able. We should think about maintaining that set closer to `AttributeName` than in `WorkspaceCreate`.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
```diff
@@ -145,4 +146,27 @@ trait LibraryServiceSupport extends LazyLogging {
     docs.copy(results = updatedResults)
   }
 
+  // this method will determine if the user is makings a change to discoverableByGroups
+  // if the attribute does not exist on the workspace, it is the same as the empty list
```
rushtong: 👍

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiService.scala
```diff
@@ -110,8 +110,21 @@ trait WorkspaceApiService extends HttpService with FireCloudRequestBuilding
...
+            get {
+              requireUserInfo() { _ =>
+                passthrough(workspacePath + "/methodconfigs", HttpMethods.GET)
+              }
+            } ~
+            post {
+              requireUserInfo() { _ =>
+                entity(as[MethodConfiguration]) { methodConfig =>
+                  requestContext =>
+                    if (methodConfig.outputs.collect { case param if param._2.value.startsWith("this.library:") => param }.isEmpty)
```
rushtong, line 122: Also probably not for this PR ... still thinking out loud. This feels like it is more a utility or library support kind of check and seems like too much logic to be putting inside of a route.

rushtong (approved):
👍  See comments inline. Should not hold up the PR.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
```diff
@@ -145,4 +146,27 @@ trait LibraryServiceSupport extends LazyLogging {
...
+
+    if (current.isEmpty && newvals.isEmpty)
+      false
+    else if (current.nonEmpty && newvals.nonEmpty) {
+      current.toSet != newvals.toSet
+    } else
+      // one of the values is the empty list and the other is not
+      true
+  }
+
```
davidangb: if you ignored the corner case where the user is changing the order of values within the list - and it's definitely a corner case because the UI doesn't allow lists - this would be as simple as:

```
  def isDiscoverableDifferent(workspaceResponse: WorkspaceResponse, userAttrs: AttributeMap): Boolean = {

    val current = workspaceResponse.workspace.attributes.get(LibraryService.discoverableWSAttribute)
    val newvals = userAttrs.get(LibraryService.discoverableWSAttribute)

    current != newvals
  }
```

is the corner case worth the extra code?

If you _do_ feel adamant about ignoring order, how about simplifying to this:

```
  def isDiscoverableDifferent(workspaceResponse: WorkspaceResponse, userAttrs: AttributeMap): Boolean = {

    def convert(list: Option[Attribute]): Seq[AttributeValue] = {
      list match {
        case Some(x:AttributeValueList) => x.list
        case _ => Seq.empty[AttributeValue]
      }
    }
    val current = convert(workspaceResponse.workspace.attributes.get(LibraryService.discoverableWSAttribute))
    val newvals = convert(userAttrs.get(LibraryService.discoverableWSAttribute))

    current.toSet != newvals.toSet

  }
```

davidangb (commented):
looking good, one refactor inline!

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
ahaessly: so AttributeValueList == will compare the data and not just the identity?

davidangb: yup ... this is one of the wonderful benefits of using case classes. Scala auto-generates equals functions for case classes! (plus, I can see it passing unit tests)

ahaessly: yeah, i feel better if it doesn't assume order - did change to your second suggestion

davidangb (approved):

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -140,21 +142,20 @@ class LibraryService (protected val argUserInfo: UserInfo,
...
+  // should only be used to change published state
   def setWorkspaceIsPublished(ns: String, name: String, value: Boolean): Future[PerRequestMessage] = {
     rawlsDAO.getWorkspace(ns, name) flatMap { workspaceResponse =>
       val pub = isPublished(workspaceResponse)
-      val requiredLevel = if (!pub || (pub && !value)) WorkspaceAccessLevels.Owner else WorkspaceAccessLevels.Write
-      if (workspaceResponse.accessLevel >= requiredLevel) {
-        val operations = updatePublishAttribute(value)
-        rawlsDAO.patchWorkspaceAttributes(ns, name, operations) map { ws =>
+      if (pub == value)
+        Future(RequestComplete(NoContent))
```
davidangb, line 150: if pub == value, but value is true, we still need to re-publish to ES, right? To ensure that the latest metadata gets to ES.

ahaessly: the re-publish (i.e. when library attributes have been modified for a published dataset) happens in the method "internalPatchWorkspaceAndRepublish". the method "setWorkspaceIsPublished" should only be invoked when changing the published state. i think the reason we have these 2 methods is that the api treats them separately and when we were checking permissions in orch, they had different requirements

davidangb: ah, ok, I see that now. all good.

davidangb (approved):
👍

PR #418, opened 2017-03-31
GAWB-1734: Use new Thurloe endpoint for saving KVPs in bulk + add endpoint for saving preferences

MatthewBemis:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1125,7 +1125,6 @@ paths:
       responses:
         200:
           description: OK
-
           schema:
             type: array
             items:
@@ -1161,6 +1160,24 @@ paths:
             - profile
             - https://www.googleapis.com/auth/devstorage.full_control
             - https://www.googleapis.com/auth/cloud-billing
+
+  /api/profile/preferences:
+    post:
+      tags:
+        - Profile
+      operationId: billing
+      summary: Update the preferences for the user
+      parameters:
+        - name: preferences
+          in: body
+          description: The preferences to set as a Map[String, String]
+          required: true
+          type: Map[String, String]
+      responses:
+        204:
+          description: Success (No Content)
+        500:
+          description: Internal Server Error
 
   /api/refresh-token-status:
     get:
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
@@ -45,7 +45,7 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
   }
 
   override def saveKeyValues(userInfo: UserInfo, keyValues: Map[String, String]): Future[Try[Unit]] = {
-    val thurloeKeyValues = keyValues.map { case (key, value) => ThurloeKeyValue(Some(userInfo.getUniqueId), Some(FireCloudKeyValue(Some(key), Some(value)))) }
+    val thurloeKeyValues = ThurloeKeyValues(Some(userInfo.getUniqueId), Some(keyValues.map { case (key, value) => FireCloudKeyValue(Some(key), Some(value)) }.toSeq))
     wrapExceptions {
       userAuthedRequest(Post(UserService.remoteSetKeyURL, thurloeKeyValues), false, true)(userInfo) map { response =>
         Try(response.status match {
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
@@ -45,7 +45,7 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
   }
 
   override def saveKeyValues(userInfo: UserInfo, keyValues: Map[String, String]): Future[Try[Unit]] = {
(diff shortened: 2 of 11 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
```diff
@@ -44,24 +44,19 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
     }
   }
 
-  override def saveKeyValue(userInfo: UserInfo, key: String, value: String): Future[Try[Unit]] = {
+  override def saveKeyValues(userInfo: UserInfo, keyValues: Map[String, String]): Future[Try[Unit]] = {
+    val thurloeKeyValues = ThurloeKeyValues(Some(userInfo.getUniqueId), Some(keyValues.map { case (key, value) => FireCloudKeyValue(Some(key), Some(value)) }.toSeq))
```
dvoet: use Option instead of Some to avoid the potential of Some(null)

dvoet (approved):
1 minor knit

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1162,6 +1161,24 @@ paths:
             - https://www.googleapis.com/auth/devstorage.full_control
             - https://www.googleapis.com/auth/cloud-billing
 
+  /api/profile/preferences:
```
dmohs, line 1164: Not your fault, but this file is over 4k lines. That is unacceptable. It needs to be divided up and, ideally, each section moved as close as possible to the code that implements the documented functionality.

https://broadinstitute.atlassian.net/browse/GAWB-1911

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
```diff
@@ -44,24 +44,19 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
     }
   }
 
-  override def saveKeyValue(userInfo: UserInfo, key: String, value: String): Future[Try[Unit]] = {
+  override def saveKeyValues(userInfo: UserInfo, keyValues: Map[String, String]): Future[Try[Unit]] = {
```
dmohs, line 47: A set of keys with associated values already has a name: a map. So, perhaps, instead of `keyValues`, `map`? Unless there is some value in inventing our own term for this.

src/main/scala/org/broadinstitute/dsde/firecloud/model/Profile.scala
```diff
@@ -13,6 +13,10 @@ case class ThurloeKeyValue(
   userId: Option[String] = None,
   keyValuePair: Option[FireCloudKeyValue] = None)
 
+case class ThurloeKeyValues(
+  userId: Option[String] = None,
+  keyValuePairs: Option[Seq[FireCloudKeyValue]] = None)
```
dmohs, line 18: `FireCloudKeyValue` is a bit misleading. The name makes me think it a special FireCloud object, but really it's just a key/value pair where both the key and value are optional. So, two questions, (1) would `OptionalKeyValue` be a better name and (2) why is `key` optional?

src/main/scala/org/broadinstitute/dsde/firecloud/model/Profile.scala
```diff
@@ -13,6 +13,10 @@ case class ThurloeKeyValue(
   userId: Option[String] = None,
   keyValuePair: Option[FireCloudKeyValue] = None)
 
+case class ThurloeKeyValues(
```
dmohs, line 16: Similar to my other comment, `UserMap` might be a better name here, since it's basically just a map (a sequence of key/value pairs) tied to a userId. An optional userId?

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/RegisterApiService.scala
```diff
@@ -33,4 +34,19 @@ trait RegisterApiService extends HttpService with PerRequestCreator with FireClo
         }
       }
     }
+
+  val profileRoutes: Route =
+    pathPrefix("profile") {
+      path("preferences") {
+        post {
+          requireUserInfo() { userInfo =>
```
dmohs, line 42: Yay for preserving good HTTP error messages!

dmohs (approved):
You have the misfortune of being a better Scala developer than me, so I'm asking harder questions and holding your code to a higher standard than I would myself. Sorry. Feel free to ignore and just submit as-is.

src/main/scala/org/broadinstitute/dsde/firecloud/model/Profile.scala
MatthewBemis: Most of your suggestions have a similar theme so I'll just respond to one of them with my thoughts:

1) I'm not sure why userId or the keyValues are optional. I was just following the already established pattern. I think that it may have something to do with JSON unmarshalling? I'll do some more digging to understand why since "I was just following the pattern" is a crappy answer
2) Most of the terminology used here was chosen to be consistent with Thurloe- to make the mental mapping a bit easier. 
3) I agree with your point about `FireCloudKeyValue`. I think that a rename would be useful there.

dmohs: And just to be clear, I'd be perfectly happy if this PR went in as-is and the research (with a potential follow-up PR) came later.

MatthewBemis: Cool, I'm down for that as a follow-up. I'll merge this as soon as the next alpha snapshot is taken

PR #424, opened 2017-04-13
/status endpoint for Firecloud Orchestration [GAWB-771]

aednichols:
The test exercises the routing behavior, but it's not very helpful yet because I haven't set up the mock DAOs to return failure responses.

Also, I aimed for the acceptance criteria, but once we get through review, it would be good to tackle the nice-to-haves as well.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [ ] **LR** sign off


src/main/scala/org/broadinstitute/dsde/firecloud/webservice/StatusApiService.scala
```diff
@@ -0,0 +1,46 @@
...
+import java.util.Date
+
+import org.broadinstitute.dsde.firecloud.FireCloudConfig
+import org.broadinstitute.dsde.firecloud.service._
+import spray.json._
+import spray.routing._
+
+trait StatusApiService extends HttpService with PerRequestCreator with FireCloudDirectives {
+
+  private final val ApiPrefix = "status"
```
rushtong: Two nits. 
1. `camelCase`
2. Both route vals use the same prefix, either use it in both places or drop it altogether.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/StatusApiService.scala
```diff
@@ -0,0 +1,46 @@
...
+  private final val dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSZ")
+  private implicit val executionContext = actorRefFactory.dispatcher
+
+  val statusServiceConstructor: () => StatusService
+
+  val statusRoutes: Route =
```
rushtong: naming suggestion ... these routes are under `/api` so I think it should be named similarly (i.e. `apiStatusRoutes`) to keep the confusion level down between why these two routes are separate. A comment here would also be good explaining why.

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
```diff
@@ -33,7 +34,7 @@ class FireCloudServiceActor extends HttpServiceActor with FireCloudDirectives
     def actorRefFactory = context
   }
 
-  val agoraDAO:AgoraDAO = new HttpAgoraDAO(FireCloudConfig.Agora.authUrl)
+  val agoraDAO:AgoraDAO = new HttpAgoraDAO(FireCloudConfig.Agora.authUrl, FireCloudConfig.Agora.baseUrl)
```
rushtong: Suggestion: Instead of adding a second url to the constructor, change the constructor to take the whole Agora config. Then, inside HttpAgoraDAO, do whatever you need to do with the different urls. This frees us up from expanding the constructor and opens up some freedom inside the dao class when we need to add more functionality.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpAgoraDAO.scala
```diff
@@ -26,4 +27,15 @@ class HttpAgoraDAO(agoraUrl: String)(implicit val system: ActorSystem, implicit
   override def postNamespacePermissions(ns: String, entity: String, perms: List[AgoraPermission])(implicit userInfo: UserInfo): Future[List[AgoraPermission]] =
     authedRequestToObject[List[AgoraPermission]]( Post(getNamespaceUrl(ns, entity), perms) )
 
+  def status: Future[(Boolean, Option[String])] = {
+    val response = unAuthedRequestToObject[AgoraStatus](Get(Uri(agoraBaseUrl).withPath(Uri.Path("/status"))))
+
+    response map { response =>
+      response.status match {
+        case "up" => (true, None)
+        case "down" => (false, if (response.message.length > 0) Some(response.message.mkString("\r\n")) else None)
```
rushtong: Probably unlikely, but you might want to handle `case _` with a custom error message to catch anything else.

rushtong (changes requested):
Sorry this isn't complete. Will re-review if possible. 
Also ... Tests!!! :-)

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpAgoraDAO.scala
```diff
@@ -26,4 +27,15 @@ class HttpAgoraDAO(agoraUrl: String)(implicit val system: ActorSystem, implicit
   override def postNamespacePermissions(ns: String, entity: String, perms: List[AgoraPermission])(implicit userInfo: UserInfo): Future[List[AgoraPermission]] =
     authedRequestToObject[List[AgoraPermission]]( Post(getNamespaceUrl(ns, entity), perms) )
 
+  def status: Future[(Boolean, Option[String])] = {
+    val response = unAuthedRequestToObject[AgoraStatus](Get(Uri(agoraBaseUrl).withPath(Uri.Path("/status"))))
+
+    response map { response =>
+      response.status match {
```
davidangb: nitpick: the variable naming here is super confusing, as this shadows `HttpResponse`. Click through on `unAuthedRequestToObject` and then `requestToObject`, and look at the match clause there that also uses `response.status` ... but for a different class.

src/main/scala/org/broadinstitute/dsde/firecloud/service/StatusService.scala
```diff
@@ -1,43 +1,61 @@
...
+
+      if (!rawlsOK) messages += "Problem with Rawls"
+      if (!thurloeOK) messages += "Problem with Thurloe: " + thurloeMessage.getOrElse("(No further information available)")
+      if (!agoraOK) messages += "Problem with Agora: " + agoraMessage.getOrElse("(No further information available)")
+      if (!searchOK) messages += "Problem with Search"
+
+      RequestComplete(SystemStatus(allOK, messages.mkString("; ")))
     }
-}
+  }
```
davidangb: I'd love to see this return a `Seq` of error messages, instead of a delimited string; this would make it easier to parse by a machine. It could even return an overall "OK" for the system, as well as individual oks/messages for each subservice.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
```diff
@@ -71,4 +71,15 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
...
     }
   }
+
+  def status: Future[(Boolean, Option[String])] = {
+    val response = unAuthedRequestToObject[ThurloeStatus](Get(Uri(FireCloudConfig.Thurloe.baseUrl).withPath(Uri.Path("/status"))), useFireCloudHeader = true)
+
+    response map { response =>
+      response.status match {
+        case "up" => (true, None)
+        case "down" => (false, response.error)
```
rushtong: similar case handling concern - might need something like  `case _ => (false, "Something unexpected")`

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ThurloeDAO.scala
```diff
@@ -19,4 +19,6 @@ trait ThurloeDAO extends LazyLogging {
   def saveKeyValue(userInfo: UserInfo, key: String, value: String): Future[Try[Unit]]
   def saveProfile(userInfo: UserInfo, profile: BasicProfile): Future[Unit]
   def saveKeyValues(userInfo: UserInfo, keyValues: Map[String, String]): Future[Iterable[Try[Unit]]]
+
```
rushtong: nit: extra line

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/StatusApiService.scala
```diff
@@ -0,0 +1,46 @@
...
+      pathEnd {
+        respondWithJSON {
+          complete {
+            JsObject(
+              "workspacesUrl" -> JsString(FireCloudConfig.Rawls.baseUrl + "/workspaces"),
+              "methodsUrl" -> JsString(MethodsService.remoteMethodsUrl),
+              "timestamp" -> JsString(dateFormat.format(new Date()))
+            ).toString
+          }
+        }
```
davidangb, line 30: this one can go away entirely, I'm pretty sure - nothing calls this (I think, must verify with devops)

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/StatusApiService.scala
```diff
@@ -0,0 +1,46 @@
...
+      } ~ path("ping") {
+        complete {
+          dateFormat.format(new Date())
+        }
+      }
```
davidangb, line 35: this one can go away with only a very minor change to the UI.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -2648,6 +2648,20 @@ paths:
...
+  /status:
+    get:
+      operationId: status
+      tags:
+        - Status
+      summary: |
+        An orchestrated status check for underlying systems: Agora, Thurloe, Elasticsearch, Rawls
+      responses:
+        200:
+          description: Success
```
davidangb: this should define the type of object it returns.

src/main/scala/org/broadinstitute/dsde/firecloud/service/StatusService.scala
```diff
@@ -1,43 +1,61 @@
...
+import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.ExecutionContext
+
+/**
+ * Created by anichols on 4/5/17.
+ */
+object StatusService {
+  def props(statusServiceConstructor: () => StatusService): Props = {
+    Props(statusServiceConstructor())
+  }
```
rushtong: Nit, this could be a one-liner
```
def props(statusServiceConstructor: () => StatusService): Props = Props(statusServiceConstructor())
```

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpAgoraDAO.scala
davidangb: you could just have `case "up"` and `case _`. Also is there a reason to add linebreaks? I'd rather pass messages through unfiltered, and leave any rendering niceness to whatever renders it.

src/main/scala/org/broadinstitute/dsde/firecloud/service/StatusService.scala
```diff
@@ -1,43 +1,61 @@
...
+
+  def collectStatusInfo(): Future[PerRequestMessage] = {
+    for {
+      rawlsOK <- rawlsDAO.status
+      (thurloeOK, thurloeMessage) <- thurloeDAO.status
+      (agoraOK, agoraMessage) <- agoraDAO.status
+      searchOK <- Future(searchDAO.indexExists())
+    } yield {
+      val allOK = rawlsOK && thurloeOK && agoraOK && searchOK
+      var messages: ListBuffer[String] = ListBuffer()
```
rushtong: `vars` in scala make me queasy. You're doing this because each of the status methods return something different and you need to do some finagling to get the right data out.

Suggestion: Change the status to return a case class, i.e. `StatusResponseResult` which has not only the boolean, but the response message. Do the same thing for `SearchDAO`. Maybe add a status trait that each of the daos can call that does this.

Then, here in this collect status info message, you can flatmap over a list of all of the status calls. Disclaimer, haven't tested this
```
    val messages: Seq[String] = List(rawlsDAO.status, thurloeDAO.status, agoraDAO.status, searchDAO.status) map { statusCall =>
      statusCall.status match {
        case false => statusCall.message
      }
    }
```

rushtong (changes requested):
Like this overall. I think we can improve the consistency with a common response from the status calls. Also, (not for this story or PR) we need to consider status checks for consent and ontology.

src/main/scala/org/broadinstitute/dsde/firecloud/service/StatusService.scala
```diff
@@ -1,43 +1,61 @@
...
+                     val rawlsDAO: RawlsDAO)
+                    (implicit protected val executionContext: ExecutionContext) extends Actor with SprayJsonSupport {
+
+  def collectStatusInfo(): Future[PerRequestMessage] = {
+    for {
+      rawlsOK <- rawlsDAO.status
+      (thurloeOK, thurloeMessage) <- thurloeDAO.status
+      (agoraOK, agoraMessage) <- agoraDAO.status
+      searchOK <- Future(searchDAO.indexExists())
+    } yield {
```
davidangb: in the case where one of the service calls throws an error, or returns a non-200 response, this will fail-fast and bubble the exception up to orch's /status response. In an ideal situation - and I'd entertain doing this as a follow-on story - each service call would handle exceptions, so we're always trying each one and returning all errors.

To make it concrete: if the Rawls check throws an error or returns a 500, we won't check Thurloe. Thus, we won't report error message for both Rawls and Thurloe.

davidangb (commented):
comments inline!

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/StatusApiService.scala
aednichols: Dropped the `val` altogether - the Spray DSL lends itself to reading with inline literals, and that's what we do everywhere else

aednichols:
- No more `var`s
- Test does something useful
- Created a trait to standardize the interface
- Exceptions from one service no longer bubble up and cause other checks to be skipped
- Updated Swagger docs to match object definitions in the code

Likely improvements:
- Remove currently-unused `/api/status`endpoint
- Update UI to use new endpoint and remove newly-unused `/api/status/ping` endpoint

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala
```diff
@@ -5,12 +5,13 @@ import org.broadinstitute.dsde.rawls.model._
...
 import spray.http.StatusCodes.BadRequest
 import org.broadinstitute.dsde.firecloud.model.MethodRepository.{AgoraPermission, FireCloudPermission}
-import org.broadinstitute.dsde.firecloud.model.Ontology.{TermResource, TermParent, ESTermParent}
+import org.broadinstitute.dsde.firecloud.model.Ontology.{ESTermParent, TermParent, TermResource}
 import spray.json._
 import spray.routing.{MalformedRequestContentRejection, RejectionHandler}
 import spray.routing.directives.RouteDirectives.complete
 import org.broadinstitute.dsde.rawls.model.UserModelJsonSupport._
 import org.broadinstitute.dsde.rawls.model.WorkspaceACLJsonSupport.WorkspaceAccessLevelFormat
+import spray.json.DefaultJsonProtocol.jsonFormat2
```
davidangb: nit: import not needed

src/main/scala/org/broadinstitute/dsde/firecloud/model/SystemStatus.scala
```diff
@@ -0,0 +1,18 @@
+package org.broadinstitute.dsde.firecloud.model
+
+import scala.concurrent.Future
+
+/**
+  * Created by anichols on 4/7/17.
+  */
+case class SystemStatus(ok: Boolean, systems: Map[String, SubsystemStatus])
+
+case class SubsystemStatus(ok: Boolean, messages: Option[Array[String]])
```
davidangb: to be more Scala-native, use a `List` or a `Seq` - one of the immutable collections. `Array`s are mutable.

optional feedback: by adding a default value for `messages`:
```
case class SubsystemStatus(ok: Boolean, messages: Option[List[String]] = None)
```

you can simplify calling code in a few places. Calling code can be just:
```
SubsystemStatus(true)
```

src/main/scala/org/broadinstitute/dsde/firecloud/model/SystemStatus.scala
```diff
@@ -0,0 +1,18 @@
...
+
+case class AgoraStatus(status: String, message: Array[String])
+case class ThurloeStatus(status: String, error: Option[String])
+case class RawlsStatus(version: Option[String])
+
+trait ReportsSubsystemStatus {
+  def status: Future[SubsystemStatus]
+}
```
davidangb: I'm a bit wonky on defining a DAO trait in the model package. I'd rather see this in the dataaccess package, though I understand the argument that this trait is closely tied to the models.

src/main/scala/org/broadinstitute/dsde/firecloud/service/StatusService.scala
```diff
@@ -1,43 +1,62 @@
...
+import scala.concurrent.ExecutionContext
+
+/**
+ * Created by anichols on 4/5/17.
+ */
+object StatusService {
+  def props(statusServiceConstructor: () => StatusService): Props = Props(statusServiceConstructor())
+
+  def constructor(app: Application)()(implicit executionContext: ExecutionContext): StatusService = {
+    new StatusService(app.searchDAO, app.agoraDAO, app.thurloeDAO, app.rawlsDAO)
```
davidangb: optional feedback: at this point, you might just want to pass the entire `app` object, instead of extracting parts.

In looking at the `Application` object, I realize we need additional status checks! Those are for a future story.

src/main/scala/org/broadinstitute/dsde/firecloud/service/StatusService.scala
```diff
@@ -1,43 +1,62 @@
...
+    }
 
-import org.broadinstitute.dsde.firecloud.FireCloudConfig
+    for {
+      rawlsStatus <- rawlsDAO.status recover subsystemExceptionHandler
+      thurloeStatus <- thurloeDAO.status recover subsystemExceptionHandler
+      agoraStatus <- agoraDAO.status recover subsystemExceptionHandler
+      searchStatus <- searchDAO.status recover subsystemExceptionHandler
+    } yield {
+      val statusMap = Map("Rawls" -> rawlsStatus, "Thurloe" -> thurloeStatus, "Agora" -> agoraStatus, "Search" -> searchStatus)
```
davidangb: Instead of hardcoding names here, let's have each DAO define the service name. So, in RawlsDAO.scala:
```
object RawlsDAO {
 ...
  lazy val serviceName = "Rawls"
  ...
}

trait RawlsDAO extends LazyLogging with ReportsSubsystemStatus {

  implicit val errorReportSource = ErrorReportSource(RawlsDAO.serviceName)
```

and here:
```
val statusMap = Map(RawlsDAO.serviceName -> rawlsStatus, ...
```

src/main/scala/org/broadinstitute/dsde/firecloud/service/StatusService.scala
```diff
@@ -1,43 +1,62 @@
...
 
-class StatusServiceActor extends Actor with StatusService {
-  def actorRefFactory = context
-  def receive = runRoute(routes)
-}
+      if (statusMap.values.count(_.ok == false) == 0)
```
davidangb: a bit more Scala-esque: `if (statusMap.values.forall(_.ok))`

src/main/scala/org/broadinstitute/dsde/firecloud/service/StatusService.scala
```diff
@@ -1,43 +1,62 @@
...
-          dateFormat.format(new Date())
-        }
-      }
     }
-}
+  }
+
+  def receive = {
+    case CollectStatusInfo => collectStatusInfo() pipeTo sender
+  }
```
davidangb: by convention, we put the `receive` method at the top of the class, and specify `override`.

src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/ESIntegrationSupport.scala
```diff
@@ -7,6 +7,7 @@ import com.typesafe.config.ConfigFactory
 import org.broadinstitute.dsde.firecloud.FireCloudConfig
 import org.broadinstitute.dsde.firecloud.dataaccess.{ElasticSearchDAO, SearchDAO}
 import org.broadinstitute.dsde.firecloud.model.LibrarySearchParams
+import scala.concurrent.ExecutionContext.Implicits.global
```
davidangb: unnecessary import?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
```diff
@@ -66,4 +66,15 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
...
     }
   }
+
+  override def status: Future[SubsystemStatus] = {
+    val thurloeStatus = unAuthedRequestToObject[ThurloeStatus](Get(Uri(FireCloudConfig.Thurloe.baseUrl).withPath(Uri.Path("/status"))), useFireCloudHeader = true)
+
+    thurloeStatus map { thurloeStatus =>
+      thurloeStatus.status match {
+        case "up" => SubsystemStatus(true, None)
+        case "down" => SubsystemStatus(false, if (thurloeStatus.error.isDefined) Some(Array(thurloeStatus.error.get)) else None)
```
davidangb: You can write this more Scala-like as:
```
case "down" => SubsystemStatus(false, thurloeStatus.error.map(List(_)))
```

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpAgoraDAO.scala
```diff
@@ -26,4 +27,15 @@ class HttpAgoraDAO(agoraUrl: String)(implicit val system: ActorSystem, implicit
   override def postNamespacePermissions(ns: String, entity: String, perms: List[AgoraPermission])(implicit userInfo: UserInfo): Future[List[AgoraPermission]] =
     authedRequestToObject[List[AgoraPermission]]( Post(getNamespaceUrl(ns, entity), perms) )
 
+  override def status: Future[SubsystemStatus] = {
+    val agoraStatus = unAuthedRequestToObject[AgoraStatus](Get(Uri(config.baseUrl).withPath(Uri.Path("/status"))))
+
+    agoraStatus map { agoraStatus =>
+      agoraStatus.status match {
+        case "up" => SubsystemStatus(true, None)
+        case _ => SubsystemStatus(false, if (agoraStatus.message.length > 0) Some(agoraStatus.message) else None)
```
davidangb: use `.nonEmpty` instead of `.length > 0`

davidangb (commented):
Comments inline. This is looking good; the majority of my comments are ways to make the code more native Scala. I've marked feedback as optional where I consider it ... well, optional.

I suggest finishing this PR and handling the removal of obsolete endpoints + updates to UI as a separate ticket.

I'm also okay with the unit test as-is, but either for another story or in this one, I'd like to see some more extensive tests. The additional tests will probably require `mockserver`. I'd like to see tests that check:
* error messages from subservices are properly propagated to the end user
* timeouts or exceptions contacting subservices properly show errors to the end user
* timeouts or exceptions in one subservice do not interfere with status of other subservices

And for a future story, we'll also need status checks for ontology, consent, and possibly google!

src/main/scala/org/broadinstitute/dsde/firecloud/service/StatusService.scala
aednichols: Makes sense. Is there a specific reason that `serviceName` is a `lazy val`?

aednichols: Do you think it would be useful to create a `BaseDAO` to define `serviceName` and house the small amount of code?

davidangb: a `BaseDAO` is a great idea! No particular reason to make the serviceName lazy; that's just habit, and especially in this case doesn't make much of a difference at all.

src/main/scala/org/broadinstitute/dsde/firecloud/service/StatusService.scala
```diff
@@ -1,43 +1,63 @@
...
 
-import org.broadinstitute.dsde.firecloud.FireCloudConfig
+  def collectStatusInfo(): Future[PerRequestMessage] = {
+    val subsystemExceptionHandler: PartialFunction[Any, SubsystemStatus] = {
+      case fcExceptionWithError: FireCloudExceptionWithErrorReport => SubsystemStatus(false, Some(List(fcExceptionWithError.errorReport.message)))
+      case fcException: FireCloudException => SubsystemStatus(false, Some(List(fcException.toString)))
+      case e: Exception => SubsystemStatus(false, Some(List(e.getMessage)))
+      case _: Any => SubsystemStatus(false)
```
davidangb: nit: be consistent about `getMessage` vs. `toString` on the exceptions; I think you should use `fcException.getMessage` here.

Also, the last case could be:
```
case x => SubsystemStatus(false, x.toString)
```
if you wanted to populate *some* kind of message - optional, don't hold the PR for that.

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockAgoraDAO.scala
```diff
@@ -24,4 +24,8 @@ class MockAgoraDAO extends AgoraDAO {
     Future(List(MockAgoraDAO.agoraPermission))
   }
 
+  def status: Future[SubsystemStatus] = {
+    Future(throw new Exception("Agora Mock DAO exception"))
+  }
+
```
davidangb, line 30: Any reason MockAgoraDAO throws an exception for status, but other daos succeed? EDIT: ah, I see, you use this later in tests.

src/test/scala/org/broadinstitute/dsde/firecloud/service/StatusServiceSpec.scala
```diff
@@ -0,0 +1,149 @@
...
+  )
+
+  "StatusService indicates Agora failure when \"down\" response" in {
+    Get("/status") ~> sealRoute(publicStatusRoutes) ~> check {
+      status.intValue should be(InternalServerError.intValue)
+      val response = responseAs[SystemStatus]
+      response.ok should be(false)
+      response.systems("Agora").ok should be(false)
+      response.systems("Agora").messages should be(Some(List("Agora is down")))
+      response.systems.size should be(4)
```
davidangb: use `AgoraDAO.serviceName` instead of `"Agora"` here. Also for good measure check that the other subsystems return true for their `ok` val?

src/test/scala/org/broadinstitute/dsde/firecloud/service/StatusServiceSpec.scala
```diff
@@ -0,0 +1,149 @@
+package org.broadinstitute.dsde.firecloud.service
```
davidangb: because these tests run through the orch api - `Get("/status") ~> sealRoute(publicStatusRoutes) ~> check {`, this file should really be named `StatusApiServiceSpec`, with appropriate class renaming inside

src/test/scala/org/broadinstitute/dsde/firecloud/service/StatusServiceSpec.scala
```diff
@@ -0,0 +1,149 @@
...
+
+  agoraServer.when(
+    request().withMethod("GET").withPath("/status")
+  ).respond(
+    org.mockserver.model.HttpResponse.response()
+      .withHeaders(MockUtils.header)
+      .withBody("""{ "status": "down", "message": ["Agora is down"] }""")
+  )
+
+  "StatusService indicates Agora failure when \"down\" response" in {
```
davidangb: no need to change the PR, but for your edification: you can use triple-quotes to avoid escaping inner quotes:
```
  """StatusService indicates Agora failure when "down" response""" in {
```

src/test/scala/org/broadinstitute/dsde/firecloud/service/StatusServiceSpec.scala
```diff
@@ -0,0 +1,149 @@
...
+    Get("/status") ~> sealRoute(publicStatusRoutes) ~> check {
+      status.intValue should be(InternalServerError.intValue)
+      val response = responseAs[SystemStatus]
```
davidangb: don't use `response` here as a variable name - spray-routing creates a `response` variable automatically (similar to the `status` variable you're using), and an explicit `val response` here shadows that.

src/test/scala/org/broadinstitute/dsde/firecloud/service/StatusServiceSpec.scala
```diff
@@ -0,0 +1,149 @@
...
+  "StatusService indicates OK" in {
+    Get("/status") ~> sealRoute(publicStatusRoutes) ~> check {
+      status.intValue should be(OK.intValue)
+      val response = responseAs[SystemStatus]
+      response.ok should be(true)
+      response.systems("Agora").ok should be(true)
+      response.systems("Agora").messages should be(None)
+      response.systems.size should be(4)
+    }
+  }
```
davidangb: you're modifying what the mock server returns for each test in this class. I'd prefer to see the mockserver expectation set up inside of each test's closure, like this:
```
  "StatusService indicates OK" in {
    agoraServer.when(
      request().withMethod("GET").withPath("/status")
    ).respond(
      org.mockserver.model.HttpResponse.response()
        .withHeaders(MockUtils.header)
        .withBody("""{ "status": "up", "message": ["No problems with Agora"] }""")
    )
    Get("/status") ~> sealRoute(publicStatusRoutes) ~> check {
      status.intValue should be(OK.intValue)
      val r = responseAs[SystemStatus]
      r.ok should be(true)
      r.systems("Agora").ok should be(true)
      r.systems("Agora").messages should be(None)
      r.systems.size should be(4)
    }
```

BUT, this causes failures because tests are also not cleaning up their expectations. Try mixing in the `BeforeAndAfterEach` trait, and add:
```
  override def beforeEach(): Unit = {
    agoraServer.reset()
  }
```
to ensure the server is clean.

src/test/scala/org/broadinstitute/dsde/firecloud/service/StatusServiceSpec.scala
```diff
@@ -0,0 +1,149 @@
...
+    request().withMethod("GET").withPath("/status")
+  ).error((new HttpError).withDropConnection(true))
+
+  "StatusService reports failure when Agora drops connection" in {
+    Get("/status") ~> sealRoute(publicStatusRoutes) ~> check {
+      status.intValue should be(InternalServerError.intValue)
+      val response = responseAs[SystemStatus]
+      response.ok should be(false)
+      response.systems("Agora").ok should be(false)
+      response.systems("Agora").messages should be(Some(List("Premature connection close (the server doesn't appear to support request pipelining)")))
```
davidangb: I don't think it's worth testing the exact error message here, since it's out of our control. Maybe just check that messages is some non-empty list?

davidangb (commented):
feedback inline, nothing critical, but a few changes ...

src/test/scala/org/broadinstitute/dsde/firecloud/service/StatusServiceSpec.scala
aednichols: Ooh, this is nice. Thanks.

src/test/scala/org/broadinstitute/dsde/firecloud/service/StatusServiceSpec.scala
aednichols: Is there a general way to pick out when this might be a problem? Going to declaration/finding usages in IntelliJ does not reveal the problem

davidangb (approved):
👍 wfm, thanks for sticking with this!

PR #428, opened 2017-04-24
gawb-1697: same incomplete library metadata

ahaessly:
lets you save incomplete metadata; cant publish a dataset with incomplete metadata saved

broken tests, fixing issue where no invalidDataset attribute

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -98,32 +98,36 @@ class LibraryService (protected val argUserInfo: UserInfo,
...
-                  // if discoverable by groups is not being changed, then skip it (i.e. don't delete from ws)
-                  Seq(publishedFlag, discoverableWSAttribute)
-
-              // this is technically vulnerable to a race condition in which the workspace attributes have changed
-              // between the time we retrieved them and here, where we update them.
-              val allOperations = generateAttributeOperations(workspaceResponse.workspace.attributes, userAttrs,
-                k => k.namespace == AttributeName.libraryNamespace && !skipAttributes.contains(k))
-              internalPatchWorkspaceAndRepublish(ns, name, allOperations, isPublished(workspaceResponse)) map (RequestComplete(_))
-            }
+            (true, Some(errorMessages.mkString("; ")))
```
rushtong: Nit ... this failure case could be a one-liner:
```
case Failure(ve: ValidationException) => (true, Some(getSchemaValidationMessages(ve).mkString("; ")))
```

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -146,17 +150,20 @@ class LibraryService (protected val argUserInfo: UserInfo,
   def setWorkspaceIsPublished(ns: String, name: String, value: Boolean): Future[PerRequestMessage] = {
     rawlsDAO.getWorkspace(ns, name) flatMap { workspaceResponse =>
       val pub = isPublished(workspaceResponse)
+      val invalidMetadata = workspaceResponse.workspace.attributes.get(
+        AttributeName(AttributeName.libraryNamespace, "invalidDataset")).orElse(Some(AttributeBoolean(false)))
```
rushtong: Nit ... a line break after the open paren is confusing here. Maybe this is more legible?
```
      val invalidMetadata = workspaceResponse.workspace.attributes.
        get(AttributeName(AttributeName.libraryNamespace, "invalidDataset")).
        orElse(Some(AttributeBoolean(false)))
```

rushtong (approved):
👍  pending the resolution of the question about returning validation issues when saving incomplete data.

ahaessly:
thanks greg - didn't even  see that comment about whether we should return the validation errors. i think that we don't need to return them from the updateAttributes method. but I was thinking about the api and realized that we don't check to make sure we have all the required fields in the setPublished call. So, when handling that, if the invalidData flag has not been set, we should validate the metadata and return the specific errors there. 
any other suggestions?

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -142,19 +148,38 @@ class LibraryService (protected val argUserInfo: UserInfo,
...
-          RequestComplete(ws)
+        val (invalid, errorMessage) = if (value)
+          // only need to check for valid attributes if we are actually publishing
+          isInvalidWorkspace(workspaceResponse.workspace)
+        else (false, None)
+        if (invalid) {
+          Future(RequestCompleteWithErrorReport(BadRequest, errorMessage.getOrElse(BadRequest.defaultMessage)))
+        } else {
+          rawlsDAO.updateLibraryAttributes(ns, name, updatePublishAttribute(value)) map { ws =>
+            if (value)
```
rushtong: Nitpick: The variable name `value` is pretty ambiguous. I prefer something more specific, like `toBePublishedFlag` or something like that, so when reading through the method, you don't have to go back and look at what it is meant to represent.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -142,19 +148,38 @@ class LibraryService (protected val argUserInfo: UserInfo,
...
+              publishDocument(ws)
+            else
+              removeDocument(ws)
+            RequestComplete(ws)
+          }
```
rushtong: Stylistic nitpick. There are four if conditions, three of them nested in this method. I would try to make this more scala-idiomatic. Haven't tested this, but something like this reads more cleanly to me. Definitely open to criticism on this.
```
      isPublished(workspaceResponse) match {
        case x if x == value => Future(RequestComplete(NoContent))
        case _ =>
          // only need to check for valid attributes if we are actually publishing
          val (invalid, errorMessage) = value match {
            case x if x => isInvalidWorkspace(workspaceResponse.workspace)
            case _ => (false, None)
          }
          invalid match {
            case x if x => Future(RequestCompleteWithErrorReport(BadRequest, errorMessage.getOrElse(BadRequest.defaultMessage)))
            case _ =>
              rawlsDAO.updateLibraryAttributes(ns, name, updatePublishAttribute(value)) map { ws =>
                if (value)
                  publishDocument(ws)
                else
                  removeDocument(ws)
                RequestComplete(ws)
              }
          }
      }
```

rushtong (approved):
👍 Minor style comments inline - feel free to address or ignore at your preference. 

I agree with your suggestions for NO to returning missing field info on `update` attributes and YES to returning missing field info on `publish` ... that makes sense to me.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
rushtong: I really wish scala used the java ternary, I absolutely disagree with the readability comments from scala folks on this. `value ?  publishDocument(ws) : removeDocument(ws)` reads so much more cleanly to me than the `if`-version

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockRawlsDAO.scala
```diff
@@ -70,6 +70,32 @@ class MockRawlsDAO  extends RawlsDAO {
...
+      AttributeName("default", "e") -> AttributeString("v1"),
+      AttributeName("default", "f") -> AttributeValueList(Seq(
+        AttributeString("v6"),
+        AttributeNumber(999),
+        AttributeBoolean(true)
+      ))),
+    Map(), //acls
+    Map(), //realm acls,
+    false
+  )
```
davidangb: optional feedback: since `Workspace` is a case class, you can use `copy()` ... something like:
```
  private val unpublishedRawlsWorkspaceLibraryValid = publishedRawlsWorkspaceWithAttributes.copy(
    attributes = publishedRawlsWorkspaceWithAttributes.attributes +
      (AttributeName.withLibraryNS("invalidDataset") -> AttributeBoolean(false)) - 
      AttributeName.withLibraryNS("published"))
```

also, I know you're just copying/pasting this code, but I prefer `AttributeName.withDefaultNS` / `AttributeName.withLibraryNS` instead of hardcoding "library" and "default".

davidangb (changes requested):
One code comment inline. I don't like the usage of a persisted `library: invalidDataset` attribute; I'd prefer to see us re-validate every time - at least until we prove that re-validating every time is a burden.

A cached `invalidDataset` flag is 1) an extra attribute we don't really need, but more importantly 2) will be immediately out of date the next time we change the schema - we'd need a process to re-validate and update each dataset upon schema change. If we re-validate every time, we won't have that problem.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -58,6 +58,7 @@ class LibraryService (protected val argUserInfo: UserInfo,
   lazy val log = LoggerFactory.getLogger(getClass)
 
   implicit val userInfo = argUserInfo
+  implicit val impAttributeFormat: AttributeFormat = new AttributeFormat with PlainArrayAttributeListSerializer
```
rushtong: There are comments for this line that did not move with the implicit.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
rushtong: Still a nitpick :-)

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
rushtong: This probably does't apply any longer

rushtong (approved):
👍  Having trouble testing through the ui, but code looks right. Minor comments inline.

davidangb (approved):

PR #433, opened 2017-04-30
GAWB-1811: enable gzip by default for calls to external services

davidangb:
if a service under orchestration (e.g. rawls) supports gzip, orch can take advantage of it. This PR requests gzip by default from the external services. This enables gzip for the workspace listing - as well as most other calls.

I could have satisfied GAWB-1811 by enabling gzip for the workspaces list only, but it felt good to enable it by default everywhere - everything still seems to work!

## Debugging/testing notes
It's pretty difficult to verify that gzip is working between orch and another service. I verified it in two ways:

1. I modified `logback.xml` and set the log level for `org.broadinstitute.dsde` to debug. By tweaking the code, I toggled back and forth between enabling/disabling gzip via the `requestCompression` argument. I saw that, for the same request, the response's `Content-Length` header value would change - in my test, gzipped had a value of 906, and uncompressed had a value of 2847. I double-checked these values by calling rawls directly from curl.

2. I again modified the code and added the following in `HttpClient.scala`:
```
trait LogResponseEncoding extends spray.httpx.ResponseTransformation {
  def logResponseEncoding(log: Logger): ResponseTransformer = { response =>
    log.warn(s"response.encoding is: ${response.encoding}")
    response
  }
}
```
and then tweaked the pipeline to include:
```
sendReceive ~> logResponseEncoding(log) ~> decode(Gzip)
```
And made some requests through swagger. I saw that it logged as expected:
```
response.encoding is: gzip
```

Note that this is necessary, instead of relying on our existing debug logging of the response, because the `decode(Gzip)` directive actually REMOVES the content-encoding header! See https://github.com/spray/spray/blob/master/spray-httpx/src/main/scala/spray/httpx/encoding/Decoder.scala#L27 for that.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/FireCloudDirectives.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/FireCloudDirectives.scala
@@ -25,14 +25,14 @@ trait FireCloudDirectives extends spray.routing.Directives with PerRequestCreato
   def respondWithJSON = respondWithMediaType(`application/json`)
 
   def passthrough(unencodedPath: String, methods: HttpMethod*) = methods map { inMethod =>
-    generateExternalHttpPerRequestForMethod(requestCompression = false, unencodedPath, inMethod)
+    generateExternalHttpPerRequestForMethod(requestCompression = true, unencodedPath, inMethod)
   } reduce (_ ~ _)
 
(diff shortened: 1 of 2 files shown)


rushtong (approved):
👍  Verified that with these changes, workspace list is indeed gzipped. I saw a 10-fold decrease in size.

PR #440, opened 2017-05-08
Polish Swagger for /status endpoint [GAWB-771]

aednichols:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -4115,10 +4115,30 @@ definitions:
       ok:
         type: boolean
         description: whether any system(s) need attention
-
+        default: false
       systems:
         type: object
         description: Map[String, SubsystemStatus]
+        default: {
+                   "ok": false,
+                   "systems": {
+                     "Rawls": {
+                       "ok": true
+                     },
+                     "Thurloe": {
+                       "ok": false,
+                       "messages": [
+                         "Thurloe misbehavior message"
+                       ]
+                     },
+                     "Agora": {
+                       "ok": true
+                     },
+                     "Search": {
+                       "ok": true
+                     }
+                   }
+                 }
 
   SubsystemStatus:
     description: specifics for each subsystem
@@ -4425,6 +4445,8 @@ definitions:
       tag:
         type: string
         description: tag string
+        default: "a tag"
       count:
-        type: number
         description: number of usages of the tag across Firecloud
+        default: 5
+        type: number
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -4447,6 +4447,6 @@ definitions:
         description: tag string
         default: "a tag"
       count:
+        type: number
         description: number of usages of the tag across Firecloud
         default: 5
-        type: number
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -4115,11 +4115,11 @@ definitions:
       ok:
         type: boolean
         description: whether any system(s) need attention
-        default: false
+        example: false
       systems:
         type: object
         description: Map[String, SubsystemStatus]
-        default: {
+        example: {
                    "ok": false,
                    "systems": {
                      "Rawls": {
@@ -4445,8 +4445,8 @@ definitions:
       tag:
         type: string
         description: tag string
-        default: "a tag"
+        example: "a tag"
       count:
         type: number
         description: number of usages of the tag across Firecloud
-        default: 5
+        example: 5
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -4119,26 +4119,17 @@ definitions:
       systems:
         type: object
         description: Map[String, SubsystemStatus]
-        example: {
-                   "ok": false,
-                   "systems": {
-                     "Rawls": {
-                       "ok": true
-                     },
-                     "Thurloe": {
-                       "ok": false,
-                       "messages": [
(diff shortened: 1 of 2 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -4115,10 +4115,30 @@ definitions:
       ok:
         type: boolean
         description: whether any system(s) need attention
-
+        default: false
```
rtitle: I'm new to swagger, but I'm wondering if these is the right use for the default keyword. From [swagger docs](http://swagger.io/unlocking-the-spec-the-default-keyword/):

> Common mistakes when using the default value
>
>There are two possible discrepancies when using the default value:
>
> Defining a default value for a container and marking that container as required is somewhat of a contradiction. If the container is required, it will always be sent and the default value will never take effect.
>
> The second case is using the default as a way to mention a sample value for its container. That is not the proper use of the field and can lead to unexpected behavior in some of the Swagger tools. Some containers offer alternative ways to setting a sample value.

Since this is a required field, it may not need a default?

src/main/resources/swagger/api-docs.yaml
```diff
@@ -4115,10 +4115,30 @@ definitions:
...
       systems:
         type: object
         description: Map[String, SubsystemStatus]
+        default: {
```
rtitle: See above. Also I'm wondering if this is correct YAML syntax? For a string spanning several lines could use
```
default: |
  This
  is
  a 
  long 
  string
```

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
```diff
@@ -71,7 +71,9 @@ class FireCloudServiceActor extends HttpServiceActor with FireCloudDirectives
     log.debug(requestContext.request.toString)
     route(requestContext)
   }
-  val appendTimestamp = mapHttpResponse { response =>
+
+  // So we have the time when users send us error screenshots
```
rtitle, line 75: +1, that's a good reason to include the timestamp.

rtitle:
Looks good, a couple swagger comments, could just be my swagger inexperience talking.

src/main/resources/swagger/api-docs.yaml
rushtong: The syntax does work - it's what is displayed as an example value. I'm pretty sure there's a yaml way to do this instead of json though ... looking into it.

rushtong: This also works and is pure yaml:
```
    example:
      ok: false
      systems:
        - Agora:
            ok: true
        - Rawls:
            ok: true
        - Search:
            ok: true
        - Thurloe:
            ok: false
            messages:
              - "Thurloe misbehavior message"
```

src/main/resources/swagger/api-docs.yaml
rushtong: I don't think you need `default` here at all - it doesn't add any value to the swagger UI.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -4425,6 +4445,8 @@ definitions:
       tag:
         type: string
         description: tag string
+        default: "a tag"
```
rushtong: I think `example` is more appropriate here

src/main/resources/swagger/api-docs.yaml
```diff
@@ -4425,6 +4445,8 @@ definitions:
...
       count:
         type: number
         description: number of usages of the tag across Firecloud
+        default: 5
```
rushtong: Same as above ... I think `example` is more appropriate here

rushtong (changes requested):
I know you didn't ask, but I reviewed anyway :-) 
Please see comments inline.

src/main/resources/swagger/api-docs.yaml
aednichols: The idea is that because the `systems` structure includes a failing system, the `ok` boolean should be set to `false`.

I agree that `example` is correct versus `default`, we seem to make that mistake a lot actually.

dvoet:
if this is a hotfix, this PR needs to be applied to the RC_Pampangan_hotfix branch

src/main/resources/swagger/api-docs.yaml
aednichols: We do this in a couple places and it does work, but those aren't great reasons.

It does seem helpful to me that the response example in the codebase looks exactly like the live version, but I'm happy to switch it over if people feel strongly.

aednichols:
@dvoet not hotfix-level priority, just doc clarification. Updated the title to be more helpful in that respect

src/main/resources/swagger/api-docs.yaml
rushtong: @aednichols - Thank you! My thoughts are that we should try to stick to yaml for the yaml file. It's cleaner, more concise, and consistent with the rest of the file. Previously, we've had even more json in the file and have been weeding that out over time.

rushtong (approved):
👍

src/main/resources/swagger/api-docs.yaml
aednichols: Makes sense, changed over.

Further learning: YAML is actually a superset of JSON, with the braces and brackets optionally (and typically) replaced with indentation. YAML's entire purpose in life is to be more human-readable.

rtitle:
👍

davidangb (approved):
looks fine to me as TL; did not test. 👍

PR #444, opened 2017-05-10
GAWB-1863: upgrade to ES 5 (compatibility)

davidangb:
Requires fc-develop config changes:
* broadinstitute/firecloud-develop#440 (local)
* broadinstitute/firecloud-develop#441 (dev)
* broadinstitute/firecloud-develop#442 (FiaB)

This PR is aimed at making our codebase *compatible* with ES5. There is still work left to be done to optimize our usage of ES5, such as switching our mappings to use `text`/`keyword` instead of `string`, and checking our queries for optimization. I wanted to keep PRs as small as possible, so this is the first and there will be at least one more.

Goals of this PR:
* codebase compiles after upgrading to ES5 client library
* unit and integration tests pass
* end user functionality appears to be correct

Note that I am intentionally using ES client version 5.3.0 here, though our cluster is 5.4. The 5.3 client is compatible, and there appears to be a problem in 5.4's published dependencies that I haven't figured out yet - simply going to 5.4 causes a whole bunch of dependencies to be missed, causing compile/runtime errors.

This PR has a lot of commits and I've tried to keep them all very small for easy consumption.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off


project/Dependencies.scala
```diff
@@ -9,36 +9,45 @@ object Dependencies {
...
+    "org.webjars"                    % "swagger-ui"          % "2.2.5",
+    "com.jason-goodwin"             %% "authentikat-jwt"     % "0.4.1",
+    "com.sun.mail"                   % "javax.mail"          % "1.5.6",
+    "com.univocity"                  % "univocity-parsers"   % "2.4.1",
+    "org.ocpsoft.prettytime"         % "prettytime"          % "4.0.1.Final",
+    "org.everit.json"                % "org.everit.json.schema" % "1.4.1",
+
+    "org.specs2"                    %% "specs2-core"         % "3.7"     % "test",
+    "org.scalatest"                 %% "scalatest"           % "2.2.6"   % "test",
+    "org.mock-server"                % "mockserver-netty"    % "3.10.2"  % "test"
```
davidangb, line 51: this is mostly whitespace/formatting changes - try the `?w=1` github trick to ignore whitespace!

project/Merging.scala
```diff
@@ -2,7 +2,7 @@ import sbtassembly.{MergeStrategy, PathList}
 
 object Merging {
   def customMergeStrategy(oldStrategy: (String) => MergeStrategy):(String => MergeStrategy) = {
-    case PathList("org", "joda", "time", "base", "BaseDateTime.class") => MergeStrategy.first
+    case x if x.endsWith("io.netty.versions.properties") => MergeStrategy.discard
```
davidangb, line 5: in ES 2.4, we had to ignore joda classes in `sbt assembly` because ES bundled them. Now, we have to deal with some netty conflicts.

project/Dependencies.scala
rushtong: Nice trick, but it also removes comments :-(

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
```diff
@@ -68,13 +74,14 @@ trait ElasticSearchDAOQuerySupport extends ElasticSearchDAOSupport {
       case Some(searchTerm) if searchTerm.trim == "" => matchAllQuery
       case Some(searchTerm) =>
         val fieldSearch = if (phrase) {
-          matchPhraseQuery(searchField, searchTerm).minimumShouldMatch("2<67%")
+          matchPhraseQuery(searchField, searchTerm)
+            //.minimumShouldMatch("2<67%")
```
rushtong, line 78: Looking at `matchPhraseQuery` and `matchQuery` below, it seems like they do not behave the same way with respect to `minimumShouldMatch` ? Or, does `matchQuery` also not need `minimumShouldMatch`?

rushtong (approved):
👍  Looks good, works great! Minor question inline, but can be part of the optimization component if it really is a legit question.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
davidangb: Great question. All our test cases are passing - but I also have this on my list for follow-up, in the optimization phase once we get through compatibility.

PR #447, opened 2017-05-12
GAWB-1917: Support multiple NIH whitelists in preparation of TARGET

MatthewBemis:
Plug a new whitelist into config everything should magically work:

```
nih {
        whitelistBucket = "firecloud-whitelist-dev"
        whitelists = {
                       "TARGET" {
                         "fileName":"target-whitelist.txt",
                         "rawlsGroup":"TARGET-dbGaP-Authorized"
                       },
                       "TCGA" {
                         "fileName":"tcga-whitelist.txt",
                         "rawlsGroup":"TCGA-dbGaP-Authorized"
                       }
                     }
}
```

UI: https://github.com/broadinstitute/firecloud-ui/pull/790
ncbiaccess: https://github.com/broadinstitute/ncbiaccess/pull/18

Note that this also fixes the "Failed to link NIH account" when a user links but isn't on the dbGaP whitelist.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
@@ -103,7 +103,7 @@ trait NihService extends LazyLogging {
     val whitelistSyncResults = Future.traverse(nihWhitelists)(syncNihWhitelistAllUsers)
 
     whitelistSyncResults map { response =>
-      if(response.forall(x => x)) RequestComplete(NoContent)
+      if(response.forall(isSuccess => isSuccess)) RequestComplete(NoContent)
       else RequestCompleteWithErrorReport(InternalServerError, "Error synchronizing NIH whitelist")
     }
   }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
@@ -6,7 +6,7 @@ import com.typesafe.scalalogging.slf4j.LazyLogging
 import org.broadinstitute.dsde.firecloud.{Application, FireCloudConfig, FireCloudExceptionWithErrorReport}
 import org.broadinstitute.dsde.firecloud.dataaccess.{GoogleServicesDAO, RawlsDAO, ThurloeDAO}
 import org.broadinstitute.dsde.firecloud.model._
-import org.broadinstitute.dsde.firecloud.service.NihService.{GetNihStatus, SyncWhitelist, UpdateNIHLinkAndSyncSelf}
+import org.broadinstitute.dsde.firecloud.service.NihService.{GetNihStatus, SyncWhitelist, UpdateNihLinkAndSyncSelf}
 import org.broadinstitute.dsde.firecloud.service.PerRequest.{PerRequestMessage, RequestComplete}
 import org.broadinstitute.dsde.firecloud.utils.DateUtils
 import org.broadinstitute.dsde.rawls.model.ErrorReport
@@ -21,22 +21,24 @@ import scala.util.{Failure, Success, Try}
 
 case class NihStatus(
   linkedNihUsername: Option[String] = None,
-  isDbgapAuthorized: Option[Boolean] = None,
-  linkExpireTime: Option[Long] = None,
-  descriptionUntilExpires: Option[String] = None
-)
+  whitelistAuthorization: Set[NihWhitelistAccess],
+  linkExpireTime: Option[Long] = None)
+
+case class NihWhitelist(
+  name: String,
+  groupToSync: String,
+  whitelistFile: String)
+
+case class NihWhitelistAccess(name: String, authorized: Boolean)
 
 object NihStatus {
-  implicit val impNihStatus = jsonFormat4(NihStatus.apply)
+  implicit val impNihWhitelistAccess = jsonFormat2(NihWhitelistAccess)
+  implicit val impNihStatus = jsonFormat3(NihStatus.apply)
 
-  def apply(profile: Profile): NihStatus = {
-    apply(profile, profile.isDbgapAuthorized)
-  }
-
-  def apply(profile: Profile, isDbGapAuthorized: Option[Boolean]): NihStatus = {
+  def apply(profile: Profile, whitelistAccess: Set[NihWhitelistAccess]): NihStatus = {
     new NihStatus(
       profile.linkedNihUsername,
-      isDbGapAuthorized,
+      whitelistAccess,
       profile.linkExpireTime
     )
   }
@@ -45,7 +47,7 @@ object NihStatus {
 object NihService {
   sealed trait ServiceMessage
   case class GetNihStatus(userInfo: UserInfo) extends ServiceMessage
-  case class UpdateNIHLinkAndSyncSelf(userInfo: UserInfo, nihLink: NIHLink) extends ServiceMessage
+  case class UpdateNihLinkAndSyncSelf(userInfo: UserInfo, nihLink: NihLink) extends ServiceMessage
   case object SyncWhitelist extends ServiceMessage
 
   def props(service: () => NihServiceActor): Props = {
@@ -61,8 +63,8 @@ class NihServiceActor(val rawlsDao: RawlsDAO, val thurloeDao: ThurloeDAO, val go
 
   override def receive = {
     case GetNihStatus(userInfo: UserInfo) => getNihStatus(userInfo) pipeTo sender
-    case UpdateNIHLinkAndSyncSelf(userInfo: UserInfo, nihLink: NIHLink) => updateNihLinkAndSyncSelf(userInfo: UserInfo, nihLink: NIHLink) pipeTo sender
-    case SyncWhitelist => syncWhitelistFull() pipeTo sender
+    case UpdateNihLinkAndSyncSelf(userInfo: UserInfo, nihLink: NihLink) => updateNihLinkAndSyncSelf(userInfo: UserInfo, nihLink: NihLink) pipeTo sender
+    case SyncWhitelist => syncNihWhitelistsFull pipeTo sender
   }
 }
 
@@ -72,13 +74,15 @@ trait NihService extends LazyLogging {
   val thurloeDao: ThurloeDAO
   val googleDao: GoogleServicesDAO
 
+  val nihWhitelists: Set[NihWhitelist] = FireCloudConfig.Nih.whitelists
+
   def getNihStatus(userInfo: UserInfo): Future[PerRequestMessage] = {
     thurloeDao.getProfile(userInfo) flatMap {
       case Some(profile) =>
         profile.linkedNihUsername match {
           case Some(_) =>
-            rawlsDao.isDbGapAuthorized(userInfo) map { isDbGapAuthorized =>
-              RequestComplete(NihStatus(profile.copy(linkExpireTime = Option(profile.linkExpireTime.getOrElse(0L))), Some(isDbGapAuthorized)))
+            Future.sequence(nihWhitelists.map(whitelistDef => rawlsDao.isGroupMember(userInfo, whitelistDef.groupToSync).map(isMember => NihWhitelistAccess(whitelistDef.name, isMember)))).map { whitelistMembership =>
+              RequestComplete(NihStatus(profile.copy(linkExpireTime = Option(profile.linkExpireTime.getOrElse(0L))), whitelistMembership))
             }
           case None => Future.successful(RequestComplete(NotFound))
         }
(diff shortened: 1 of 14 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
```diff
@@ -72,78 +74,87 @@ trait NihService extends LazyLogging {
...
+  val nihWhitelists: Set[NihWhitelist] = FireCloudConfig.Nih.whitelists
+
   def getNihStatus(userInfo: UserInfo): Future[PerRequestMessage] = {
     thurloeDao.getProfile(userInfo) flatMap {
       case Some(profile) =>
         profile.linkedNihUsername match {
           case Some(_) =>
-            rawlsDao.isDbGapAuthorized(userInfo) map { isDbGapAuthorized =>
-              RequestComplete(NihStatus(profile.copy(linkExpireTime = Option(profile.linkExpireTime.getOrElse(0L))), Some(isDbGapAuthorized)))
+            Future.sequence(nihWhitelists.map(whitelistDef => rawlsDao.isGroupMember(userInfo, whitelistDef.groupToSync).map(isMember => NihWhitelistStatus(whitelistDef.name, isMember)))).map { whitelistMembership =>
```
rtitle: Minor, `Future.traverse` might be a little cleaner:
```
Future.traverse(nihWhitelists) { whitelistDef => 
  rawlsDao.isGroupMember(userInfo, whitelistDef.groupToSync).map(isMember => NihWhitelistStatus(whitelistDef.name, isMember)
}.map { whitelistMembership =>
  RequestComplete(NihStatus(profile.copy(linkExpireTime = Option(profile.linkExpireTime.getOrElse(0L))), whitelistMembership))
}
```

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
```diff
@@ -72,78 +74,87 @@ trait NihService extends LazyLogging {
...
-    syncWhiteListResult flatMap { _ => profileResponse } recover { case t => RequestCompleteWithErrorReport(InternalServerError, "Error updating NIH link", t) }
+    usersList.getLines().toSet
   }
 
-  private def getWhitelist(): Set[String] = {
-    val (bucket, file) = (FireCloudConfig.Nih.whitelistBucket, FireCloudConfig.Nih.whitelistFile)
-    val whitelist = Source.fromInputStream(googleDao.getBucketObjectAsInputStream(bucket, file))
+  // This syncs all of the whitelists in full
+  def syncNihWhitelistsFull: Future[PerRequestMessage] = {
+    val whitelistSyncResults = Future.sequence(nihWhitelists.map(syncNihWhitelistAllUsers))
```
rtitle: Same as `Future.traverse(nihWhitelists)(syncNihWhitelistAllUsers)`

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
```diff
@@ -72,78 +74,87 @@ trait NihService extends LazyLogging {
...
+
+    thurloeDao.saveKeyValues(userInfo, profilePropertyMap)
+  }
+
+  def updateNihLinkAndSyncSelf(userInfo: UserInfo, nihLink: NihLink): Future[PerRequestMessage] = {
+    //first, link the NIH account. it doesn't matter if they're on a whitelist or not
+    //next, sync the appropriate rawls groups based on which whitelists the user is a member of
+    val linkResult = for {
+      linkNihResult <- linkNihAccount(userInfo, nihLink)
+      _ <- Future.sequence(nihWhitelists.map(syncNihWhitelistForUser(userInfo.getUniqueId, nihLink.linkedNihUsername, _)))
```
rtitle: Another candidate for `traverse`. 

Also, do you really want to fire and forget the second operation?

MatthewBemis: I think of syncing the various whitelists as side effects of NIH-linking. The result of the overall operation doesn't depend on what happens with the whitelists- so I think fire and forget is OK here.

rtitle: OK. Another way to do side effects with futures is `andThen`:
https://www.scala-lang.org/api/current/scala/concurrent/Future.html#andThen[U](pf:PartialFunction[scala.util.Try[T],U])(implicitexecutor:scala.concurrent.ExecutionContext):scala.concurrent.Future[T]

For example:
```
val linkResult = linkNihAccount(userInfo, nihLink).andThen { case _ =>
  nihWhitelists.foreach(syncNihWhitelistForUser(userInfo.getUniqueId, nihLink.linkedNihUsername, _))
}
```
You avoid needing sequence/map in that case.

src/test/scala/org/broadinstitute/dsde/firecloud/service/NihApiServiceSpec.scala
```diff
@@ -89,7 +88,7 @@ class NihApiServiceSpec extends ApiServiceSpec {
     Post("/sync_whitelist") ~> sealRoute(services.syncRoute) ~> check {
       status should equal(NoContent)
     }
-    assert(services.rawlsDao.groups(dbGapAuthorizedUsersGroupName).equals(Set(services.thurloeDao.linkedUserId)))
+    assert(services.rawlsDao.groups(dbGapAuthorizedUsersGroupName).contains(services.thurloeDao.linkedUserId))
```
dvoet: these tests check to see if the user is in the dbGapAuthorizedUsersGroupName group, what about target?

dvoet (changes requested):
functionality looks ok to me but I would like more tests

src/test/scala/org/broadinstitute/dsde/firecloud/service/NihApiServiceSpec.scala
```diff
@@ -35,63 +46,85 @@ class NihApiServiceSpec extends ApiServiceSpec {
...
      1 user that is linked and has active TCGA access. they should remain in the dbGapAuthorizedUsers group
    */
-  it should "return NoContent and properly sync the whitelist for users of different link statuses" in withDefaultApiServices { services =>
+  it should "return NoContent and properly sync the whitelist for users of different link statuses across whitelists" in withDefaultApiServices { services =>
     Post("/sync_whitelist") ~> sealRoute(services.syncRoute) ~> check {
       status should equal(NoContent)
     }
-    assert(services.rawlsDao.groups(dbGapAuthorizedUsersGroupName).equals(Set(services.thurloeDao.linkedUserId)))
+    assert(services.rawlsDao.groups(tcgaDbGaPAuthorized).equals(Set(services.thurloeDao.TCGA_AND_TARGET_LINKED, services.thurloeDao.TCGA_LINKED)))
+    assert(services.rawlsDao.groups(targetDbGaPAuthorized).equals(Set(services.thurloeDao.TCGA_AND_TARGET_LINKED, services.thurloeDao.TARGET_LINKED)))
```
jmthibault79: This test fails for me locally

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
```diff
@@ -72,78 +74,88 @@ trait NihService extends LazyLogging {
...
     thurloeDao.getProfile(userInfo) flatMap {
       case Some(profile) =>
         profile.linkedNihUsername match {
           case Some(_) =>
-            rawlsDao.isDbGapAuthorized(userInfo) map { isDbGapAuthorized =>
-              RequestComplete(NihStatus(profile.copy(linkExpireTime = Option(profile.linkExpireTime.getOrElse(0L))), Some(isDbGapAuthorized)))
+            Future.traverse(nihWhitelists) { whitelistDef =>
+              rawlsDao.isGroupMember(userInfo, whitelistDef.groupToSync).map(isMember => NihWhitelistStatus(whitelistDef.name, isMember))
+            }.map { whitelistMembership =>
+              RequestComplete(NihStatus(profile.copy(linkExpireTime = Option(profile.linkExpireTime.getOrElse(0L))), whitelistMembership))
```
jmthibault79: Not new in this PR but this is a code smell: why are we turning `None` into `Some(0)` here?

MatthewBemis: This is the weird situation where a user has a linked NIH account but we never saved an expiration date for them. We set it to the beginning of time to force them to re-link.

It should never actually happen

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
```diff
@@ -72,78 +74,88 @@ trait NihService extends LazyLogging {
...
-
-    syncWhiteListResult flatMap { _ => profileResponse } recover { case t => RequestCompleteWithErrorReport(InternalServerError, "Error updating NIH link", t) }
+    usersList.getLines().toSet
   }
 
-  private def getWhitelist(): Set[String] = {
-    val (bucket, file) = (FireCloudConfig.Nih.whitelistBucket, FireCloudConfig.Nih.whitelistFile)
-    val whitelist = Source.fromInputStream(googleDao.getBucketObjectAsInputStream(bucket, file))
+  // This syncs all of the whitelists in full
+  def syncNihWhitelistsFull: Future[PerRequestMessage] = {
```
jmthibault79: I prefer the `AllUsers` name.  It's not clear what a non-full sync would be, but `AllUsers` is an obvious contrast to a single user sync.

src/test/scala/org/broadinstitute/dsde/firecloud/service/NihApiServiceSpec.scala
```diff
@@ -35,63 +46,85 @@ class NihApiServiceSpec extends ApiServiceSpec {
...
     }
-    assert(services.rawlsDao.groups(dbGapAuthorizedUsersGroupName).contains(services.thurloeDao.linkedButExpiredUserId))
+    assert(services.rawlsDao.groups(tcgaDbGaPAuthorized).contains(services.thurloeDao.TCGA_AND_TARGET_LINKED_EXPIRED))
+    assert(services.rawlsDao.groups(targetDbGaPAuthorized).contains(services.thurloeDao.TCGA_AND_TARGET_LINKED_EXPIRED))
   }
 
   /* Test scenario:
      1 user that is linked but their TCGA access has expired. they should be removed
      1 user that is linked but they have no expiration date stored in their profile. they should be removed
      1 user that is linked and has active TCGA access. they should remain in the dbGapAuthorizedUsers group
```
jmthibault79: each group

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockRawlsDAO.scala
```diff
@@ -141,6 +144,7 @@ class MockRawlsDAO  extends RawlsDAO {
   override def adminAddMemberToGroup(groupName: String, memberList: RawlsGroupMemberList): Future[Boolean] = {
     val userEmailsToAdd = memberList.userSubjectIds.getOrElse(Seq[String]()).toSet
     val groupWithNewMembers = (groupName -> ((groups(groupName).filterNot(userEmailsToAdd.contains)) ++ userEmailsToAdd))
+    println(s"${groupName} + ${memberList}")
```
jmthibault79: Do we want this?

src/test/scala/org/broadinstitute/dsde/firecloud/mock/MockGoogleServicesDAO.scala
```diff
@@ -14,7 +14,13 @@ import scala.concurrent.{ExecutionContext, Future}
 
 class MockGoogleServicesDAO extends GoogleServicesDAO {
   override def getAdminUserAccessToken: String = ""
-  override def getBucketObjectAsInputStream(bucketName: String, objectKey: String): InputStream = new ByteArrayInputStream("firecloud-dev".getBytes("UTF-8"))
+  override def getBucketObjectAsInputStream(bucketName: String, objectKey: String): InputStream = {
```
jmthibault79, line 17: This can throw an `IOException` so I would like a key that will cause one, and tests for how we handle it

src/test/scala/org/broadinstitute/dsde/firecloud/service/NihApiServiceSpec.scala
```diff
@@ -35,63 +46,85 @@ class NihApiServiceSpec extends ApiServiceSpec {
...
+    Post("/nih/callback", JWTWrapper("bad-token")) ~> dummyUserIdHeaders(services.thurloeDao.TCGA_AND_TARGET_UNLINKED) ~> sealRoute(services.nihRoutes) ~> check {
       status should equal(BadRequest)
     }
-    assert(!services.rawlsDao.groups(dbGapAuthorizedUsersGroupName).contains(services.thurloeDao.normalUserId))
+    assert(!services.rawlsDao.groups(tcgaDbGaPAuthorized).contains(services.thurloeDao.TCGA_AND_TARGET_UNLINKED))
+    assert(!services.rawlsDao.groups(targetDbGaPAuthorized).contains(services.thurloeDao.TCGA_AND_TARGET_UNLINKED))
+  }
+
+  it should "link and sync when user is on TCGA whitelist but not TARGET" in withDefaultApiServices { services =>
+    assert(!services.rawlsDao.groups(targetDbGaPAuthorized).contains(services.thurloeDao.TCGA_UNLINKED))
```
jmthibault79: Please add something like `val toLink = services.thurloeDao.TCGA_UNLINKED` to make the assertions read more naturally.

jmthibault79: Similar comments for the others.

jmthibault79 (changes requested):
Some in-line comments.  I would also like to see tests for how we handle `syncNihWhitelistForUser` failures.

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
jmthibault79: OK but `None` would express that just as well, right?  It seems to me that this logic should live in the consumer of this message instead of here.  If it really does need to work this way, then `NihStatus.linkExpireTime` should not be an `Option`.

MatthewBemis: Yes, I agree that this is a thing that would be better handled in the UI

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
```diff
@@ -72,78 +76,97 @@ trait NihService extends LazyLogging {
...
-      else RequestCompleteWithErrorReport(InternalServerError, "Error updating NIH link.")
+  private def downloadNihWhitelist(whitelist: NihWhitelist): Set[String] = {
+    val usersList = try {
+      Source.fromInputStream(googleDao.getBucketObjectAsInputStream(FireCloudConfig.Nih.whitelistBucket, whitelist.fileName))
+    } catch {
+      case ioe: IOException => {
+        logger.error(s"Unable to download ${whitelist.name} whitelist from Google Bucket (${FireCloudConfig.Nih.whitelistBucket}/${whitelist.fileName})")
+        //In the rare case where we cannot read the whitelist from the bucket, this will act as a fail-safe and
+        //act as if the whitelist is empty
+        Source.fromInputStream(new ByteArrayInputStream("".getBytes("UTF-8")))
```
MatthewBemis: I now ask: is this the right behavior? It seems aggressive but maybe that's the right approach given what we're dealing with

src/test/scala/org/broadinstitute/dsde/firecloud/service/NihApiServiceSpec.scala
```diff
@@ -35,63 +48,101 @@ class NihApiServiceSpec extends ApiServiceSpec {
...
+     1 user that is linked and has active TARGET access. they should remain in the TARGET group
+     1 user that is linked and has active TARGET & TCGA access. they should remain in the TARGET and TCGA groups
    */
-  it should "return NoContent and properly sync the whitelist for users of different link statuses" in withDefaultApiServices { services =>
+  it should "return NoContent and properly sync the whitelist for users of different link statuses across whitelists" in withDefaultApiServices { services =>
     Post("/sync_whitelist") ~> sealRoute(services.syncRoute) ~> check {
       status should equal(NoContent)
     }
-    assert(services.rawlsDao.groups(dbGapAuthorizedUsersGroupName).equals(Set(services.thurloeDao.linkedUserId)))
+    services.rawlsDao.groups(tcgaDbGaPAuthorized) should contain theSameElementsAs Set(services.thurloeDao.TCGA_AND_TARGET_LINKED, services.thurloeDao.TCGA_LINKED)
```
jmthibault79: `assertSameElements` please, even though it doesn't make a difference in this case

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
jmthibault79: Good question.  I recall debating this when we first implemented this, but not the details.  Do you have information on what NCI agreed to @abaumann or @dvoet ?

MatthewBemis: I talked this over with Doug. We will leave it the way it was (essentially letting the exception bubble up) because the sync will retry in 6 or 8 hours (forget which interval). We are only required to have this stay in sync every 24 hours, so we should get a Loggly alert before then and be able to diagnose what happens and handle it. But it seems like setting the whitelist to blank is overly aggressive because we still have 3-4 more retries and 24 hours before it becomes an actual problem. And from what I know, this has yet to happen

dvoet (approved):

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
jmthibault79: That sounds good to me.  Thanks!

src/test/scala/org/broadinstitute/dsde/firecloud/service/ApiServiceSpec.scala
```diff
@@ -38,4 +38,10 @@ trait ApiServiceSpec extends FlatSpec with Matchers with HttpService with Scalat
     )_
 
   }
+
+  // lifted from rawls. prefer this to using theSameElementsAs directly, because its functionality depends on whitespace
+  def assertSameElements[T](expected: TraversableOnce[T], actual: TraversableOnce[T]): Unit = {
```
jmthibault79, line 43: whoops forgot this wasn't Rawls.  Thanks!

jmthibault79 (approved):

PR #448, opened 2017-05-18
[NOT YET] Increase default chunk size so large downloads are not blocked.

rushtong:
Addresses https://broadinstitute.atlassian.net/browse/GAWB-1998

Large downloads from other services to orchestration hit the default max size of 1M. Increasing this to 5M was enough to handle a 20K entity download.

**Testing:**
Generate a large entity set and upload to a workspace. Here's a groovy script that worked for me. More than 2500 rows starts to generate timeout errors which is not the focus of this PR.
```
import static java.util.UUID.randomUUID

String entity = "participant"
Integer num_columns = 150
Integer num_rows = 2500
Integer num_files = 10

static String uuid() { randomUUID() as String }

static String header(String entity, Integer num_columns) {
    "entity:${entity}_id" +
            (1..num_columns).collect { "\tprop_" + it }.join("")
}

static String rows(Integer num_rows, Integer num_columns) {
    (0..num_rows).collect {
        (0..num_columns).collect { uuid() }.join("\t")
    }.join("\n")

}

(1..num_files).each {
    File file = new File(entity + "s_" + it + ".tsv")
    file.write header(entity, num_columns) + "\n" + rows(num_rows, num_columns)
}
```

---

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -37,3 +37,11 @@ userprofile {
 googlecloud {
   priceListUrl: "https://cloudpricingcalculator.appspot.com/static/data/pricelist.json"
 }
+
+spray {
+  can {
+    client {
+      response-chunk-aggregation-limit = 5m
+    }
+  }
+}


rtitle:
Aha, nice find.
http://spray.io/documentation/1.2.4/spray-can/configuration/
```
# If this setting is non-zero the HTTP client connections automatically
# aggregate incoming response chunks into full HttpResponses before
# dispatching them to the application.
# If the size of the aggregated response surpasses the specified limit the
# HTTP client connection is closed and an error returned.
# Set to zero to disable automatic request chunk aggregation and have
# ChunkedResponseStart, MessageChunk and ChunkedMessageEnd messages be
# dispatched to the application.
response-chunk-aggregation-limit = 1m
```

So I guess you don't need to implement streaming data after all (as long as the application has enough memory to buffer the entire response..).

👍

rushtong:
Right ... we can avoid the code-based solution as long as the hosts
are beefy enough. I still think the streaming solution is more
future-proof, but in the effort to move this story along and unblock
users, I am leaning toward this solution.

> So I guess you don't need to implement streaming data after all (as long as the application as memory to buffer the entire response..).
>

rushtong:
👎  This approach doesn't work for even bigger downloads - rawls begins to time out when the entity size gets closer to 20K+ with 150+ columns.

PR #450, opened 2017-05-22
unit test sketch for GAWB-2059

davidangb:
Ideas and shell/framework for `AgoraEntityService` unit tests.

I know I scoffed at a stateful mock dao previously, but after hacking, it seemed like a good idea. I've left a bunch of tests incomplete, and we might need additional tests, but hopefully this gives you a good start.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/Application.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/Application.scala
@@ -6,11 +6,11 @@ import org.broadinstitute.dsde.firecloud.dataaccess._
   * Created by davidan on 9/23/16.
   */
 
-class Application (val agoraDAO: AgoraDAO,
-                   val googleServicesDAO: GoogleServicesDAO,
-                   val ontologyDAO: OntologyDAO,
-                   val rawlsDAO: RawlsDAO,
-                   val searchDAO: SearchDAO,
-                   val thurloeDAO: ThurloeDAO) {
+case class Application (agoraDAO: AgoraDAO,
+                   googleServicesDAO: GoogleServicesDAO,
+                   ontologyDAO: OntologyDAO,
+                   rawlsDAO: RawlsDAO,
+                   searchDAO: SearchDAO,
+                   thurloeDAO: ThurloeDAO) {
 
 }
--- a/src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockAgoraDAO.scala
+++ b/src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockAgoraDAO.scala
@@ -1,26 +1,18 @@
 package org.broadinstitute.dsde.firecloud.dataaccess
 
+import org.broadinstitute.dsde.firecloud.FireCloudExceptionWithErrorReport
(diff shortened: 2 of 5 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/Application.scala
```diff
@@ -6,11 +6,11 @@ import org.broadinstitute.dsde.firecloud.dataaccess._
...
-                   val ontologyDAO: OntologyDAO,
-                   val rawlsDAO: RawlsDAO,
-                   val searchDAO: SearchDAO,
-                   val thurloeDAO: ThurloeDAO) {
+case class Application (agoraDAO: AgoraDAO,
+                   googleServicesDAO: GoogleServicesDAO,
+                   ontologyDAO: OntologyDAO,
+                   rawlsDAO: RawlsDAO,
+                   searchDAO: SearchDAO,
+                   thurloeDAO: ThurloeDAO) {
```
davidangb, line 14: just so we can do `Application.copy(...)`, which is used in unit tests below.

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockAgoraDAO.scala
```diff
@@ -1,27 +1,19 @@
...
 import org.broadinstitute.dsde.firecloud.model.{SubsystemStatus, UserInfo}
+import org.broadinstitute.dsde.rawls.model.ErrorReport
 import spray.http.HttpResponse
 
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
 
 object MockAgoraDAO {
-  def agoraPermission = AgoraPermission(
+  val agoraPermission = AgoraPermission(
```
davidangb, line 13: `def` -> `val` so exterior classes can read it

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockAgoraDAO.scala
```diff
@@ -1,27 +1,19 @@
...
-
-  def method = Method(
-    namespace = Some("test-namespace"),
-    name = Some("test-name"),
-    snapshotId = Some(1),
-    synopsis = Some("test-synopsis"),
-    documentation = Some("test-documentation"),
-    createDate = Some("01-01-1970"),
-    entityType = Some("Workflow")
-  )
```
davidangb, line 24: moved to the interior of `postMethod`

rtitle:
Have you used [Mockito](http://site.mockito.org/) at all? I find it's useful for mocking behavior for tests, without needing to actually implement concrete mock objects. Furthermore, each test can define its mocks in the way that it needs -- so if a certain test needs stateful mocks, then that test would set them up itself, and not interfere with other tests.

Here's an example where I used it recently:

https://github.com/broadinstitute/rawls/blob/rt-gawb-769/core/src/test/scala/org/broadinstitute/dsde/rawls/monitor/HealthMonitorSpec.scala#L215-L293

and

https://github.com/broadinstitute/rawls/blob/develop/util/src/test/scala/org/broadinstitute/dsde/rawls/util/RetrySpec.scala#L154-L169

This might be a personal preference, but I thought I'd drop a comment anyway -- feel free to ignore.

PR #456, opened 2017-05-31
Synchronize maps when modifying groups in MockRawlsDAO

MatthewBemis:
  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockRawlsDAO.scala
+++ b/src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockRawlsDAO.scala
@@ -144,7 +144,7 @@ class MockRawlsDAO  extends RawlsDAO {
   override def adminAddMemberToGroup(groupName: String, memberList: RawlsGroupMemberList): Future[Boolean] = {
     val userEmailsToAdd = memberList.userSubjectIds.getOrElse(Seq[String]()).toSet
     val groupWithNewMembers = (groupName -> ((groups(groupName).filterNot(userEmailsToAdd.contains)) ++ userEmailsToAdd))
-    groups = groups.synchronized { groups + groupWithNewMembers }
+    this.synchronized { groups = groups + groupWithNewMembers }
 
     Future.successful(true)
   }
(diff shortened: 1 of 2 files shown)


davidangb (approved):
👍 works for me.

I'd prefer to see a `ConcurrentHashMap` for the groups var instead of `synchronized` ... but it's just a mock dao used for unit tests, so I don't care all that much.

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockRawlsDAO.scala
```diff
@@ -144,15 +144,15 @@ class MockRawlsDAO  extends RawlsDAO {
   override def adminAddMemberToGroup(groupName: String, memberList: RawlsGroupMemberList): Future[Boolean] = {
     val userEmailsToAdd = memberList.userSubjectIds.getOrElse(Seq[String]()).toSet
     val groupWithNewMembers = (groupName -> ((groups(groupName).filterNot(userEmailsToAdd.contains)) ++ userEmailsToAdd))
-    groups = groups + groupWithNewMembers
+    groups = groups.synchronized { groups + groupWithNewMembers }
```
rtitle: But the assignment to `groups` still happens outside of the synchronized block? This would synchronize both reads and writes to `groups`:
```
groups.synchronized { groups = groups + groupWithNewMembers }
```
Or even:
```
this.synchronized { groups = groups + groupWithNewMembers }
```

breilly2:
The thing about `synchronized` is that (at least in Java) it's protection by convention. If you want to protect some data using `synchronized`, every potential access to that data needs to synchronize on the same object (the object itself or a companion lock object). If you forget to do this anywhere, you'll be vulnerable to concurrent access problems. (Even reads outside of `synchronized` blocks can be stale due to lack of CPU cache flush.)

That said, if it's obvious this solution works AND it will be easy for anyone working with this class to also follow the conventions then this solution is fine. If it would be more obvious and easier to maintain to use a concurrent collection then that would be preferred.

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockRawlsDAO.scala
MatthewBemis: nice catch, I've updated this in the latest commit

rtitle:
👍

PR #458, opened 2017-05-31
[NOTYET] GAWB-2059: orch endpoint for method edit

davidangb:
I made the branch into a PR to make it easier to comment and pinpoint specific lines of code. Not yet ready for full review or merging.

I also rebased to develop since there was a conflict.

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -3446,7 +3446,6 @@ definitions:
       - synopsis
       - documentation
       - payload
-      - redactOldSnapshot
 
   EditMethodResponse:
     description: response of an edit method operation
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1118,7 +1118,9 @@ paths:
         - $ref: '#/parameters/editMethodRequest'
       responses:
         200:
-          description: Creating the new method succeeded.
+          description: Creating the new method succeeded. Check the response for any partial failures.
+          schema:
+            $ref: '#/definitions/EditMethodResponse'
         500:
           description: Internal Server Error
 
@@ -3445,6 +3447,58 @@ definitions:
       - documentation
       - payload
       - redactOldSnapshot
+
+  EditMethodResponse:
+    description: response of an edit method operation
+    type: object
+    properties:
+      method:
+        description: the created Method
+        type: object
+        properties:
+          namespace:
+            description: Method namespace
+            type: string
+          name:
+            description: Method name
+            type: string
+          snapshotId:
+            description: Method snapshot ID
+            type: integer
+          synopsis:
+            description: Method synopsis
+            type: string
+          documentation:
+            description: Method documentation
+            type: string
+          createDate:
+            description: Timestamp of Method creation
+            type: string
+          url:
+            description: Method URL
+            type: string
+          payload:
+            description: Method WDL
+            type: string
+          entityType:
+            description: Method type
+            type: string
+        required:
+          - namespace
+          - name
+          - snapshotId
+          - synopsis
+          - documentation
+          - createDate
+          - url
+          - payload
+          - entityType
+      message:
+        description: info about partial failures, if present
+        type: string
+    required:
+      - method
+
 
   Enabled:
     type: object
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1108,6 +1108,20 @@ paths:
         500:
           description: Internal Server Error
 
+  /api/methods/edit:
+    post:
+      tags:
+        - Method Repository
+      operationId: editMethod
+      summary: edit Method to create a new snapshot
+      parameters:
+        - $ref: '#/parameters/editMethodRequest'
+      responses:
+        200:
+          description: Creating the new method succeeded.
+        500:
+          description: Internal Server Error
+
   /api/nih/callback:
     post:
       tags:
@@ -3016,6 +3030,13 @@ parameters:
     required: true
     type: string
     in: path
+  editMethodRequest:
+    name: editMethodRequest
+    description: Edit Method Request
+    required: true
+    in: body
+    schema:
+      $ref: '#/definitions/EditMethodRequest'
 
 ##########################################################################################
 ## DEFINITIONS
@@ -3383,6 +3404,47 @@ definitions:
       authorized:
         type: boolean
         description: true if you have access to the dataset, false otherwise
+
+  EditMethodRequest:
+    description: info to edit a method
+    type: object
+    properties:
+      source:
+        description: ID of base method to edit
+        type: object
+        properties:
+          namespace:
+            type: string
+            description: Method namespace
+          name:
+            type: string
+            description: Method name
+          snapshotId:
+            type: integer
+            default: 0
+            description: Method version
+        required:
+          - namespace
+          - name
+          - snapshotId
+      synopsis:
+        description: synopsis for new snapshot
+        type: string
+      documentation:
+        description: documentation for new snapshot
+        type: string
+      payload:
+        description: WDL for new snapshot
+        type: string
+      redactOldSnapshot:
+        description: whether or not to redact the source snapshot upon success
+        type: boolean
+    required:
+      - source
+      - synopsis
+      - documentation
+      - payload
+      - redactOldSnapshot
 
   Enabled:
     type: object
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/model/MethodRepository.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/model/MethodRepository.scala
@@ -54,6 +54,12 @@ object MethodRepository {
     redactOldSnapshot: Boolean
   )
 
+  case class EditMethodResponse(
+    method: EntityId,
+    message: Option[String] = None,
+    error: Option[Throwable] = None
+  )
+
   // represents a method/config permission as exposed to the user from the orchestration layer
   case class FireCloudPermission(
     user: String,
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/model/MethodRepository.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/model/MethodRepository.scala
@@ -56,8 +56,7 @@ object MethodRepository {
 
   case class EditMethodResponse(
     method: MethodId,
-    message: Option[String] = None,
-    error: Option[Throwable] = None
+    message: Option[String] = None
   )
 
   // represents a method/config permission as exposed to the user from the orchestration layer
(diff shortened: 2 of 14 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/model/MethodRepository.scala
```diff
@@ -25,12 +25,41 @@ object MethodRepository {
     name: Option[String] = None,
     snapshotId: Option[Int] = None,
     synopsis: Option[String] = None,
-    owner: Option[String] = None,
```
davidangb, line 28: why was owner removed?

src/main/scala/org/broadinstitute/dsde/firecloud/model/MethodRepository.scala
```diff
@@ -25,12 +25,41 @@ object MethodRepository {
...
+    synopsis: String,
+    documentation: String,
+    payload: String,
+    redactOldSnapshot: Boolean
+  )
+
+  case class EditMethodResponse(
+    method: Method,
+    message: Option[String] = None
+  )
```
davidangb, line 61: do we really need all these new model classes? For instance, since the `Method` class uses `Options`, we could possibly replace `MethodCreate` and `MethodId` with `Method`. This is a brainstorm, not an explicit directive.

src/main/scala/org/broadinstitute/dsde/firecloud/service/AgoraEntityService.scala
```diff
@@ -0,0 +1,69 @@
...
+  extends Actor with SprayJsonSupport {
+
+  implicit val userInfo = argUserInfo
+
+  override def receive: Receive = {
+    case EditMethod(editMethodRequest: EditMethodRequest) => { editMethod(editMethodRequest) } pipeTo sender
+  }
+
+  def editMethod(req: EditMethodRequest): Future[PerRequestMessage] = {
+    val source = req.source
```
davidangb, line 40: nowhere in here do we actually verify that the source is a valid namespace/name/snapshot. We should probably make sure it exists before attempting to edit it.

src/main/scala/org/broadinstitute/dsde/firecloud/service/AgoraEntityService.scala
```diff
@@ -0,0 +1,69 @@
...
+          agoraDAO.redactMethod(source.namespace, source.name, source.snapshotId) map { _ =>
+            RequestComplete(OK, EditMethodResponse(newMethod))
+          } recover {
+            case e: Exception => RequestCompleteWithErrorReport(NonAuthoritativeInformation, "Error occurred while redacting previous snapshot", e)
+          }
+        } else {
+          Future(RequestComplete(OK, EditMethodResponse(newMethod)))
+        }
+      } recover {
+        case e: Exception => RequestCompleteWithErrorReport(NonAuthoritativeInformation, "Error occurred while copying permissions", e)
```
davidangb: I used `NonAuthoritativeInformation` here and above as a prototype; I'm open to other suggestions.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/AgoraEntityApiService.scala
```diff
@@ -0,0 +1,33 @@
...
+  with StandardUserInfoDirectives{
+
+  private implicit val ec: ExecutionContext = actorRefFactory.dispatcher
+
+  val agoraEntityServiceConstructor: UserInfo => AgoraEntityService
+
+  val agoraEntityRoutes: Route = {
+    path("methods" / "edit") {
+      requireUserInfo() { userInfo =>
+        entity(as[EditMethodRequest]) { editMethodRequest =>
```
davidangb, line 23: This requires that the end user specify a fully-populated `EditMethodRequest`, which contains a synopsis, documentation, and payload. Since we are editing a pre-existing method, I think it would make more sense to make synopsis, documentation, and payload optional, and copy them from the pre-existing method if the user didn't specify them.

Similarly, let's make `redactOldSnapshot` optional and give it a default of false.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -3384,6 +3407,99 @@ definitions:
...
+      - synopsis
+      - documentation
+      - payload
+      - redactOldSnapshot
+
+  EditMethodResponse:
+    description: response of an edit method operation
+    type: object
+    properties:
+      method:
```
davidangb, line 3454: can't we just reference an existing schema def, something like:
```
      method:
        $ref: '#/definitions/MethodResponse'
```
instead of manually listing out all the properties?

(not sure that's the right definition to reference)

src/main/scala/org/broadinstitute/dsde/firecloud/service/AgoraEntityService.scala
```diff
@@ -0,0 +1,69 @@
...
+      }
+    } recover {
+      case e: Exception => RequestCompleteWithErrorReport(InternalServerError, "An internal error occurred on method edit", e)
+    }
+  }
```
davidangb: now that all the routing is correct, adding individual `recover` clauses to the various futures works just fine.

aside: I don't love the way that `requestToObject` and its ilk obfuscates the underlying errors/responses. That's my code in the first place and fixing it is scope creep for this PR, but it becomes obvious in situations like this where we try to propagate the exception ... and end up displaying something pretty unhelpful.

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockAgoraDAO.scala
```diff
@@ -24,6 +26,44 @@ class MockAgoraDAO extends AgoraDAO {
     Future(List(MockAgoraDAO.agoraPermission))
   }
 
+  override def postMethod(ns: String, name: String, synopsis: String, documentation: String, payload: String)(implicit userInfo: UserInfo): Future[Method] = {
+    // based on what the unit test passed in, return success or failure
+    (ns, name) match {
+      case ("exceptions", "postError") => Future.failed(new AgoraException("postMethod"))
```
davidangb: It's important to use `Future.failed(...)` here instead of directly throwing an exception; otherwise `recover` doesn't catch it in the right way. Think of it this way:

throw Exception: we blew up generating the request, for instance a NPE building the request payload.
Future.failed(Exception): the service we called returned an error response.

src/test/scala/org/broadinstitute/dsde/firecloud/service/AgoraEntityServiceSpec.scala
```diff
@@ -0,0 +1,174 @@
...
+import scala.concurrent.duration._
+
+class AgoraEntityServiceSpec extends BaseServiceSpec with BeforeAndAfterEach {
+
+  val dur = Duration(2, MINUTES)
+
+  var dao: StatefulMockAgoraDAO = _
+  var aes: AgoraEntityService = _
+
+  val validPayload = "task wc {File in_file command { cat ${in_file} | wc -l } output { Int count = read_int(stdout()) }} workflow www {call wc}"
```
davidangb, line 24: in any test that actually talks to agora - including when we as humans are testing via swagger - beware the payload. Agora validates the payload when creating a method, and an invalid payload will result in an error.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/AgoraEntityApiService.scala
```diff
@@ -0,0 +1,33 @@
...
+
+trait AgoraEntityApiService extends HttpService with FireCloudRequestBuilding with FireCloudDirectives
+  with StandardUserInfoDirectives{
+
+  private implicit val ec: ExecutionContext = actorRefFactory.dispatcher
+
+  val agoraEntityServiceConstructor: UserInfo => AgoraEntityService
+
+  val agoraEntityRoutes: Route = {
+    path("methods" / "edit") {
```
davidangb, line 21: I removed the `api` from here, and added this to the api-protected routes in `FireCloudServiceActor`.

src/main/scala/org/broadinstitute/dsde/firecloud/service/AgoraEntityService.scala
```diff
@@ -0,0 +1,69 @@
...
+    case EditMethod(editMethodRequest: EditMethodRequest) => { editMethod(editMethodRequest) } pipeTo sender
+  }
+
+  def editMethod(req: EditMethodRequest): Future[PerRequestMessage] = {
+    val source = req.source
+    agoraDAO.postMethod(source.namespace, source.name, req.synopsis, req.documentation, req.payload) flatMap { newMethod =>
+      val newId = MethodId(newMethod.namespace.get, newMethod.name.get, newMethod.snapshotId.get)
+      copyMethodPermissions(source, newId) flatMap { _ =>
+        if (req.redactOldSnapshot) {
+          agoraDAO.redactMethod(source.namespace, source.name, source.snapshotId) map { _ =>
```
davidangb: Under the covers, `redactMethod` uses `userAuthedRequest` instead of `authedRequestToObject`, and therefore has very different error-handling. I don't think this will actually cause a failure if the redact call fails, and therefore won't get caught by its `recover` clause. If you dig in, note how `authedRequestToObject` checks the status of the response ... but `userAuthedRequest` doesn't. This probably will need to change.

src/main/scala/org/broadinstitute/dsde/firecloud/model/MethodRepository.scala
aednichols: Agora does not return it. The `managers` array seems to have taken its place (though I do not know when or why this changed).
```
{
  "managers": [
    "oednichols@gmail.com"
  ],
  "name": "Adam-Method-2",
  "createDate": "2017-06-01T18:20:03Z",
  "payload": "task wc {File in_file command { cat ${in_file} | wc -l } output { Int count = read_int(stdout()) }} workflow www {call wc}",
  "url": "http://agora.dsde-dev.broadinstitute.org/api/v1/methods/Adam-Namespace/Adam-Method-2/1",
  "documentation": "Test docs",
  "synopsis": "Test synopsis",
  "entityType": "Workflow",
  "snapshotId": 1,
  "namespace": "Adam-Namespace"
}
```

src/main/scala/org/broadinstitute/dsde/firecloud/model/MethodRepository.scala
aednichols: Let's say we're completing the argument list for a hypothetical `EditMethodRequest(source: Method, [...])`. We need to supply a `Method` instance - but there's no easy way to tell that exactly three of the fields are needed for this to work.

So, they do seem somewhat helpful to me.

src/main/scala/org/broadinstitute/dsde/firecloud/service/AgoraEntityService.scala
aednichols: You're right - the redaction fails silently when the request is invalid. Fixed.

PR #462, opened 2017-06-10
GAWB-2122: GET endpoints for library discoverable groups and metadata

davidangb:
Permission-wise, both new endpoints only require read privileges on the target workspace. This is consistent with the GET on the workspace itself, since all the library attributes are returned as part of the workspace.

I've manually tested the round-trip on these: GET a payload, then PUT it back using the corresponding endpoint.

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -831,6 +831,10 @@ paths:
           description: The workspace, after assigning groups to be able to discover the dataset
           schema:
             $ref: '#/definitions/Workspace'
+        400:
+          description: Invalid input
+          schema:
+            $ref: '#/definitions/ErrorReport'
         403:
           description: Insufficient permissions to change catalog permissions (must be owner or have grant or catalog with access to the workspace)
           schema:
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -785,6 +785,32 @@ paths:
       operationId: isCurator
 
   /api/library/{workspaceNamespace}/{workspaceName}/discoverableGroups:
+    get:
+      tags:
+        - Library
+      operationId: getDiscoverableGroups
+      summary: |
+        get the groups that can discover this library dataset
+      parameters:
+        - $ref: '#/parameters/workspaceNamespaceParam'
+        - $ref: '#/parameters/workspaceNameParam'
+      responses:
+        200:
+          description: The groups that can discover the dataset
+          schema:
+            $ref: '#/definitions/StringArray'
+        403:
+          description: Insufficient permissions
+          schema:
+            $ref: '#/definitions/ErrorReport'
+        404:
+          description: Workspace not found
+          schema:
+            $ref: '#/definitions/ErrorReport'
+        500:
+          description: Error
+          schema:
+            $ref: '#/definitions/ErrorReport'
     put:
       tags:
         - Library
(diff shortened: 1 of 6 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -795,16 +821,16 @@ paths:
...
+            $ref: '#/definitions/StringArray'
       responses:
         200:
-          description: The groups that have been assigned to be able to discover the dataset
+          description: The workspace, after assigning groups to be able to discover the dataset
           schema:
-            type: array
-            items:
-              type: string
+            $ref: '#/definitions/Workspace'
```
davidangb, line 833: I did not change the API response; I just updated swagger to reflect what the API actually does.

src/test/resources/reference.conf
```diff
@@ -58,7 +58,7 @@ elasticsearch {
   urls = "localhost:9292"
   clusterName = "elasticsearch5a"
   index = "unittest"
-  discoverGroupNames = ["all_broad_users"]
+  discoverGroupNames = ["all_broad_users","demo_users"]
```
davidangb, line 61: better reflects what is used at runtime

src/main/resources/swagger/api-docs.yaml
```diff
@@ -785,6 +785,32 @@ paths:
...
+            $ref: '#/definitions/ErrorReport'
+        404:
+          description: Workspace not found
+          schema:
+            $ref: '#/definitions/ErrorReport'
+        500:
+          description: Error
+          schema:
+            $ref: '#/definitions/ErrorReport'
     put:
```
aednichols, line 814: When specifying `discoverableGroups` outside of `["all_broad_users", "demo_users"]`, the API returns a `400`, which is not documented here

Instead, the existing Swagger (misleadingly?) suggests that would be a 404 ("Workspace, user or group not found")

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -133,6 +149,18 @@ class LibraryService (protected val argUserInfo: UserInfo,
     }
   }
 
+  def getAttributes(ns: String, name: String): Future[PerRequestMessage] = {
```
aednichols: It's somewhat confusing that this response is different than the `attributes` key of the `updateAttributes` response - because we're limiting to attributes where`k.namespace == AttributeName.libraryNamespace`.

Is it correct to say that the predicate above defines the set of attributes we call metadata? I've gotten a bit lost in the naming.

aednichols (changes requested):
Works nicely, noted two minor clarity things

src/main/resources/swagger/api-docs.yaml
davidangb: Good catch. You'll also get a 400 if you input something that isn't a `List[String]`. I added the 400 response code.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
davidangb: Yeah, I agree that the asymmetry here is a bit ugly. I didn't change the response of `updateAttributes` (now `updateLibraryMetadata`) because that could break compatibility for anyone using it. But at the same time, it makes no sense for the put-metadata endpoint to accept an entire workspace object.

I did rename the case classes and methods to try to keep code understandable.

"metadata" is short for "Library metadata" which is anything in the library namespace. The put-metadata endpoint enforces that at line 145.

aednichols (approved):

PR #466, opened 2017-06-22
Enable Large TSV Downloads

rushtong:
Addresses https://broadinstitute.atlassian.net/browse/GAWB-1998

# Changes in this PR:
- Use paginated endpoint from Rawls to get all entities
- Use akka streams to process large entity downloads
- Use a streaming actor to chunk content directly to the browser for entities.

# Notes:
- Pulled the endpoint handling out of `EntityService` due to problems with passthroughs. (see https://broadinstitute.atlassian.net/browse/GAWB-1303)
- Cannot be tested through Swagger due to current bug (see https://broadinstitute.atlassian.net/browse/GAWB-2191)

# Testing:
See google drive doc: https://docs.google.com/document/d/1vlXjU7hVRkFhVzWWOfQO9wTbiQVPABJC3b1CUs8qOxs/edit#heading=h.gaeu9xnlvenh

## UI Tests

- Generate a workspace and upload lots of entities. I have a groovy script that does this (see below)
- Try to Export entities from data tab.
- Test both small and large datasets.
- Test with both set and individual entity types

## API Tests

- APIs: Test both the entity service and the cookie-authed service endpoints. 
- Test adding filterable attributes and ensure that they work.
- Test both small and large downloads
- Test with both set (sample_set, participant_set, etc.) and individual entity (sample, participant, etc.) types
- Test multiple downloads at once and compare files afterward to ensure correctness

---

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off
- [x] **Submitter**: Inform other teams of any API changes via Slack and/or email


src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
```diff
@@ -113,10 +115,8 @@ class FireCloudServiceActor extends HttpServiceActor with FireCloudDirectives
         pathPrefix("api") {
           apiRoutes
         } ~
-        pathPrefix("cookie-authed") {
-          // insecure cookie-authed routes
-          cookieAuthedRoutes
-        }
+        // insecure cookie-authed routes
+        cookieAuthedRoutes
```
rushtong, line 120: I specifically pulled this `pathPrefix` out to make the cookie-authed endpoint more consistent with our other routes. This also clarifies the test case so we don't have to guess which root path this belongs to.

src/main/scala/org/broadinstitute/dsde/firecloud/service/EntityService.scala
```diff
@@ -71,16 +69,6 @@ trait EntityService extends HttpService with PerRequestCreator with FireCloudDir
...
-                  parameters('attributeNames.?) { attributeNamesString =>
-                    path("tsv") {
-                      requireUserInfo() { userInfo => requestContext =>
-                        val filename = entityType + ".txt"
-                        val attributeNames = attributeNamesString.map(_.split(",").toIndexedSeq)
-                        perRequest(requestContext, ExportEntitiesByTypeActor.props(exportEntitiesByTypeConstructor, userInfo),
-                          ExportEntitiesByTypeActor.ExportEntities(workspaceNamespace, workspaceName, filename, entityType, attributeNames))
-                      }
-                    }
-                  } ~
```
rushtong, line 83: Specifically pulled the tsv download out of this service due to interference from passthrough route problems. Ticket for that is here: https://broadinstitute.atlassian.net/browse/GAWB-1303

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -1,79 +1,224 @@
...
-    val bos = new ByteArrayOutputStream()
-    val zos = new ZipOutputStream(bos)
-    zos.putNextEntry(new ZipEntry(entityType + "_membership.tsv"))
-    zos.write(membershipData.getBytes)
-    zos.closeEntry()
-    zos.putNextEntry(new ZipEntry(entityType + "_entity.tsv"))
-    zos.write(entityData.getBytes)
-    zos.closeEntry()
-    zos.finish()
-    bos.toByteArray
```
rushtong, line 76: This is replaced by `writeFilesToZip` on line 237

src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingActor.scala
```diff
@@ -0,0 +1,52 @@
...
+    // This case comes back from the client
+    // all chunks were sent. stop.
+    case Ok(remaining: Int) =>
+      if (remaining == 0) {
+        ctx.responder ! ChunkedMessageEnd
+        context.stop(self)
+      }
+
+    //
+    case x => unhandled(x)
```
rushtong: Note to self - need an error log statement here.

rushtong:
`CookieAuthedExportEntitiesByTypeServiceSpec.scala` and `ExportEntitiesByTypeServiceSpec.scala` are virtual copies. Would love to know how to refactor them as one - too much duplication.

project/Dependencies.scala
```diff
@@ -2,6 +2,7 @@ import sbt._
 
 object Dependencies {
   val akkaV = "2.4.1"
+  val akkaStreamV = "2.4.17"
```
davidangb: could we use a consistent version for all the akka artifacts? We could ugprade akka-actor/akka-testkit/akka-slf4j to 2.4.17, or even move everything to latest 2.4.x. 

Is akka 2.5 the version that doesn't play well with Spray?

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -1,79 +1,224 @@
...
-  case class ExportEntities(workspaceNamespace: String, workspaceName: String, filename: String, entityType: String, attributeNames: Option[IndexedSeq[String]]) extends ExportEntitiesByTypeMessage
+  case class ExportEntities(ctx: RequestContext, workspaceNamespace: String, workspaceName: String, entityType: String, attributeNames: Option[IndexedSeq[String]]) extends ExportEntitiesByTypeMessage
 
   def props(exportEntitiesByTypeConstructor: UserInfo => ExportEntitiesByTypeActor, userInfo: UserInfo): Props = {
     Props(exportEntitiesByTypeConstructor(userInfo))
   }
 
   def constructor(app: Application)(userInfo: UserInfo)(implicit executionContext: ExecutionContext) =
-    new ExportEntitiesByTypeActor(app.rawlsDAO, userInfo)
+    new ExportEntitiesByTypeActor(app.rawlsDAO, app.googleServicesDAO, userInfo)
```
davidangb: remove `GoogleServicesDAO`, assuming it's not used anymore.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -1,79 +1,224 @@
...
-            HttpHeaders.`Content-Disposition`.apply("attachment", Map("filename" -> zippedFileName)))
-        case _ =>
-          val data = TSVFormatter.makeEntityTsvString(entities, entityType, attributeNames)
-          RequestCompleteWithHeaders(
-            (OK, data),
-            HttpHeaders.`Content-Disposition`.apply("attachment", Map("filename" -> filename)),
-            HttpHeaders.`Content-Type`(`text/plain`))
+  implicit val timeout = Timeout(1 minute)
+
+  lazy val logger: Logger = LoggerFactory.getLogger(getClass)
```
davidangb: style: I really like pulling in the `LazyLogging` trait instead of manually creating our own logger.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -1,79 +1,224 @@
...
+        getEntities(workspaceNamespace, workspaceName, entityType, query)
       }
+    ) map(_.flatten)
+  }
+
+  private def getEntityQueries(metadata: EntityTypeMetadata, entityType: String): Seq[EntityQuery] = {
+    val pageSize = FireCloudConfig.Rawls.defaultPageSize
+    val filteredCount = metadata.count
+    val sortField = entityType + "_id"
+    val pages = filteredCount % pageSize match {
```
davidangb: I think `val pages = Math.ceil(filteredCount / pageSize).toInt` is equivalent, and shorter

src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingActor.scala
```diff
@@ -0,0 +1,52 @@
...
+
+object StreamingActor {
+  case class FirstChunk(httpData: HttpData, remaining: Int)
+  case class NextChunk(httpData: HttpData, remaining: Int)
+  case class Ok(remaining: Int)
+}
+
+class StreamingActor(ctx: RequestContext, contentType: ContentType, contentDisposition: String) extends Actor with HttpService {
+
+  private lazy val logger: Logger = LoggerFactory.getLogger(getClass)
```
davidangb: consider `LazyLogging`

src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingActor.scala
```diff
@@ -0,0 +1,52 @@
...
+import spray.http._
+import spray.routing.{HttpService, RequestContext}
+
+object StreamingActor {
+  case class FirstChunk(httpData: HttpData, remaining: Int)
+  case class NextChunk(httpData: HttpData, remaining: Int)
+  case class Ok(remaining: Int)
+}
+
+class StreamingActor(ctx: RequestContext, contentType: ContentType, contentDisposition: String) extends Actor with HttpService {
```
davidangb: I find the `contentDisposition` variable name confusing here - the value is actually the attachment filename. (I know it goes inside the content-disposition header, but it's still the filename)

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/CookieAuthedApiService.scala
```diff
@@ -1,34 +1,37 @@
...
   val exportEntitiesByTypeConstructor: UserInfo => ExportEntitiesByTypeActor
 
   private implicit val executionContext = actorRefFactory.dispatcher
   lazy val log = LoggerFactory.getLogger(getClass)
 
-  def cookieAuthedRoutes: Route =
+  val cookieAuthedRoutes: Route =
     // download "proxy" for TSV files
-    path("workspaces" / Segment / Segment/ "entities" / Segment/ "tsv") {
+    path( "cookie-authed" / "workspaces" / Segment / Segment/ "entities" / Segment / "tsv" ) {
```
davidangb: the `cookie-authed` prefix needs to be on both the TSV path and the GCS download path.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -1,79 +1,224 @@
...
+    case ExportEntities(ctx, workspaceNamespace, workspaceName, entityType, attributeNames) => streamEntities(ctx, workspaceNamespace, workspaceName, entityType, attributeNames) pipeTo sender
   }
 }
 
 trait ExportEntitiesByType extends FireCloudRequestBuilding {
   val rawlsDAO: RawlsDAO
+  val googleDAO: GoogleServicesDAO
   implicit val userInfo: UserInfo
   implicit protected val executionContext: ExecutionContext
+  implicit def actorRefFactory: ActorRefFactory
```
rtitle: It seems kind of strange (to me) to put stuff in the trait which require an implicit `executionContext`, `actorRefFactory`, etc. Why not just put the actor-specific stuff in the actor class and do `import context._` at the top?

davidangb (commented):
some first-pass feedback - I am definitely going to take a second pass, as well as run it myself.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
rtitle: Tip: if you mix in `ActorLogging` to your actor you can get a logger provided for you:

http://doc.akka.io/docs/akka/current/scala/logging.html

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -1,79 +1,224 @@
...
+    * For Collection types, write the content to temp files, zip and return.
+    *
+    * For Singular types, pipe the content from `Source` -> `Flow` -> `Sink`
+    *   Source generates the entity queries
+    *   Flow executes the queries and sends formatted content to chunked response handler
+    *   Sink finishes the execution pipeline
+    *
+    * Handle exceptions directly by completing the request.
+    */
+  def streamEntities(ctx: RequestContext, workspaceNamespace: String, workspaceName: String, entityType: String, attributeNames: Option[IndexedSeq[String]]): Future[Any] = {
```
rtitle: Returning `Future[Any]` doesn't seem right to me - it should be a more specific type.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -1,79 +1,224 @@
...
+  /**
+    * General Approach
+    * 1. Define a `Source` of entity queries
+    * 2. Run the source events through a `Flow`.
+    * 3. Flow sends events (batch of entities) to a streaming output actor
+    * 4. Return a Done to the calling route when complete.
+    */
+  private def streamSingularType(ctx: RequestContext, workspaceNamespace: String, workspaceName: String, entityType: String, entityQueries: Seq[EntityQuery], metadata: EntityTypeMetadata, headers: IndexedSeq[String], attributeNames: Option[IndexedSeq[String]]): Future[Done] = {
+    // Akka Streams Support
+    implicit val system: ActorSystem = ActorSystem("Streaming-Entity-Exporter")
```
rtitle: Creating a new `ActorSystem` and `ActorMaterializer` for every method invocation doesn't seem right. These are expensive objects to create and usually you want to reuse them throughout your application. I think you already have an implicit ActorSystem in scope -- and the class/trait could be modified to take an implicit `ActorMaterializer` as well.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -1,79 +1,224 @@
...
+    Future(Done)
+  }
+
+  private def sendRowsAsChunks(actorRef: ActorRef, query: EntityQuery, querySize: Int, entityType: String, headers: IndexedSeq[String], entities: Seq[Entity]): Unit = {
+    val rows = TSVFormatter.makeEntityRows(entityType, entities, headers)
+    val remaining = querySize - query.page + 1
+    // Send headers as the first chunk of data
+    if (query.page == 1) { actorRef ! FirstChunk(HttpData(headers.mkString("\t") + "\n"), remaining)}
+    // Send entities
+    actorRef ! NextChunk(HttpData(rows.map { _.mkString("\t") }.mkString("\n") + "\n"), remaining - 1)
```
rtitle: Would it be possible to use [TSVParser](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVParser.scala) which uses [Univocity](https://github.com/uniVocity/univocity-parsers) under the hood? This might fail for certain exotic TSVs which contain tabs inside quotes, for example. Also it handles things like auto-detecting line endings (`\n` vs \r\n`).

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -1,79 +1,224 @@
...
+  }
+
+  private def writeCollectionTypeZipFile(workspaceNamespace: String, workspaceName: String, entityType: String, entityQueries: Seq[EntityQuery], metadata: EntityTypeMetadata, attributeNames: Option[IndexedSeq[String]]): Future[File] = {
+    val entityWriter: ActorRef = actorRefFactory.actorOf(TSVWriterActor.props(entityType, metadata.attributeNames, attributeNames, entityQueries.size))
+    val membershipWriter: ActorRef = actorRefFactory.actorOf(TSVWriterActor.props(entityType, metadata.attributeNames, attributeNames, entityQueries.size))
+    val foldOperation = entityQueries.foldLeft(Future.successful(0, Seq[File]())) { (accumulator, queryGroup) =>
```
rtitle: Maybe
```
Future.fold(entityQueries)((0, Seq.empty[File])) { (accumulator, queryGroup) =>
  // etc
}
```
?

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -1,79 +1,224 @@
...
+      val membership = zipFile/s"${entityType}_membership.tsv"
+      val entity = zipFile/s"${entityType}_entity.tsv"
+      membershipTSV.copyTo(membership)
+      entityTSV.copyTo(entity)
+      zipFile.zip()
+    }
+  }
+
+  private def getEntityBatchFromQueries(queryGroup: Seq[EntityQuery], workspaceNamespace: String, workspaceName: String, entityType: String): Future[Seq[Entity]] = {
+    Future.sequence(
```
rtitle: If you're doing a `Future.sequence(something.map(...))` it can usually be written as a `Future.traverse`. E.g.:
```
Future.traverse(queryGroup) { query =>
  getEntities(workspaceNamespace, workspaceName, entityType, query)
} map(_.flatten)
```

src/main/scala/org/broadinstitute/dsde/firecloud/service/TSVWriterActor.scala
```diff
@@ -0,0 +1,59 @@
...
+
+  def entityType: String
+  def originalHeaders: Seq[String]
+  def requestedHeaders: Option[IndexedSeq[String]]
+  def pages: Int
+
+  lazy val file: File = File.newTemporaryFile(UUID.randomUUID().toString, ".tsv")
+
+  def receive: Receive = {
+    case WriteEntityTSV(page: Int, entities: Seq[Entity]) => val receiver = sender; receiver ! writeEntityTSV(page, entities)
```
rtitle: Think you could just do `sender ! writeEntityTSV(page, entities)`

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -1,79 +1,224 @@
...
+
+    // The output to the user
+    val streamingActorRef = actorRefFactory.actorOf(Props(new StreamingActor(ctx, ContentTypes.`text/plain`, entityType + ".txt")))
+
+    // The Source
+    val entityQuerySource = Source(entityQueries.toStream)
+
+    // MapAsync should preserve order.
+    val flow = Flow[EntityQuery].mapAsync(1) { query =>
+      getEntities(workspaceNamespace, workspaceName, entityType, query) map { entities =>
```
rtitle: What happens if one of these futures fail? Will it abort the whole stream? (and what will the user see in that case?)

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
rushtong: I was actually confused by what we should be doing. In some cases we do both, `LazyLogging` and `LoggerFactory`. Will switch to `LazyLogging`

project/Dependencies.scala
rushtong: Will test out to see what works. 
Yes, 2.5 (which is latest akka streams) evicts the 2.4 akka code in spray.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
rushtong: This works for `filteredCount / pageSize` is > 1, but fails for < 1. I think I can simplify though, thanks.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
rushtong: I think our preference is for `LazyLogging`, which seems equivalent.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/CookieAuthedApiService.scala
rushtong: 👍

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
rushtong: So ... this is where I get lost in actor-land. Conceptually, I agree, but I don't know what the right type should be. We're either streaming content to the browser, so the return doesn't matter (but it's a `akka.Done` in successful cases), or we're completing the request with an `HttpResponse` that holds the error.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
rushtong: That looks cleaner, I wish it compiled.

rushtong: Ah, this compiles `Future.fold(entityQueries.map(Future(_)))((0, Seq.empty[File]))`. The extra map->future kind of loses the cleanliness though.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
rushtong: 👍

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
rushtong: I have to look more into this. If anything fails, the request should complete with a firecloud error, but I'm not confident that is working the way I intended it to. It did at one point in the development.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TSVWriterActor.scala
rushtong: 👍

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
rtitle: Oh yeah, maybe I should have tried it before commenting (I forgot Future.fold takes a list of futures). The current code is fine.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
rushtong: @rtitle - if you have some time, I'd love to sit with you and understand what this means. At present, I'm not grasping the details of this.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
rtitle: I see -- no one is even using the return value of this method because you're completing the RequestContext directly. It looks like the error case returns a `Future[Unit]` (because `ctx.complete` returns Unit), and the success case returns a `Future[Done]`. Maybe you could map the `Done` to a `Unit` and have the whole method return a `Future[Unit]`.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
rushtong: I think I figured this out what your intention is. In the existing style, we have actor classes that are super thin and then any real work is done in a trait. In this case, that doesn't make as much sense since we have so much actor-based work to do. Pulling all of the helper methods into the actor class cleans this up a lot.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
rushtong: Was able to get rid of `ActorSystem` by pulling helpers into actor class. Looking into moving `ActorMaterializer` to the app level. That's only used by `akka.stream` so we have never needed that before.

rtitle: Yep, when using akka streams you usually need to propagate _both_ an `ActorSystem` and an `ActorMaterializer` throughout the app.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
rtitle: I _think_ the stream will abort based on reading the `mapAsync` documentation, but it's not clear to me what the user will see in that case.

project/Dependencies.scala
rushtong: Upgrading akka to the latest (turns out it's 2.4.19 since I last checked) works fine. 2.5 is a somewhat non-trivial change (see http://doc.akka.io/docs/akka/snapshot/scala/project/migration-guide-2.4.x-2.5.x.html) and definitely does not work without code changes ... I tried.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
rushtong: 👍 next on my list.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/RawlsDAO.scala
```diff
@@ -41,6 +41,24 @@ trait RawlsDAO extends LazyLogging with ReportsSubsystemStatus {
...
+    val filteredQMap = query.filterTerms match {
+      case Some(f) => qMap + ("filterTerms" -> f)
+      case _ => qMap
+    }
+    val baseRawlsEntityQueryUrl = FireCloudConfig.Rawls.entityQueryPathFromWorkspace(workspaceNamespace, workspaceName)
+    val baseEntityQueryUri = Uri(baseRawlsEntityQueryUrl)
+    val entityQueryUri = baseEntityQueryUri.
+      withPath(baseEntityQueryUri.path ++ Uri.Path.SingleSlash ++ Uri.Path(entityType)).
+      withQuery(filteredQMap)
+    entityQueryUri.toString()
```
davidangb: There's a bunch of duplication between here and `EntityService.scala` starting around line 80. Could that be cleaned up somehow? Note that the comment in EntityService "passthrough does not handle query params well" is no longer true, afaik.

project/Dependencies.scala
```diff
@@ -51,6 +52,7 @@ object Dependencies {
     "com.univocity"                  % "univocity-parsers"   % "2.4.1",
     "org.ocpsoft.prettytime"         % "prettytime"          % "4.0.1.Final",
     "org.everit.json"                % "org.everit.json.schema" % "1.4.1",
+    "com.github.pathikrit"          %% "better-files"        % "2.17.1",
```
davidangb, line 55: better-files is currently at 3.0.0. Are you using 2.17.1 explicitly?

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
davidangb: wouldn't a failure in one of the futures here get caught in `streamEntities..recoverWith`? End result being that we close the connection to the end user, and their download aborts.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -1,79 +1,208 @@
...
+    // The output file
+    lazy val zipFile = writeCollectionTypeZipFile(workspaceNamespace, workspaceName, entityType, entityQueries, metadata, attributeNames)
+
+    // The output to the user
+    lazy val streamingActorRef = actorRefFactory.actorOf(Props(new StreamingActor(ctx, ContentTypes.`application/octet-stream`, entityType + ".zip")))
+    zipFile map { f =>
+      streamingActorRef ! FirstChunk(HttpData.apply(f.byteArray), 0)
+    }
+
+    Future(Done)
```
davidangb: is there a way to respond with the result of the `FirstChunk` invocation, instead of hardcoding `Future(Done)` here?

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -1,79 +1,208 @@
...
+  }
+
+  private def streamCollectionType(ctx: RequestContext, workspaceNamespace: String, workspaceName: String, entityType: String, entityQueries: Seq[EntityQuery], metadata: EntityTypeMetadata, attributeNames: Option[IndexedSeq[String]]): Future[Done] = {
+    // The output file
+    lazy val zipFile = writeCollectionTypeZipFile(workspaceNamespace, workspaceName, entityType, entityQueries, metadata, attributeNames)
+
+    // The output to the user
+    lazy val streamingActorRef = actorRefFactory.actorOf(Props(new StreamingActor(ctx, ContentTypes.`application/octet-stream`, entityType + ".zip")))
+    zipFile map { f =>
+      streamingActorRef ! FirstChunk(HttpData.apply(f.byteArray), 0)
```
davidangb: this is interesting. The zipfile might be quite large itself - larger than the chunk size. In that situation, does the `FirstChunk` codepath actually send multiple chunks, because it knows the entire `HttpEntity`?

src/main/resources/reference.conf
```diff
@@ -36,4 +40,6 @@ userprofile {
 
 googlecloud {
   priceListUrl: "https://cloudpricingcalculator.appspot.com/static/data/pricelist.json"
-}
\ No newline at end of file
+}
+
+spray.can.client.response-chunk-aggregation-limit = 5m
```
davidangb: I'm confused about what this actually does in practice. Does it affect how we send chunks to the browser? From the docs, I'm not sure it does. (this might be a face-to-face conversation, unless it's easy)

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
davidangb: total nitpick, hardly worth mentioning: I find `Seq.empty[File]` to be more readable than `Seq[File]()`. Your call on changing or not.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -1,79 +1,208 @@
...
+    // Send entities
+    actorRef ! NextChunk(HttpData(rows.map { _.mkString("\t") }.mkString("\n") + "\n"), remaining - 1)
+  }
+
+  private def writeCollectionTypeZipFile(workspaceNamespace: String, workspaceName: String, entityType: String, entityQueries: Seq[EntityQuery], metadata: EntityTypeMetadata, attributeNames: Option[IndexedSeq[String]]): Future[File] = {
+    val entityWriter: ActorRef = actorRefFactory.actorOf(TSVWriterActor.props(entityType, metadata.attributeNames, attributeNames, entityQueries.size))
+    val membershipWriter: ActorRef = actorRefFactory.actorOf(TSVWriterActor.props(entityType, metadata.attributeNames, attributeNames, entityQueries.size))
+    val foldOperation = entityQueries.foldLeft(Future.successful(0, Seq[File]())) { (accumulator, queryGroup) =>
+      for {
+        acc <- accumulator
```
davidangb: you could do `(count, files) <- accumulator` or probably even `(count, _) <- accumulator` here, and avoid the ugly `acc._1` in the remainder of the code

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -1,79 +1,208 @@
...
+        entityBatch <- getEntityBatchFromQueries(entityQueries, workspaceNamespace, workspaceName, entityType)
+        membershipTSV <- (membershipWriter ? WriteMembershipTSV(acc._1, entityBatch)).mapTo[File]
+        entityTSV <- (entityWriter ? WriteEntityTSV(acc._1, entityBatch)).mapTo[File]
+        zip <- writeFilesToZip(entityType, membershipTSV, entityTSV)
+      } yield (acc._1 + 1, Seq(zip))
+    }
+    foldOperation map { files => files._2.head }
```
davidangb: architecturally, I'm not understanding why this is a `foldLeft`. What I'm seeing is:
* `streamEntities` generates the entire `Seq[EntityQuery]` that represents all of the paginated queries that get sent to rawls
* here, we `foldLeft` on that Seq ... but inside the function that we're folding with, we call `getEntityBatchFromQueries` and pass it the *entire* `Seq[EntityQuery]`.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -1,79 +1,208 @@
...
+  }
+
+  private def writeFilesToZip(entityType: String, membershipTSV: File, entityTSV: File): Future[File] = {
+    Future {
+      val zipFile = File.newTemporaryDirectory()
+      val membership = zipFile/s"${entityType}_membership.tsv"
+      val entity = zipFile/s"${entityType}_entity.tsv"
+      membershipTSV.copyTo(membership)
+      entityTSV.copyTo(entity)
```
davidangb: should these be `moveTo` to avoid leaving extra temp files behind?

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -1,79 +1,208 @@
...
+    val sortField = entityType + "_id"
+    val pages = Math.ceil(filteredCount.toDouble / pageSize.toDouble).toInt
+    (1 to pages) map { page =>
+      EntityQuery(page = page, pageSize = pageSize, sortField = sortField, sortDirection = SortDirections.Ascending, filterTerms = None)
+    }
+  }
+
+  private def getEntityTypeMetadata(workspaceNamespace: String, workspaceName: String, entityType: String): Future[EntityTypeMetadata] = {
+    rawlsDAO.getEntityTypes(workspaceNamespace, workspaceName).
+      map { metadata => metadata.get(entityType) }.
```
davidangb: you could probably just do:
```
map { _.getOrElse(entityType, throw new FireCloudExceptionWithErrorReport(...)) }
```
because the exception should be logged by default

src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingActor.scala
```diff
@@ -0,0 +1,53 @@
...
+
+  import StreamingActor._
+
+  def actorRefFactory: ActorContext = context
+
+  def receive: Receive = {
+
+    // send first chunk to client
+    case FirstChunk(httpData: HttpData, remaining: Int) =>
+      logger.debug(s"Writing first chunk: ${httpData.length}, remaining: $remaining")
```
davidangb: I'm sure these log statements helped during development, but I doubt they'll be helpful for debugging production issues, because they have too little info (and it would be convoluted to get more info into them)

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
```diff
@@ -4,74 +4,19 @@ import org.broadinstitute.dsde.rawls.model._
 import org.broadinstitute.dsde.firecloud.model._
 import org.broadinstitute.dsde.firecloud.service.TsvTypes
 
-import scala.collection.immutable
-import scala.util.{Success, Try}
-
-import spray.json.DefaultJsonProtocol._
+//import spray.json.DefaultJsonProtocol._
```
davidangb: remove if commented out

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/CookieAuthedApiService.scala
```diff
@@ -1,38 +1,40 @@
...
           post { requestContext =>
-            val filename = entityType + ".tsv"
             val attributeNames = attributeNamesString.map(_.split(",").toIndexedSeq)
             val userInfo = UserInfo("dummy", OAuth2BearerToken(tokenValue), -1, "dummy")
-            perRequest(requestContext, ExportEntitiesByTypeActor.props(exportEntitiesByTypeConstructor, userInfo),
-              ExportEntitiesByTypeActor.ExportEntities(workspaceNamespace, workspaceName, filename, entityType, attributeNames))
+            val exportProps: Props = ExportEntitiesByTypeActor.props(exportEntitiesByTypeConstructor, userInfo)
+            val exportMessage = ExportEntitiesByTypeActor.ExportEntities(requestContext, workspaceNamespace, workspaceName, entityType, attributeNames)
+            val exportActor = actorRefFactory.actorOf(exportProps)
+            exportActor ! exportMessage
```
davidangb: I'm curious why this changed to eliminate PerRequest?

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/ExportEntitiesApiService.scala
```diff
@@ -0,0 +1,38 @@
...
+import org.broadinstitute.dsde.firecloud.utils.StandardUserInfoDirectives
+import spray.routing.{HttpService, Route}
+
+import scala.language.postfixOps
+
+
+/**
+  * Created by grushton on 6/20/17.
+  */
+trait ExportEntitiesApiService extends HttpService with PerRequestCreator with FireCloudDirectives with FireCloudRequestBuilding with StandardUserInfoDirectives with LazyLogging {
```
davidangb: remove `PerRequestCreator` if you don't need it (but also see previous comment)

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/ExportEntitiesApiService.scala
```diff
@@ -0,0 +1,38 @@
...
+    path( "api" / "workspaces" / Segment / Segment / "entities" / Segment / "tsv" ) { (workspaceNamespace, workspaceName, entityType) =>
+      parameters('attributeNames.?) { attributeNamesString =>
+        requireUserInfo() { userInfo =>
+          get {
+            requestContext =>
+              val attributeNames = attributeNamesString.map(_.split(",").toIndexedSeq)
+              val exportProps: Props = ExportEntitiesByTypeActor.props(exportEntitiesByTypeConstructor, userInfo)
+              val exportMessage = ExportEntitiesByTypeActor.ExportEntities(requestContext, workspaceNamespace, workspaceName, entityType, attributeNames)
+              val exportActor = actorRefFactory.actorOf(exportProps)
+              exportActor ! exportMessage
```
davidangb: There's duplication between here and `CookieAuthedApiService.scala` ... but not enough to do anything about, imho. Maybe just add comments to both places to reference each other, so future developers know they need to change both places?

src/test/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeServiceSpec.scala
```diff
@@ -1,98 +1,248 @@
...
+
+  private def validateLineCount(chunks: List[MessageChunk], count: Int): Unit = {
+    val lineCount = chunks.map(c => scala.io.Source.fromString(c.data.asString).getLines().size).sum
+    lineCount should equal(count)
+  }
+
+  private def validateProps(entity: HttpEntity): Unit = {
+    val entityHeaderString = entity.asString
+    filterProps.map { h => entityHeaderString.contains(h) should be(true) }
+    missingProps.map { h => entityHeaderString.contains(h) should be(false) }
```
davidangb, line 281: I'd love to see one (or a few) unit tests that check what happens if one of the paginated queries to rawls fails. Maybe you could handle this in `MockRawlsDAO.queryEntitiesOfType` ... something like "if page == 3 then throw exception"?

davidangb (commented):
@rushtong second round of code review back to you - still haven't done extensive testing, that's next ...

project/Dependencies.scala
rushtong: 3.0.0 isn't in maven yet: https://mvnrepository.com/artifact/com.github.pathikrit

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
rushtong: I don't handle that explicitly. To date, I've relied on spray to do the right thing here. In the case of FirstChunk being the only thing we send, we have a single response:
`HttpResponse(entity = HttpEntity(contentType, httpData), headers = List(keepAlive, disposition))`  I don't know how Spray works internally with a large blob of `HttpData`.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
rushtong: I spent a lot of time trying to make the chunks an `ask` instead of a `tell` and failed for reasons I don't understand. If I could get the ask pattern to work here, it would definitely clarify the upstream handling.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/RawlsDAO.scala
rushtong: I was trying to touch `EntityService` as little as possible due to the weird bugs I saw when the export endpoint was inside of it. I can take a stab at refactoring it to use the dao for some basic uri calculations.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
rushtong: 👍

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
rushtong: Yeah, I see that. I'm not even using the `queryGroup`. Definitely need to refactor this.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
rushtong: Will try that out. During development, I remember running into trouble with `moveTo`, but will revisit.

src/main/resources/reference.conf
rushtong: This affects the size of allowable chunks coming from rawls. This number is balanced against the default rawls batch size too. The default is 1m, so bumping this up allows us to grab more content from Rawls in one request.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingActor.scala
rushtong: True.

project/Dependencies.scala
davidangb: I think it is: https://mvnrepository.com/artifact/com.github.pathikrit/better-files_2.12

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/CookieAuthedApiService.scala
rushtong: It wasn't being used any longer. If there's a way to wrap the streaming download into PerRequest, I'd love to see how to do it. I can't envision how that would work.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/ExportEntitiesApiService.scala
rushtong: 👍

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/ExportEntitiesApiService.scala
rushtong: 👍

src/test/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeServiceSpec.scala
rushtong: 👍

project/Dependencies.scala
rushtong: That's for scala 2.12 and up. We're still on 2.11.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
rushtong: 👍

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
rushtong: 👍

src/test/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeServiceSpec.scala
rushtong: I could use some help sorting through this case. In testing, the route gets completed/rejected more than once, which makes sense. We've started a response already, and then mid-stream, try to complete the same response with a firecloud error.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -1,79 +1,260 @@
...
   }
 
-  def constructor(app: Application)(userInfo: UserInfo)(implicit executionContext: ExecutionContext) =
-    new ExportEntitiesByTypeActor(app.rawlsDAO, userInfo)
+  def constructor(app: Application, materializer: ActorMaterializer)(userInfo: UserInfo)(implicit executionContext: ExecutionContext) =
+    new ExportEntitiesByTypeActor(app.rawlsDAO, userInfo, materializer)
 }
 
-class ExportEntitiesByTypeActor(val rawlsDAO: RawlsDAO, val userInfo: UserInfo)(implicit protected val executionContext: ExecutionContext) extends Actor with ExportEntitiesByType {
+class ExportEntitiesByTypeActor(val rawlsDAO: RawlsDAO, val argUserInfo: UserInfo, argMaterializer: ActorMaterializer)(implicit protected val executionContext: ExecutionContext) extends Actor with LazyLogging {
```
rtitle: I think you could probably define the implicit `UserInfo` and `ActorMaterializer` in the constructor like so:
```
class ExportEntitiesByTypeActor(val rawlsDAO: RawlsDAO)(implicit protected val executionContext: ExecutionContext, val userInfo: UserInfo, val materializer: ActorMaterializer) extends Actor with LazyLogging {
```
and remove the implicit vals on lines 48-49.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -1,79 +1,260 @@
...
+
+  // Requires its own actor context to work with downstream actors: TSVWriterActor and StreamingActor
+  def actorRefFactory: ActorContext = context
```
rtitle: Don't really see why this is needed -- can't you just use `context` directly where you're currently using `actorRefFactory`?

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -1,79 +1,260 @@
...
+        streamCollectionType(ctx, workspaceNamespace, workspaceName, entityType, entityQueries, metadata, attributeNames).onFailure {
+          case t: Throwable => handleStreamException(ctx, t)
+        }
+      } else {
+        val headers = TSVFormatter.makeEntityHeaders(entityType, metadata.attributeNames, attributeNames)
+        streamSingularType(ctx, workspaceNamespace, workspaceName, entityType, entityQueries, metadata, headers, attributeNames).onFailure {
+          case t: Throwable => handleStreamException(ctx, t)
+        }
+      }
+      Future(())
```
rtitle: I'm not sure this does what you want -- what's happening here is you're creating 2 _independent_ futures and only returning the second one. Thus the Future returned by this method is not attached in any way to the Futures returned by `streamCollectionType`/`streamSingularType`.

Maybe something like this would be preferable?
```
  def streamEntities(ctx: RequestContext, workspaceNamespace: String, workspaceName: String, entityType: String, attributeNames: Option[IndexedSeq[String]]): Future[Unit] = {
    getEntityTypeMetadata(workspaceNamespace, workspaceName, entityType) flatMap { metadata =>
      val entityQueries = getEntityQueries(metadata, entityType)

      val streamFuture = if (TSVFormatter.isCollectionType(entityType)) {
        streamCollectionType(ctx, workspaceNamespace, workspaceName, entityType, entityQueries, metadata, attributeNames)
      } else {
        val headers = TSVFormatter.makeEntityHeaders(entityType, metadata.attributeNames, attributeNames)
        streamSingularType(ctx, workspaceNamespace, workspaceName, entityType, entityQueries, metadata, headers, attributeNames)
      }

      streamFuture.mapTo[Unit].recover { case t: Throwable =>
        // Stream exceptions have to be handled by directly closing out the RequestContext responder
        handleStreamException(ctx, t)
      }
    }
  }.recoverWith {
    // Standard exceptions have to be handled as a completed request
    case t: Throwable => handleStandardException(ctx, t)
  }
```

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -1,79 +1,260 @@
...
+  /**
+    * General Approach
+    * 1. Define a `Source` of entity queries
+    * 2. Run the source events through a `Flow`.
+    * 3. Flow sends events (batch of entities) to a streaming output actor
+    * 4. Return a Done to the calling route when complete.
+    */
+  private def streamSingularType(ctx: RequestContext, workspaceNamespace: String, workspaceName: String, entityType: String, entityQueries: Seq[EntityQuery], metadata: EntityTypeMetadata, headers: IndexedSeq[String], attributeNames: Option[IndexedSeq[String]]): Future[Done] = {
+    // The output to the user
+    val streamingActorRef = actorRefFactory.actorOf(Props(new StreamingActor(ctx, ContentTypes.`text/plain`, entityType + ".txt")))
```
rtitle: This is pretty minor, but a usual Akka convention is to put a `props` method in the actor companion object, so you can instantiate it like this:
```
StreamingActor.props(ctx, ...)
```

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -1,79 +1,260 @@
...
+  private def streamCollectionType(ctx: RequestContext, workspaceNamespace: String, workspaceName: String, entityType: String, entityQueries: Seq[EntityQuery], metadata: EntityTypeMetadata, attributeNames: Option[IndexedSeq[String]]): Future[Done] = {
+
+    // Future of all entities
+    val entityBatches = Future.traverse(entityQueries) { query =>
+      getEntitiesFromQuery(workspaceNamespace, workspaceName, entityType, query)
+    } map (_.flatten)
+
+    // Two File sinks, one for each kind of entity set file needed.
+    // The temp files will end up zipped and streamed when complete.
+    val tempEntityFile: File = File.newTemporaryFile()
```
rtitle: Worth giving these tmp files a prefix? Like:
```
val tempEntityFile: File = File.newTemporaryFile(prefix = "entity_")
```
It might help with debugging down the road.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -1,79 +1,260 @@
...
+          })
+          val membershipFlow: FlowShape[Seq[Entity], ByteString] = builder.add(Flow[Seq[Entity]].map { entities =>
+            val rows = TSVFormatter.makeMembershipRows(entityType, entities)
+            ByteString(rows.map { _.mkString("\t")}.mkString("\n") + "\n")
+          })
+          val eConcat: UniformFanInShape[ByteString, ByteString] = builder.add(Concat[ByteString]())
+          val mConcat: UniformFanInShape[ByteString, ByteString] = builder.add(Concat[ByteString]())
+
+          // Graph
+          entityHeaderSource                         ~> eConcat
```
rtitle, line 207: Cool :)

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -1,79 +1,260 @@
...
+        lazy val streamingActorRef = actorRefFactory.actorOf(Props(new StreamingActor(ctx, ContentTypes.`application/octet-stream`, entityType + ".zip")))
+        zipFile map { f =>
+          streamingActorRef ! FirstChunk(HttpData.apply(f.byteArray), 0)
+        }
+      } else {
+        throw new FireCloudExceptionWithErrorReport(ErrorReport(s"Unable to collect entity metadata for $workspaceNamespace:$workspaceName:$entityType"))
+      }
+    }
+
+    Future(Done)
```
rtitle: Again this returns a completely unrelated Future to your operation. Maybe:
```
  private def streamCollectionType(...) = {
    // snip

    // And then map those files to a ZIP.
    fileStreamResult flatMap { s =>
      if (s) {
        val zipFile = writeFilesToZip(entityType, tempEntityFile, tempMembershipFile)
        // The output to the user
        lazy val streamingActorRef = actorRefFactory.actorOf(Props(new StreamingActor(ctx, ContentTypes.`application/octet-stream`, entityType + ".zip")))
        zipFile map { f =>
          streamingActorRef ! FirstChunk(HttpData.apply(f.byteArray), 0)
        }
      } else {
        Future.failed(new FireCloudExceptionWithErrorReport(ErrorReport(s"Unable to collect entity metadata for $workspaceNamespace:$workspaceName:$entityType")))
      }
    }.mapTo[Done]
  }
```

rtitle: ... or just have this method and `streamSingularType` return a `Future[Unit]`, not sure we need `Future[Done]`.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -1,79 +1,260 @@
...
+    // Check that each file is completed
+    val fileStreamResult = for {
+      eResult <- fileStreamIOResults._1
+      mResult <- fileStreamIOResults._2
+    } yield eResult.wasSuccessful && mResult.wasSuccessful
+
+    // And then map those files to a ZIP.
+    fileStreamResult map { s =>
+      if (s) {
+        val zipFile = writeFilesToZip(entityType, tempEntityFile, tempMembershipFile)
```
rtitle: Also maybe you should add a recovery to `writeFilesToZip` since it's doing file I/O and could potentially throw an `IOException`. Might want to map errors to a `FireCloudExceptionWithErrorReport`.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingActor.scala
```diff
@@ -0,0 +1,51 @@
...
+  case class NextChunk(httpData: HttpData, remaining: Int)
+  case class Ok(remaining: Int)
+}
+
+class StreamingActor(ctx: RequestContext, contentType: ContentType, fileName: String) extends Actor with HttpService with LazyLogging {
+
+  private val keepAlive = HttpHeaders.Connection("Keep-Alive")
+  private val disposition = HttpHeaders.`Content-Disposition`.apply("attachment", Map("filename" -> fileName))
+
+  import StreamingActor._
```
rtitle: Minor, maybe put this import at the top level.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingActor.scala
```diff
@@ -0,0 +1,51 @@
...
+
+  def actorRefFactory: ActorContext = context
```
rtitle: Again, wondering if this is needed

src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingActor.scala
```diff
@@ -0,0 +1,51 @@
...
+    // This case comes back from the client
+    // all chunks were sent. stop.
+    case Ok(remaining: Int) =>
+      if (remaining == 0) {
+        ctx.responder ! ChunkedMessageEnd
+        context.stop(self)
+      }
+
+    //
+    case x =>
```
rtitle: Could probably remove this, the superclass calls `unhandled` on unhandled messages (and I think logs them as well).

rtitle:
Nice! Made a few comments, see above. The akka-stream stuff is cool :)

project/Dependencies.scala
rtitle: I'd love to upgrade to 2.12 but unfortunately there's no spray build for 2.12, so upgrading to akka-http is a dependency..

davidangb: yeah, my bad on that, I wasn't paying attention to scala versions!

src/main/resources/reference.conf
davidangb: got it, thanks for the explanation.

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
```diff
@@ -59,11 +60,29 @@ object FireCloudConfig {
...
     val notificationsPath = workspace.getString("notificationsPath")
     val notificationsUrl = authUrl + notificationsPath
+    val defaultPageSize = rawls.getInt("defaultPageSize")
 
     def entityPathFromWorkspace(namespace: String, name: String) = authUrl + entitiesPath.format(namespace, name)
     def entityQueryPathFromWorkspace(namespace: String, name: String) = authUrl + entityQueryPath.format(namespace, name)
     def importEntitiesPathFromWorkspace(namespace: String, name: String) = authUrl + importEntitiesPath.format(namespace, name)
     def overwriteGroupMembershipUrlFromGroupName(groupName: String) = authUrl + overwriteGroupMembershipPath.format(groupName)
+    def entityQueryUriFromWorkspaceAndQuery(workspaceNamespace: String, workspaceName: String, entityType: String, query: Option[EntityQuery] = None): Uri = {
+      val baseEntityQueryUri = Uri(entityQueryPathFromWorkspace(workspaceNamespace, workspaceName) + s"/$entityType")
```
davidangb: nitpick, and not a blocker: this is a weird use of string interpolation. I'd do either:
`entityQueryPathFromWorkspace(workspaceNamespace, workspaceName) + "/" + entityType`
or
`s"${entityQueryPathFromWorkspace(workspaceNamespace, workspaceName)}/$entityType"`
but not mix the two approaches.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingActor.scala
davidangb: `spray.routing.HttpService` requires a `def actorRefFactory` implementation ... but I don't think this actor needs to mix in `HttpService` at all!

src/test/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeServiceSpec.scala
davidangb: I think this is the nature of http chunking, though, isn't it? Each chunk sends its own response code/payload.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -1,79 +1,260 @@
...
+  // Requires its own actor context to work with downstream actors: TSVWriterActor and StreamingActor
+  def actorRefFactory: ActorContext = context
+
+  implicit val timeout: Timeout = Timeout(1 minute)
+  implicit val userInfo: UserInfo = argUserInfo
+  implicit val materializer: ActorMaterializer = argMaterializer
+
   override def receive: Receive = {
-    case ExportEntities(workspaceNamespace, workspaceName, filename, entityType, attributeNames) => exportEntities(workspaceNamespace, workspaceName, filename, entityType, attributeNames) pipeTo sender
+    case ExportEntities(ctx, workspaceNamespace, workspaceName, entityType, attributeNames) => streamEntities(ctx, workspaceNamespace, workspaceName, entityType, attributeNames) pipeTo sender
```
davidangb: Would it be possible to move workspaceNamespace, workspaceName, entityType, attributeNames (and possibly ctx) into `props` / constructor? If those exist at the instance level, then you wouldn't have to pass them from `streamEntities` to `streamCollectionType`/`streamSingularType ` to `getEntityTypeMetadata` and `getEntitiesFromQuery` ... etc. - it would clean up a bunch of signatures.

src/main/resources/reference.conf
davidangb: based on the release-debugging I just did, this needs to move out of reference.conf and into firecloud-orchestration.conf, so it won't get clobbered by spray defaults. Note that our other spray config overrides are in firecloud-orchestration.conf too.

The way to test this is to call `sbt assembly` and then run the resultant jar, instead of running from the sbt shell.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
davidangb: yeah, this is some wack magic dsl!

davidangb (commented):
a few more comments inline, getting very close!

src/test/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeServiceSpec.scala
rushtong: Pretty sure my question is resolved. It's true that once a response is started via the chunking process, we no longer have control over the response code. What I chose to do, after we start streaming, is to put any error that we get from either Rawls or Orchestration into the output. That way, there is at least some indication for the user that something went wrong. The tests we have check for this too.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
rushtong: I'd love to simplify this, but I can't get the syntax/compilation to work right. Making those two vals implicit breaks the constructor in the companion object. Will look for a way to make this work.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
rushtong: This compiles `streamFuture.mapTo[Unit]` but fails with a runtime error `Error generating entity download: Cannot cast akka.Done$ to scala.runtime.BoxedUnit[info]` :-( 
The only way I can think of to get around this is to explicitly return a `Future[Unit]`

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
rushtong: @davidangb - Let's discuss this when you can. I'd love to simplify the construction here, but I'm completely clueless about how. Rob's suggestion wrt implicits is also relevant here - I don't have the scala chops to make this convoluted construction work any other way than it does now.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -1,79 +1,272 @@
...
+    // Check that each file is completed
+    val fileStreamResult = for {
+      eResult <- fileStreamIOResults._1
+      mResult <- fileStreamIOResults._2
+    } yield eResult.wasSuccessful && mResult.wasSuccessful
+
+    // And then map those files to a ZIP.
+    val zipResult = fileStreamResult flatMap { s =>
+      if (s) {
+        val zipFile: Future[File] = writeFilesToZip(tempEntityFile, tempMembershipFile)
```
aednichols, line 224: Interesting, for some reason I thought that the streaming architecture would not wait until the file is complete, but rather start streaming to the user as soon as possible while entities are still being retrieved. But I guess we don't start streaming to the user until the file is complete (line 219)?

What is it about this mechanism that helps us get around the timeout?

aednichols:
I think it would be helpful to add a brief "why?" comment to explain the reasoning for choosing this architecture. Maintainers might think it's too complicated, when in fact it's exactly right to solve a hard problem.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
rushtong: There are two cases - singular and set types. 
* For singular entity downloads, we do start streaming right away since we don't have to do any file operations. This case is where we can avoid timeout problems with large sets. 
* For set types, we can't stream directly since we need to write files first. Once those files are written, we can then send them. I'm pretty sure there's no way to both start writing the files *AND* stream them at the same time. This case does *not* allow us to avoid timeouts. However, timeouts are not really a problem in this case because set downloads are generally *much* smaller (orders of magnitude smaller) than singular entities.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -64,26 +77,23 @@ class ExportEntitiesByTypeActor(val rawlsDAO: RawlsDAO, val argUserInfo: UserInf
     *
     * Handle exceptions directly by completing the request.
     */
-  def streamEntities(ctx: RequestContext, workspaceNamespace: String, workspaceName: String, entityType: String, attributeNames: Option[IndexedSeq[String]]): Future[Unit] = {
-    getEntityTypeMetadata(workspaceNamespace, workspaceName, entityType) flatMap { metadata =>
+  def streamEntities(): Future[Unit] = {
+    getEntityTypeMetadata(entityType) flatMap { metadata =>
```
davidangb: could remove `entityType` argument here and read from the class level

davidangb (approved):
one small tweak inline, otherwise 👍 !!!!!!!!!!!11!1!1!!

davidangb (approved):

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
rushtong: Will do. Also adding a little class-level doc to describe why this approach.

PR #489, opened 2017-08-16
TSV downloads performance improvement [GAWB-2350]

aednichols:
Orchestration specifies a sort field in its queries to Rawls, while Alex's Python code does not. Rawls does an extra join when the sort is applied, blowing up execution time.

We can [avoid the join by sorting specifically on entity name](https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala#L138). It's already the default in at least one [place](https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/webservice/EntityApiService.scala#L41).

I don't suppose the sort order matters in the final product, but the MD5 of the 20k workspace file is the same before and after this PR (`c38e074a85d88537b8327e60ee1d005b`).

The running time for 20k is down from ~10 minutes to ~2 minutes.

---
This PR represents the smallest possible change. We could also...
1. Make `sortField` and `sortDirection` into `Option` types on `EntityQuery`, with more far-reaching consequences.
2. Overload `entityQueryUriFromWorkspaceAndQuery` in Orchestration with a version that takes a `Map` instead of an `EntityQuery` to construct the URL.

---
  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off
- [x] **Submitter**: Inform other teams of any API changes via Slack and/or email

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
@@ -254,7 +254,7 @@ class ExportEntitiesByTypeActor(rawlsDAO: RawlsDAO,
   private def getEntityQueries(metadata: EntityTypeMetadata, entityType: String): Seq[EntityQuery] = {
     val pageSize = FireCloudConfig.Rawls.defaultPageSize
     val filteredCount = metadata.count
-    val sortField = entityType + "_id"
+    val sortField = "name" // Anything else and the Rawls execution time blows up with a join
     val pages = Math.ceil(filteredCount.toDouble / pageSize.toDouble).toInt
     (1 to pages) map { page =>
       EntityQuery(page = page, pageSize = pageSize, sortField = sortField, sortDirection = SortDirections.Ascending, filterTerms = None)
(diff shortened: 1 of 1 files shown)


rushtong:
The time savings is exactly what Alex was hoping for. That buys us a good deal of time until we start hitting the current 15 minute LB timeout settings.

rushtong (approved):
👍  LGTM, and also for keeping this small. I prefer that over the more complicated model changes.

rushtong:
Any chance we can debug this failing travis test:
```
[info] HttpOntologyDAOConcurrencySpec:
[info] HttpOntologyDAO concurrency
[info] - should obey the spray max-clients throttle *** FAILED ***
[info]   Expected 8, but got 4 request count should have reached its maximum throttle limit (HttpOntologyDAOConcurrencySpec.scala:68)
```

aednichols:
I inspected that test and couldn't come up with much; could not get it to fail again. It seems unlikely that out of 40 requests (`throttleCount * 5`) 36 would somehow get lost, so maybe the problem is with counting those received?

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -254,7 +254,7 @@ class ExportEntitiesByTypeActor(rawlsDAO: RawlsDAO,
   private def getEntityQueries(metadata: EntityTypeMetadata, entityType: String): Seq[EntityQuery] = {
     val pageSize = FireCloudConfig.Rawls.defaultPageSize
     val filteredCount = metadata.count
-    val sortField = entityType + "_id"
+    val sortField = "name" // Anything else and the Rawls execution time blows up with a join
```
ahaessly: looks good. maybe reference the ticket number?

ahaessly (approved):

PR #500, opened 2017-09-08
Un-publish on Workspace Delete

rushtong:
## Addresses 
https://broadinstitute.atlassian.net/browse/GAWB-2352
Requires corresponding UI PR: https://github.com/broadinstitute/firecloud-ui/pull/931

## Changes:
Update the delete workspace logic to include un-publishing from Library if necessary.

## Testing:
* Delete a normal, un-published workspace and ensure no regressions
* Delete a published workspace and ensure that it does not show up in Library.
* Try to delete a published workspace with users that are not allowed.
  * Workspace "READER"
  * Workspace "OWNER" but not "CATALOG" permissions
  * "CATALOG" user, but not "OWNER"

---

- [x] I've followed [the instructions](../CONTRIBUTING.md#api-changes) if I've made any changes to the API, _especially_ if they're breaking changes
- [x] I've updated the [FISMA documentation](../CONTRIBUTING.md#fisma-documentation-changes) if I've made any security-related changes, including auth, encryption, or auditing

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspacePublishingSupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspacePublishingSupport.scala
@@ -1,10 +1,13 @@
 package org.broadinstitute.dsde.firecloud.service
 
+import org.broadinstitute.dsde.firecloud.FireCloudException
 import org.broadinstitute.dsde.firecloud.dataaccess.{OntologyDAO, RawlsDAO, SearchDAO}
+import org.broadinstitute.dsde.firecloud.model.WithAccessToken
 import org.broadinstitute.dsde.firecloud.service.LibraryService.publishedFlag
-import org.broadinstitute.dsde.rawls.model.{AttributeBoolean, Workspace, WorkspaceResponse}
+import org.broadinstitute.dsde.rawls.model.{AttributeBoolean, ErrorReport, Workspace, WorkspaceResponse}
 
-import scala.concurrent.ExecutionContext
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.Try
 
 trait WorkspacePublishingSupport extends LibraryServiceSupport {
 
@@ -35,6 +38,19 @@ trait WorkspacePublishingSupport extends LibraryServiceSupport {
 
   def isPublished(workspace: Workspace): Boolean = {
     workspace.attributes.get(publishedFlag).fold(false)(_.asInstanceOf[AttributeBoolean].value)
+  }
+
+  def setWorkspacePublishedStatus(ws: Workspace, publishArg: Boolean, rawlsDAO: RawlsDAO, ontologyDAO: OntologyDAO, searchDAO: SearchDAO)(implicit userToken: WithAccessToken): Future[Workspace] = {
+    rawlsDAO.updateLibraryAttributes(ws.namespace, ws.name, updatePublishAttribute(publishArg)) map { workspace =>
+      val docPublishResult = if (publishArg)
+        Try(publishDocument(workspace, ontologyDAO, searchDAO)).isSuccess
+      else
+        Try(removeDocument(workspace, searchDAO)).isSuccess
+      if (docPublishResult)
+        workspace
+      else
+        throw new FireCloudException(s"Unable to update this workspace, ${ws.namespace}:${ws.name}, to $publishArg in elastic search.")
+    }
   }
 
 }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspacePublishingSupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspacePublishingSupport.scala
@@ -42,14 +42,19 @@ trait WorkspacePublishingSupport extends LibraryServiceSupport {
 
   def setWorkspacePublishedStatus(ws: Workspace, publishArg: Boolean, rawlsDAO: RawlsDAO, ontologyDAO: OntologyDAO, searchDAO: SearchDAO)(implicit userToken: WithAccessToken): Future[Workspace] = {
     rawlsDAO.updateLibraryAttributes(ws.namespace, ws.name, updatePublishAttribute(publishArg)) map { workspace =>
-      val docPublishResult = if (publishArg)
-        Try(publishDocument(workspace, ontologyDAO, searchDAO)).isSuccess
-      else
-        Try(removeDocument(workspace, searchDAO)).isSuccess
+      val docPublishResult = Try {
+        if (publishArg)
+          publishDocument(workspace, ontologyDAO, searchDAO)
+        else
+          removeDocument(workspace, searchDAO)
+      }.isSuccess
       if (docPublishResult)
         workspace
-      else
+      else {
+        val message = s"Unable to update this workspace, ${ws.namespace}:${ws.name}, to $publishArg in elastic search."
+        logger.error(message)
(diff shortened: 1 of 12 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -222,6 +218,30 @@ class WorkspaceService(protected val argUserToken: WithAccessToken, val rawlsDAO
...
+          case e => RequestComplete(InternalServerError, ErrorReport(message = e.getMessage))
+        }
+      } else {
+        rawlsDAO.deleteWorkspace(ws.workspace.namespace, ws.workspace.name) map { response =>
+          RequestComplete(response)
+        }
+      }
+    }
+  }
+
```
davidangb: this duplicates a bunch of logic with `LibraryService.setWorkspaceIsPublished`. Could you call that method instead of duplicating here? You might have to move `setWorkspaceIsPublished` into `WorkspacePublishingSupport` or `LibraryServiceSupport` to achieve the reuse.

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -222,6 +218,30 @@ class WorkspaceService(protected val argUserToken: WithAccessToken, val rawlsDAO
...
+        rawlsDAO.updateLibraryAttributes(ns, name, updatePublishAttribute(false)) flatMap { f1 =>
+          Future(removeDocument(ws.workspace, searchDAO)).flatMap { f2 =>
+            rawlsDAO.deleteWorkspace(ns, name) map { f3 =>
+              RequestComplete(f3.copy(message = Some(f3.message.getOrElse("") + unPublishSuccessMessage(ns, name))))
+            }
+          }
+        } recover {
+          case e: FireCloudExceptionWithErrorReport => RequestComplete(e.errorReport.statusCode.getOrElse(InternalServerError), e.errorReport)
+          case e => RequestComplete(InternalServerError, ErrorReport(message = e.getMessage))
+        }
```
davidangb: by replaying the error message from rawls, the end user - in the case where the user is owner but not curator - gets a message "You must be a curator and either be an owner or have catalog with read+." That's correct, but not necessarily helpful. Maybe change to something like `s"You cannot delete this workspace: ${e.getMessage}"`

davidangb (commented):
two comments inline.

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
rushtong: I started out on that road, but that method is doing much more than what I want to do and is completing the request in cases that I don't want it to. I think I can extract out the last `else` condition in that method to look like what I have.

MattPutnam (approved):
👍 besides what David said

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
rushtong: @davidangb - Since `setWorkspaceIsPublished` has a good chunk of preliminary logic checking that this feature doesn't need to go through, I ended up extracting what I needed from `setWorkspaceIsPublished` into another method in `WorkspacePublishingSupport`. Please have a second look.

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -222,6 +218,28 @@ class WorkspaceService(protected val argUserToken: WithAccessToken, val rawlsDAO
...
+          case e: FireCloudExceptionWithErrorReport => RequestComplete(e.errorReport.statusCode.getOrElse(InternalServerError), ErrorReport(message = s"You cannot delete this workspace: ${e.errorReport.message}"))
+          case e: Throwable => RequestComplete(InternalServerError, ErrorReport(message = s"You cannot delete this workspace: ${e.getMessage}"))
+        }
+      } else {
+        rawlsDAO.deleteWorkspace(ws.workspace.namespace, ws.workspace.name) map { response =>
+          RequestComplete(response)
+        }
+      }
+    }
+  }
```
davidangb: The other way I've seen this kind of logic flow handled is:
```
      val unpublishFuture = if (isPublished(ws))
        setWorkspacePublishedStatus(ws.workspace, publishArg = false, rawlsDAO, ontologyDAO, searchDAO)
      else
        Future.successful(ws)

      unpublishFuture flatMap { _ =>
        rawlsDAO.deleteWorkspace(ns, name) map { wsResponse =>
          RequestComplete(wsResponse.copy(message = Some(wsResponse.message.getOrElse("") + unPublishSuccessMessage(ns, name))))
        } recover {
          case e: FireCloudExceptionWithErrorReport => RequestComplete(e.errorReport.statusCode.getOrElse(InternalServerError), ErrorReport(message = s"You cannot delete this workspace: ${e.errorReport.message}"))
          case e: Throwable => RequestComplete(InternalServerError, ErrorReport(message = s"You cannot delete this workspace: ${e.getMessage}"))
        }
      }
```
I think it's slightly cleaner, but I'm not officially requesting a change for this PR; your call if you want to adopt different syntax or not.

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspacePublishingSupport.scala
```diff
@@ -37,4 +40,17 @@ trait WorkspacePublishingSupport extends LibraryServiceSupport {
...
+    rawlsDAO.updateLibraryAttributes(ws.namespace, ws.name, updatePublishAttribute(publishArg)) map { workspace =>
+      val docPublishResult = if (publishArg)
+        Try(publishDocument(workspace, ontologyDAO, searchDAO)).isSuccess
+      else
+        Try(removeDocument(workspace, searchDAO)).isSuccess
+      if (docPublishResult)
+        workspace
+      else
+        throw new FireCloudException(s"Unable to update this workspace, ${ws.namespace}:${ws.name}, to $publishArg in elastic search.")
+    }
```
davidangb: scope creep warning!

If the `updateLibraryAttributes ` succeeds, but the `Try(publishDocument)` or `Try(removeDocument)` fails, we're out of sync. Now that you have these in a `Try`, it would be easy enough to look for failures and reset the published attribute. This is GAWB-1683, so feel free to push it off.

davidangb (approved):
two comments inline; both optional. I like the refactoring! :+1:

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
rushtong: 👍  I like that approach better, more granular and less nest-y.

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspacePublishingSupport.scala
rushtong: Yes, that is a problem. I'm going to add some error logging around this so we can get alerted to it, but leave the real fix for 1683.

davidangb (approved):
:+1:

rushtong:
Getting totally unrelated travis errors. Need to sort it out before merging.
```
[info] HttpOntologyDAO concurrency
[info] - should obey the spray max-clients throttle *** FAILED ***
[info]   Expected 8, but got 4 request count should have reached its maximum throttle limit (HttpOntologyDAOConcurrencySpec.scala:68)
```

PR #502, opened 2017-09-11
Fix Profile definition in api-docs.yaml

dmohs:
The definition was badly out-of-date.

- [x] I've followed [the instructions](../CONTRIBUTING.md#api-changes) if I've made any changes to the API, _especially_ if they're breaking changes
- [x] I've updated the [FISMA documentation](../CONTRIBUTING.md#fisma-documentation-changes) if I've made any security-related changes, including auth, encryption, or auditing

In all cases:

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -4517,7 +4517,7 @@ definitions:
         description: User's title
       contactEmail:
         type: string
-        description: User's title
+        description: User's contact email (if different from account email)
       institute:
         type: string
         description: User's home institution
@@ -4535,7 +4535,7 @@ definitions:
         description: User's program location country
       pi:
         type: string
-        description: Principle Investigator
+        description: Principal Investigator
       nonProfitStatus:
         type: string
         description: User's program non-profit status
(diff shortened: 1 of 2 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -4506,18 +4506,50 @@ definitions:
...
+      programLocationState:
+        type: string
+        description: User's program location state
+      programLocationCountry:
+        type: string
+        description: User's program location country
       pi:
         type: string
-        description: Primary Investigator
+        description: Principle Investigator
```
danqrodney: Principal?

src/main/resources/swagger/api-docs.yaml
```diff
@@ -4506,18 +4506,50 @@ definitions:
...
+        type: string
+        description: User's last name
+      title:
         type: string
-        description: User's email address
-      institution:
+        description: User's title
+      contactEmail:
+        type: string
+        description: User's title
```
danqrodney: Wrong description

dmohs:
@dvoet says: If this is right, merge it. I'm as confident as I can be that this is right.

PR #503, opened 2017-09-20
GAWB-2601: passthrough for ga4gh tool descriptor api in agora

davidangb:
see broadinstitute/agora#212 - this just adds the required passthrough to orchestration.

In all cases:

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -3315,6 +3315,57 @@ paths:
         200:
           description: OK
 
+  /ga4gh/tools/{id}/versions/{version-id}/{type}/descriptor:
+    get:
+      summary: Get the tool descriptor (CWL/WDL) for the specified tool.
+      description: Returns the CWL or WDL descriptor for the specified tool.
+      tags:
+        - GA4GH Tool Registry
+      parameters:
+        - name: type
+          required: true
+          in: path
+          description: >
+            The output type of the descriptor. If not specified it is up to the underlying
+            implementation to determine which output type to return. Plain types return the
+            bare descriptor while the "non-plain" types return a descriptor wrapped
+            with metadata.
+
+              *In FireCloud, only WDL and plain-WDL are supported.*
+          type: string
+          enum:
+            - CWL
+            - WDL
+            - plain-CWL
+            - plain-WDL
+        - name: id
+          in: path
+          description: >
+            A unique identifier of the tool, scoped to this registry, for example `123456`.
+
+              *In FireCloud, this must be a namespace + ":" + name. For instance, if your namespace
+              is 'foo' and name is 'bar', this must be 'foo:bar'.*
+          required: true
+          type: string
+        - name: version-id
+          in: path
+          required: true
+          type: string
+          description: >
+            An identifier of the tool version for this particular tool registry, for example `v1`.
+
+              *In FireCloud, this must be an integer representing the FireCloud snapshot id.*
+      responses:
+        '200':
+          description: The tool descriptor.
+          schema:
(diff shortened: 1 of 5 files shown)


src/test/resources/reference.conf
```diff
@@ -70,7 +70,7 @@ duos {
 # these are the settings currently used at runtime; copy them here
 # so we're testing under similar conditions.
 spray.can.host-connector {
-  max-connections = 8
+  max-connections = 4
   max-retries = 5
-  pipelining = true
+  pipelining = off
```
davidangb, line 75: these changes are not associated with this ticket, but they fix the flaky HttpOntologyDAO concurrency unit test. We made changes a while back to the runtime conf, but didn't keep the reference.conf in sync.

I don't know why unit tests pass locally (for me) but fail in Travis - either way, this seems like a decent change to make.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -3315,6 +3315,57 @@ paths:
...
+          description: >
+            The output type of the descriptor. If not specified it is up to the underlying
+            implementation to determine which output type to return. Plain types return the
+            bare descriptor while the "non-plain" types return a descriptor wrapped
+            with metadata.
+
+              *In FireCloud, only WDL and plain-WDL are supported.*
+          type: string
+          enum:
+            - CWL
```
aednichols, line 3337: I see there is a "this service only supports" message, but seems strange to me to suggest unsupported formats in the dropdown

src/main/resources/swagger/api-docs.yaml
```diff
@@ -3315,6 +3315,57 @@ paths:
...
+            A unique identifier of the tool, scoped to this registry, for example `123456`.
+
+              *In FireCloud, this must be a namespace + ":" + name. For instance, if your namespace
+              is 'foo' and name is 'bar', this must be 'foo:bar'.*
+          required: true
+          type: string
+        - name: version-id
+          in: path
+          required: true
+          type: string
```
aednichols, line 3353: Maybe `type: integer`? Depends on whether we're replicating the spec or describing our implementation, I lean mildly towards the latter since it is a little more restrictive

src/main/resources/swagger/api-docs.yaml
```diff
@@ -3315,6 +3315,57 @@ paths:
...
+          description: >
+            An identifier of the tool version for this particular tool registry, for example `v1`.
+
+              *In FireCloud, this must be an integer representing the FireCloud snapshot id.*
+      responses:
+        '200':
+          description: The tool descriptor.
+          schema:
+            $ref: '#/definitions/ToolDescriptor'
```
aednichols, line 3362: This looks right syntactically, but I can't find it in the UI

aednichols (approved):
LGTM, noted some minor things in Swagger

src/main/resources/swagger/api-docs.yaml
davidangb: For this - and for the WDL/CWL enum comment before - I kept the swagger definitions as close as possible to the GA4GH spec, and only added comments about the differences. This was intentional, to show our compliance with the interface, even if our underlying implementation doesn't support everything. Think that makes sense?

src/main/resources/swagger/api-docs.yaml
davidangb: Do you mean in the swagger UI? It appears as the model for the 200 response, above the parameters. If you meant the FireCloud UI, that's correct - it's not used there.

aednichols: Ah yes, missed it because the Swagger examples aren't labelled with the object (have to switch to model)

PR #506, opened 2017-09-27
GAWB-2579: passthrough query params for redacting methods while editing

davidangb:
Three commits in this PR:
1. The actual bug: make sure to pass query params to agora for the edit endpoint
2. Total rewrite of the MethodsApiServiceSpec unit tests, which were woefully inadequate. This includes a fix to the runtime routes for issues found in testing
3. Move MethodsApiServiceSpec from the service package to the webservice package

Note: I am currently unable to verify this locally due to problems with sam.

I chose to go down the implementation route of only passing query params where absolutely necessary, with the principle of exposing the smallest surface area possible for security. The other option would have been to rewrite `passthrough` or provide another `passthrough` option to always pass query params; I didn't like that because it has implications elsewhere.

**Please see my comment on the Jira ticket for testing locally!!!**

----

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/MethodsApiService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/MethodsApiService.scala
@@ -38,7 +38,12 @@ trait MethodsApiService extends HttpService
         (get | post) {
           extract(_.request.method) { method =>
             extract(_.request.uri.query) { query =>
-              passthrough(Uri(passthroughBase).withQuery(query), method)
+              // only pass query params for GETs
+              val targetUri = if (method == HttpMethods.GET)
+                Uri(passthroughBase).withQuery(query)
+              else
+                Uri(passthroughBase)
+              passthrough(targetUri, method)
             }
           }
         }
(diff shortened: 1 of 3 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/webservice/MethodsApiService.scala
```diff
@@ -38,7 +38,12 @@ trait MethodsApiService extends HttpService
         (get | post) {
           extract(_.request.method) { method =>
             extract(_.request.uri.query) { query =>
-              passthrough(Uri(passthroughBase).withQuery(query), method)
+              // only pass query params for GETs
```
rushtong, line 41: I am worried that we're doing some non-standard things with POSTs and query params, but I don't have any clear evidence on hand.

rushtong (approved):
👍 LGTM - love the test refactor.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/MethodsApiService.scala
davidangb: we certainly don't document it in swagger for this endpoint ...

PR #507, opened 2017-09-29
Agora snapshot comments [GAWB-2032]

aednichols:
Requires https://github.com/broadinstitute/agora/pull/214

In all cases:


src/main/resources/swagger/api-docs.yaml
```diff
@@ -221,6 +221,10 @@ paths:
           description: Snapshot ID of configuration.
           required: false
           type: number
+        - name: snapshotComment
```
rushtong, line 224: This doesn't appear to be `required`, similar to the Agora PR. Can this be annotated similarly?

rushtong (approved):
👍  One minor suggestion inline.

src/main/resources/swagger/api-docs.yaml
rushtong: My bad, this is required - ignore comment. Pretty sure this also constitutes a breaking change to the API - we will need to handle that with a release email.

aednichols: It should not be `required`, snapshot comments are designed to be very optional. Will fix & check consistency across PRs.

rushtong: 👍  - Yes, as I dig more into the agora PR, it is becoming more clear, thanks!

src/main/resources/swagger/api-docs.yaml
```diff
@@ -975,6 +979,10 @@ paths:
           description: Snapshot ID of method.
           required: false
           type: number
+        - name: snapshotComment
+          in: query
+          type: string
+          description: Snapshot comment of AgoraEntity
         - name: synopsis
           in: query
           description: Synopsis of method.
```
davidangb: need to add snapshotComment to the `POST /api/methods/{namespace}/{name}/{snapshotId}` endpoint, around line 1192. That's the copy-snapshot endpoint.

davidangb (commented):
did not test; code looks good modulo one comment inline.

src/main/resources/swagger/api-docs.yaml
aednichols: The default value of `required` is `false`, so this was correct, but I'm going to make it explicit.

davidangb (approved):
👍 updates look good to me. I am not at a place where I can run this locally yet, sorry!

rushtong (approved):
👍  LGTM

PR #513, opened 2017-10-13
GAWB-849: point to updated agora url

rushtong:
Partially addresses https://broadinstitute.atlassian.net/browse/GAWB-849 
Required for https://github.com/broadinstitute/agora/pull/218

Changes the public-facing API to point to the same version that exists in the agora PR.

---

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/Ga4ghApiService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/Ga4ghApiService.scala
@@ -15,11 +15,13 @@ trait Ga4ghApiService extends HttpService with FireCloudDirectives {
 
   val ga4ghRoutes: Route =
     pathPrefix("ga4gh") {
-      pathPrefix("tools") {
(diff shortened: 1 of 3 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/webservice/Ga4ghApiService.scala
```diff
@@ -18,7 +18,7 @@ trait Ga4ghApiService extends HttpService with FireCloudDirectives {
       pathPrefix("tools") {
         path (Segment / "versions" / Segment / Segment / "descriptor") { (id, versionId, descriptorType) =>
           get {
-            val targetUri = Uri(s"$agora/ga4gh/tools/$id/versions/$versionId/$descriptorType/descriptor")
+            val targetUri = Uri(s"$agora/ga4gh/v1/tools/$id/versions/$versionId/$descriptorType/descriptor")
```
davidangb: please move orch's api to /ga4gh/v1 too! Yes this is technically a breaking API change, but we warned people during the demo, and we know that nobody is using it yet - it hasn't even been released yet.

davidangb (approved):
looks right to me, did not test :+1:

PR #516, opened 2017-10-18
GAWB-1303: Remove passthroughAllPaths

kyuksel:

In all cases:


src/main/scala/org/broadinstitute/dsde/firecloud/webservice/NotificationsApiService.scala
```diff
@@ -3,14 +3,27 @@ package org.broadinstitute.dsde.firecloud.webservice
...
+    get {
+      pathPrefix(separateOnSlashes(ApiPrefix)) {
+        unmatchedPath { remaining =>
+          val encodedTargetUri = Uri(encodeUri(FireCloudConfig.Rawls.notificationsUrl + remaining))
+          path(General) {
+            passthrough(encodedTargetUri, HttpMethods.GET)
+          } ~
+          path(Workspace / Segment / Segment) { (_, _) =>
+            passthrough(encodedTargetUri, HttpMethods.GET)
+          }
```
kyuksel: Instead of
```
path(General) {
  passthrough(encodedTargetUri, HttpMethods.GET)
} ~
path(Workspace / Segment / Segment) { (_, _) =>
  passthrough(encodedTargetUri, HttpMethods.GET)
}
```
I wanted to have something like 
```
path(General) | path(Workspace / Segment / Segment) {
  passthrough(encodedTargetUri, HttpMethods.GET)
}
```
but types don't work out. I'm guessing it's possible to have a `regex` on the URIs but not sure if it's worth it.

@davidangb

davidangb:
Looks good so far, with two tweaks. I reviewed at the design level, not at a detailed syntax level.

1. I'd love to see negative unit tests that check that post, put, etc are rejected for these routes. We have examples of that in a lot of places. The positive tests you modified are good.

2. I don't like the use of `unmatchedPath` inside the route. I 100% agree it is functionally correct and that it makes for concise code. However, the purpose of this refactor is to remove that kind of genericness and write out the routing a bit more longhand to avoid confusion ... when longhand, the routes are more self-documenting and easy to understand. So I'd prefer something like:
```
  final val notificationsRoutes: Route = {
    get {
      pathPrefix(separateOnSlashes(ApiPrefix)) {
        path(General) {
          passthrough(s"${FireCloudConfig.Rawls.notificationsUrl}/$General", HttpMethods.GET)
        } ~
        path(Workspace / Segment / Segment) { (namespace, name) =>
          passthrough(s"${FireCloudConfig.Rawls.notificationsUrl}/$Workspace/$namespace/$name" , HttpMethods.GET)
        }
      }
    }
  }
```
note: I didn't test or compile that, it could be wrong.

I think that last point also eliminates your question on regexes/ORs in routes.

src/main/scala/org/broadinstitute/dsde/firecloud/service/FireCloudDirectives.scala
```diff
@@ -27,23 +26,22 @@ trait FireCloudDirectives extends spray.routing.Directives with PerRequestCreato
   def passthrough(unencodedPath: String, methods: HttpMethod*): Route =
     passthrough(Uri(unencodedPath), methods:_*)
 
-  def passthrough(uri: Uri, methods: HttpMethod*): Route =
-    passthrough(requestCompression = true, uri, methods:_*)
-
   def passthrough(requestCompression: Boolean, unencodedPath: String, methods: HttpMethod*): Route =
```
davidangb: this is the one you (rightfully) suggest should go away, correct? Fine to leave it in for now - either clean it up as another PR on this story, or file a new tech debt story to clean up later (I can help with filing a ticket if you need it)

src/test/scala/org/broadinstitute/dsde/firecloud/service/NotificationsApiServiceSpec.scala
```diff
@@ -22,23 +18,29 @@ class NotificationsApiServiceSpec extends FlatSpec with BeforeAndAfterAll with H
...
-    Get(s"/api/notifications/general") ~> dummyAuthHeaders ~>
-      sealRoute(notificationsRoutes) ~>
-      check {
-        assertResult(StatusCodes.OK, response.entity.asString) {
-          status
-        }
-      }
+  private def doAssert(method: HttpMethod, uri: String, response: StatusCode): Unit = {
+    new RequestBuilder(method)(uri) ~> dummyAuthHeaders ~> sealRoute(notificationsRoutes) ~> check {
+      status should be(response)
```
davidangb: nit: avoid using "response" as a variable name (or "request" or "context" and almost certainly others). Various spray libraries create those variable names implicitly. How about `expectedStatus`?

src/test/scala/org/broadinstitute/dsde/firecloud/service/ServiceSpec.scala
```diff
@@ -15,8 +16,15 @@ trait ServiceSpec extends FreeSpec with ScalaFutures with ScalatestRouteTest wit
 
   implicit val routeTestTimeout = RouteTestTimeout(5.seconds)
 
+  // we don't include OPTIONS here because we accept all OPTIONS requests to all endpoints for CORS.
+  // we also don't include TRACE or CONNECT or any of the more obscure verbs.
+  val allHttpMethods = Seq(GET, POST, PUT, PATCH, DELETE, HEAD)
+
+  def allHttpMethodsExcept(method: HttpMethod, methods: HttpMethod*): Seq[HttpMethod] = allHttpMethodsExcept(method +: methods)
```
davidangb: nice syntactic sugar for callers here!

davidangb (commented):
looking good, minor comments inline.

src/main/scala/org/broadinstitute/dsde/firecloud/service/FireCloudDirectives.scala
kyuksel: I already did the work to remove it as it was straightforward.

Just so I know, do we have a criterion for situations like this, such as 'do it if it takes less than an hour'? Should they be accompanied by a ticket even if they take less than an hour?

davidangb: There's no concrete guideline for this. It's really developer prerogative, and you'll get a good feel for this over time. In the meantime, reach out to peers and and get opinions! As far as a separate ticket ... to me, that's based on the risk of regression and pervasiveness of the change. If it's risky and needs deeper QA/review, make a ticket. If it's simple, just include it in some other PR.

davidangb (approved):
:+1: with latest code changes

src/test/scala/org/broadinstitute/dsde/firecloud/service/PassthroughDirectivesSpec.scala
```diff
@@ -179,24 +129,38 @@ class PassthroughDirectivesSpec extends BaseServiceSpec with HttpServiceBase
...
-                                       inroot:String, outroot:String): Unit = {
-    val actualInRoot =    if (inroot.equals("")) "" else "/"+inroot
-    val expectedOutRoot = if (outroot.equals("")) "" else "/"+outroot
-    val specRoute = passthroughAllPaths(inroot, echoUrl + expectedOutRoot)
+  private def validateAllPathsWithRoots(inpath: String, outpath: String,
+                                        queryParams: Option[Map[String, String]],
+                                        inroot: String, outroot: String): Unit = {
+    val actualInRoot = if (inroot.equals("")) "" else "/" + inroot
+    val expectedOutRoot = if (outroot.equals("")) "" else "/" + outroot
+
```
kyuksel: @davidangb I pretty much included the code that used to be in `passthroughAllPaths()` below as part of `val specRoute = ...`. If I missed something or there's a better way to do it, let me know.

davidangb: where are the `validateAllPaths` functions actually used now? Since you're removing the `"passthroughAllPaths() directive"` test, I don't see them in use and these can just be deleted (therefore answering your question)

kyuksel: Great, thanks!

src/test/scala/org/broadinstitute/dsde/firecloud/service/SubmissionServiceNegativeSpec.scala
```diff
@@ -0,0 +1,50 @@
...
+      allHttpMethodsExcept(GET, POST) foreach { method =>
+        checkIfPassedThrough(routes, method, localSubmissionsPath, toBeHandled = false)
+      }
+    }
+
+//    "non-POST requests hitting the /workspaces/*/*/submissions/validate path are not passed through path are not passed through" in {
+//      allHttpMethodsExcept(POST) foreach { method =>
+//        checkIfPassedThrough(routes, method, s"$localSubmissionsPath/validate", toBeHandled = false)
+//      }
+//    }
```
kyuksel: Do you have any thoughts on why the test commented out above may be failing?
@aednichols @davidangb

aednichols: It's strange that the test still fails even if I delete the whole block
```
path("validate") {
  post {
    passthrough(s"$workspacesUrl/$namespace/$name/submissions/validate", POST)
  }
} ~
```
from `SubmissionService.scala`

src/test/scala/org/broadinstitute/dsde/firecloud/service/ServiceSpec.scala
```diff
@@ -15,11 +17,22 @@ trait ServiceSpec extends FreeSpec with ScalaFutures with ScalatestRouteTest wit
 
   implicit val routeTestTimeout = RouteTestTimeout(5.seconds)
 
+  // OPTIONS is not included because we accept all OPTIONS requests to all endpoints for CORS.
```
aednichols: I don't think this is true... according to David, those requests are handled by the proxy and never even hit Scala code

src/test/scala/org/broadinstitute/dsde/firecloud/service/SubmissionServiceNegativeSpec.scala
aednichols: I believe the path `/workspaces/namespace/name/submissions/validate` is being passed through on line 40 with a `submissionId` of `"validate"`

kyuksel: You're right on. Thank you!

src/test/scala/org/broadinstitute/dsde/firecloud/service/ServiceSpec.scala
kyuksel: What's the implication, remove `OPTIONS` from `allHttpMethods`?

aednichols: Yup, remove `OPTIONS` from `allHttpMethods` and remove the comment

aednichols: Check with David to be sure, though

src/test/scala/org/broadinstitute/dsde/firecloud/service/SubmissionServiceNegativeSpec.scala
```diff
@@ -0,0 +1,52 @@
...
+      allHttpMethodsExcept(GET, POST) foreach { method =>
+        checkIfPassedThrough(routes, method, localSubmissionsPath, toBeHandled = false)
+      }
+    }
+
+    "non-POST requests hitting the /workspaces/*/*/submissions/validate path are not passed through" in {
+      // Also excluding DELETE and GET since the path matches
+      // /workspaces/*/*/submissions/{submissionId} APIs as well
+      allHttpMethodsExcept(DELETE, GET, POST) foreach { method =>
+        checkIfPassedThrough(routes, method, s"$localSubmissionsPath/validate", toBeHandled = false)
```
aednichols, line 41: Submission IDs are UUIDs of form 8-4-4-4-12, I wonder if there's a way to reject `"validate"` as a `submissionID` on that basis. Then you wouldn't need the special handling in this test, and the code would be more bulletproof

src/test/scala/org/broadinstitute/dsde/firecloud/service/ServiceSpec.scala
kyuksel: @davidangb Does Adam's suggestion sound good to you, too?

src/test/scala/org/broadinstitute/dsde/firecloud/service/SubmissionServiceNegativeSpec.scala
kyuksel: Good idea. Is it documented somewhere that `submissionId`s are UUIDs?

aednichols: I don't know about documented, but I'm pretty sure this is where they come from: https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala#L1463

Though Orchestration tests may need to be updated: https://github.com/broadinstitute/firecloud-orchestration/blob/develop/src/test/scala/org/broadinstitute/dsde/firecloud/mock/MockWorkspaceServer.scala#L38

kyuksel: Thanks for the pointer! 

I don't feel comfortable changing the code based on the implementation.

src/test/scala/org/broadinstitute/dsde/firecloud/service/ServiceSpec.scala
davidangb: agree - remove `OPTIONS` from the list of `allHttpMethods`, no need to test it here. (plus, removing it from the list makes it consistent with the comment)

src/test/scala/org/broadinstitute/dsde/firecloud/service/SubmissionServiceNegativeSpec.scala
kyuksel: I've filed [GAWB-2807](https://broadinstitute.atlassian.net/browse/GAWB-2807) to keep track of the `submissionId` issue.

src/main/scala/org/broadinstitute/dsde/firecloud/service/FireCloudDirectives.scala
```diff
@@ -24,31 +23,18 @@ object FireCloudDirectiveUtils {
...
-    passthrough(requestCompression = true, uri, methods:_*)
+  def passthrough(unencodedPath: String, methods: HttpMethod*): Route = {
+    passthrough(Uri(unencodedPath), methods: _*)
+  }
 
-  def passthrough(requestCompression: Boolean, unencodedPath: String, methods: HttpMethod*): Route =
-    passthrough(requestCompression, Uri(unencodedPath), methods:_*)
+  def passthrough(requestCompression: Boolean, unencodedPath: String, methods: HttpMethod*): Route = {
+    passthrough(Uri(unencodedPath), methods: _*)
+  }
```
davidangb: this version of `passthrough` doesn't make sense anymore. It accepts a `requestCompression` argument and then immediately discards it. From a quick search, it's only called from 4 places, all in `EntityService`. Could you update those 4 calls and remove this signature?

src/main/scala/org/broadinstitute/dsde/firecloud/service/BillingService.scala
```diff
@@ -1,29 +1,35 @@
...
   val routes: Route =
     pathPrefix("billing") {
-      passthrough(rawlsBillingsRoot, HttpMethods.POST)
-    } ~
-    pathPrefix("billing" / RestPath) { restPath =>
-      passthrough(s"$rawlsBillingsRoot/${restPath.toString}", HttpMethods.GET, HttpMethods.DELETE, HttpMethods.POST, HttpMethods.PUT)
+      pathEnd {
+        post {
+          extract(_.request.uri.query) { query =>
+            passthrough(Uri(s"$billingUrl").withQuery(query), POST)
```
davidangb: I don't think this endpoint uses any query params. No need to `extract(...)` and `withQuery` it.

src/main/scala/org/broadinstitute/dsde/firecloud/service/SubmissionService.scala
```diff
@@ -15,20 +14,49 @@ abstract class SubmissionServiceActor extends Actor with SubmissionService {
...
+        (get | post) {
+          extract(_.request.method) { method =>
+            passthrough(s"$workspacesUrl/$namespace/$name/submissions", method)
+          }
+        }
+      } ~
+      path("validate") {
+        post {
+          extract(_.request.uri.query) { query =>
+            passthrough(Uri(s"$workspacesUrl/$namespace/$name/submissions/validate").withQuery(query), POST)
```
davidangb: this also doesn't accept query params. Maybe you are mixing up query params with post body?

davidangb (changes requested):
some small changes inline, this is sooooo close

src/main/scala/org/broadinstitute/dsde/firecloud/service/SubmissionService.scala
kyuksel: Exactly. Thanks for pointing it out!

src/main/scala/org/broadinstitute/dsde/firecloud/service/FireCloudDirectives.scala
kyuksel: Great! The same seemed to be the case for `generateExternalHttpPerRequestForMethod` and `externalHttpPerRequest`, so I cleaned up those as well.

davidangb (approved):
:+1: lgtm!

src/test/scala/org/broadinstitute/dsde/firecloud/service/SubmissionServiceNegativeSpec.scala
aednichols: Can you put a reference to this ticket in the code as well?

aednichols (approved):
Looks good! Thank you for undertaking this necessary but finicky set of changes.

One general comment (no need to change) - it would be helpful if you put batches of purely cosmetic changes in their own commits - makes reviewing easier. Also, I try to avoid PRs that make cosmetic changes to files that don't also have logical changes - keeps noise down in the git history.

PR #517, opened 2017-10-18
GAWB-849: Agora supported GA4GH APIs

rushtong:
Partially addresses https://broadinstitute.atlassian.net/browse/GAWB-849

## Changes
* Update swagger doc with new endpoints supported by Agora
* Unimplemented GA4GH endpoints not documented, but handled, similarly to what Agora does.

Goes hand in hand with https://github.com/broadinstitute/agora/pull/221, but both PRs can be reviewed in isolation.

---

In all cases:

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -3426,8 +3426,7 @@ paths:
     get:
       summary: List all tools
       description: >
-        This endpoint returns all tools available or a filtered subset using
-        metadata query parameters.
+        This endpoint returns all tools available.
       tags:
         - GA4GH Tool Registry
       responses:
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -5130,13 +5130,27 @@ definitions:
       meta-version:
         type: string
         description: 'The version of this tool in the registry. Iterates when fields like the description, author, etc. are updated.'
+      contains:
+        description: An array of IDs for the applications that are stored inside this tool (for example `https://bio.tools/tool/mytum.de/SNAP2/1`). This always returns an empty array.
+        type: array
+        items:
+           type: string
+      verified:
+        type: boolean
+        description: Reports whether this tool has been verified by a specific organization or individual.  This always returns false.
+      verified-source:
+        type: string
+        description: Source of metadata that can support a verified tool, such as an email or URL
+      signed:
+        type: boolean
+        description: Reports whether this tool has been signed.  This always returns false.
       versions:
         description: A list of versions for this tool
         type: array
         items:
           $ref: '#/definitions/ToolVersion'
   ToolVersion:
-    description: A tool version describes a particular iteration of a tool.
+    description: A tool version describes a particular iteration of a tool as described by a reference to a specific image and dockerfile.
     required:
       - url
       - id
@@ -5151,6 +5165,9 @@ definitions:
       id:
         type: string
         description: An identifier of the version of this tool for this particular tool registry, for example `v1`
+      image:
+        type: string
+        description: The docker path to the image (and version) for this tool. (e.g. quay.io/seqware/seqware_full/1.1). This always returns an empty string.
       descriptor-type:
         type: array
         description: The type (or types) of descriptors available.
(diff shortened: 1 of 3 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/webservice/Ga4ghApiService.scala
```diff
@@ -11,17 +11,48 @@ trait Ga4ghApiService extends HttpService with FireCloudDirectives {
...
+          } ~
+          path("tools" / Segment / "versions" / Segment / Segment / "descriptor") { (id, versionId, descriptorType) =>
+            val targetUri = Uri(s"$agoraGA4GH/tools/$id/versions/$versionId/$descriptorType/descriptor")
+            passthrough(targetUri, HttpMethods.GET)
+          } ~
+          path("tools" / Segment / "versions" / Segment / Segment / "descriptor" / Segment) { (id, versionId, descriptorType, relativePath) =>
+            complete(spray.http.StatusCodes.NotImplemented)
+          } ~
+          path("tools" / Segment / "versions" / Segment / Segment / "tests") { (id, versionId, descriptorType) =>
+            complete(spray.http.StatusCodes.NotImplemented)
```
davidangb: for the NotImplemented endpoints, I suggest passing through to agora, and let agora generate the NotImplemented response.

davidangb (commented):
one comment inline. I didn't review the swagger defs in  depth, because those may change in agora.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/Ga4ghApiService.scala
rushtong: 👍

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/Ga4ghApiServiceSpec.scala
```diff
@@ -32,19 +54,29 @@ class Ga4ghApiServiceSpec extends BaseServiceSpec with Ga4ghApiService with Befo
 
   "GA4GH API service" - {
     "Tool Registry" - {
-      "get-descriptor passthrough API" - {
-
+      "passthrough APIs" - {
+        "should handle un-implemented paths" in {
+          unImplementedPaths.map { path =>
+            Get(path) ~> ga4ghRoutes ~> check {
```
aednichols: My version of IntelliJ warns to use `foreach` instead of `map` because there's no return value... I'm not concerned, but you could make IntelliJ's day if you like.

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/Ga4ghApiServiceSpec.scala
```diff
@@ -11,19 +12,40 @@ import spray.http.StatusCodes._
...
-      )
+    toolPaths.map { path =>
+      toolRegistryServer.when(request().withMethod(HttpMethods.GET.name).withPath(path))
+        .respond(
+          org.mockserver.model.HttpResponse.response()
+            .withStatusCode(OK.intValue)
+        )
+    }
+    // Currently, Agora returns NotImplemented for these paths.
+    unImplementedPaths.map { path =>
```
aednichols: It seems strange to test an Agora behavior in Orchestration. But I see David's point, and this may be the best way to do it.

aednichols (approved):
👍

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/Ga4ghApiServiceSpec.scala
rushtong: It was strange that I chose to return `NotImplemented` ... I chose to let Agora do the work and instead just check that we actually handle the endpoint and disregard the actual status.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -3358,10 +3358,91 @@ paths:
...
+          schema:
+              $ref: '#/definitions/ToolVersion'
+      security: []
+
+  /ga4gh/v1/tools:
+    get:
+      summary: List all tools
+      description: >
+        This endpoint returns all tools available or a filtered subset using
+        metadata query parameters.
```
davidangb: remove the "or a filtered subset" part

davidangb (approved):
one comment inline, otherwise :+1:. I don't need to re-review if you address my comment.

src/main/resources/swagger/api-docs.yaml
rushtong: I think I need to remove the following words too - metadata query parameters don't make sense here.

rushtong: And same in agora.

davidangb (approved):
:+1: again!

PR #519, opened 2017-10-27
GAWB-2746: DUL search API

davidangb:
API support for the first stage of DUL search.

See the tech doc at https://docs.google.com/a/broadinstitute.org/document/d/1P70Gt5wCru0YzvJWNru9Nt4tCADhEsBWpKPp7qs_n2M/edit?usp=sharing for discussion of the API changes and their effect on backwards compatibility (summary: I tried real hard to make it backwards compatible).

You should definitey look at this PR and the tech doc if:
* you are working on the UI and need to understand how the underlying APIs changed (see the swagger)
* you are working on indexing consent codes and need to know what the runtime queries expect (see ElasticSearchDAOResearchPurposeSupport.scala)
* you are working on indexing consent codes and are curious about how your model objects might interact with mine

In all cases:

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/ResearchPurposeSearchTestFixtures.scala
+++ b/src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/ResearchPurposeSearchTestFixtures.scala
@@ -17,20 +17,40 @@ object ResearchPurposeSearchTestFixtures {
 
   implicit val impRestrictionFormat = jsonFormat5(Restriction)
 
-  val datasetTuples:Seq[(String,String,Restriction)] = Seq(
-    ("CSA_9220", "Kombucha vaporware flexitarian chambray bespoke", Restriction(GRU=true)),
-    ("L_1240", "3 wolf moon vape try-hard knausgaard", Restriction(HMB=true)),
-    ("HC_2531", "wolf freegan irony lomo", Restriction(NCU=true)),
-    ("FASD_0050696", "gastropub tattooed hammock mustache", Restriction(NAGR=true)),
-    ("D_4", "Neutra selvage chicharrones, prism taxidermy cray squid", Restriction(NPU=true))
+  // tuples in this order:
+  //   dataset name: used to validate results
+  //   project name: used to test facet filters
+  //   indication: used to test suggestions
+  //   datasetCustodian: used to test text search
+  //   restriction: used to test research purpose matching
+  val datasetTuples:Seq[(String,String,String,String,Restriction)] = Seq(
+    ("one",      "hydrogen", "antiaging",     "the quick brown fox", Restriction(GRU=true, NCU=true)),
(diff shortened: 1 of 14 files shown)


src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/ResearchPurposeSearchTestFixtures.scala
```diff
@@ -0,0 +1,57 @@
...
+
+  )
+
+  val fixtureDocs:Seq[Document] = datasetTuples map {x:(String,String,String,String,Restriction) =>
+    Document(x._1, Map(
+      AttributeName("library","datasetName") -> AttributeString(x._1),
+      AttributeName("library","projectName") -> AttributeString(x._2),
+      AttributeName("library","indication") -> AttributeString(x._3),
+      AttributeName("library","datasetCustodian") -> AttributeString(x._4),
+      AttributeName.withDefaultNS("dataUseRestriction") -> AttributeValueRawJson(x._5.toJson)
```
rushtong: I called this `structuredUseRestriction` in my PR because `dataUseRestriction` is already a library field. I could, however, change mine to not be a library field to just be a default namespaced `dataUseRestriction` instead.

davidangb: nah, leave yours as-is, I'll update this PR once yours lands!

rushtong (approved):
👍  LGTM. Will revisit after my PR feedback is addressed.

src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/ResearchPurposeSearchTestFixtures.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  )
+
+  val fixtureDocs:Seq[Document] = datasetTuples map {x:(String,String,String,String,Restriction) =>
+    Document(x._1, Map(
+      AttributeName("library","datasetName") -> AttributeString(x._1),
+      AttributeName("library","projectName") -> AttributeString(x._2),
+      AttributeName("library","indication") -> AttributeString(x._3),
+      AttributeName("library","datasetCustodian") -> AttributeString(x._4),
+      structuredUseRestrictionAttributeName -> AttributeValueRawJson(x._5.toJson)
```
rushtong, line 44: This is a little bit confusing, but I'm pretty sure I see what's happening. By using `structuredUseRestrictionAttributeName`, it looks like this **should** be a structured data use restriction, but it isn't, its actually a research purpose. Since this is used for query tests, I think this is fine - just wanted to make sure I'm reading this right.

rushtong (approved):
👍 Still LGTM. One comment/clarification inline but should not hold anything up.

rushtong:
👍

PR #520, opened 2017-10-31
GAWB-2767 Index consent codes as a structured object

rushtong:
Addresses https://broadinstitute.atlassian.net/browse/GAWB-2767

## Changes
* This PR adds one more element to the workspace document that is indexed to ES. In the case where there are consent code library attributes, we collect the valid ones and flesh that out into a full object. 

## Testing
* Run this against a custom index if running locally.
* Make sure there is at least one workspace that is catalogued with answers to the data use questionnaire.
* Run a reindex
* Search your local index for something like this:
```
http://elasticsearch5a3.dsde-dev.broadinstitute.org:9200/librarylocalgr/dataset/_search?q=_id=df7d1e4a-0956-436c-bf29-5d17a6af9483
```
You should see an indexed workspace with the following object:
```
          "library:structuredUseRestriction": {
            "DS": [
              "Diabetes Mellitus"
            ], 
            "GRU": false, 
            "HMB": true, 
            "NAGR": false, 
            "NCTRL": false, 
            "NCU": false, 
            "NDMS": false, 
            "NPU": false, 
            "RS-FM": false, 
            "RS-G": false, 
            "RS-M": false, 
            "RS-PD": false, 
            "RS-POP": [
              "N/A"
            ]
          }, 
``` 

---

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
@@ -11,6 +11,8 @@ trait DataUseRestrictionSupport {
   private val listCodes = List("DS", "RS-POP")
   private val durNames = booleanCodes ++ listCodes ++ List("RS-G")
 
+  val dataUseRestrictionAttributeName: AttributeName = AttributeName.withLibraryNS("dataUseRestriction")
+
   /**
     * This method looks at all of the library attributes that are associated to Consent Codes and
     * builds a set of Data Use Restriction attributes from the values. Most are boolean values, some are lists
@@ -27,7 +29,7 @@ trait DataUseRestrictionSupport {
     val libraryAttrs = workspace.attributes.filter { case (attr, value) => durNames.contains(attr.name) }
 
     if (libraryAttrs.isEmpty) {
-      Map(AttributeName.withLibraryNS("dataUseRestriction") -> AttributeNull)
+      Map.empty
     } else {
       val existingAttrs: Map[AttributeName, Attribute] = libraryAttrs.flatMap {
         // Handle the String->Boolean case first
@@ -52,14 +54,14 @@ trait DataUseRestrictionSupport {
 
       val existingKeyNames = existingAttrs.keys.map(_.name).toSeq
 
-      // Missing boolean codes with default false
+      // Missing boolean codes default to false
       val booleanAttrs: Map[AttributeName, AttributeBoolean] = (booleanCodes ++ List("RS-G", "RS-FM", "RS-M")).
         filter(!existingKeyNames.contains(_)).
         flatMap { code =>
           Map(AttributeName.withDefaultNS(code) -> AttributeBoolean(false))
         }.toMap
 
-      // Missing list codes with default empty lists
+      // Missing list codes default to empty lists
       val listAttrs = listCodes.
         filter(!existingKeyNames.contains(_)).
         flatMap { code =>
@@ -68,8 +70,7 @@ trait DataUseRestrictionSupport {
 
       val allAttrs = existingAttrs ++ booleanAttrs ++ listAttrs
 
-      Map(AttributeName.withLibraryNS("dataUseRestriction") ->
-        AttributeValueRawJson.apply(allAttrs.toJson.compactPrint))
+      Map(dataUseRestrictionAttributeName -> AttributeValueRawJson.apply(allAttrs.toJson.compactPrint))
     }
 
   }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
@@ -1,15 +1,16 @@
 package org.broadinstitute.dsde.firecloud.service
 
+import com.typesafe.scalalogging.slf4j.LazyLogging
 import org.broadinstitute.dsde.rawls.model.WorkspaceJsonSupport.AttributeNameFormat
 import org.broadinstitute.dsde.rawls.model._
 import spray.json.DefaultJsonProtocol._
 import spray.json._
 
-trait DataUseRestrictionSupport {
+trait DataUseRestrictionSupport extends LazyLogging {
 
-  private val booleanCodes = List("GRU", "HMB", "NCU", "NPU", "NDMS", "NAGR", "NCTRL", "RS-PD")
-  private val listCodes = List("DS", "RS-POP")
-  private val durNames = booleanCodes ++ listCodes ++ List("RS-G")
+  private val booleanCodes: Seq[String] = Seq("GRU", "HMB", "NCU", "NPU", "NDMS", "NAGR", "NCTRL", "RS-PD")
+  private val listCodes: Seq[String] = Seq("DS", "RS-POP")
+  val durFieldNames: Seq[String] = booleanCodes ++ listCodes ++ Seq("RS-G")
 
   val structuredUseRestrictionAttributeName: AttributeName = AttributeName.withLibraryNS("structuredUseRestriction")
 
@@ -30,10 +31,10 @@ trait DataUseRestrictionSupport {
     implicit val impAttributeFormat: AttributeFormat with PlainArrayAttributeListSerializer =
       new AttributeFormat with PlainArrayAttributeListSerializer
 
-    val libraryAttrs = workspace.attributes.filter { case (attr, value) => durNames.contains(attr.name) }
+    val libraryAttrs = workspace.attributes.filter { case (attr, value) => durFieldNames.contains(attr.name) }
 
     if (libraryAttrs.isEmpty) {
-      Map.empty
+      Map.empty[AttributeName, Attribute]
     } else {
       val existingAttrs: Map[AttributeName, Attribute] = libraryAttrs.flatMap {
         // Handle the known String->Boolean conversion cases first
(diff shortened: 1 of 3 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -0,0 +1,93 @@
...
+    */
+  def generateStructuredUseRestriction(workspace: Workspace): Map[AttributeName, Attribute] = {
+
+    implicit val impAttributeFormat: AttributeFormat with PlainArrayAttributeListSerializer =
+      new AttributeFormat with PlainArrayAttributeListSerializer
+
+    val libraryAttrs = workspace.attributes.filter { case (attr, value) => durNames.contains(attr.name) }
+
+    if (libraryAttrs.isEmpty) {
+      Map.empty
```
davidangb: syntax: elsewhere you use `Map.empty[AttributeName, Attribute]`, use it here too for consistency

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -0,0 +1,93 @@
...
+                case v if v.equalsIgnoreCase("male") =>
+                  Map(AttributeName.withDefaultNS("RS-G") -> AttributeBoolean(true),
+                    AttributeName.withDefaultNS("RS-FM") -> AttributeBoolean(false),
+                    AttributeName.withDefaultNS("RS-M") -> AttributeBoolean(true))
+                case _ =>
+                  Map(AttributeName.withDefaultNS("RS-G") -> AttributeBoolean(false),
+                    AttributeName.withDefaultNS("RS-FM") -> AttributeBoolean(false),
+                    AttributeName.withDefaultNS("RS-M") -> AttributeBoolean(false))
+              }
+            case _ => Map.empty[AttributeName, Attribute]
```
davidangb: here and in the `case _` at the end, log a message with the attr/value so we can catch this dirty data

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -0,0 +1,93 @@
...
+        case (attr: AttributeName, value: AttributeBoolean) => Map(AttributeName.withDefaultNS(attr.name) -> value)
+        case (attr: AttributeName, value: AttributeValueList) => Map(AttributeName.withDefaultNS(attr.name) -> value)
+        case _ => Map.empty[AttributeName, Attribute]
+      }
+
+      val existingKeyNames = existingAttrs.keys.map(_.name).toSeq
+
+      // Missing boolean codes default to false
+      val booleanAttrs: Map[AttributeName, AttributeBoolean] = (booleanCodes ++ List("RS-G", "RS-FM", "RS-M")).
+        filter(!existingKeyNames.contains(_)).
```
davidangb: this is mostly style, but I prefer using the Scala collection set APIs instead of filter. Equivalent code is:
```
((booleanCodes ++ List("RS-G", "RS-FM", "RS-M")) diff existingKeyNames).flatMap { ... }
```

also, you can do away with the `flatMap` of nested maps, and instead call `map` and create tuples:
```
 ((booleanCodes ++ List("RS-G", "RS-FM", "RS-M")) diff existingKeyNames).map { code =>
          AttributeName.withDefaultNS(code) -> AttributeBoolean(false)
        }.toMap
```

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
```diff
@@ -82,7 +82,9 @@ trait LibraryServiceSupport extends LazyLogging {
       case None => Map()
     }
 
-    val fields = attrfields ++ idfields ++ tagfields
+    val dur: Map[AttributeName, Attribute] = generateStructuredUseRestriction(workspace)
+
+    val fields = attrfields ++ idfields ++ tagfields ++ dur
```
davidangb: could you remove anything that's in `dur` from `attrfields`? Else we are indexing those twice.

This might mean that `generateStructuredUseRestriction` needs to return the lower level of consent codes, so you have access to their key names, and `indexableDocument` is the thing that puts them inside an `AttributeValueRawJson` nested under "structuredUseRestriction"

If this becomes super difficult we could merge this PR and do that work in a follow-on.

src/test/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupportSpec.scala
```diff
@@ -0,0 +1,211 @@
...
+      f.setAccessible(true)
+      getFieldName(f) -> f.get(dur)
+    } toMap
+  }
+
+  // Since we have dashes in DUR field names, the value that comes back from Field.getName
+  // looks like "RS$minusPOP" instead of "RS-POP"
+  private def getFieldName(f: Field): String = { f.getName.replace("$minus", "-") }
+
+}
```
davidangb, line 211: I'd love to see integration tests, too: actually publish these workspaces to ES, then retrieve them, and make sure the indexed format is as expected. The unit test verifies that we are correctly producing what we think ES expects - but an integration test would verify that what we send to ES is actually usable by ES.

Again - because this PR is blocking other work, I'm ok if we write integration tests for a future story.

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -0,0 +1,93 @@
...
+          Map(AttributeName.withDefaultNS(code) -> AttributeBoolean(false))
+        }.toMap
+
+      // Missing list codes default to empty lists
+      val listAttrs = listCodes.
+        filter(!existingKeyNames.contains(_)).
+        flatMap { code =>
+          Map(AttributeName.withDefaultNS(code) -> AttributeValueList(Seq.empty))
+        }.toMap
+
```
davidangb: @ahaessly in your work for GAWB-2758 you may need to massage the `DS` consent code somewhere in this trait - just a heads up for multiple code branches in progress!

davidangb (commented):
looking good, and produces the expected result locally.

Some minor syntax comments inline; two larger comments that are deferrable.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
rushtong: This is reasonable, but is slightly different with respect to the codes that we do transforms on. For example, gender and `NAGR` codes go through a string translation to a boolean value. We will lose those string values and only get the boolean. I think this is OK, but to be clear, it's different from what we're doing now.

src/test/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupportSpec.scala
rushtong: Yeah ... I didn't consider this and agree that it's necessary.

ahaessly:
should we add the "indexable":false attribute in the attribute-definitions.json file for these DULs in this PR or should I do it with GAWB-2758?

rushtong:
@ahaessly - I don't think we want to do that universally because we want them indexed if, for some reason, we don't have a valid structured data use object to index. This PR is updated so that we remove them from the indexing when we do have one.

The only time this should happen is when there are improperly tagged attributes, i.e. `"library:GRU":"Yes"`

rushtong:
@davidangb - PR updated with feedback, except for the integration tests which I'm looking at now. If I don't make progress today, I'll move that to a new story.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
```diff
@@ -82,7 +82,15 @@ trait LibraryServiceSupport extends LazyLogging {
...
-    val fields = attrfields ++ idfields ++ tagfields
+    val dur: Map[AttributeName, Attribute] = generateStructuredUseRestriction(workspace)
+
+    val fields = dur match {
+      case x if x.isEmpty => attrfields ++ idfields ++ tagfields
+      case _ =>
+        // remove the data use attributes so we don't index them twice
+        val durAttributeNames = durFieldNames.map(AttributeName.withLibraryNS)
+        attrfields.filterKeys { key => !durAttributeNames.contains(key) } ++ idfields ++ tagfields ++ dur
+    }
```
davidangb: `filterKeys` is unsafe in the same way `mapValues` is: https://issues.scala-lang.org/browse/SI-4776. I'm guessing the use of `filterKeys` here is okay, but I'd like to keep it out of the codebase altogether for sanity's sake. With that in mind, I think you can simplify this to:

```
(attrfields -- durAttributeNames) ++ idfields ++ tagfields ++ dur
```

Also - I don't see any reason to index the consent codes at all, if `dur` is empty. If we don't have a valid dur, there's no point in indexing them, since we can't rely on their values, and we don't have any code that looks for them outside of the dur. I say remove the match/case and always use the second codepath.

rushtong: Argh ... yes, I should have remembered the filterKeys issue. Will fix.

davidangb (approved):
:+1:

src/test/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupportSpec.scala
rushtong: See https://broadinstitute.atlassian.net/browse/GAWB-2812 for follow-on story

PR #521, opened 2017-11-01
GAWB-2812 Integration Tests for DURs

rushtong:
Addresses https://broadinstitute.atlassian.net/browse/GAWB-2812

Adds integration tests to validate that indexed workspaces with structured data use restrictions are indeed indexed, and when searched, are found and can be deconstructed with the same values they were created with. 

---

In all cases:


davidangb (changes requested):
Tests pass and this makes a lot of progress on test coverage, yay!

That said, I'd love to see these tests validate the entire restriction indexed for each dataset, instead of just validating one or two codes in each test. Or, in the case of DS, validating the values instead of just `nonEmpty`. #522 has suggestions for this.

I'd also love to see one or two tests that set multiple values in the DUR, and then we validate the final result! These would verify that the DUR-indexing code doesn't have any problems turning on multiple codes at once.

davidangb (approved):
:+1: thank you!

ahaessly (approved):
👍

PR #524, opened 2017-11-03
GAWB-2829: Add passthrough for submissionsCount endpoint in Rawls

kyuksel:
This is to fix a regression introduced by #515. 

Having gone through the three `Orchestration` services (i.e. `NotificationsApiService`, `BillingService`,  `SubmissionService`) that used to use `passthroughAllPaths`, the `/api/workspaces/{workspaceNamespace}/{workspaceName}/submissionsCount` endpoint appears to be the only one that got lost in transition, which is now added.

Verified that 
- the Swagger page lists `submissionsCount` 
- workspaces can load their `Analysis Submissions` section (which uses `submissionsCount` endpoint)

In all cases:

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -2714,10 +2714,16 @@ paths:
       responses:
         '200':
           description: Successful Request
+          schema:
+            $ref: '#/definitions/SubmissionsCountResponse'
         '404':
           description: Workspace not found
+          schema:
+            $ref: '#/definitions/ErrorReport'
         '500':
           description: Internal Error
+          schema:
+            $ref: '#/definitions/ErrorReport'
       parameters:
         - $ref: '#/parameters/workspaceNamespaceParam'
         - $ref: '#/parameters/workspaceNameParam'
@@ -4997,6 +5003,24 @@ definitions:
         description: What happens after a task fails. Choose from ContinueWhilePossible and NoNewCalls. Defaults to NoNewCalls if not specified. See Cromwell docs for more information.
         default: NoNewCalls
         enum: ["NoNewCalls", "ContinueWhilePossible"]
+
+  SubmissionsCountResponse:
+      type: object
+      properties:
+        Accepted:
+          type: integer
+        Evaluating:
+          type: integer
+        Submitting:
+          type: integer
+        Submitted:
+          type: integer
+        Aborting:
+          type: integer
+        Aborted:
+          type: integer
+        Done:
+          type: integer
 
   SubmissionValidationEntityInputs:
     description: the results of parsing each of the inputs for one entity
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -2704,6 +2704,44 @@ paths:
         500:
           description: Internal Error
 
+  /api/workspaces/{workspaceNamespace}/{workspaceName}/submissionsCount:
+      get:
+        responses:
+          '200':
+            description: Successful Request
+            schema:
+              type: object
+              description: Map[String,Int]
+          '404':
+            description: Workspace not found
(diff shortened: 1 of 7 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -2704,6 +2704,44 @@ paths:
         500:
           description: Internal Error
```
kyuksel, line 2702: The Swagger update below is mostly copied from Rawls'. I'm not sure if that was the right thing to do. It has the `Model` and `Example Value` fields empty for `submissionsCount` on the Swagger page.

src/main/scala/org/broadinstitute/dsde/firecloud/service/SubmissionService.scala
```diff
@@ -18,36 +18,43 @@ trait SubmissionService extends HttpService with PerRequestCreator with FireClou
         passthrough(submissionQueueStatusUrl, GET)
       }
     } ~
```
kyuksel, line 20: Sorry that the diff below looks painfully difficult to parse. All I did was turning `"workspaces" / Segment / Segment / "submissions"` prefix into `"workspaces" / Segment / Segment` and moving `submissions` underneath along with the newly added `submissionsCount`.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -2704,6 +2704,39 @@ paths:
...
+          - in: path
+            description: Workspace Namespace
+            name: workspaceNamespace
+            required: true
+            type: string
+          - in: path
+            description: Workspace Name
+            name: workspaceName
+            required: true
+            type: string
```
davidangb: use
```
```
instead of typing out the namespace/name parameters. This is good for centralization, but most importantly it reduces the number of lines of code per instance of these params (and there are a lot). This file is already big, we try to keep it as small as reasonably possible.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -2704,6 +2704,39 @@ paths:
...
           description: Internal Error
 
+  /api/workspaces/{workspaceNamespace}/{workspaceName}/submissionsCount:
+      get:
+        responses:
+          '200':
+            description: Successful Request
+            schema:
+              type: object
+              description: Map[String,Int]
```
davidangb: The model & example fields are empty in swagger-ui because swagger only knows this as an `object`, and all it can render is `{}`.  To populate it, you'd have to create something in `definitions` like:
```
  SubmissionCount:
    type: object
    properties:
      Accepted:
        type: integer
      Evaluating:
        type: integer
      Submitting:
        type: integer
      Submitted:
        type: integer
      Aborting:
        type: integer
      Aborted:
        type: integer
      Done:
        type: integer
```
and then reference it from here with:
```
          '200':
            description: Successful Request
            schema:
              $ref: '#/definitions/SubmissionCount'
```
While we put effort into documenting Rawls, since Rawls isn't public-facing (yet), we've cut some corners. Since orchestration is our public-facing API, we should do our best to make doc comprehensive.

davidangb (commented):
Scala code looks good, passes tests locally. I have some comments on the swagger yaml, inline.

aednichols (commented):
- The regression I noticed last week is fixed (submission count not working) 👍 
- Local tests pass 👍 

I second David's Swagger suggestions, will re-review when addressed. It might be annoying to get right, but a ton of people use the API so it's really valuable!

davidangb (approved):
:+1: code looks good, did not test.

aednichols:
Travis tests have failed twice in a row (I re-ran after the first failure), so I think this is worthy of some investigation before merge

kyuksel:
The third time is the charm apparently 🤷🏻‍♂️

aednichols (approved):
👍

PR #525, opened 2017-11-06
GAWB-2793: disease-focused research purpose search

davidangb:
This PR adds support for the DS, NDMS, and NCTRL clauses in research purpose search - all of which involve ontology-related business logic.

PR is rather large. For compatibility, I had to:
1. pass the app's `OntologyDAO` all the way down into the `ElasticSearchDAO`, which required a lot of code to be updated. I'm not super happy about one dao knowing about another, but it seemed the best of a few ugly solutions.
2. change the `DataUseRestriction` model class to use a `Seq[Int]` instead of `Seq[String]`, which required a lot of code to be updated.
3. change the `MockOntologyDAO` to behave a bit more like the real dao, which required a lot of code to be updated.

I added a bunch of tests, including the test cases defined by PO. You'll note there are a few TODOs in the tests where I need to talk to PO about the test cases, because I'm not sure they're accurate. I'll do that as part of PO review for this story.

`ElasticSearchDAOResearchPurposeSupport.scala` has the salient changes in this PR. Pretty much everything else is either testing or compatibility updates.

In all cases:

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/ResearchPurposeSearchSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/ResearchPurposeSearchSpec.scala
@@ -165,32 +165,86 @@ class ResearchPurposeSearchSpec extends FreeSpec with SearchResultValidation wit
           searchResponse
         )
       }
-      "should intersect with a standard facet filter" ignore {
-        fail("test not implemented")
+      "should intersect with a standard facet filter" in {
+        val researchPurpose = ResearchPurpose.default.copy(DS = Seq(DiseaseOntologyNodeId("http://purl.obolibrary.org/obo/DOID_535")))
+        val filter = Map("library:projectName" -> Seq("beryllium"))
+        val searchResponse = searchWithPurpose(researchPurpose, filter)
+        validateResultNames(
+          Set("sixteen", "eighteen"),
+          searchResponse
+        )
       }
-      "should intersect with a text search" ignore {
-        fail("test not implemented")
+      "should intersect with a text search" in {
+        val researchPurpose = ResearchPurpose.default.copy(DS = Seq(DiseaseOntologyNodeId("http://purl.obolibrary.org/obo/DOID_535")))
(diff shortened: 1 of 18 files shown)


src/test/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionTestFixtures.scala
```diff
@@ -13,7 +13,7 @@ object DataUseRestrictionTestFixtures {
   case class DataUseRestriction(
     GRU: Boolean = false,
     HMB: Boolean = false,
-    DS: Seq[String] = Seq.empty[String],
+    DS: Seq[Int] = Seq.empty[Int],
```
rushtong, line 16: This will need coordination with GAWB-2758 and GAWB-2769 since they both look at the values stored for this field.

src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/ResearchPurposeSearchSpec.scala
```diff
@@ -139,6 +139,168 @@ class ResearchPurposeSearchSpec extends FreeSpec with SearchResultValidation wit
...
+      "should return any dataset where NDMS is true and the disease matches exactly" in {
+        val researchPurpose = ResearchPurpose.default.copy(NDMS=true, DS = Seq(DiseaseOntologyNodeId("http://purl.obolibrary.org/obo/DOID_535")))
+        val searchResponse = searchWithPurpose(researchPurpose)
+        validateResultNames(
+          Set("one", "six", "eleven", "sixteen", "eighteen"), // NDMS=false or (NDMS=true and DS match)
+          searchResponse
+        )
+        // NB: this doesn't match fixture "twenty" because even though the NDMS clauses are satisfied, the DS
+        // clause is not. In other words, if you made this search without specifying NDMS=true, you wouldn't
+        // match on "twenty". TODO: verify this is correct business logic.
```
rushtong, line 206: This seems like the correct logic to me.

rushtong (approved):
👍  Looks good to me. I don't have any real code suggestions, just some comments. Also agree about the ontology dao, but I also don't have a better approach.

PR #527, opened 2017-11-06
GAWB-2834: Index data use restriction display codes as a list of easily displayed terms

rushtong:
Addresses https://broadinstitute.atlassian.net/browse/GAWB-2834

Builds and indexes a display version of the data use restriction in a list format. 

Note that there are a bunch of new logger warnings that we might need to add loggly alerts for. Confirm with David An.

---

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
@@ -16,8 +16,7 @@ trait DataUseRestrictionSupport extends LazyLogging {
   private val genderCodes: Seq[String] = Seq("RS-G", "RS-FM", "RS-M")
   val durFieldNames: Seq[String] = booleanCodes ++ listCodes ++ genderCodes
 
-  // TODO: This value will change depending on what is implemented for GAWB-2758
-  val diseaseLabelsAttributeName: AttributeName = AttributeName.withLibraryNS("DS-labels")
+  val diseaseLabelsAttributeName: AttributeName = AttributeName.withLibraryNS("DS_URL")
   val structuredUseRestrictionAttributeName: AttributeName = AttributeName.withLibraryNS("structuredUseRestriction")
   val dataUseDisplayAttributeName: AttributeName = AttributeName.withLibraryNS("dataUseDisplay")
 
@@ -37,7 +36,7 @@ trait DataUseRestrictionSupport extends LazyLogging {
     implicit val impAttributeFormat: AttributeFormat with PlainArrayAttributeListSerializer =
       new AttributeFormat with PlainArrayAttributeListSerializer
 
-    getDataUseAttributes(workspace).filter { t => durFieldNames.contains(t._1.name) } match {
+    getDataUseAttributes(workspace) match {
       case x if x.isEmpty => Map.empty[AttributeName, Attribute]
       case existingAttrs =>
 
@@ -67,27 +66,29 @@ trait DataUseRestrictionSupport extends LazyLogging {
     * @return An Attribute Map representing a data use display
     */
   def generateUseRestrictionDisplayAttribute(workspace: Workspace): Map[AttributeName, Attribute] = {
-    getDataUseAttributes(workspace).flatMap {
-      case (attr: AttributeName, value: AttributeBoolean) if value.value =>  Seq(attr.name)
-      case (attr: AttributeName, value: AttributeValueList) if attr.name.equals(diseaseLabelsAttributeName.name) => value.list.map {
-        case avl@(a: AttributeString) => "DS:" + a.value
-        case _ =>
-          logger.warn(s"Attribute value list is of the wrong type (workspace-id: ${workspace.workspaceId}, attribute name: ${attr.name}, value type: ${value.getClass})")
-          ""
+
+    val booleanCodes:Seq[String] = getDataUseAttributes(workspace).collect {
+      case (attr: AttributeName, AttributeBoolean(true)) => attr.name
+    }.toSeq
+
+    val dsLabels:Seq[String] = (workspace.attributes.get(diseaseLabelsAttributeName) collect {
+      case value: AttributeValueList => value.list.collect {
+        case a: AttributeString => "DS:" + a.value
       }
-      case (attr: AttributeName, value: Attribute) =>
-        logger.debug(s"Ignored attribute: ${attr.name}; value: ${value.getClass.getName}")
-        Seq.empty[String]
-    }.toSeq.filter(_.nonEmpty) match {
-      case x if x.nonEmpty => Map(dataUseDisplayAttributeName -> AttributeValueList(x.map(AttributeString)))
-      case _ => Map.empty[AttributeName, Attribute]
-    }
+    }).getOrElse(Seq.empty[String])
+
+    val displayCodes = booleanCodes ++ dsLabels
+
+    if (displayCodes.nonEmpty)
+      Map(dataUseDisplayAttributeName -> AttributeValueList(displayCodes.map(AttributeString)))
+    else
+      Map.empty[AttributeName, Attribute]
   }
 
   private def getDataUseAttributes(workspace: Workspace): Map[AttributeName, Attribute] = {
 
     // Find all library attributes that contribute to data use restrictions
-    val dataUseAttributes = workspace.attributes.filter { case (attr, value) => (durFieldNames ++ Seq(diseaseLabelsAttributeName.name)).contains(attr.name) }
+    val dataUseAttributes = workspace.attributes.filter { case (attr, value) => durFieldNames.contains(attr.name) }
 
     if (dataUseAttributes.isEmpty) {
       Map.empty[AttributeName, Attribute]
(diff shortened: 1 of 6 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -1,18 +1,25 @@
...
 trait DataUseRestrictionSupport extends LazyLogging {
 
   private val booleanCodes: Seq[String] = Seq("GRU", "HMB", "NCU", "NPU", "NDMS", "NAGR", "NCTRL", "RS-PD")
   private val listCodes: Seq[String] = Seq("DS", "RS-POP")
-  val durFieldNames: Seq[String] = booleanCodes ++ listCodes ++ Seq("RS-G")
+  private val genderCodes: Seq[String] = Seq("RS-G", "RS-FM", "RS-M")
+  val durFieldNames: Seq[String] = booleanCodes ++ listCodes ++ genderCodes
 
+  // TODO: This value will change depending on what is implemented for GAWB-2758
+  val diseaseLabelsAttributeName: AttributeName = AttributeName.withLibraryNS("DS-labels")
```
rushtong: This will need to change depending on what the field is actually named in GAWB-2758

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -26,17 +33,66 @@ trait DataUseRestrictionSupport extends LazyLogging {
...
+          ""
+      }
+      case (attr: AttributeName, value: Attribute) =>
+        logger.debug(s"Ignored attribute: ${attr.name}; value: ${value.getClass.getName}")
+        Seq.empty[String]
+    }.toSeq.filter(_.nonEmpty) match {
+      case x if x.nonEmpty => Map(dataUseDisplayAttributeName -> AttributeValueList(x.map(AttributeString)))
+      case _ => Map.empty[AttributeName, Attribute]
+    }
+  }
```
davidangb: I would change things in here so that:
* `getDataUseAttributes` returns only those attrs in `durFieldNames` (not also `diseaseLabelsAttributeName`)
* `generateStructuredUseRestrictionAttribute` doesn't re-filter the result of `getDataUseAttributes`, since it can now assume it is receiving the right content. Otherwise, we're filtering almost the same thing twice.
* `generateUseRestrictionDisplayAttribute` internally calls `getDataUseAttributes` and then reaches out itself to get `workspace.attributes.get(diseaseLabelsAttributeName)`.

If that's the case - AND if we don't care about logging warnings - then `generateUseRestrictionDisplayAttribute` can simplify a bit:
```
 def generateUseRestrictionDisplayAttribute(workspace: Workspace): Map[AttributeName, Attribute] = {

    val booleanCodes:Seq[String] = getDataUseAttributes(workspace).collect {
      case (attr: AttributeName, AttributeBoolean(true)) => attr.name
    }.toSeq

    val dsLabels:Seq[String] = (workspace.attributes.get(diseaseLabelsAttributeName) collect {
      case value: AttributeValueList => value.list.collect {
        case a: AttributeString => "DS:" + a.value
      }
    }).getOrElse(Seq.empty[String])

    val displayCodes = booleanCodes ++ dsLabels

    if (displayCodes.nonEmpty)
      Map(dataUseDisplayAttributeName -> AttributeValueList(displayCodes.map(AttributeString)))
    else
      Map.empty[AttributeName, Attribute]
  }
```

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -66,27 +122,31 @@ trait DataUseRestrictionSupport extends LazyLogging {
...
+                }
+              case avl@(a: AttributeNumber) => AttributeNumber(a.value.toInt)
+              case _ => AttributeNumber(0)
+            }
+            case _ => Seq.empty[AttributeNumber]
+          }
+          diseaseNumericIdValues.filter(_.value > 0) match {
+            case values if values.nonEmpty => Map(AttributeName.withDefaultNS(attr.name) -> AttributeValueList(values))
+            case _ => Map.empty[AttributeName, Attribute]
+          }
```
davidangb: again if we didn't care about logging, this case could be a lot more terse:
```
        // Turn DS string ids into numeric IDs for ES indexing
        case (attr: AttributeName, value: AttributeValueList) if attr.name.equals("DS") =>
          val diseaseNumericIdValues = value.list.collect {
            case a: AttributeString => Try(DiseaseOntologyNodeId(a.value)).toOption.map(_.numericId)
          }.flatten

          if (diseaseNumericIdValues.nonEmpty)
            Map(AttributeName.withDefaultNS(attr.name) -> AttributeValueList(diseaseNumericIdValues.map(n => AttributeNumber(n))))
          else
            Map.empty[AttributeName, Attribute]
```

I'm debating in my own head how important the logging is.

src/test/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupportSpec.scala
```diff
@@ -13,105 +13,178 @@ class DataUseRestrictionSupportSpec extends FreeSpec with Matchers with DataUseR
...
+          }
+        }
+
+        "'EVERYTHING' dataset should have the right codes" in {
+          val attrs: Map[AttributeName, Attribute] = generateUseRestrictionDisplayAttribute(everythingDataset.head)
+          val codes: Seq[String] = getValuesFromAttributeValueListAsAttribute(attrs.get(dataUseDisplayAttributeName))
+          (booleanCodes ++ Seq("RS-G", "RS-FM", "NAGR")).map { c => codes should contain(c) }
+          diseaseValuesLabels.map { c => codes should contain("DS:" + c) }
+          Seq("RS-M", "RS-POP").map { c => codes shouldNot contain(c) }
+        }
```
davidangb: I generally don't like using `contain` tests, because it's too easy to miss a value on the left or right. Could you instead create the set of expected codes, and compare equality to (or `theSameElementsAs`) the actual codes?

src/test/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupportSpec.scala
```diff
@@ -120,8 +193,23 @@ class DataUseRestrictionSupportSpec extends FreeSpec with Matchers with DataUseR
...
+          case x: AttributeValueList => x.list.map {
+            case avl@(a: AttributeString) => a.value
+            case _ => ""
+          }
+          case unhandled => Seq.empty[String]
+        }
+      case None => Seq.empty[String]
+    }
+    results.distinct.filter(_.nonEmpty)
+  }
```
davidangb: more simplification:
```
  private def getValuesFromAttributeValueListAsAttribute(attrs: Option[Attribute]): Seq[String] = {
    (attrs collect {
      case x: AttributeValueList => x.list.collect {
        case a: AttributeString => a.value
      }
    }).getOrElse(Seq.empty[String])
  }
```

davidangb (commented):
Looks good functionally. I have comments inline on Scala syntax/structure. Big question is how much we care about logging bad data - if we don't, the code can become smaller.

davidangb (approved):
:+1: nice!

( boy, I hope I don't regret recommending removing the logging )

rushtong:
I think we'll be OK without the logging. For things that don't show up, we
need to communicate to Moran that they need to be updated and re-published.

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -1,18 +1,24 @@
...
+
 trait DataUseRestrictionSupport extends LazyLogging {
 
   private val booleanCodes: Seq[String] = Seq("GRU", "HMB", "NCU", "NPU", "NDMS", "NAGR", "NCTRL", "RS-PD")
   private val listCodes: Seq[String] = Seq("DS", "RS-POP")
-  val durFieldNames: Seq[String] = booleanCodes ++ listCodes ++ Seq("RS-G")
+  private val genderCodes: Seq[String] = Seq("RS-G", "RS-FM", "RS-M")
+  val durFieldNames: Seq[String] = booleanCodes ++ listCodes ++ genderCodes
 
+  val diseaseLabelsAttributeName: AttributeName = AttributeName.withLibraryNS("DS_URL")
```
ahaessly: DS is where i'm displaying the label and DS_URL is the DOID URL. i think the use of these in this code may be switched. for list codes, if you want the DOID, you should use DS_URL. and for the display name you should use DS. i can change the names to something less confusing. do you have a suggestion?

actually,it may be confusing that there is a DS in the data use restriction object and a library:DS attribute.

rushtong: I might be confused. Currently, we store the regular old string value in the `library:DS` array. New behavior will be:
`library:DS` -> holds the string arrays of disease labels
`library:DS_URL` -> holds the full urls of the disease ontology terms.

If that is correct (and I think it is), then I do need to update this.

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -7,16 +7,16 @@ import org.broadinstitute.dsde.rawls.model._
...
+import scala.util.Try
 
 trait DataUseRestrictionSupport extends LazyLogging {
 
   private val booleanCodes: Seq[String] = Seq("GRU", "HMB", "NCU", "NPU", "NDMS", "NAGR", "NCTRL", "RS-PD")
-  private val listCodes: Seq[String] = Seq("DS", "RS-POP")
   private val genderCodes: Seq[String] = Seq("RS-G", "RS-FM", "RS-M")
-  val durFieldNames: Seq[String] = booleanCodes ++ listCodes ++ genderCodes
+  private val duRestrictionFieldNames: Seq[String] = booleanCodes ++ genderCodes ++ Seq("DS_URL", "RS-POP")
+  val allDurFieldNames: Seq[String] = booleanCodes ++ genderCodes ++ Seq("DS", "DS_URL", "RS-POP")
```
davidangb: `val allDurFieldNames: Seq[String] = duRestrictionFieldNames :+ "DS"` ?

davidangb (approved):
:+1: again!

ahaessly (approved):
👍

PR #528, opened 2017-11-07
GAWB-2795: user-initiated endpoint to enroll in free trial

davidangb:
Basic endpoint, model objects, some scaffolding for a user to initiate their own start in the free trial.

This does NOT include billing account/project work - that's for a future story. This is a pretty rudimentary endpoint - the big goal of this story is to deal with daos and model objects and such, so that future stories are easier.

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ThurloeDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ThurloeDAO.scala
@@ -24,6 +24,7 @@ trait ThurloeDAO extends LazyLogging with ReportsSubsystemStatus {
   def saveProfile(userInfo: UserInfo, profile: BasicProfile): Future[Unit]
   def saveKeyValues(userInfo: UserInfo, keyValues: Map[String, String]): Future[Try[Unit]]
 
+  // methods to work with free-trial objects
   def getTrialStatus(userInfo: UserInfo): Future[Option[UserTrialStatus]]
   def saveTrialStatus(userInfo: UserInfo, trialStatus: UserTrialStatus): Future[Unit]
 
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ThurloeDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ThurloeDAO.scala
@@ -1,6 +1,7 @@
 package org.broadinstitute.dsde.firecloud.dataaccess
 
 import com.typesafe.scalalogging.slf4j.LazyLogging
+import org.broadinstitute.dsde.firecloud.model.Trial.UserTrialStatus
 import org.broadinstitute.dsde.firecloud.model.{BasicProfile, Profile, UserInfo}
 import org.broadinstitute.dsde.rawls.model.ErrorReportSource
 
@@ -22,5 +23,10 @@ trait ThurloeDAO extends LazyLogging with ReportsSubsystemStatus {
   def getAllUserValuesForKey(key: String): Future[Map[String, String]]
   def saveProfile(userInfo: UserInfo, profile: BasicProfile): Future[Unit]
   def saveKeyValues(userInfo: UserInfo, keyValues: Map[String, String]): Future[Try[Unit]]
+
+  def getTrialStatus(userInfo: UserInfo): Future[Option[UserTrialStatus]]
+  def saveTrialStatus(userInfo: UserInfo, trialStatus: UserTrialStatus): Future[Unit]
+
+
   override def serviceName:String = ThurloeDAO.serviceName
 }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/model/Trial.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/model/Trial.scala
@@ -1,0 +1,85 @@
+package org.broadinstitute.dsde.firecloud.model
+
+import org.broadinstitute.dsde.firecloud.FireCloudException
+import org.broadinstitute.dsde.firecloud.model.Trial.TrialStates.TrialState
+
+import scala.util.Try
+
+object Trial {
+
+  object TrialStates {
+    sealed trait TrialState  {
+      override def toString: String = TrialStates.stringify(this)
+      def withName(name: String): TrialState = TrialStates.withName(name)
+    }
+
+    def withName(name: String): TrialState = {
+      name match {
+        case "Enabled" => Enabled
+        case "Enrolled" => Enrolled
+        case "Terminated" => Terminated
+        case _ => throw new FireCloudException(s"invalid TrialState [$name]")
+      }
+    }
+
+    def stringify(state: TrialState): String = {
+      state match {
+        case Enabled => "Enabled"
+        case Enrolled => "Enrolled"
+        case Terminated => "Terminated"
+        case _ => throw new FireCloudException(s"invalid TrialState [${state.getClass.getName}]")
+      }
+    }
+
+    case object Enabled extends TrialState
+    case object Enrolled extends TrialState
+    case object Terminated extends TrialState
+  }
+
+  case class UserTrialStatus(
+    userId: String,
+    currentState: Option[TrialState],
+    enabledDate: Long,    // timestamp a campaign manager granted the user trial permissions
+    enrolledDate: Long,   // timestamp user started their trial
+    terminatedDate: Long, // timestamp user was actually terminated
+    expirationDate: Long  // timestamp user is due to face termination
+    )
+
+  object UserTrialStatus {
+    def apply(profileWrapper:ProfileWrapper) = {
+
+      def profileDate(key: String, kvps: Map[String,String], default: Int = 0): Int = {
+        Try(kvps.getOrElse(key, default.toString).toInt).toOption.getOrElse(default)
+      }
+
+      val mappedKVPs:Map[String,String] = (profileWrapper.keyValuePairs collect {
+        case kvp if kvp.key.nonEmpty && kvp.value.nonEmpty => kvp.key.get -> kvp.value.get
+      }).toMap
+
+      val enabledDate = profileDate("trialEnabledDate", mappedKVPs)
+      val enrolledDate = profileDate("trialEnrolledDate", mappedKVPs)
+      val terminatedDate = profileDate("trialTerminatedDate", mappedKVPs)
+      val expirationDate = profileDate("trialExpirationDate", mappedKVPs)
+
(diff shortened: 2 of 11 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/model/Trial.scala
```diff
@@ -0,0 +1,85 @@
...
+      val stateKV:Map[String,String] = userTrialStatus.currentState match {
+        case Some(state) => Map("trialCurrentState" -> state.toString)
+        case None => Map.empty[String,String]
+      }
+      Map(
+        "trialEnabledDate" -> userTrialStatus.enabledDate.toString,
+        "trialEnrolledDate" -> userTrialStatus.enrolledDate.toString,
+        "trialTerminatedDate" -> userTrialStatus.terminatedDate.toString,
+        "trialExpirationDate" -> userTrialStatus.expirationDate.toString
+      ) ++ stateKV
```
davidangb: the `mappedKVPs` and `toKVPs` parallel similar code in `Profile` and `mappedPropVals`. But those are really old classes we wrote when we had less experience with Scala or with spray-json, and I didn't want to reuse that code.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -0,0 +1,90 @@
...
+  // TODO: implement fully! Check that the user does not already have a state, before overwriting.
+  // this method exists solely for developer-testing purposes right now.
+  private def enableUser(userInfo: UserInfo): Future[PerRequestMessage] = {
+    // build the state that we want to persist to indicate the user is enabled
+    val now = System.currentTimeMillis
+    val enabledStatus = UserTrialStatus(userInfo.id, Some(TrialStates.Enabled), now, 0, 0, 0)
+    thurloeDao.saveTrialStatus(userInfo, enabledStatus) map { _ =>
+      RequestComplete(spray.http.StatusCodes.OK)
+    }
+  }
```
davidangb: As a developer running your own local instance of orchestration, you can call `enableUser` on yourself by hacking the code ... which puts you in a position for `enrollUser` to succeed.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -0,0 +1,90 @@
...
+                // TODO: add user to free-trial billing project / create said project if necessary
+                RequestComplete(NoContent)
+              }
+            }
+            // user in some other state; don't enroll
+            case _ => Future(RequestCompleteWithErrorReport(BadRequest, "You are not eligible for a free trial."))
+          }
+      }
+    }
+  }
```
davidangb, line 87: I'm being lazy and not dealing with conf updates. We'll need a conf update soon to hold the name of the campaign manager group ... so I'm hoping to do them all at the same time.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
```diff
@@ -71,7 +65,9 @@ trait UserApiService extends HttpService with PerRequestCreator with FireCloudRe
 
   lazy val log = LoggerFactory.getLogger(getClass)
 
-  val routes =
+  val trialServiceConstructor: () => TrialService
```
davidangb, line 68: refactoring UserApiService was necessary to get the `trialServiceConstructor` properly injected from `FireCloudServiceActor` ... unfortunately this meant I had to rename `routes` to `userServiceRoutes` and that cascaded changes elsewhere.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ThurloeDAO.scala
```diff
@@ -22,5 +23,10 @@ trait ThurloeDAO extends LazyLogging with ReportsSubsystemStatus {
   def getAllUserValuesForKey(key: String): Future[Map[String, String]]
   def saveProfile(userInfo: UserInfo, profile: BasicProfile): Future[Unit]
   def saveKeyValues(userInfo: UserInfo, keyValues: Map[String, String]): Future[Try[Unit]]
+
```
kyuksel, line 26: I wonder if `*TrialStatus()` are too (Free Trial) application-specific to be in this file.

Comments would also be helpful.

src/main/scala/org/broadinstitute/dsde/firecloud/model/Trial.scala
```diff
@@ -0,0 +1,85 @@
...
+    enabledDate: Long,    // timestamp a campaign manager granted the user trial permissions
+    enrolledDate: Long,   // timestamp user started their trial
+    terminatedDate: Long, // timestamp user was actually terminated
+    expirationDate: Long  // timestamp user is due to face termination
+    )
+
+  object UserTrialStatus {
+    def apply(profileWrapper:ProfileWrapper) = {
+
+      def profileDate(key: String, kvps: Map[String,String], default: Int = 0): Int = {
```
kyuksel: Would `java.time.Instant` or such be better to represent dates?

src/main/scala/org/broadinstitute/dsde/firecloud/model/Trial.scala
```diff
@@ -0,0 +1,85 @@
...
+        Try(kvps.getOrElse(key, default.toString).toInt).toOption.getOrElse(default)
+      }
+
+      val mappedKVPs:Map[String,String] = (profileWrapper.keyValuePairs collect {
+        case kvp if kvp.key.nonEmpty && kvp.value.nonEmpty => kvp.key.get -> kvp.value.get
+      }).toMap
+
```
kyuksel, line 73: I wonder if `trial*Date` strings below could be more strongly-typed than `String` to catch errors in compile-time.

src/main/scala/org/broadinstitute/dsde/firecloud/model/Trial.scala
```diff
@@ -0,0 +1,85 @@
+package org.broadinstitute.dsde.firecloud.model
+
+import org.broadinstitute.dsde.firecloud.FireCloudException
+import org.broadinstitute.dsde.firecloud.model.Trial.TrialStates.TrialState
+
+import scala.util.Try
+
+object Trial {
+
```
kyuksel, line 11: Not a big deal but `Trial` in `TrialStates` and `UserTrialStatus` could probably be dropped for more conciseness since the outer `object Trial` already provides the namespacing.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -0,0 +1,90 @@
...
+  private def enableUser(userInfo: UserInfo): Future[PerRequestMessage] = {
+    // build the state that we want to persist to indicate the user is enabled
+    val now = System.currentTimeMillis
+    val enabledStatus = UserTrialStatus(userInfo.id, Some(TrialStates.Enabled), now, 0, 0, 0)
+    thurloeDao.saveTrialStatus(userInfo, enabledStatus) map { _ =>
+      RequestComplete(spray.http.StatusCodes.OK)
+    }
+  }
+
+  private def enrollUser(userInfo:UserInfo): Future[PerRequestMessage] = {
```
kyuksel, line 56: The comments are very helpful 👍🏼

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -0,0 +1,90 @@
...
+            // user already enrolled; don't re-enroll
+            case Some(TrialStates.Enrolled) => Future(RequestCompleteWithErrorReport(BadRequest, "You are already enrolled in a free trial."))
+            // user enabled (eligible) for trial, enroll!
+            case Some(TrialStates.Enabled) => {
+              // build the new state that we want to persist to indicate the user is enrolled
+              val now = System.currentTimeMillis
+              val enrolledStatus = status.copy(
+                currentState = Some(TrialStates.Enrolled),
+                enrolledDate = now,
+                expirationDate = now + (60 * 24 * 60 * 60 * 1000) // add 60 days from now
```
kyuksel: Should `60` (days) be set in a config, or at least somewhere higher-level? 

Using a date/time library may also save the conversion work and make it less error-prone.

src/main/scala/org/broadinstitute/dsde/firecloud/model/Trial.scala
davidangb: Thurloe only knows how to work with Strings, so everything up to and including `ProfileWrapper` must be a String. It's here - in the `apply` method that converts `ProfileWrapper` to `UserTrialStatus` - that I can finally type them.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
davidangb: yes, the 60-days should be stored in config! That's my other comments ... I'm trying to batch up the config updates because they're a pain.

rushtong (approved):
👍  Looks good pending stories to cover the to-dos and config updates.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ThurloeDAO.scala
davidangb: By convention, all our daos work with higher-level business objects. The `getProfile` a few lines above returns a `Profile`, for instance.

src/main/scala/org/broadinstitute/dsde/firecloud/model/Trial.scala
davidangb: I wanted to be clear here for cases like in `ThurloeDAO`, where you have:
```
import org.broadinstitute.dsde.firecloud.model.Trial.UserTrialStatus
...
def getTrialStatus(userInfo: UserInfo): Future[Option[UserTrialStatus]]
```

just looking at the signature for the method, without reading the imports, you don't necessarily understand the `Trial` namespacing.

davidangb:
@kyuksel @rushtong back to you for re-review. I've updated code to address some comments and replied to others. I will handle config updates as part of GAWB-2827 (I claimed it), and will be writing stories for the rest of this epic - including the TODOs - today.

kyuksel (approved):

rushtong (approved):
👍

aednichols (approved):
Looks good to me. No objections to merging partial implementation, as no production codepaths are affected yet.

PR #531, opened 2017-11-10
GAWB-2860: query param to conditionally validate Library metadata

davidangb:
Addresses an issue Kathleen ran into.

This PR adds a `validate` query param to the PUT-library metadata endpoint, so users can conditionally turn on validation. Previously, we always validated for published workspaces, never validated for unpublished workspaces. This means that API users (like Kathleen and team) who were trying to upload Library metadata had no programmatic guidance for their API calls.

Now, they can toggle `validate=true` and get feedback.

if the user does not specify a value for `validate` or specifies a value that doesn't parse to a Boolean, behavior is as before: don't validate for unpublished, do validate for published.

In all cases:

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -895,8 +895,8 @@ paths:
         - $ref: '#/parameters/workspaceNameParam'
         - name: validate
           description: >
-            should the payload be validated for correctness and completeness? Defaults to false if unpublished
-            and true if published.
+            Should the payload be validated? Validation is always enabled
+            for published datasets and defaults to false for unpublished datasets.
           in: query
           type: boolean
           required: false
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -893,6 +893,13 @@ paths:
       parameters:
         - $ref: '#/parameters/workspaceNamespaceParam'
         - $ref: '#/parameters/workspaceNameParam'
+        - name: validate
+          description: >
+            should the payload be validated for correctness and completeness? Defaults to false if unpublished
+            and true if published.
+          in: query
+          type: boolean
+          required: false
         - name: libraryMetadataJson
           description: Library metadata
           required: true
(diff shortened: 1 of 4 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -893,6 +893,13 @@ paths:
       parameters:
         - $ref: '#/parameters/workspaceNamespaceParam'
         - $ref: '#/parameters/workspaceNameParam'
+        - name: validate
+          description: >
+            should the payload be validated for correctness and completeness? Defaults to false if unpublished
```
kyuksel: Capitalize `s` in `should`

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/LibraryApiService.scala
```diff
@@ -81,10 +81,14 @@ trait LibraryApiService extends HttpService with FireCloudRequestBuilding
           pathPrefix(Segment / Segment) { (namespace, name) =>
             path("metadata") {
               put {
-                entity(as[String]) { rawAttrsString => requestContext =>
-                  perRequest(requestContext,
-                    LibraryService.props(libraryServiceConstructor, userInfo),
-                    LibraryService.UpdateLibraryMetadata(namespace, name, rawAttrsString))
+                parameter("validate" ? "false") { validationParam =>
+                  val doValidate = java.lang.Boolean.valueOf(validationParam) // for lenient parsing
```
kyuksel, line 85: How is the case handled when the value of `validate` within the query is a string other than `true` or `false`?

kyuksel:
Verified that the unit tests pass and the API behaves as expected on Swagger. Couldn't be sure about the case where the value of `validate` is not a boolean string (see the question above).

davidangb:
I updated the parameter description in swagger.

Regarding `validate` when the querystring other than `true` or `false`: the `doValidate` boolean is false in this case - that's exactly why I used `java.lang.Boolean.valueOf` instead of Scala's `String.toBoolean`: the former is lenient and the latter throws an exception.

The non-boolean string cases are covered in the new unit tests that include "if user specifies a non-boolean value".

EDIT: oh, and if you want to test non-boolean values yourself, you can copy the curl command from swagger, mod it, and run it with your mods!

kyuksel (approved):

PR #534, opened 2017-11-14
GAWB-2889: lenient library sort + helpful errors on search

davidangb:
no jira, found by Andrew. Two changes in this PR for the library search endpoint:

1. treat `"sortField": ""` in the payload the same as if the user omitted `"sortField"`: the search applies no explicit sort.
2. when ES encounters an exception in its query, make an attempt to bubble the ES error message up to the user. Previously, this endpoint just gave a plain 500 error with no ErrorReport.

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
@@ -245,7 +245,7 @@ class LibraryService (protected val argUserInfo: UserInfo,
 
   private def errorMessageFromSearchException(ex: Throwable): String = {
     // elasticsearch errors are often nested, try to dig into them safely to find a message
-    if (ex.getCause != null) {
+    val message = if (ex.getCause != null) {
       val firstCause = ex.getCause
       if (firstCause.getCause != null) {
         firstCause.getCause.getMessage
@@ -255,6 +255,12 @@ class LibraryService (protected val argUserInfo: UserInfo,
     } else {
       ex.getMessage
     }
+
+    Option(message) match {
+      case Some(m:String) => m
+      case _ => "Unknown error during search."
+    }
+
   }
 
 }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
@@ -3,7 +3,7 @@ package org.broadinstitute.dsde.firecloud.service
 import akka.actor.{Actor, Props}
 import akka.pattern._
 import com.typesafe.scalalogging.slf4j.LazyLogging
-import org.broadinstitute.dsde.firecloud.{Application, FireCloudConfig, FireCloudException}
+import org.broadinstitute.dsde.firecloud.{Application, FireCloudConfig, FireCloudException, FireCloudExceptionWithErrorReport}
 import org.broadinstitute.dsde.firecloud.dataaccess.{OntologyDAO, RawlsDAO, SearchDAO}
 import org.broadinstitute.dsde.rawls.model._
 import org.broadinstitute.dsde.firecloud.model._
@@ -218,14 +218,17 @@ class LibraryService (protected val argUserInfo: UserInfo,
   }
 
   def findDocuments(criteria: LibrarySearchParams): Future[PerRequestMessage] = {
-    getEffectiveDiscoverGroups(rawlsDAO) map { userGroups =>
+    getEffectiveDiscoverGroups(rawlsDAO) flatMap { userGroups =>
       val docsFuture = searchDAO.findDocuments(criteria, userGroups)
       val idsFuture = rawlsDAO.getWorkspaces
-      val searchResults = for {
+
+      (for {
         docs <- docsFuture
         ids <- idsFuture
-      } yield updateAccess(docs, ids)
-      RequestComplete(searchResults)
+      } yield RequestComplete(updateAccess(docs, ids))) recover {
+        case ex =>
+          throw new FireCloudExceptionWithErrorReport(ErrorReport(errorMessageFromSearchException(ex)))
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -218,14 +218,17 @@ class LibraryService (protected val argUserInfo: UserInfo,
...
         docs <- docsFuture
         ids <- idsFuture
-      } yield updateAccess(docs, ids)
-      RequestComplete(searchResults)
+      } yield RequestComplete(updateAccess(docs, ids))) recover {
+        case ex =>
+          throw new FireCloudExceptionWithErrorReport(ErrorReport(errorMessageFromSearchException(ex)))
+      }
     }
   }
```
kyuksel, line 235: In case it's clearer:

```scala
  def findDocuments(criteria: LibrarySearchParams): Future[PerRequestMessage] = {
    val request = for {
      userGroups <- getEffectiveDiscoverGroups(rawlsDAO)
      docs <- searchDAO.findDocuments(criteria, userGroups)
      ids <- rawlsDAO.getWorkspaces
    } yield RequestComplete(updateAccess(docs, ids))

    request recover {
      case ex =>
        throw new FireCloudExceptionWithErrorReport(ErrorReport(errorMessageFromSearchException(ex)))
    }
  }
```

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -239,4 +242,19 @@ class LibraryService (protected val argUserInfo: UserInfo,
       case e: FireCloudException => Future(RequestCompleteWithErrorReport(BadRequest, s"suggestions not available for field %s".format(field)))
     }
   }
+
+  private def errorMessageFromSearchException(ex: Throwable): String = {
+    // elasticsearch errors are often nested, try to dig into them safely to find a message
```
kyuksel, line 249: Are the errors at most two-level-deep? I'm not familiar with ElasticSearch but I wonder if there's a generic (and hopefully more idiomatic) way of retrieving the message.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -239,4 +242,19 @@ class LibraryService (protected val argUserInfo: UserInfo,
...
+    if (ex.getCause != null) {
+      val firstCause = ex.getCause
+      if (firstCause.getCause != null) {
+        firstCause.getCause.getMessage
+      } else {
+        firstCause.getMessage
+      }
+    } else {
+      ex.getMessage
```
kyuksel, line 258: Should we worry about `getMessage`s returning `null` as well?

kyuksel:
There's a typo in `sortField` descriptions of `/api/library/search` and `/api/library/suggest` Swaggers' `Model` sections where `so` should be `to`. It may be good to correct that while at it.

src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/SortSpec.scala
```diff
@@ -36,7 +36,14 @@ class SortSpec extends FreeSpec with Matchers with BeforeAndAfterAll with LazyLo
...
     }
     "search with no sort (or filter) criteria" - {
-      "returns all results" in {
+      "returns all results in engine-defined order" in {
+        val searchResponse = sortBy(None, None)
+        assertResult(IntegrationTestFixtures.datasetTuples.size) {searchResponse.total}
+        // results are sorted by relevancy/native index order, which we won't test here
+      }
+    }
+    "search with empty string as sort criteria" - {
```
kyuksel: `"search with empty string as sort criteria"` seems to be a duplicate of `"search with no sort (or filter) criteria"`. Was it meant to input empty strings instead of `None`s?

kyuksel (commented):
`http://local.broadinstitute.org/#library` on my UI pointing to the changes results in `POST http://localhost:8080/api/library/search 400 (Bad Request)`. This doesn't happen with the `develop` branch so something might have been broken but I didn't get a chance to investigate.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
davidangb: The current syntax is intentional (and I've added a code comment to that effect). For-yields are, if I understand correctly, serialized. By declaring the `docsFuture` and `idsFuture` before referencing them in the for-yield, they can be parallelized. The `getEffectiveDiscoverGroups` call *needs* to come first, because its output is used as input to `findDocuments` - so it goes outside the for-yield entirely.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
davidangb: I honestly don't know. I tried all the error cases I could think of, like setting a negative number for the search `size` param, or setting a `sortField` that isn't a real field in ES, and in those attempts I saw ES errors nested two-deep. The first level is "all shards failed" and then the second level indicates *how* the shard failed. So it seems better, but I haven't done a comprehensive audit.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
davidangb: I don't think it will be null - but good call and I've updated the code to be super safe.

src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/SortSpec.scala
davidangb: aaah good catch! Fixed.

davidangb:
@kyuksel back to you for re-review. I've addressed all your comments either with code change or with a reply. I did fix the typo in swagger too. 

Re: the 400 error you were getting - this branch needed a rebase (develop was moving pretty fast!). It's working for me now.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
kyuksel: I see. Thanks for the clarification!

kyuksel (approved):
Looks great!

PR #536, opened 2017-11-15
GAWB-2758: capture ontology node info

ahaessly:

In all cases:

### DIFF
--- a/src/main/resources/library/attribute-definitions.json
+++ b/src/main/resources/library/attribute-definitions.json
@@ -240,59 +240,77 @@
       "title": "Future use is limited to research involving the following disease area(s)",
       "type": "array",
       "items": { "type": "string" },
-      "consentCode": "DS"
+      "typeahead": "ontology",
+      "consentCode": "DS",
+      "relatedID": "library:DS_URL",
+      "relatedLabel": "library:DS",
+      "indexable": false
+    },
+    "library:DS_URL": {
+      "type": "array",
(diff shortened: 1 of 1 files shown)


src/main/resources/library/attribute-definitions.json
```diff
@@ -240,59 +240,77 @@
...
+    "library:DS_URL": {
+      "type": "array",
+      "items": { "type": "string" },
+      "hidden": true
     },
     "library:NCU": {
       "title": "Future commercial use is prohibited",
       "type": "boolean",
       "renderHint": {"wording": "yes/no"},
+      "indexable": false,
```
rushtong, line 283: Do we need to specify indexable false here? See https://github.com/broadinstitute/firecloud-orchestration/blob/3df70194b3ac98b3134ac33a6cc810784e3884a2/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala#L88

On that line, we remove the indexable fields from the document that we've already used for generating anything data-use related. That includes this one, and several others that are also annotated below.

davidangb: Greg, I think you're right that it's not needed, but it also doesn't hurt to have indexable: false. Now it's *really* not indexed.

rushtong (approved):
👍

davidangb (approved):
:+1:, assuming the UI PR doesn't require any changes here.

PR #537, opened 2017-11-15
GAWB-2863: de-duplicate per-field suggestions for catalog wizard

davidangb:
11 characters for the runtime fix, then a couple tests ...

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
@@ -233,7 +233,7 @@ trait ElasticSearchDAOQuerySupport extends ElasticSearchDAOSupport with ElasticS
     results map { suggestResult =>
       val compSugg: CompletionSuggestion = suggestResult.getSuggest.getSuggestion(suggestionName)
       val options : Seq[CompletionSuggestion.Entry.Option] = compSugg.getEntries.get(0).getOptions.asScala
-      options map { option: CompletionSuggestion.Entry.Option => option.getText.string }
+      (options map { option: CompletionSuggestion.Entry.Option => option.getText.string }).distinct
     }
   }
 
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
@@ -232,8 +232,16 @@ trait ElasticSearchDAOQuerySupport extends ElasticSearchDAOSupport with ElasticS
 
     results map { suggestResult =>
       val compSugg: CompletionSuggestion = suggestResult.getSuggest.getSuggestion(suggestionName)
-      val options : Seq[CompletionSuggestion.Entry.Option] = compSugg.getEntries.get(0).getOptions.asScala
-      (options map { option: CompletionSuggestion.Entry.Option => option.getText.string }).distinct
+      if (compSugg != null) {
+        val entries = compSugg.getEntries
+        if (!entries.isEmpty) {
+          entries.get(0).getOptions.asScala.map(_.getText.string).distinct
+        } else {
+          Seq.empty[String] // getEntries returned empty
+        }
+      } else {
+        Seq.empty[String] // getSuggestion(suggestionName) returned null
+      }
     }
   }
 
--- a/src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/IntegrationTestFixtures.scala
+++ b/src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/IntegrationTestFixtures.scala
@@ -43,6 +43,7 @@ object IntegrationTestFixtures {
     Document(x._1, Map(
       AttributeName("library","datasetName") -> AttributeString(x._1),
       AttributeName("library","indication") -> AttributeString(x._2),
+      AttributeName("library","datasetOwner") -> AttributeString(x._2),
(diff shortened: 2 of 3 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
```diff
@@ -233,7 +233,7 @@ trait ElasticSearchDAOQuerySupport extends ElasticSearchDAOSupport with ElasticS
     results map { suggestResult =>
       val compSugg: CompletionSuggestion = suggestResult.getSuggest.getSuggestion(suggestionName)
       val options : Seq[CompletionSuggestion.Entry.Option] = compSugg.getEntries.get(0).getOptions.asScala
-      options map { option: CompletionSuggestion.Entry.Option => option.getText.string }
+      (options map { option: CompletionSuggestion.Entry.Option => option.getText.string }).distinct
```
kyuksel:
```
options.map(_.getText.string).distinct
```
reads easier to me in this case. 

It could be even clearer if the whole block (lines 233-238) were turned into a `for comprehension`.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
```diff
@@ -233,7 +233,7 @@ trait ElasticSearchDAOQuerySupport extends ElasticSearchDAOSupport with ElasticS
     results map { suggestResult =>
       val compSugg: CompletionSuggestion = suggestResult.getSuggest.getSuggestion(suggestionName)
       val options : Seq[CompletionSuggestion.Entry.Option] = compSugg.getEntries.get(0).getOptions.asScala
```
kyuksel: Should we guard against and/or handle `compSugg.getEntries` returning an empty `List`? I couldn't see a test for it, either.

src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/IntegrationTestFixtures.scala
```diff
@@ -43,6 +43,7 @@ object IntegrationTestFixtures {
     Document(x._1, Map(
       AttributeName("library","datasetName") -> AttributeString(x._1),
       AttributeName("library","indication") -> AttributeString(x._2),
+      AttributeName("library","datasetOwner") -> AttributeString(x._2),
       AttributeName("library","numSubjects") -> AttributeNumber(x._3)
     ))
   }
```
kyuksel: I'd suggest something like the following for readability:

```
val fixtureDocs:Seq[Document] = datasetTuples map { 
    case (access, phenotype, count) =>
      Document(access, Map(
        AttributeName("library","datasetName") -> AttributeString(access),
        AttributeName("library","indication") -> AttributeString(phenotype),
        AttributeName("library","datasetOwner") -> AttributeString(phenotype),
        AttributeName("library","numSubjects") -> AttributeNumber(count)
      ))
  }```

kyuksel (commented):
Verified that the unit tests pass and the fix works locally. 

A few comments inline...

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
aednichols: I doubt it would get that far, because there are a couple ways for `getSuggestion()` to return `null` first.
https://github.com/elastic/elasticsearch/blob/093218e052acc84338e90e1ab77ed413843a6244/core/src/main/java/org/elasticsearch/search/suggest/Suggest.java#L106

We also have a top-level exception handler, though it's specific to `FireCloudException`s
https://github.com/broadinstitute/firecloud-orchestration/blob/develop/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala#L242

It could just be that this case is undefined behavior where our system is concerned, and we don't care what happens ("Don't do that.").

aednichols (approved):
LGTM 👍

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
davidangb: I've added some null/empty protection for the sake of safety, though it does make the code more verbose.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
davidangb: I took the `map` suggestion - good call!

This code doesn't lend well to a for comp, because it mixes future resolution and non-futures, as well as mixing Scala and Java (the ES API is pure Java), so I didn't rewrite it.

src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/IntegrationTestFixtures.scala
davidangb: good call, I've made this change - though we have the same pattern in other test fixtures, and I did NOT fix there.

davidangb:
@kyuksel @aednichols back to you for re-review, and thanks!

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
aednichols: Doesn't seem too necessary to me (would this ever actually happen?), but appears correct

aednichols (approved):

kyuksel (approved):

PR #538, opened 2017-11-15
GAWB-2820: Add dataAccessInstructions property and question

MattPutnam:
Pretty straightforward.

In all cases:

### DIFF
--- a/src/main/resources/library/attribute-definitions.json
+++ b/src/main/resources/library/attribute-definitions.json
@@ -214,6 +214,12 @@
       "default": false,
       "inputHint": "i.e. approval from a second source is required in order to access the data"
     },
+    "library:dataAccessInstructions": {
+      "title": "Data Access Instructions",
+      "type": "string",
+      "renderHint": {"datatype": "markdown"},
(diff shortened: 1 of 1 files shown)


src/main/resources/library/attribute-definitions.json
```diff
@@ -214,6 +214,12 @@
       "default": false,
       "inputHint": "i.e. approval from a second source is required in order to access the data"
     },
+    "library:dataAccessInstructions": {
+      "title": "Data Access Instructions",
+      "type": "string",
+      "renderHint": {"datatype": "markdown"},
+      "inputHint": "Users will see this message when attempting to access a workspace they are unauthorized to enter. Leave blank for a default message with the Contact Email above."
```
davidangb: add `"indexable": false` ... no reason for these instructions to be searchable.

aednichols (approved):
👍 

Tested that a markdown message shows up after setting one for a test dataset.

davidangb (approved):
I don't think the `indexable` change made a difference; the data use instructions are still searchable from the main search box. BUT I also don't think it's a fault of this PR, and that we may have a lower level problem in the system. :+1: and we'll follow up on indexable later.

PR #544, opened 2017-11-19
GAWB-2781, GAWB-2782: read ontology ES index directly, bypass consent-ontology REST API

davidangb:
Orchestration queries consent-ontology REST API when:
* getting ontology autocomplete results (research purpose search, dataset cataloging)
* indexing a dataset or reindexing all datasets

we've seen some instability in these requests - intermittent errors. This PR bypasses the consent-ontology REST API and queries the ontology Elasticsearch index directly. Hopefully this is more performant and stable.

I have NOT deleted obsolete code in this PR - `HttpOntologyDAO`, `HttpOntologyDAOConcurrencySpec` to make it easy to reverse this change/compare. But maybe I should just delete them and rely on git history if we need to reverse. Thoughts?

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchOntologyDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchOntologyDAO.scala
@@ -1,0 +1,53 @@
+package org.broadinstitute.dsde.firecloud.dataaccess
+import org.broadinstitute.dsde.firecloud.model.Ontology.TermResource
+import org.broadinstitute.dsde.firecloud.model.SubsystemStatus
+import org.elasticsearch.action.admin.indices.exists.indices.{IndicesExistsRequest, IndicesExistsRequestBuilder, IndicesExistsResponse}
+import org.elasticsearch.action.search.{SearchRequest, SearchRequestBuilder, SearchResponse}
+import org.elasticsearch.client.transport.TransportClient
+import org.elasticsearch.index.query.QueryBuilders._
+import spray.http.Uri.Authority
+import spray.json._
+
+import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol.impOntologyTermResource
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+
+class ElasticSearchOntologyDAO(servers: Seq[Authority], clusterName: String, indexName: String) extends OntologyDAO with ElasticSearchDAOSupport {
+
+  private val client: TransportClient = buildClient(servers, clusterName)
+
+  private final val datatype = "ontology_term"
+
+
+  override def search(term: String): Future[Option[List[TermResource]]] = Future(None)
+
+  override def autocomplete(term: String): List[TermResource] = {
+    val prefix = term.toLowerCase
+    // user's term must be a prefix in either label or synonyms
+    val query = boolQuery()
+      .should(prefixQuery("label", prefix))
+      .should(prefixQuery("synonyms", prefix))
+
+    val searchRequest = client.prepareSearch(indexName)
+      .setQuery(query)
+        .setFetchSource(List("id","ontology","usable","label","synonyms","definition").toArray, null)
+
+    val autocompleteResults = executeESRequest[SearchRequest, SearchResponse, SearchRequestBuilder](searchRequest)
+
+    val allHits = autocompleteResults.getHits.getHits
+    allHits.map { hit =>
+      hit.getSourceAsString.parseJson.convertTo[TermResource]
+    }.toList
+  }
+
+  private def indexExists: Boolean = {
+    executeESRequest[IndicesExistsRequest, IndicesExistsResponse, IndicesExistsRequestBuilder](
+      client.admin.indices.prepareExists(indexName)
+    ).isExists
+  }
+
+  override def status: Future[SubsystemStatus] = {
+    Future(SubsystemStatus(indexExists, None))
+  }
+}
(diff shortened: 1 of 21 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
```diff
@@ -102,16 +98,14 @@ trait LibraryServiceSupport extends DataUseRestrictionSupport with LazyLogging {
...
         terms.head.parents.getOrElse(Seq.empty)
-      case None => Seq.empty[TermParent]
-    } recoverWith {
-      case ex:Exception => {
+      case Success(empty) => Seq.empty[TermParent]
+      case Failure(ex) =>
         logger.warn(s"exception getting term and parents from ontology: ${ex.getMessage}")
-        Future(Seq.empty[TermParent])
-      }
+        Seq.empty[TermParent]
```
davidangb, line 108: A better solution is to change the ontology index itself, such that it holds both the label and the id for any node's parents. Right now it only holds the id - so we have to make these extra queries to populate the labels.

rushtong: See https://github.com/broadinstitute/consent/pull/247

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchOntologyDAO.scala
```diff
@@ -0,0 +1,82 @@
...
+      } else {
+        term
+      }
+      List(annotatedTerm)
+    }
+  }
+
+  override def autocomplete(term: String): List[TermResource] = {
+    val prefix = term.toLowerCase
+    // user's term must be a prefix in either label or synonyms
```
rushtong, line 50: I think this is a little stricter than what ontology is doing, but I don't have any objections to that.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchOntologyDAO.scala
```diff
@@ -0,0 +1,82 @@
...
+      .setQuery(query)
+        .setSize(20)
+        .setFetchSource(List("id","ontology","usable","label","synonyms","definition").toArray, null)
+
+    val autocompleteResults = executeESRequest[SearchRequest, SearchResponse, SearchRequestBuilder](searchRequest)
+
+    val allHits = autocompleteResults.getHits.getHits
+    allHits.map { hit =>
+      hit.getSourceAsString.parseJson.convertTo[TermResource]
+    }.toList
```
rushtong: More scala-ish? I don't have a strong opinion either way.
`allHits.map(_.getSourceAsString.parseJson.convertTo[TermResource]).toList`

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpOntologyDAO.scala
```diff
@@ -35,9 +35,12 @@ class HttpOntologyDAO(implicit val system: ActorSystem, implicit val executionCo
...
 
-  override def search(term: String): Future[Option[List[TermResource]]] = {
-    searchAsync(term)
-  }
+  override def search(term: String): List[TermResource] = List.empty[TermResource]
+  override def autocomplete(term: String) = List.empty[TermResource]
+
+//  override def search(term: String): Future[Option[List[TermResource]]] = {
+//    searchAsync(term)
+//  }
```
rushtong: Remove?

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/LibraryApiService.scala
```diff
@@ -25,15 +25,17 @@ trait LibraryApiService extends HttpService with FireCloudRequestBuilding
   lazy val rawlsCuratorUrl = FireCloudConfig.Rawls.authUrl + "/user/role/curator"
 
   val libraryServiceConstructor: UserInfo => LibraryService
+  val ontologyServiceConstructor: () => OntologyService
 
   val duosAutocompleteUrl = FireCloudConfig.Duos.baseOntologyUrl + "/autocomplete"
```
rushtong: Pretty sure this can go away.

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpOntologyDAOConcurrencySpec.scala
```diff
@@ -42,31 +42,31 @@ class HttpOntologyDAOConcurrencySpec extends FreeSpec with Matchers with BeforeA
...
+//        assertResult(1) { sr.size }
+//        assertResult("disease") { sr.head.label }
+//        val responseCount = sr.head.id.toInt
+//        if (responseCount > maxCount) maxCount = responseCount
+//        if (responseCount < minCount) minCount = responseCount
+//      }
+//
+//      assert(maxCount <= throttleCount, s"request count should always be under $throttleCount; found $maxCount")
+//      assertResult(throttleCount, s"request count should have reached its maximum throttle limit") { maxCount }
+//      assert(0 < minCount, s"request count should always be > 0; found $minCount")
```
rushtong, line 69: Do we need this?

rushtong (changes requested):
Code looks good, but I think we need to work on the query format. During functional testing, searching on `cancer` doesn't always return `DOID_162`. It usually does, but roughly 1 out of 5 queries don't have it in the result list.

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpOntologyDAOConcurrencySpec.scala
davidangb: We don't - I left all this in the code in case we need to revert. Do you think I should just remove it all?

davidangb:
@rushtong back to you for re-review. I applied your suggested syntax tweaks, and changed the matching/scoring on the query itself - hopefully it works a bit better now.

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpOntologyDAOConcurrencySpec.scala
rushtong: It's so much fun to delete code ... and we have it in history :-)

rushtong (approved):
👍 Much more consistent results!

PR #545, opened 2017-11-21
GAWB-2863: better generation of per-field population suggestions

davidangb:
I was unhappy with the end-user functionality of my pr #537 ... that other PR could easily result in showing only a single suggestion to the end user, even when multiple suggestions exist. I totally rewrote the elasticsearch query to make the functionality better; here's my new PR.

Old behavior in #537: query ES for any document that matches the user's term (as a prefix), return the top 10 values, then `distinct` them in Scala. If the top 10 ES values all have the same value, this means that the result of `distinct` would be a single value.

New behavior in this PR: in the ES query, use a matchPhrasePrefix to filter to the set of documents that match the user's term (as a prefix). Then, still inside the ES query, use a terms aggregation to return the top 10 unique values. No post-processing in Scala.

*important*: this PR requires a full library reindex to work.

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
@@ -9,17 +9,16 @@ import org.elasticsearch.client.transport.TransportClient
 import org.elasticsearch.action.search.{SearchRequest, SearchRequestBuilder, SearchResponse}
 import org.elasticsearch.index.query.{BoolQueryBuilder, QueryBuilder}
 import org.elasticsearch.index.query.QueryBuilders._
-import org.elasticsearch.search.aggregations.bucket.terms.Terms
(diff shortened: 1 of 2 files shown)


kyuksel (approved):
Verified that the tests pass and the functionality works as desired on local UI. Might be good to get someone familiar with ES to review, too.

aednichols:
When I type in a DUL of "a", I don't get "a" as a suggestion. Are we filtering out single-letter suggestions? (I don't see anything in the code that would suggest this.)

screen shot 2017-11-22 at 10 14 52 am

I see that we do still show suggestions that are an exact match for what's already typed, so that's not it.

screen shot 2017-11-22 at 10 18 03 am

davidangb:
We weren't matching "a" because "a" is a stop word, and therefore excluded from consideration in searches. I've updated the filtering logic to be (match-phrase-prefix OR prefix) ... the `match-phrase-prefix` query honors stop words and does other token analysis; the `prefix` query matches purely on prefix and thus will match "a". We can't use `prefix` alone because it misses a lot of stuff that `match-phrase-prefix` will catch. @aednichols back to you.

aednichols (approved):
Update makes sense, thanks 👍

PR #549, opened 2017-11-27
GAWB-2858: health monitor

davidangb:
Remove orchestration's bespoke health-check code, and adopt the HealthMonitor from workbench-libs. The main reason for this is to gain the HealthMonitor's execution-context isolation and its timeouts. Previously, the orchestration health checks operated in the same execution context as other runtime requests. Any failures in subsystems could - and did - negatively affect runtime performance.

There's a lot of lines of code that have simple changes due to moving from `o.b.d.firecloud.model.SubsystemStatus` to `o.b.d.workbench.util.health.SubsystemStatus`, especially because the latter has a different constructor signature.

Also resolves _GAWB-2513: error not handled correctly for failed google bucket access on status endpoint_.

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
@@ -277,10 +277,13 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
 
   def status: Future[SubsystemStatus] = {
     val storage = new Storage.Builder(httpTransport, jsonFactory, getBucketServiceAccountCredential).setApplicationName("firecloud").build()
-    val bucketResponse = storage.buckets().list(FireCloudConfig.FireCloud.serviceProject).executeUsingHead()
-    bucketResponse.getStatusCode match {
-      case x if x == 200 => Future(SubsystemStatus(ok = true, messages = None))
-      case _ => Future(SubsystemStatus(ok = false, messages = Some(List(bucketResponse.parseAsString()))))
+    val bucketResponseTry = Try(storage.buckets().list(FireCloudConfig.FireCloud.serviceProject).executeUsingHead())
+    bucketResponseTry match {
+      case scala.util.Success(bucketResponse) => bucketResponse.getStatusCode match {
+        case x if x == 200 => Future(SubsystemStatus(ok = true, messages = None))
+        case _ => Future(SubsystemStatus(ok = false, messages = Some(List(bucketResponse.parseAsString()))))
+      }
+      case Failure(ex) => Future(SubsystemStatus(ok = false, messages = Some(List(ex.getMessage))))
     }
   }
 
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
@@ -10,9 +10,10 @@ import org.broadinstitute.dsde.firecloud.model.ErrorReportExtensions.FCErrorRepo
 import org.broadinstitute.dsde.firecloud.{FireCloudConfig, FireCloudExceptionWithErrorReport}
 import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol.impGoogleObjectMetadata
 import org.broadinstitute.dsde.rawls.model.{ErrorReport, ErrorReportSource}
-import org.broadinstitute.dsde.firecloud.model.{OAuthUser, ObjectMetadata, SubsystemStatus}
+import org.broadinstitute.dsde.firecloud.model.{OAuthUser, ObjectMetadata}
 import org.broadinstitute.dsde.firecloud.service.FireCloudRequestBuilding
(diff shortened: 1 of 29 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -276,10 +277,13 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
...
-    bucketResponse.getStatusCode match {
-      case x if x == 200 => Future(SubsystemStatus(ok = true, messages = None))
-      case _ => Future(SubsystemStatus(ok = false, messages = Some(List(bucketResponse.parseAsString()))))
+    val bucketResponseTry = Try(storage.buckets().list(FireCloudConfig.FireCloud.serviceProject).executeUsingHead())
+    bucketResponseTry match {
+      case scala.util.Success(bucketResponse) => bucketResponse.getStatusCode match {
+        case x if x == 200 => Future(SubsystemStatus(ok = true, messages = None))
+        case _ => Future(SubsystemStatus(ok = false, messages = Some(List(bucketResponse.parseAsString()))))
+      }
+      case Failure(ex) => Future(SubsystemStatus(ok = false, messages = Some(List(ex.getMessage))))
```
davidangb, line 286: The inclusion of `Try` here is what addresses GAWB-2513.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpSamDAO.scala
```diff
@@ -27,7 +28,7 @@ class HttpSamDAO( implicit val system: ActorSystem, implicit val executionContex
     val pipeline = sendReceive
     pipeline(Get(samStatusUrl)) map { response =>
       val ok = response.status.isSuccess
-      SubsystemStatus(ok, if (ok) Option(List(response.entity.asString)) else None)
+      SubsystemStatus(ok, if (!ok) Option(List(response.entity.asString)) else None)
```
davidangb, line 31: I caught this during development: previously, if Sam returned an OK status, we'd display messages, but if it returned not-OK, we'd suppress them. That's backwards.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/StatusApiService.scala
```diff
@@ -14,8 +14,9 @@ trait StatusApiService extends HttpService with PerRequestCreator with FireCloud
 
   val statusRoutes: Route = {
     path("status") {
-      requestContext =>
+      get { requestContext =>
         perRequest(requestContext, StatusService.props(statusServiceConstructor), StatusService.CollectStatusInfo)
+      }
```
davidangb, line 19: our swagger documents GET; I made the routing strict to only support GET.

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockAgoraDAO.scala
```diff
@@ -29,7 +30,7 @@ class MockAgoraDAO extends AgoraDAO {
   }
 
   def status: Future[SubsystemStatus] = {
-    Future(throw new Exception("Agora Mock DAO exception"))
+    Future(SubsystemStatus(ok = true, None))
```
davidangb, line 33: to make unit tests happy.

src/main/scala/org/broadinstitute/dsde/firecloud/service/StatusService.scala
```diff
@@ -1,63 +1,45 @@
...
+    new StatusService(healthMonitor)
   }
 
   case class CollectStatusInfo()
 }
 
-class StatusService (val app: Application)
+class StatusService (val healthMonitor: ActorRef)
                     (implicit protected val executionContext: ExecutionContext) extends Actor with SprayJsonSupport {
+  implicit val timeout = Timeout(1.minute)
```
aednichols: Is this a timeout on getting a response from the `healthMonitor`, or does it apply to how long the health monitor waits for HTTP requests? Would be a nice comment to have.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpSamDAO.scala
aednichols: Nice catch!

aednichols (approved):
Looks good to me, just some questions/observations.

Tested by running Orchestration against local Agora, then shutting down local Agora.

As expected, I get (excerpt)
```
"Agora": {
  "ok": false,
  "messages": [
    "Connection attempt to local.broadinstitute.org:30443 failed"
  ]
}
```
and the response is instant, which is very nice.

It does take a few seconds for the initial status to be populated; I see
```"ok": false,
"systems": {
  "Thurloe": {
    "ok": false, 
    "messages": [
      "Unknown status"
    ]
  } // etc.
```
for a short while after starting Orch. I assume this is the initial delay of `10.seconds` - is there a specific reason not to have it be zero? If the system takes a few seconds to "warm up" before returning true, I would think it should still report a usable `"ok": false` status.

davidangb:
about the 10 second delay ... 10 seconds is probably overkill. There are a couple things that happen at startup, like the elasticsearch client connecting to the ES cluster, that made me afraid of a zero delay. I changed it to 3 seconds ... how's that?

PR #550, opened 2017-11-28
GAWB-2912: free trial project-pool dao

davidangb:
A dao to manage a pool of free-tier billing projects. Backed by elasticsearch.

Not used anywhere yet - this is prospective coding, trying to get infrastructure in place for when we do start building features.

Will require fc-develop PRs:
broadinstitute/firecloud-develop#843
broadinstitute/firecloud-develop#844
broadinstitute/firecloud-develop#845

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
@@ -1,9 +1,10 @@
 package org.broadinstitute.dsde.firecloud.dataaccess
 import org.broadinstitute.dsde.firecloud.FireCloudException
 import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol.impTrialProject
-import org.broadinstitute.dsde.firecloud.model.{SubsystemStatus, WorkbenchUserInfo}
+import org.broadinstitute.dsde.firecloud.model.WorkbenchUserInfo
 import org.broadinstitute.dsde.firecloud.model.Trial.TrialProject
 import org.broadinstitute.dsde.rawls.model.RawlsBillingProjectName
+import org.broadinstitute.dsde.workbench.util.health.SubsystemStatus
 import org.elasticsearch.action.admin.indices.exists.indices.{IndicesExistsRequest, IndicesExistsRequestBuilder, IndicesExistsResponse}
 import org.elasticsearch.action.get.{GetRequest, GetRequestBuilder, GetResponse}
 import org.elasticsearch.action.index.{IndexRequest, IndexRequestBuilder, IndexResponse}
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
@@ -1,0 +1,170 @@
+package org.broadinstitute.dsde.firecloud.dataaccess
+import org.broadinstitute.dsde.firecloud.FireCloudException
+import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol.impTrialProject
+import org.broadinstitute.dsde.firecloud.model.{SubsystemStatus, WorkbenchUserInfo}
+import org.broadinstitute.dsde.firecloud.model.Trial.TrialProject
+import org.broadinstitute.dsde.rawls.model.RawlsBillingProjectName
+import org.elasticsearch.action.admin.indices.exists.indices.{IndicesExistsRequest, IndicesExistsRequestBuilder, IndicesExistsResponse}
+import org.elasticsearch.action.index.{IndexRequest, IndexRequestBuilder, IndexResponse}
+import org.elasticsearch.action.search.{SearchRequest, SearchRequestBuilder, SearchResponse}
+import org.elasticsearch.action.support.WriteRequest.RefreshPolicy
+import org.elasticsearch.action.update.{UpdateRequest, UpdateRequestBuilder, UpdateResponse}
+import org.elasticsearch.client.transport.TransportClient
+import org.elasticsearch.common.xcontent.XContentType
+import org.elasticsearch.index.query.QueryBuilders._
+import org.elasticsearch.search.sort.SortOrder
+import spray.json._
+
+import scala.concurrent.Future
+import scala.concurrent.ExecutionContext.Implicits.global
+
+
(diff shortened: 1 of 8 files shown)


davidangb:
@andy7i do you care to review test coverage for stories like this?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/TrialDAO.scala
```diff
@@ -0,0 +1,66 @@
...
+  def getProject(projectName: RawlsBillingProjectName): TrialProject
+
+  /**
+    * Create a record for the specified project. Throws error if name
+    * already exists or could not be otherwise created.
+    *
+    * @param projectName name of the project to use when creating a record
+    * @return the created project record
+    */
+  def createProject(projectName: RawlsBillingProjectName): TrialProject
```
aednichols: I struggled briefly to understand that this class is only recording the results of operations that happen elsewhere (d'oh).

Totally optional, but I think we might have better luck with e.g. `recordProjectCreation` and`setProjectVerifiedStatus`

src/main/scala/org/broadinstitute/dsde/firecloud/model/Trial.scala
```diff
@@ -98,4 +99,9 @@ object Trial {
 
   }
 
+  case class TrialProject(name: RawlsBillingProjectName, verified: Boolean, user: Option[WorkbenchUserInfo])
```
aednichols, line 102: We only set `verified` in one place (`verifyProject`). Would it make sense to make `verified` an `Option[Boolean]` to better distinguish the cases "it is not verified" and "we have not checked yet"?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
```diff
@@ -0,0 +1,208 @@
...
+      .setSize(1)
+      .setVersion(true)
+
+    val nextProjectResponse = executeESRequest[SearchRequest, SearchResponse, SearchRequestBuilder](nextProjectRequest)
+
+    if (nextProjectResponse.getHits.totalHits == 0)
+      throw new FireCloudException("no available projects")
+
+    val hit = nextProjectResponse.getHits.getAt(0)
+    val version = hit.getVersion
```
aednichols, line 144: How does `version` work? I see that we get it and set it in a few places, but how is it incremented?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
```diff
@@ -0,0 +1,208 @@
...
+      .prepareSearch(indexName)
+      .setQuery(nextProjectQuery)
+      .addSort("name.keyword", SortOrder.ASC)
+      .setSize(1)
+      .setVersion(true)
+
+    val nextProjectResponse = executeESRequest[SearchRequest, SearchResponse, SearchRequestBuilder](nextProjectRequest)
+
+    if (nextProjectResponse.getHits.totalHits == 0)
+      throw new FireCloudException("no available projects")
```
aednichols, line 141: I'm guessing it's left to the calling code (that doesn't exist yet) to catch this exception, create a new BP, and immediately claim it? Would it make sense to have a companion method `claimProject(project: TrialProject, userInfo: WorkbenchUserInfo)`?

Maybe we distrust BP creation enough that it has to be done in a `while` loop, and a "second chance" codepath isn't good enough.

src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/ElasticSearchTrialDAOSpec.scala
```diff
@@ -0,0 +1,138 @@
...
+    }
+    logger.info("... fixtures indexed.")
+  }
+
+  override def afterAll = {
+    // using the delete from search dao, because we don't have recreate in trial dao.
+    searchDAO.deleteIndex()
+  }
+
+  "ElasticSearchTrialDAO" - {
```
aednichols, line 36: It would be nice to have some concurrency scenarios. On the other hand, I have no idea how to set that up, so I might be sending you on an impossible quest

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
davidangb: correct, I was leaving it up to calling code to catch & handle any exceptions. Since this is prospective development and we don't know exactly what the calling code will need, I didn't build out too much ... I could certainly see changing/adding methods to the dao once we really understand the use cases.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
davidangb: `version` is an Elasticsearch feature that ensures nobody else has updated the document between the get and the update.

See https://www.elastic.co/guide/en/elasticsearch/reference/5.4/docs-update.html and search that page for "version & version_type", at the bottom of the page.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/TrialDAO.scala
davidangb: good ideas (I hate naming)

src/main/scala/org/broadinstitute/dsde/firecloud/model/Trial.scala
davidangb: I was hoping to avoid the `Option` and its corresponding requirements for `.get` and `.nonEmpty` etc ... the `TrialProject.apply` method defaults this to false.

But you're right - if we find we truly need to differentiate between "not verified" and "not checked" then we'd need to do something different. I might go with an enum in that case, with a few values like "unchecked", "creating", "verified", "failed", etc.

src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/ElasticSearchTrialDAOSpec.scala
davidangb: That was my thought process too. I bet I could figure out a way to do it, though.

davidangb:
jenkins retest

aednichols (approved):

PR #554, opened 2017-12-01
GAWB-2808: Enable free trial users

kyuksel:
Opening to get feedback on the missing pieces...

In all cases:


src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -30,26 +35,130 @@ object TrialService {
...
+    // TODO: Handle multiple users
+    require(users.size == 1, "For the time being, we can only enable one user at a time.")
+
+    val user = users.head
+
+    // Define what to overwrite in user's status
+    val statusTransition: UserTrialStatus => UserTrialStatus =
+      currentStatus =>
+        currentStatus.copy(
+          userId = managerInfo.id,
```
davidangb: why does the `userId` change (both here and in terminate & disable)? In the `UserTrialStatus`, the userid should always be the id of the free trial user, not the campaign manager making the call.

src/main/scala/org/broadinstitute/dsde/firecloud/model/Trial.scala
```diff
@@ -28,26 +53,37 @@ object Trial {
...
   // "profile" object to hold all free trial-related information for a single user
   case class UserTrialStatus(
-    userId: String,
-    currentState: Option[TrialState],
-    enabledDate: Instant,    // timestamp a campaign manager granted the user trial permissions
-    enrolledDate: Instant,   // timestamp user started their trial
-    terminatedDate: Instant, // timestamp user was actually terminated
-    expirationDate: Instant  // timestamp user is due to face termination
+      userId: String,
+      state: Option[TrialState],
```
davidangb, line 82: it would be nice - but not a blocker for this PR - that if renaming `currentState` to `state`, you also rename it in the argument to the `apply` methods below, and in the KVP names used by Thurloe.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -30,26 +35,130 @@ object TrialService {
...
+      case StatusUpdate.Success => RequestComplete(NoContent)
+      case StatusUpdate.Failure => RequestComplete(InternalServerError)
+    }
+  }
+
+  private def updateTrialStatus(managerInfo: UserInfo,
+                                userInfo: WorkbenchUserInfo,
+                                statusTransition: UserTrialStatus => UserTrialStatus): Future[StatusUpdate.Attempt] = {
+    // Create hybrid UserInfo with the managerInfo's credentials and users' subjectIds
+    val sudoUserInfo = managerInfo.copy(id = userInfo.userSubjectId)
```
davidangb, line 117: I think it would be cleaner to change the signature of the `getTrialStatus` and `saveTrialStatus` in `*ThurloeDAO` so that we have cleanly separated arguments for the caller (campaign manager) and target (free trial user), and thus avoid creating this hybrid.

I know this was my code in the first place - very much a hack.

Again, not a blocker for this PR.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -30,26 +35,130 @@ object TrialService {
...
+          require(newState.nonEmpty, "Cannot transition to an unspecified state")
+
+          // TODO Handle invalid initial trial status by
+          //    -adding another case to Trial.StatusUpdate
+          //    -using that case to return a BadRequest from the parent method (i.e. updateUserState())
+          assert(newState.get.isAllowedFrom(currentState),
+            s"Cannot transition from $currentState to $newState")
+
+          // Save updates to user's trial status
+          thurloeDao.saveTrialStatus(sudoUserInfo, newStatus)
```
davidangb: map this `Future` and inspect its result before declaring success

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -30,26 +35,130 @@ object TrialService {
...
+          logger.warn(s"Updated profile saved as $newState; we are done!")
+
+          StatusUpdate.Success
+        }
+      // TODO: Handle case where TrialStatusOpt is None, i.e., if the user profile doesn't exist
+      // in Thurloe, create one and set to the new state along with the other tags
```
davidangb: I don't think we should automatically create keys in Thurloe in this instance. Given the leniency of the conversion from `ProfileWrapper` to `UserTrialStatus`, I believe you should only get a `None` here if something went wrong. Also, based on the business logic, the only time we should be creating a trial status is during an enable operation; not disable, terminate, or enroll.

davidangb (commented):
See comments inline - I tried to be clear about when my comments should not hold the PR.

As far as feedback on missing pieces: I'd love to see more tests and more graceful error-handling; you've called out a bunch of TODOs around those error cases already.

I think we need to support changing state for multiple users at once, but - as you said in email - we can handle that in scripts if need be ... we'd just have to write those scripts.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
rushtong: 👍  to David's comment, the hybrid user is very confusing, especially that it requires one to follow this logic backwards to the callers generating a `UserTrialStatus`

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -30,26 +35,130 @@ object TrialService {
...
+        )
+
+    updateUserState(managerInfo, user, statusTransition)
+  }
+
+  private def updateUserState(managerInfo: UserInfo,
+                              user: String,
+                              statusTransition: UserTrialStatus => UserTrialStatus): Future[PerRequestMessage] = {
+    // TODO: Handle unregistered users which get 404 from Sam causing adminGetUserByEmail to throw
+    // TODO: Handle errors that may come up while querying Sam
```
rushtong: You could do something like this to handle all SAM-related errors, and add more specific error messages (i.e. 404's):
```
    samDao.adminGetUserByEmail(RawlsUserEmail(user)).flatMap { regInfo =>
      val workbenchUserInfo = WorkbenchUserInfo(regInfo.userInfo.userSubjectId, regInfo.userInfo.userEmail)
      updateTrialStatus(managerInfo, workbenchUserInfo, statusTransition).map {
        case StatusUpdate.Success => RequestComplete(NoContent)
        case StatusUpdate.Failure => RequestComplete(InternalServerError)
      }
    }.recover {
      case f: FireCloudExceptionWithErrorReport if f.errorReport.statusCode.isDefined =>
        RequestComplete(f.errorReport.statusCode.get)
      case t: Throwable =>
        throw new FireCloudExceptionWithErrorReport(ErrorReport.apply(StatusCodes.InternalServerError, t))
    }
```

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiService.scala
```diff
@@ -0,0 +1,41 @@
...
+
+trait TrialApiService extends HttpService with PerRequestCreator with FireCloudDirectives
+  with StandardUserInfoDirectives with SprayJsonSupport {
+
+  private implicit val executionContext = actorRefFactory.dispatcher
+  val trialServiceConstructor: () => TrialService
+
+  val trialApiServiceRoutes: Route = {
+    post {
+      path("trial" / "manager" / "enable|disable|terminate".r) { (operation: String) =>
```
rushtong: Very minor, probably out of scope for this PR, and likely defined in a spec doc elsewhere, but I am curious why we aren't using HTTP verbs here, specifically `DELETE` for `terminate`. Arguably, we could be using `PUT` for disable.

src/main/scala/org/broadinstitute/dsde/firecloud/model/Trial.scala
kyuksel: Thanks for the catch! I went ahead with the renamings thinking it could create an unnecessary burden of merge conflicts for the subsequent PRs.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
kyuksel: Just to confirm: We want to return error if `thurloeDao.getTrialStatus()` returns `None` because we require users to have registered in FC, which implies `thurloeDao.getTrialStatus()` shouldn't have returned `None`, correct?

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiService.scala
kyuksel: That may be a good idea. Let me look into that as a potential follow-on PR.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -20,12 +20,13 @@ import spray.json.DefaultJsonProtocol._
...
+// TODO: Contain userEmail in value class for stronger type safety without incurring performance penalty
 object TrialService {
   sealed trait TrialServiceMessage
-  case class EnableUsers(userInfo:UserInfo, users: Seq[String]) extends TrialServiceMessage
-  case class DisableUsers(userInfo:UserInfo, users: Seq[String]) extends TrialServiceMessage
-  case class EnrollUser(userInfo:UserInfo) extends TrialServiceMessage
-  case class TerminateUsers(userInfo:UserInfo, users: Seq[String]) extends TrialServiceMessage
+  case class EnableUsers(managerInfo: UserInfo, userEmails: Seq[String]) extends TrialServiceMessage
+  case class DisableUsers(managerInfo: UserInfo, userEmails: Seq[String]) extends TrialServiceMessage
+  case class EnrollUser(managerInfo: UserInfo) extends TrialServiceMessage
```
davidangb: `enrollUser` is different: end users enroll themselves, so this should be `userInfo`. For enroll, disable, and terminate, it's a manager making the call, so `managerInfo` is correct there.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -42,74 +43,60 @@ final class TrialService
...
 
   override def receive = {
-    case EnableUsers(userInfo, users) => enableUsers(userInfo, users) pipeTo sender
-    case DisableUsers(userInfo, users) => disableUsers(userInfo, users) pipeTo sender
-    case EnrollUser(userInfo) => enrollUser(userInfo) pipeTo sender
-    case TerminateUsers(userInfo, users) => terminateUsers(userInfo, users) pipeTo sender
+    case EnableUsers(managerInfo, userEmails) => enableUsers(managerInfo, userEmails) pipeTo sender
+    case DisableUsers(managerInfo, userEmails) => disableUsers(managerInfo, userEmails) pipeTo sender
+    case EnrollUser(managerInfo) => enrollUser(managerInfo) pipeTo sender
+    case TerminateUsers(managerInfo, userEmails) => terminateUsers(managerInfo, userEmails) pipeTo sender
```
davidangb: This is where to enforce that only campaign managers can execute these actions. This is a must-do before merging this PR, and sorry for missing it in previous review. Mix in `PermissionsSupport`, and then wrap the methods in `asTrialCampaignManager`:

```
case EnableUsers(managerInfo, userEmails) => asTrialCampaignManager{enableUsers(managerInfo, userEmails)} pipeTo sender
case DisableUsers(managerInfo, userEmails) => asTrialCampaignManager{disableUsers(managerInfo, userEmails)} pipeTo sender
case TerminateUsers(managerInfo, userEmails) => asTrialCampaignManager{terminateUsers(managerInfo, userEmails)} pipeTo sender
```
search the codebase for `asAdmin` to see a parallel usage. And note that `enrollUser` should NOT be wrapped in `asTrialCampaignManager`.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
kyuksel: How would you do an inspection given `thurloeDao.saveTrialStatus` returns a `Future` of type `Unit`? We could modify `saveTrialStatus` to return something meaningful. 

Either way, should this hold the PR?

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
kyuksel: I agree. Hoping to address it in a separate PR.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
kyuksel: Thanks for the suggestion! Hoping to do something like that in a follow-on PR.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
davidangb: Bah, under the covers in `HttpThurloeDao` this is actually returning `Future[Try[Unit]]`, and you could be checking the `Try`. Yes, we would have to update the `ThurloeDao` trait to support this. In my opinion this is required for release but I guess not for this PR.

kyuksel:
jenkins retest

davidangb (approved):
:+1: let's get this in!

what I see reserved for follow-on work:
- bunch more test coverage
- separate manager vs. target user in Thurloe calls
- more graceful catching/recovery of possible exceptions, and better feedback to caller
- handling multiple users at once (if we decide to handle multiple outside of orch code, at least update swagger)
- re-review TODOs to see if there's anything else

Hopefully we can break these tasks out into separate PRs and potentially multiple people!

aednichols (approved):

PR #556, opened 2017-12-03
GAWB-2909: bulk creation of projects under a billing acct

davidangb:
See https://docs.google.com/a/broadinstitute.org/document/d/1Yp4ljrYyh7pw6Bcv_zkV3uBUymvF7Rnkx5LAhVS0BBI/edit?usp=sharing for high-ish level documentation of how this works.

This doesn't have tests yet, so it is incomplete. Opening now to get as much feedback as possible. My primary objective was to get projects created in the dev env to unblock other stories; my secondary objective was to productionize this code so it can be run again in prod once we have the appropriate billing account. I accomplished the first, and am now working on the second.

The commit history on this PR is a mess - sorry about that! You'll have to review it as a whole.

Requires
broadinstitute/firecloud-develop#856
broadinstitute/firecloud-develop#857
broadinstitute/firecloud-develop#858

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
@@ -25,22 +25,21 @@ import scala.util.{Failure, Success, Try}
 
 trait TrialQueries {
 
-  val Unverified = termQuery("verified", false)
+  val Unverified:QueryBuilder = termQuery("verified", false)
 
-  val Errored = boolQuery()
+  val Errored:QueryBuilder = boolQuery()
     .must(termQuery("verified", true))
     .must(termQuery("status.keyword", CreationStatuses.Error.toString))
 
-  val Available = boolQuery()
+  val Available:QueryBuilder = boolQuery()
     .must(termQuery("verified", true))
     .must(termQuery("status.keyword", CreationStatuses.Ready.toString))
     .mustNot(existsQuery("user.userSubjectId.keyword"))
 
-  val Claimed = boolQuery()
+  val Claimed:QueryBuilder = boolQuery()
     .must(termQuery("verified", true))
     .must(termQuery("status.keyword", CreationStatuses.Ready.toString))
     .must(existsQuery("user.userSubjectId.keyword"))
-
 }
 
 class ElasticSearchTrialDAO(client: TransportClient, indexName: String, refreshMode: RefreshPolicy = RefreshPolicy.NONE)
@@ -195,7 +194,6 @@ class ElasticSearchTrialDAO(client: TransportClient, indexName: String, refreshM
       Seq.empty[TrialProject]
     else
       unverifiedResponse.getHits.getHits.toSeq map ( _.getSourceAsString.parseJson.convertTo[TrialProject] )
-
   }
 
   /**
@@ -203,17 +201,11 @@ class ElasticSearchTrialDAO(client: TransportClient, indexName: String, refreshM
     * @return counts of project records.
     */
   override def countProjects: Map[String,Long] = {
-    // unverified, errored, available, claimed
-    val unverified = count(Unverified)
-    val errored = count(Errored)
-    val available = count(Available)
-    val claimed = count(Claimed)
-
     Map(
-      "unverified" -> unverified,
-      "errored" -> errored,
-      "available" -> available,
-      "claimed" -> claimed
+      "unverified" -> count(Unverified),
+      "errored" -> count(Errored),
+      "available" -> count(Available),
+      "claimed" -> count(Claimed)
     )
   }
 
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
@@ -136,7 +136,7 @@ class ElasticSearchTrialDAO(client: TransportClient, indexName: String, refreshM
   override def claimProjectRecord(userInfo: WorkbenchUserInfo): TrialProject = {
     val nextProjectQuery = boolQuery()
       .must(termQuery("verified", true))
-      .must(termQuery("status.keyword", CreationStatuses.Ready))
+      .must(termQuery("status.keyword", CreationStatuses.Ready.toString))
       .mustNot(existsQuery("user.userSubjectId.keyword"))
 
     // if we find regular race conditions in which multiple users attempt to claim the "next" project,
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
@@ -72,7 +72,6 @@ class ElasticSearchTrialDAO(client: TransportClient, indexName: String, refreshM
     project
   }
 
-
   /**
     * Check to see if the project record exists.
     *
@@ -82,7 +81,6 @@ class ElasticSearchTrialDAO(client: TransportClient, indexName: String, refreshM
   override def projectRecordExists(projectName: RawlsBillingProjectName): Boolean = {
     Try(getProjectInternal(projectName)).isSuccess
   }
-
 
   /**
     * Create a record for the specified project. Throws error if name
@@ -164,7 +162,6 @@ class ElasticSearchTrialDAO(client: TransportClient, indexName: String, refreshM
     updatedProject
   }
 
-
   /**
     * Returns a list of project records in the pool that are unverified.
     *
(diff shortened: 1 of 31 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/trial/ProjectManager.scala
```diff
@@ -0,0 +1,153 @@
...
+    case Verify(projectName: String) => verify(projectName)
+  }
+
+  private def startCreation(count: Int): CreateProjectsResponse = {
+    getCurrentStatus match {
+      case `idle` =>
+        self ! Create(1, count)
+        CreateProjectsResponse(success = true, count, Some(s"$count projects are queued for creation."))
+      case x =>
+        CreateProjectsResponse(success = false, 0, Some("ProjectManager is already creating projects; don't try to create more!"))
```
davidangb, line 66: this "already creating" protection only works for the current orch instance. Since orch is behind a load balancer, it is possible for an end user to start both orch instances on creation loops ... potentially causing quota problems. Solving that is future work, not attempting to deal with it in this PR.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
```diff
@@ -121,6 +121,28 @@ trait UserApiService extends HttpService with PerRequestCreator with FireCloudRe
       }
     } ~
     pathPrefix("api") {
+      // TODO: move to TrialApiService, once that exists
```
davidangb, line 124: waiting on #554 to land, then will move this route over.

src/main/scala/org/broadinstitute/dsde/firecloud/trial/ProjectNamer.scala
```diff
@@ -0,0 +1,32 @@
...
+  private final val ra:Range = Range(1000, 9999)
+
+  private final val prefix = "FCcredits"
+
+  def randomName:String = {
+    val color = colors.apply(nextInt(colors.size))
+    val element = elements.apply(nextInt(elements.size))
+    val suffix = ra.head + nextInt(ra.end - ra.head)
+    s"$prefix-$element-$color-$suffix".replace(" ", "-").toLowerCase
+  }
```
davidangb, line 30: this naming logic discussed and approved with POs and Comms, via email.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/RawlsDAO.scala
```diff
@@ -90,6 +91,9 @@ trait RawlsDAO extends LazyLogging with ReportsSubsystemStatus {
 
   def deleteWorkspace(workspaceNamespace: String, workspaceName: String)(implicit userToken: WithAccessToken): Future[WorkspaceDeleteResponse]
 
+  def createProject(projectName: String, billingAccount: String)(implicit userToken: WithAccessToken): Future[Boolean]
```
rushtong, line 94: For this story, I think `Future[Boolean]` is good. I can imagine a case where we'd need to return more info about possible failure conditions.

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala
```diff
@@ -261,4 +262,34 @@ object ModelJsonProtocol extends WorkspaceJsonSupport {
     def read(value: JsValue) = format.read(value)
   }
 
+
+
+  // following are horribly copied-and-pasted from rawls core, since they're not available as shared models
```
rushtong, line 267: Can we add a ticket to move this to `rawls-model` ?

src/main/scala/org/broadinstitute/dsde/firecloud/model/Trial.scala
```diff
@@ -99,9 +101,68 @@ object Trial {
...
-  case class TrialProject(name: RawlsBillingProjectName, verified: Boolean, user: Option[WorkbenchUserInfo])
+  case class TrialProject(name: RawlsBillingProjectName, verified: Boolean, user: Option[WorkbenchUserInfo], status: Option[CreationStatus])
   object TrialProject {
-    def apply(name: RawlsBillingProjectName) = new TrialProject(name, verified=false, user=None)
+    def apply(name: RawlsBillingProjectName) = new TrialProject(name, verified=false, user=None, status=None)
   }
 
+  case class CreateProjectsResponse(success: Boolean, count: Int, message: Option[String])
+
+  // following are horribly copied-and-pasted from rawls core, since they're not available as shared models
```
rushtong: Same comment - let's add a ticket for this.
Additionally, would it make sense to keep the rawls case classes together? Not a huge issue.

src/main/scala/org/broadinstitute/dsde/firecloud/trial/ProjectManager.scala
rushtong: Agreed - this is my concern as well as multiple trial managers (if that's a thing) creating projects at the same time.

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockTrialDAO.scala
```diff
@@ -0,0 +1,81 @@
...
+    * @return counts of project records.
+    */
+  override def countProjects: Map[String,Long] = Map.empty[String,Long]
+
+  /**
+    * Returns a list of project records that have associated users.
+    *
+    * @return list of project records that have associated users.
+    */
+override def projectReport: Seq[Trial.TrialProject] = Seq.empty[TrialProject]
```
rushtong: nit - indentation.

rushtong (commented):
Sending current crop of comments your way now and reviewing the new tests. So far, so good. I want to read up some more on maintaining status state in `ProjectManager` to allay my own fears, but it looks pretty good.

rushtong (approved):
👍 New tests look good. No need for a re-review of minor comments. Approved with the understanding that potential conflicts could happen when creating more than one project creation process is kicked off.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
```diff
@@ -121,16 +152,12 @@ class ElasticSearchTrialDAO(client: TransportClient, indexName: String, refreshM
...
     */
   override def claimProjectRecord(userInfo: WorkbenchUserInfo): TrialProject = {
-    val nextProjectQuery = boolQuery()
-      .must(termQuery("verified", true))
-      .mustNot(existsQuery("user.userSubjectId.keyword"))
-
     // if we find regular race conditions in which multiple users attempt to claim the "next" project,
     // we could change this to return N (= ~20) available projects, then choose a random project
     // from that list.
     val nextProjectRequest = client
```
ahaessly: this still would not keep multiple clients from choosing the same project

davidangb: even if multiple users find the same record from the read query, the update query will only succeed for one of them due to the version checks. So this comment could read something like "if we find that we are regularly retrying this method in order to succeed at claiming a record, we could return 20 and pick a random one from that list, which would lessen the problem but not eliminate it entirely".

src/main/scala/org/broadinstitute/dsde/firecloud/trial/ProjectManager.scala
```diff
@@ -0,0 +1,153 @@
...
+              trialDAO.setProjectRecordVerified(rawlsName, verified = true, status = CreationStatuses.Ready)
+            case CreationStatuses.Error =>
+              logger.warn(s"Project <$projectName> errored during creation: ${proj.message}")
+              trialDAO.setProjectRecordVerified(rawlsName, verified = true, status = CreationStatuses.Error)
+          }
+      }
+    } recover {
+      case t:Throwable =>
+        logger.warn(s"Error checking status on <$projectName>: ${t.getMessage}")
+        context.system.scheduler.scheduleOnce(verifyDelay, self, Verify(projectName))
```
ahaessly: should we only retry this a few times?
also, it's useful to see the stack trace. i get that we may not want it just on warn, but maybe add it at a more verbose level for debugging purposes.

src/main/scala/org/broadinstitute/dsde/firecloud/trial/ProjectManager.scala
```diff
@@ -0,0 +1,153 @@
...
+        case scala.util.Success(project) =>
+          logger.debug(s"found unique project name in $currentAttempt attempt(s); recorded in pool.")
+          Some(project.name.value)
+        case scala.util.Failure(f) =>
+          if (currentAttempt > sanityLimit) {
+            logger.error(s"Could not generate a unique project name after $currentAttempt tries: ${f.getMessage}")
+            None
+          } else {
+            verifyUniqueProjectName(ProjectNamer.randomName, currentAttempt+1)
+          }
```
ahaessly, line 150: do we need any sort of delay before we try again?

ahaessly (approved):
just a couple of questions

src/main/scala/org/broadinstitute/dsde/firecloud/trial/ProjectManager.scala
davidangb: good call, I added a fairly lenient limit on retries for this.

src/main/scala/org/broadinstitute/dsde/firecloud/trial/ProjectManager.scala
davidangb: I don't think this needs a retry. But, I changed the default refresh policy for the dao, so that at runtime we ensure updates - from any user - are committed before trying again.

PR #558, opened 2017-12-05
GAWB-2948: move the trial/manager/projects route

davidangb:
now that 2808 and 2909 have both merged, I can move the /api/trial/manager/projects route from UserApiService to TrialApiService. This eliminates a TODO in the code. Also added route tests against the /api/trial/manager/projects route.

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
@@ -46,12 +46,12 @@ object TrialService {
   }
 
   def constructor(app: Application, projectManager: ActorRef)()(implicit executionContext: ExecutionContext) =
-    new TrialService(app.samDAO, app.thurloeDAO, app.rawlsDAO, app.trialDAO, projectManager)
+    new TrialService(app.samDAO, app.thurloeDAO, app.rawlsDAO, app.trialDAO, app.googleServicesDAO, projectManager)
 }
 
 // TODO: Remove loggers used for development purposes or lower their level
 final class TrialService
-  (val samDao: SamDAO, val thurloeDao: ThurloeDAO, val rawlsDAO: RawlsDAO, val trialDAO: TrialDAO, projectManager: ActorRef)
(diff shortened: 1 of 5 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -222,7 +222,7 @@ final class TrialService
 
   private def verifyProjects: Future[PerRequestMessage] = {
 
-    val saToken:WithAccessToken = AccessToken(OAuth2BearerToken(HttpGoogleServicesDAO.getTrialBillingManagerAccessToken))
+    val saToken:WithAccessToken = AccessToken(OAuth2BearerToken(googleDAO.getTrialBillingManagerAccessToken))
```
davidangb, line 225: it was necessary to use the google dao from the constructor instead of hardcoding the HttpGoogleServicesDAO in order for the unit tests to allow mocks.

src/main/scala/org/broadinstitute/dsde/firecloud/trial/ProjectManager.scala
```diff
@@ -60,8 +60,12 @@ class ProjectManager(val rawlsDAO: RawlsDAO, val trialDAO: TrialDAO, val googleD
...
     getCurrentStatus match {
       case `idle` =>
-        self ! Create(1, count)
-        CreateProjectsResponse(success = true, count, Some(s"$count projects are queued for creation."))
+        if (count < 1) {
+          CreateProjectsResponse(success = false, count, Some("You must specify a positive number."))
+        } else {
+          self ! Create(1, count)
+          CreateProjectsResponse(success = true, count, Some(s"$count projects are queued for creation."))
+        }
```
davidangb, line 68: found this condition while writing tests, thought it was worth handling.

rushtong (approved):
👍 I will need the google dao as well, thank you!

kyuksel (approved):
Not directly related to this PR but upon noticing another mutable state in one of the changed files, I created https://broadinstitute.atlassian.net/browse/GAWB-2949 as a tech debt ticket to consider.

PR #560, opened 2017-12-05
GAWB-2946: cleaner Thurloe DAO method signatures

davidangb:
Thurloe DAO's `getTrialStatus` and `saveTrialStatus` now accept separate arguments for the userid on which to operate and the OAuth token of the user making the call.

In the one instance - inside `TrialService.enrollUser` - where the user making the call is the same as the user being queried - I just specify both arguments. This felt safer than having multiple methods with different argument lists - too easy to confuse and get/set the wrong user.

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ThurloeDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ThurloeDAO.scala
@@ -2,7 +2,7 @@ package org.broadinstitute.dsde.firecloud.dataaccess
 
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.firecloud.model.Trial.UserTrialStatus
-import org.broadinstitute.dsde.firecloud.model.{BasicProfile, Profile, UserInfo}
+import org.broadinstitute.dsde.firecloud.model.{BasicProfile, Profile, UserInfo, WithAccessToken}
 import org.broadinstitute.dsde.rawls.model.ErrorReportSource
 
 import scala.concurrent.Future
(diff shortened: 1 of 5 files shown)


src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockThurloeDAO.scala
```diff
@@ -106,7 +106,13 @@ class MockThurloeDAO extends ThurloeDAO {
     val newKVsForUser = (userInfo.id -> (mockKeyValues(userInfo.id) ++ keyValues.map { case (key, value) => FireCloudKeyValue(Option(key), Option(value))}))
     mockKeyValues = mockKeyValues + newKVsForUser
     Future.successful(Success(()))
+  }
+
 
+  override def saveKeyValues(forUserId: String, callerToken: WithAccessToken, keyValues: Map[String, String]): Future[Try[Unit]] = {
+    val newKVsForUser = (forUserId -> (mockKeyValues(forUserId) ++ keyValues.map { case (key, value) => FireCloudKeyValue(Option(key), Option(value))}))
```
kyuksel, line 113: Splitting the line above into multiple lines could make it more self-documenting.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ThurloeDAO.scala
```diff
@@ -22,11 +22,44 @@ trait ThurloeDAO extends LazyLogging with ReportsSubsystemStatus {
...
 
+  /**
+    * Save KVPs for a different user - the KVPs will be saved to the "forUserId" user,
+    * but the call to Thurloe will be authenticated as the "callerToken" user.
+    * @param forUserId the userid of the user for which to save KVPs
+    * @param callerToken auth token of the user making the call
+    * @param keyValues the KVPs to save
+    * @return success/failure of save
+    */
+  def saveKeyValues(forUserId: String, callerToken: WithAccessToken, keyValues: Map[String, String]): Future[Try[Unit]]
```
kyuksel, line 42: It'd be good to type `forUserId` more strongly via something like a value class:

```scala
final case class UserId(value: String) extends AnyVal
```

kyuksel (approved):
Thank you for making the user/manager distinction a lot clearer!

Verified that the basic operations still work on Swagger and the tests pass. 

My comments are suggestions only and they don't need to hold up the merge.

PR #561, opened 2017-12-06
GAWB-2855: Onix Billing Report

rushtong:
## Addresses
https://broadinstitute.atlassian.net/browse/GAWB-2855

## Notes
This PR adds the ability for Orchestration to write/update a spreadsheet file to a user's google drive. The data is canned Free Trial data.

Currently, this assumes the value of the Onix user agreement value, and will need to be updated once #566 lands to take advantage of the new user trial status value. 

For a future story, I'd love for the update endpoint to be smarter. Currently, it replaces the data that it writes to. It would be much nicer if it first checked the existing values, update those in their current rows, and then append any other data that doesn't already exist (see https://broadinstitute.atlassian.net/browse/GAWB-2963).

This PR requires that we enable two APIs (Google Drive and Google Sheets APIs) in our google projects for each environment. I have done this for dev, but it needs to be done for all other environments and we need to make sure this gets onto Devops' radar so their FC project setup documentation can be updated.

---

In all cases:


src/main/resources/swagger/api-docs.yaml
```diff
@@ -1690,6 +1693,58 @@ paths:
...
 
+  /api/trial/manager/report:
+    post:
+      tags:
+        - Trial
+      operationId: createReportTrialProjects
+      summary: Create a Spreadsheet Report of free trial projects
+      responses:
+        200:
+          description: OK; report data spreadsheet is created
```
rushtong, line 1726: I am considering a better response for these endpoints. I think it makes sense to provide a link to the created/updated spreadsheet.

davidangb: if you can return the link, 200 OK is fine; if there's no body to the response, maybe 201 Created or 202 Accepted or maybe even 204 No Content

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -293,6 +295,68 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
...
+    * @param requestContext RequestContext
+    * @param userInfo       UserInfo
+    * @param props          SpreadsheetProperties
+    * @return               JsObject representing the Google Create response
+    *
+    */
+  def createSpreadsheet(requestContext: RequestContext, userInfo: UserInfo, props: SpreadsheetProperties)
+    (implicit actorRefFactory: ActorRefFactory, executionContext: ExecutionContext): Future[JsObject] = {
+
+    getUserProfile(requestContext) map { userResponse =>
```
davidangb: is this call to `getUserProfile` necessary? If the user made it this far, past our auth proxy, can't you assume they have a google account?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -293,6 +295,68 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
...
+          val response = service.spreadsheets().values().update(spreadsheetId, range, content).setValueInputOption("RAW").execute()
+          val spreadsheetJson = response.toString.parseJson.asJsObject
+          logger.debug(s"Spreadsheet Update Response: ${spreadsheetJson.toString}")
+          spreadsheetJson
+        case x =>
+          log.error(s"Failed to save a google sheets service for [$responseStr]")
+          throw new FireCloudException(s"Failed to build a google sheets service for [$responseStr]")
+      }
+    }
+  }
```
davidangb: does this overwrite existing data in the spreadsheet? In other words, if some campaign manager manually adds comments to a cell or formatting or something ... will those be lost?

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialServiceSupport.scala
```diff
@@ -0,0 +1,65 @@
...
+    val dateString = dateFormat.format(new Date())
+    new SpreadsheetProperties().setTitle(s"$title: $dateString")
+  }
+
+  // TODO: Look further into the Onix "User Agreement" field
+  def makeSpreadsheetValues(managerInfo: UserInfo, trialDAO: TrialDAO, thurloeDAO: ThurloeDAO, majorDimension: String, range: String)
+    (implicit executionContext: ExecutionContext): Future[ValueRange] = {
+    import scala.collection.JavaConverters._
+    val claimedProjects: Seq[TrialProject] = trialDAO.projectReport
+    val availableProjects: Seq[TrialProject] = trialDAO.availableProjectReport
```
davidangb: instead of two queries, why not just update `projectReport` to return all projects?

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialServiceSupport.scala
```diff
@@ -0,0 +1,65 @@
...
+    val trialStatuses = Future.sequence(claimedProjects.map { p =>
+      if (p.user.isDefined)
+        thurloeDAO.getTrialStatus(p.user.get.userSubjectId, managerInfo)
+      else
+        Future[Option[UserTrialStatus]](None)
+    })
+    trialStatuses.map { userTrialStatuses =>
+      val headers = List("Project Name", "User Subject Id", "User Email", "Enrollment Date", "User Agreement").map(_.asInstanceOf[AnyRef]).asJava
```
davidangb: I realize this isn't in the requirements, but now that I'm looking at the code: it probably makes sense to add termination date here too.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiService.scala
```diff
@@ -52,6 +54,24 @@ trait TrialApiService extends HttpService with PerRequestCreator with FireCloudD
...
+        }
+      } ~
+      path("report" / Segment) { (spreadsheetId) =>
+        put {
+          requireUserInfo() { userInfo => requestContext =>
+            perRequest(requestContext, TrialService.props(trialServiceConstructor),
+              TrialService.UpdateBillingReport(requestContext, userInfo, spreadsheetId)
+            )
+          }
+        }
```
davidangb, line 75: it looks like this supports the ability to have multiple spreadsheets. Do we really want that? I was picturing a single spreadsheet, and all involved know its location.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
rushtong: We need to build a new credential object to make the spreadsheet API call. To do that, we need the bearer token. I didn't see any other way to get that other than going through the google API for that.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
rushtong: It does overwrite the data, but does not change cell formatting. The way I envision this working would be that orch writes to the same worksheet columns every time, and all charts, additional columns (outside of what we update), etc. would remain in place. I think a best practice would be that the worksheet orch touches is never modified by users and instead, they make their changes in other worksheets.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialServiceSupport.scala
rushtong: I could do that - did not want to muck with existing queries that used `projectReport`

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialServiceSupport.scala
rushtong: Yeah, easy enough to do. I was thinking we would certainly want more info.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
rushtong: Additionally, it might make sense to write a set of instructions in one worksheet, data in another, and let the user do what they want in all of the others.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiService.scala
rushtong: The issue is that we need to write an initial spreadsheet to *someone's* google drive since FC doesn't have one. That person would then have to be the one to manage sharing. Orch would never change that, unless we want to add that kind of functionality.

Side note - I did think that perhaps the initial campaign manager could supply a list of email addresses that we add as shared users - not sure if that's technically possible yet.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
davidangb: But you're building the credential from the `userInfo` that is passed in:
```
val credential = new GoogleCredential().setAccessToken(userInfo.accessToken.token)
```

The `userResponse` is only used to make the `responseStr`, and `responseStr` is only used for logging.

rushtong: That will simplify things, yes, thank you!

kyuksel:
The corresponding JIRA tickets also includes `whether or not the user accepted Onix's user agreement` in the list of things to report. The endpoint to persist that information is not yet in place I believe. Is there a stub for that call in this PR?

kyuksel:
I couldn't hit the endpoints on Swagger due to authorization errors. When I run Orch locally with this branch and hit `Authorize` button on top right of the Swagger page, I don't see the second `Authorize` button on the upcoming dialogue box (as shown below) anymore. Am I missing something?

authorizebuttonmissing

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialServiceSupport.scala
```diff
@@ -0,0 +1,72 @@
...
+        dateFormat.format(Date.from(userTrialStatus.get.terminatedDate))
+      else
+        ""
+      (userSubjectId, userEmail, enrollmentDate, terminatedDate)
+    } else {
+      ("", "", "", "")
+    }
+  }
+
+}
```
kyuksel: `TrailServiceSupport` trait seems to exist for Onix spreadsheet reporting purposes only. Does it make sense to rename it to reflect that?

src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/ElasticSearchTrialDAOSpec.scala
```diff
@@ -120,19 +120,19 @@ class ElasticSearchTrialDAOSpec extends FreeSpec with Matchers with BeforeAndAft
       }
     }
     "projectReport" - {
-      "should return an accurate report with only verified/claimed projects" in {
+      "should return an accurate report with all verified/ready projects" in {
         // unverified/unclaimed projects listed below but commented out for developer clarity
```
kyuksel: I couldn't understand why the lines that are commented out were left over.

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
```diff
@@ -261,6 +261,62 @@ final class TrialApiServiceSpec extends BaseServiceSpec with UserApiService with
     }
   }
 
+  "Campaign manager report endpoints" - {
+
+    "Create Report endpoint" - {
+
+      allHttpMethodsExcept(POST) foreach { method =>
+        s"should reject ${method.toString} method" in {
+          new RequestBuilder(method)("/trial/manager/report") ~> dummyUserIdHeaders("campaignManager") ~> trialApiServiceRoutes ~> check {
```
kyuksel: Instead of the `campaignManager` and `nonCampaignManager` `String` literals, consider using `manager` and `unauthorizedUser` defined in `TrialApiServiceSpec` object so we have one place to modify them as necessary.

src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/ElasticSearchTrialDAOSpec.scala
rushtong: I'll fix that ... it's carry over from before I changed what `projectReport` does. That change was one of David's suggestions.

rushtong:
@kyuksel - With respect to the authentication, it's a swagger UI weirdness. You have to scroll within the list of auth scopes to get to the auth button. Definitely a swagger thing - and not intuitive.

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
rushtong: 👍  Will do.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialServiceSupport.scala
rushtong: The intention is that this supports all of `TrialService`. See `LibraryServiceSupport` and `PermissionsSupport` for how we approach this idea.

src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/ElasticSearchTrialDAOSpec.scala
rushtong: ... re-read your comment. They were left there intentionally by David so other developers would be able to see all of the potential trial projects that could be returned. I will update the comment to be more clear.

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
```diff
@@ -327,6 +383,8 @@ final class TrialApiServiceSpec extends BaseServiceSpec with UserApiService with
     override def isGroupMember(userInfo: UserInfo, groupName: String): Future[Boolean] = {
       userInfo.id match {
         case "failme" => Future.failed(new Exception("intentional exception for unit tests"))
+        case "campaignManager" => Future.successful(true)
```
kyuksel: Note that if you use the `manager` val defined in the companion object, you should be able to remove the case above since `"manager"` is defined as one of the `"trial_managers"` above. In my previous PR, I meant to turn 

```scala 
"trial_managers" -> Seq("manager")
```
into 
```scala
"trial_managers" -> Seq(manager)
```
but forgot. It will be in my next PR and you're welcome to do that here, too.

kyuksel:
> With respect to the authentication, it's a swagger UI weirdness. You have to scroll within the list of auth scopes to get to the auth button. Definitely a swagger thing - and not intuitive.

Doh!

Getting past that, my authorization attempt failed with

screen shot 2017-12-08 at 1 06 57 pm

rushtong:
@kyuksel - Technically, it's not a failure, just an extra step to go through. Click the advanced -> proceed link and it will work. I will work with devops on seeing if we can get Google to verify this when running locally.

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
rushtong: Ugh, this is tougher. Doing that breaks other tests. Looking more closely into that.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialServiceSupport.scala
kyuksel: We also have `Trial.scala` that I thought served the same purpose albeit it's an `object`. Not a big deal, though.

rushtong: The `Trial` class looks like a holder for all of the trial-related case classes/objects - the spreadsheet utilities didn't seem to fit there. If this does become a something more broadly used, it should probably go into a google support class instead of a trial support class.

kyuksel:
Hitting `/api/trial/manager/report` returns something like 

```json
{
  "spreadsheetUrl": "https://docs.google.com/spreadsheets/d/1JOdoq34pQvlnbJa-LpVnvQ"
}
```
and then the parameter to `/api/trial/manager/report/{spreadsheetId}` seems to be the last segment on the above URL, which wasn't obvious to me. I wonder if we can clarify it on the Swagger comments or something.

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
kyuksel: Probably not worth it if it ends up taking too much effort.

rushtong: I did get it working ... it was my own stupidity.

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
rushtong: 👍 Got both in, thank you!

kyuksel (approved):

rushtong:
Moved PR from `develop` to `free_tier` branch

kyuksel:
Re-verified that it still works 👍  

A few suggestions that shouldn't necessarily hold up the merge:

1. Return `404` instead of `500` when a user enters an ID that doesn't correspond to an existing document.

2. In the aforementioned scenario, extract the `Resource Not Found` message from the JSON and return that instead of the entire JSON.

3. On `/api/trial/manager/report`'s Swagger, note what part of the response URL will be the ID for subsequent document update requests.

aednichols:
@rushtong - https://github.com/broadinstitute/firecloud-orchestration/pull/566 is now on dev.

rushtong:
@aednichols - Re-based PR to`develop` and needed to do a minor merge fixup.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1712,6 +1715,67 @@ paths:
...
+  /api/trial/manager/report/{spreadsheetId}:
+    put:
+      tags:
+        - Trial
+      operationId: updateReportTrialProjects
+      summary: Update a Spreadsheet Report of free trial projects
+      parameters:
+        - name: spreadsheetId
+          description: |
+            The spreadsheet id to update. The spreadsheet id can be found in the url of the link.
```
aednichols, line 1752: I am a bit worried about users clobbering existing spreadsheets of theirs... maybe we should say something like
> This endpoint replaces the contents of the specified Google spreadsheet with the free trial report. The spreadsheetId can be found in the URL of the link. For example, in the link "https://docs.google.com/spreadsheets/d/long-random-string/edit", the string "long-random-string" is the "spreadsheetId" value. (You may wish to pre-create a blank spreadsheet to use this feature.)

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialServiceSupport.scala
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  def makeSpreadsheetProperties(title: String): SpreadsheetProperties = {
+    val dateString = dateFormat.format(new Date())
+    new SpreadsheetProperties().setTitle(s"$title: $dateString")
+  }
+
+  def makeSpreadsheetValues(managerInfo: UserInfo, trialDAO: TrialDAO, thurloeDAO: ThurloeDAO, majorDimension: String, range: String)
+    (implicit executionContext: ExecutionContext): Future[ValueRange] = {
+    import scala.collection.JavaConverters._
```
aednichols: This works as expected, but I've heard that we now prefer imports together at the top for style/consistency

aednichols (commented):
Looking good, I have to run to an interview recap so this is my feedback so far:

- This is really cool to use! I tested with the Swagger on one display and the spreadsheet in another, and the real-time update is great.
- Whenever we display `Thu, 1 Jan 1970 00:00:00` in the report, compsci people know that means `0` or `null`, but maybe we should say `(No data)` for the business people?

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialServiceSupport.scala
rushtong: I kept it local so as not to interfere with other scala-y things but will test out moving it.

src/main/resources/swagger/api-docs.yaml
rushtong: Yes, I have that task as a separate story (see https://broadinstitute.atlassian.net/browse/GAWB-2963). The current approach has Moran's blessing, as flawed as it is.

rushtong:
+1 to showing no data for `Instant`-zero cases.

aednichols (approved):

PR #563, opened 2017-12-06
GAWB-2959: startup checks

davidangb:
This adds checks to ensure the free trial service account - and the admin service account - are properly registered in FC. In our shared envs (dev, alpha, staging, prod) I manually registered the free trial SA, so checks should pass; in fiab land, where each env is new, the checks should auto-register the SAs.

This is *not* currently working correctly in my fiab, because orchestration starts before sam, and therefore orch can't register the SAs in sam. If, once my fiab is up and running, I restart orchestration (knowing that sam is up), everything works correctly.

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
@@ -38,6 +38,8 @@ class FireCloudServiceActor extends HttpServiceActor with FireCloudDirectives
   with TrialApiService
 {
 
+  override lazy val log = LoggerFactory.getLogger(getClass)
+
   implicit val system = context.system
 
   trait ActorRefFactoryContext {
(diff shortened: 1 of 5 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
```diff
@@ -36,7 +36,9 @@ class FireCloudServiceActor extends HttpServiceActor with FireCloudDirectives
   with Ga4ghApiService
   with UserApiService
   with TrialApiService
-  {
+{
+
+  override lazy val log = LoggerFactory.getLogger(getClass)
```
davidangb: needed to move this higher, up from line 91

src/main/scala/org/broadinstitute/dsde/firecloud/model/UserInfo.scala
```diff
@@ -29,6 +29,9 @@ object UserInfo {
 }
 
 case class AccessToken(accessToken: OAuth2BearerToken) extends WithAccessToken
+object AccessToken{
+  def apply(tokenStr: String) = new AccessToken(OAuth2BearerToken(tokenStr))
+}
```
davidangb, line 34: not strictly necessary - a convenience constructor.

ahaessly (approved):
👍

PR #565, opened 2017-12-07
GAWB-2911 and GAWB-2945: implement enable, disable. handle mutiple users

ahaessly:

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
@@ -91,7 +91,7 @@ class ElasticSearchTrialDAO(client: TransportClient, indexName: String, refreshM
   }
 
   // update a project, with a check on its Elasticsearch version
-  def updateProjectInternal(updatedProject: TrialProject, version: Long) = {
+  def indexProjectInternal(updatedProject: TrialProject, version: Long) = {
     val update = client
       .prepareIndex(indexName, datatype, updatedProject.name.value)
       .setSource(updatedProject.toJson.compactPrint, XContentType.JSON)
@@ -204,12 +204,16 @@ class ElasticSearchTrialDAO(client: TransportClient, indexName: String, refreshM
     *
     * @param projectName the name of the project to return to the available pool
     */
-  override def releaseProjectRecord(projectName: RawlsBillingProjectName): Unit = {
+  override def releaseProjectRecord(projectName: RawlsBillingProjectName): TrialProject = {
     val (version, project) = getProjectInternal(projectName)
     project.user match {
       // TODO should i take user info and verify that it is assigned to the user that is releasing it?
-      case Some(_) => updateProjectInternal(project.copy(user = None), version)
-      case None => Unit
+      case Some(_) => {
+        val updatedProject = project.copy(user = None)
+        indexProjectInternal(project.copy(user = None), version)
+        updatedProject
+      }
+      case None => project
     }
   }
 
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
@@ -80,7 +80,7 @@ class ElasticSearchTrialDAO(client: TransportClient, indexName: String, refreshM
   }
 
   // update a project, with a check on its Elasticsearch version
-  def updateProjectInternal(updatedProject: TrialProject, version: Long) = {
+  def mergeProjectInternal(updatedProject: TrialProject, version: Long) = {
     val update = client
       .prepareUpdate(indexName, datatype, updatedProject.name.value)
       .setDoc(updatedProject.toJson.compactPrint, XContentType.JSON)
@@ -88,6 +88,17 @@ class ElasticSearchTrialDAO(client: TransportClient, indexName: String, refreshM
       .setRefreshPolicy(refreshMode)
 
     executeESRequest[UpdateRequest, UpdateResponse, UpdateRequestBuilder](update) // will throw error if update fails
+  }
+
+  // update a project, with a check on its Elasticsearch version
+  def updateProjectInternal(updatedProject: TrialProject, version: Long) = {
+    val update = client
+      .prepareIndex(indexName, datatype, updatedProject.name.value)
+      .setSource(updatedProject.toJson.compactPrint, XContentType.JSON)
+      .setVersion(version) // guarantee nobody else has updated in the meantime
+      .setRefreshPolicy(refreshMode)
+
+    executeESRequest[IndexRequest, IndexResponse, IndexRequestBuilder](update) // will throw error if update fails
   }
 
   /**
@@ -146,7 +157,7 @@ class ElasticSearchTrialDAO(client: TransportClient, indexName: String, refreshM
       project // noop
     } else {
       val updatedProject = project.copy(verified = verified, status=Some(status))
-      updateProjectInternal(updatedProject, version) // will throw error if update fails
+      mergeProjectInternal(updatedProject, version) // will throw error if update fails
       updatedProject
     }
   }
@@ -183,7 +194,7 @@ class ElasticSearchTrialDAO(client: TransportClient, indexName: String, refreshM
     assert(project.user.isEmpty)
 
     val updatedProject = project.copy(user = Some(userInfo))
-    updateProjectInternal(updatedProject, version) // will throw error if update fails
+    mergeProjectInternal(updatedProject, version) // will throw error if update fails
     updatedProject
   }
 
@@ -196,8 +207,9 @@ class ElasticSearchTrialDAO(client: TransportClient, indexName: String, refreshM
   override def releaseProjectRecord(projectName: RawlsBillingProjectName): Unit = {
     val (version, project) = getProjectInternal(projectName)
     project.user match {
-      // should i take user info and verify that it is assigned to the user that is releasing it?
+      // TODO should i take user info and verify that it is assigned to the user that is releasing it?
       case Some(_) => updateProjectInternal(project.copy(user = None), version)
+      case None => Unit
     }
   }
 
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
@@ -80,7 +80,7 @@ class ElasticSearchTrialDAO(client: TransportClient, indexName: String, refreshM
   }
 
   // update a project, with a check on its Elasticsearch version
-  def mergeProjectInternal(updatedProject: TrialProject, version: Long) = {
+  def updateProjectInternal(updatedProject: TrialProject, version: Long) = {
     val update = client
       .prepareUpdate(indexName, datatype, updatedProject.name.value)
       .setDoc(updatedProject.toJson.compactPrint, XContentType.JSON)
@@ -157,7 +157,7 @@ class ElasticSearchTrialDAO(client: TransportClient, indexName: String, refreshM
       project // noop
     } else {
       val updatedProject = project.copy(verified = verified, status=Some(status))
-      mergeProjectInternal(updatedProject, version) // will throw error if update fails
+      updateProjectInternal(updatedProject, version) // will throw error if update fails
       updatedProject
     }
   }
@@ -194,7 +194,7 @@ class ElasticSearchTrialDAO(client: TransportClient, indexName: String, refreshM
     assert(project.user.isEmpty)
 
     val updatedProject = project.copy(user = Some(userInfo))
-    mergeProjectInternal(updatedProject, version) // will throw error if update fails
+    updateProjectInternal(updatedProject, version) // will throw error if update fails
     updatedProject
   }
 
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
@@ -188,6 +188,20 @@ class ElasticSearchTrialDAO(client: TransportClient, indexName: String, refreshM
   }
 
   /**
+    * Removes the associated user from the project record, making it available again.
+    * Throws an error if the project record could not be updated.
+    *
+    * @param projectName the name of the project to return to the available pool
+    */
+  override def releaseProjectRecord(projectName: RawlsBillingProjectName): Unit = {
+    val (version, project) = getProjectInternal(projectName)
+    project.user match {
+      // should i take user info and verify that it is assigned to the user that is releasing it?
+      case Some(_) => updateProjectInternal(project.copy(user = None), version)
+    }
+  }
+
+  /**
     * Returns a list of project records in the pool that are unverified.
     *
     * @return list of project records in the pool that are unverified.
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/TrialDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/TrialDAO.scala
@@ -65,8 +65,9 @@ trait TrialDAO extends ReportsSubsystemStatus with ElasticSearchDAOSupport {
   * Throws an error if the project record could not be updated.
   *
   * @param projectName the name of the project to return to the available pool
+  * @return the released project record
   */
-  def releaseProjectRecord(projectName: RawlsBillingProjectName): Unit
+  def releaseProjectRecord(projectName: RawlsBillingProjectName): TrialProject
 
     /**
     * Returns a list of project records in the pool that are unverified.
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/TrialDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/TrialDAO.scala
@@ -61,6 +61,14 @@ trait TrialDAO extends ReportsSubsystemStatus with ElasticSearchDAOSupport {
   def claimProjectRecord(userInfo: WorkbenchUserInfo): TrialProject
 
   /**
+  * Removes the associated user from the project record, making it available again.
+  * Throws an error if the project record could not be updated.
+  *
+  * @param projectName the name of the project to return to the available pool
+  */
+  def releaseProjectRecord(projectName: RawlsBillingProjectName): Unit
+
+    /**
     * Returns a list of project records in the pool that are unverified.
     * @return list of project records in the pool that are unverified.
     */
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
@@ -71,59 +71,96 @@ final class TrialService
     case x => throw new FireCloudException("unrecognized message: " + x.toString)
   }
 
+  private def getUserSubjectId(userEmail: String): Future[String] = {
+    for {
+      regInfo <- samDao.adminGetUserByEmail(RawlsUserEmail(userEmail))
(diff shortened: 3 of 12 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
```diff
@@ -80,7 +80,7 @@ class ElasticSearchTrialDAO(client: TransportClient, indexName: String, refreshM
   }
 
   // update a project, with a check on its Elasticsearch version
-  def updateProjectInternal(updatedProject: TrialProject, version: Long) = {
+  def mergeProjectInternal(updatedProject: TrialProject, version: Long) = {
```
davidangb: in this PR, the merge method performs an update, and the update method performs an index.

Could you change this so that the method performing an update is named update, and the method performing the index is named one of index, insert, or upsert?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/TrialDAO.scala
```diff
@@ -61,6 +61,14 @@ trait TrialDAO extends ReportsSubsystemStatus with ElasticSearchDAOSupport {
   def claimProjectRecord(userInfo: WorkbenchUserInfo): TrialProject
 
   /**
+  * Removes the associated user from the project record, making it available again.
+  * Throws an error if the project record could not be updated.
+  *
+  * @param projectName the name of the project to return to the available pool
+  */
+  def releaseProjectRecord(projectName: RawlsBillingProjectName): Unit
```
davidangb: claim and insert both return a `TrialProject`; I think this should too, for consistency

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockThurloeDAO.scala
```diff
@@ -130,9 +130,18 @@ class MockThurloeDAO extends ThurloeDAO {
 
   def status: Future[SubsystemStatus] = Future(SubsystemStatus(ok = true, None))
 
-  override def getTrialStatus(forUserId: String, callerToken: WithAccessToken) = Future.successful(Some(
-    UserTrialStatus(forUserId, Some(TrialStates.Terminated), 0, 0, 0, 0
-  )))
+  override def getTrialStatus(forUserId: String, callerToken: WithAccessToken) = {
+    println("in get trial status: " + forUserId)
```
davidangb: remove println

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -72,106 +73,137 @@ final class TrialService
...
+                s"No further action will be taken.")
+            Future.successful((userEmail, StatusUpdate.toName(StatusUpdate.NoChangeRequired)))
           }
         }
       }
-      // TODO: Respond with a more user-friendly error
-      case None => {
-        logger.warn(s"Trial status could not be found for ${userInfo.userEmail}")
+    }
+    Future.sequence(results) map { output => println(output); RequestComplete(output.toMap) }
```
davidangb: remove println

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -72,106 +73,137 @@ final class TrialService
...
+  }
```
davidangb: all the passing of anonymous functions - plus the liberal use of `Option` and `Any` - makes this feel pretty incomprehensible to me.

Could you change the signature of `executeStateTransitions` to something like this?
```
 private def executeStateTransitions(managerInfo: UserInfo, userEmails: Seq[String],
                                       statusTransition: UserTrialStatus => UserTrialStatus,
                                       beforeTransitionHandler: (UserInfo, UserTrialStatus, UserTrialStatus) => UserTrialStatus,
                                       afterTransitionHandler: (UserInfo, UserTrialStatus, UserTrialStatus) => UserTrialStatus
                                       ): Future[PerRequestMessage] = {
    val results: Seq[Future[(String, String)]] = userEmails map { userEmail =>
      getUserSubjectId(userEmail) flatMap { subId =>
        val userInfo = WorkbenchUserInfo(subId, userEmail)
        thurloeDao.getTrialStatus(subId, managerInfo) map { currentStatus =>
          val newStatus = statusTransition(currentStatus)
          val (email:String, transitionAttempt:Attempt) = if (requiresStateTransition(currentStatus, newStatus)) {
            for {
              // run custom handler to execute code before performing the transition.
              // before handler can update the state we are trying to transition to.
              statusFromBeforeHandler <- beforeTransitionHandler(managerInfo, currentStatus, newStatus)
              // update thurloe to save the target user's new state
              statusFromTransition <- updateTrialStatus (managerInfo, userInfo, statusFromBeforeHandler)
              // run custom handler to execute code after performing the transition
              // after handler sees the result of what we actually transitioned to
              afterResults <- afterTransitionHandler (managerInfo, currentStatus, statusFromTransition)
            } yield {
              (userEmail, StatusUpdate.Success)
            }
          } else {
            logger.warn(
              s"The user '${userInfo.userEmail}' is already in the trial state of '$newStatus.newState'. " +
                s"No further action will be taken.")
            (userEmail, StatusUpdate.NoChangeRequired)
          }
          (email, StatusUpdate.toName(transitionAttempt))
        }
      }
    }
    Future.sequence(results) map { output => RequestComplete(output.toMap) }
  }
```
if so, the `statusTransition` remains as before this PR, pretty simple; and `beforeTransitionHandler` and `afterTransitionHandler` can each be defined as their own function defs, then passed in. Instead of a single function `enableUsers` which defines anonymous functions inside of it, you could have `enableUsers`, `beforeEnable`, and `afterEnable`.

davidangb (commented):
I am suggesting a somewhat large change to the signature of `executeStateTransitions` and the code that feeds into it ... with that suggestion in mind I haven't reviewed a bunch of this code in detail.

( also this needs swagger updates )

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
```diff
@@ -183,10 +194,25 @@ class ElasticSearchTrialDAO(client: TransportClient, indexName: String, refreshM
...
+  /**
+    * Removes the associated user from the project record, making it available again.
+    * Throws an error if the project record could not be updated.
+    *
+    * @param projectName the name of the project to return to the available pool
+    */
+  override def releaseProjectRecord(projectName: RawlsBillingProjectName): Unit = {
+    val (version, project) = getProjectInternal(projectName)
+    project.user match {
+      // TODO should i take user info and verify that it is assigned to the user that is releasing it?
```
aednichols: It would be nice for this method to remain idempotent, so I would check that it is assigned either to the correct user, or to no user at all

aednichols (commented):
The new API works exactly as I'd hoped, very handy to work with. I like the idempotence and how it gives you a per-user confirmation in the response.

There are two cases I've found that don't work quite as expected:

1. I got at least one `"requirement failed: Cannot transition to an unspecified state"` when attempting to enable a user. Couldn't reproduce it, possibly because someone else was doing something with the same user
2. When enabling an unregistered user, we get an overall `404` response, instead of noting that the user is not registered as a line item in the standard response

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
```diff
@@ -90,6 +90,17 @@ class ElasticSearchTrialDAO(client: TransportClient, indexName: String, refreshM
     executeESRequest[UpdateRequest, UpdateResponse, UpdateRequestBuilder](update) // will throw error if update fails
   }
 
+  // update a project, with a check on its Elasticsearch version
+  def indexProjectInternal(updatedProject: TrialProject, version: Long) = {
```
rushtong, line 94: Nit: I like specified return values, i.e. `IndexResponse` in this case.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
```diff
@@ -187,6 +198,25 @@ class ElasticSearchTrialDAO(client: TransportClient, indexName: String, refreshM
...
+  /**
+    * Removes the associated user from the project record, making it available again.
+    * Throws an error if the project record could not be updated.
+    *
+    * @param projectName the name of the project to return to the available pool
+    */
+  override def releaseProjectRecord(projectName: RawlsBillingProjectName): TrialProject = {
+    val (version, project) = getProjectInternal(projectName)
+    project.user match {
+      // TODO should i take user info and verify that it is assigned to the user that is releasing it?
```
rushtong, line 210: Are we doing this in a future story or this one? Otherwise, remove.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -71,110 +72,128 @@ final class TrialService
...
-    // TODO: Handle multiple users
-    require(userEmails.size == 1, "For the time being, we can only enable one user at a time.")
-
-    val userEmail = userEmails.head
+  private def buildEnableUserStatus(userInfo: WorkbenchUserInfo, currentStatus: Option[UserTrialStatus]): UserTrialStatus = {
+    val needsProject = currentStatus match {
+      case None => true
+      case Some(statusObj) => statusObj.state match {
+        case Some(Disabled) => true
+        case _ => false // either an invalid transition or noop
```
rushtong, line 80: Technically, I don't think this is really a noop since in the false case, the current status is set to `Enabled`.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -71,110 +72,128 @@ final class TrialService
...
 
-  private def disableUsers(managerInfo: UserInfo, userEmails: Seq[String]): Future[PerRequestMessage] = {
-    // TODO: Handle multiple users
-    require(userEmails.size == 1, "For the time being, we can only disable one user at a time.")
+  private def buildDisableUserStatus(userInfo: WorkbenchUserInfo, currentStatus: Option[UserTrialStatus]): UserTrialStatus = {
+    currentStatus.get.copy(state = Some(Disabled), billingProjectName = None)
+  }
 
-    val userEmail = userEmails.head
+  private def disableUserPostProcessing(updateStatus: Attempt, prevStatus: Option[UserTrialStatus], newStatus: UserTrialStatus): Unit = {
```
rushtong, line 107: The `Unit` return type feels wrong to me. It seems like these transition steps should return at least true/false at a minimum to let the caller know that it did something or failed to do something. 

Same feedback for the other transition function `enableUserPostProcessing`.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -71,110 +72,128 @@ final class TrialService
...
-    val statusTransition: UserTrialStatus => UserTrialStatus =
-      status => status.copy(state = Some(Terminated), terminatedDate = Instant.now)
+    executeStateTransitions(managerInfo, userEmails, buildTerminateUserStatus, (_,_,_) => ())
+  }
 
-    updateUserState(managerInfo, userEmail, statusTransition)
+  private def requiresStateTransition(currentState: Option[UserTrialStatus], newState: UserTrialStatus): Boolean = {
+    // TODO: Handle invalid initial trial status by
+    //    -adding another case to Trial.StatusUpdate
+    //    -using that case to return a BadRequest from the parent method (i.e. updateUserState())
```
rushtong, line 132: Is this covered in another story?

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -71,110 +72,128 @@ final class TrialService
...
   }
 
-  private def updateUserState(managerInfo: UserInfo,
-                              userEmail: String,
-                              statusTransition: UserTrialStatus => UserTrialStatus): Future[PerRequestMessage] = {
-    // TODO: Handle unregistered users which get 404 from Sam causing adminGetUserByEmail to throw
-    // TODO: Handle errors that may come up while querying Sam
+  private def getUserSubjectId(userEmail: String): Future[String] = {
+    //    // TODO: Handle unregistered users which get 404 from Sam causing adminGetUserByEmail to throw
+    //    // TODO: Handle errors that may come up while querying Sam
```
rushtong, line 148: Nit: double commented.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -71,110 +72,128 @@ final class TrialService
...
     for {
       regInfo <- samDao.adminGetUserByEmail(RawlsUserEmail(userEmail))
-      userInfo = WorkbenchUserInfo(regInfo.userInfo.userSubjectId, regInfo.userInfo.userEmail)
-      result <- updateTrialStatus(managerInfo, userInfo, statusTransition)
-    } yield result match {
-      case StatusUpdate.Success => RequestComplete(NoContent)
-      case StatusUpdate.Failure => RequestComplete(InternalServerError)
-      case StatusUpdate.ServerError(msg) => RequestComplete(InternalServerError, msg)
-    }
+    } yield regInfo.userInfo.userSubjectId
```
rushtong, line 151: This whole for-comprehension could be a one-liner:
```
samDao.adminGetUserByEmail(RawlsUserEmail(userEmail)).map(_.userInfo.userSubjectId)
```

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -71,110 +72,128 @@ final class TrialService
...
-            case Failure(ex) =>
-              logger.warn(s"We could not have Thurloe update the new state as $newState; aborting...")
-              StatusUpdate.ServerError(ex.getMessage)
+  private def executeStateTransitions(managerInfo: UserInfo, userEmails: Seq[String],
+                                      statusTransition: (WorkbenchUserInfo, Option[UserTrialStatus]) => UserTrialStatus,
+                                      transitionPostProcessing: (Attempt, Option[UserTrialStatus], UserTrialStatus) => Unit): Future[PerRequestMessage] = {
+    var results: Seq[(String,String)] = Seq()
+    userEmails.foreach { userEmail =>
+     val finalStatus =
+      getUserSubjectId(userEmail) flatMap { subId =>
```
rushtong, line 160: Nit: single-line-worthy: `val finalStatus = getUserSubjectId(userEmail) flatMap { subId =>`

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -71,110 +72,128 @@ final class TrialService
...
+        val userInfo = WorkbenchUserInfo(subId, userEmail)
+        thurloeDao.getTrialStatus(subId, managerInfo) flatMap { userTrialStatus =>
+          val newStatus = statusTransition(userInfo, userTrialStatus)
+          if (requiresStateTransition(userTrialStatus, newStatus)) {
+            for {
```
rushtong, line 165: Nit: mixing a single-line for-comprehension seems out of place in the nested flatmaps.
```
            updateTrialStatus(managerInfo, userInfo, newStatus).map { stateResponse =>
              transitionPostProcessing(stateResponse, userTrialStatus, newStatus)
              StatusUpdate.toName(stateResponse)
            }
```

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -71,110 +72,128 @@ final class TrialService
...
+                s"No further action will be taken.")
+            Future.successful(StatusUpdate.toName(StatusUpdate.NoChangeRequired))
           }
         }
       }
-      // TODO: Respond with a more user-friendly error
-      case None => {
-        logger.warn(s"Trial status could not be found for ${userInfo.userEmail}")
+      // Use await here so that multiple Futures don't have a conflict when claiming a billing project
+      val result = (Await.result(finalStatus, scala.concurrent.duration.Duration.Inf))
```
rushtong, line 180: Nit: unnecessary parens around `Await...`

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -71,110 +72,128 @@ final class TrialService
...
+            for {
+              stateResponse <- updateTrialStatus(managerInfo, userInfo, newStatus)
+            } yield {
+              transitionPostProcessing(stateResponse, userTrialStatus, newStatus)
+              StatusUpdate.toName(stateResponse)
+            }
+          } else {
+            logger.warn(
+              s"The user '${userInfo.userEmail}' is already in the trial state of '$newStatus.newState'. " +
+                s"No further action will be taken.")
```
rushtong, line 174: Nit: extraneous `s` - there are no substitutions in the second string.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -71,110 +72,128 @@ final class TrialService
...
+    Future.successful(RequestComplete(sorted))
+  }
 
-        Future(StatusUpdate.Failure)
-      }
+  private def updateTrialStatus(managerInfo: UserInfo,
+                                userInfo: WorkbenchUserInfo,
+                                updatedTrialStatus: UserTrialStatus): Future[StatusUpdate.Attempt] = {
+
+    // TODO: Test the logic below by adding a mock ThurloeDAO whose saveTrialStatus() returns Failure
```
rushtong, line 191: Is this for a future story? There are some failure cases handled by `TrialApiServiceSpecThurloeDAO` so this may be handled.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -71,110 +72,128 @@ final class TrialService
...
+    thurloeDao.saveTrialStatus(userInfo.userSubjectId, managerInfo, updatedTrialStatus) map {
+      case Success(_) =>
+        StatusUpdate.Success
+      case Failure(ex) =>
+        StatusUpdate.ServerError(ex.getMessage)
```
rushtong, line 196: Nit: this line and `StatusUpdate.Success` can be single-lined with their case conditions.

rushtong (commented):
Have not functionally tested this yet, but I wanted to get syntax level feedback out to you.

rushtong:
Here is a failure case that doesn't show what I would expect:
```
[
    "grushton.dev@gmail.com",
    "grushton.dev2@gmail.com",
    "grushton.dev3@gmail.com"
]
```
The first two, when submitted on their own, return a correctly formatted "NoChangeRequired" block. When adding the third, it should show the first two the same way, but the third as an error, but instead the whole operation fails:
```
{
  "statusCode": 404,
  "source": "Sam",
  "timestamp": 1512769304041,
  "causes": [],
  "stackTrace": [],
  "message": ""
}
```

rushtong (approved):
👍 No need for a re-review if you want to address minor nit-picks.
* Enable/Disable works as expected.
* The Sam 404 problem is still a problem, but I see that it is not addressed in this PR.
* Terminate still throws a 500, but that is also not handled in this PR.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
ahaessly: in the false case the new status object is set to enabled. if the current status object is also enabled, then the state transition will be a noop

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
ahaessly: the post processing isn't really doing the state transition just some persistence based on the state transition state. i'm not sure what i would do with an output parameter...

aednichols (approved):
👍  to merge into branch `free_tier` pending resolution of Greg's comments.

PR #569, opened 2017-12-10
GAWB-2964: make sure health monitor repeats its checks

davidangb:
Found while debugging a problem with Thurloe in dev ... during that session, we noticed that orchestration's status never went red, and continued to show Thurloe as up. Then through code inspection, realized that the set of health checks we were passing into the health monitor were resolved the first time they were checked, and then never again.

End result is that orch's status wasn't updating over time.

Tested by running a local orch and a local agora, switched agora on and off and watched orch's status update correctly.

In all cases:


rushtong (approved):
👍 Tested by spinning up orch pointing to a local consent instance. Spinning consent up and down triggered the correct response in orchestration's status endpoint.

PR #571, opened 2017-12-11
fix dependency/assembly problems in build

davidangb:

In all cases:

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -6,6 +6,8 @@ object Dependencies {
   val jacksonV = "2.8.10"
      // note that jackson-databind overrides this below! 2.8.8.1 is not released for core or annotations.
 
+  def excludeGuavaJDK5(m: ModuleID): ModuleID = m.exclude("com.google.guava", "guava-jdk5")
+
   val rootDependencies = Seq(
     // proactively pull in latest versions of Jackson libs, instead of relying on the versions
     // specified as transitive dependencies, due to OWASP DependencyCheck warnings for earlier versions.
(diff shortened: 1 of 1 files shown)


project/Dependencies.scala
```diff
@@ -36,11 +38,13 @@ object Dependencies {
 
     "org.elasticsearch.client"       % "transport"           % "5.4.3",
 
-    "com.google.api-client"          % "google-api-client"   % "1.22.0"
-      exclude("com.google.guava", "guava-jdk5"),
-    "com.google.apis"                % "google-api-services-storage" % "v1-rev58-1.21.0",
-    "com.google.apis"                % "google-api-services-compute" % "v1-rev120-1.22.0",
-    "com.google.apis"                % "google-api-services-sheets"  % "v4-rev493-1.23.0",
+    "com.google.guava"               % "guava"               % "19.0",
```
davidangb, line 41: we could probably get away with a more recent version; this is the same one rawls uses

project/Dependencies.scala
```diff
@@ -36,11 +38,13 @@ object Dependencies {
...
+
+    excludeGuavaJDK5("com.google.api-client"          % "google-api-client"   % "1.23.0"),
+    excludeGuavaJDK5("com.google.apis"                % "google-api-services-storage" % "v1-rev116-1.23.0"),
+    excludeGuavaJDK5("com.google.apis"                % "google-api-services-compute" % "v1-rev163-1.23.0"),
+    excludeGuavaJDK5("com.google.apis"                % "google-api-services-sheets"  % "v4-rev494-1.23.0"),
+    excludeGuavaJDK5("com.google.apis"                % "google-api-services-cloudbilling" % "v1-rev14-1.23.0"),
```
davidangb, line 47: I've updated all the google dependencies so they are all on the 1.23 version; previously we had mixed versions

breilly2:
Nice! I may borrow this as I ran into something similar last week when I needed to update the version of Selenium we use. It would be great if we had a way to share this technique of importing Google libs sans guava-jdk5 in all of our projects.

rushtong (approved):
👍

PR #572, opened 2017-12-12
gawb 2947: update exception handling

ahaessly:
instead of assert and require, return errors
multi-user requests will only return an overall error code if there is an issue with the permissions of the caller. otherwise any errors will be reported within the body of the response.

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
@@ -46,7 +46,7 @@ class HttpRawlsDAO( implicit val system: ActorSystem, implicit val executionCont
     userAuthedRequest(Get(RawlsDAO.groupUrl(groupName)))(userInfo) map { response =>
       response.status match {
         case OK => true
-        case Forbidden => false
+        case NotFound => false
         case _ => throw new FireCloudExceptionWithErrorReport(FCErrorReport(response))
       }
     }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
@@ -46,7 +46,8 @@ class HttpRawlsDAO( implicit val system: ActorSystem, implicit val executionCont
     userAuthedRequest(Get(RawlsDAO.groupUrl(groupName)))(userInfo) map { response =>
       response.status match {
         case OK => true
-        case _ => false
+        case Forbidden => false
+        case _ => throw new FireCloudExceptionWithErrorReport(FCErrorReport(response))
       }
     }
   }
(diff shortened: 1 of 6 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
```diff
@@ -46,7 +46,8 @@ class HttpRawlsDAO( implicit val system: ActorSystem, implicit val executionCont
     userAuthedRequest(Get(RawlsDAO.groupUrl(groupName)))(userInfo) map { response =>
       response.status match {
         case OK => true
-        case _ => false
+        case Forbidden => false
```
aednichols: Per the docs on the Rawls endpoint `GET /api/user/group/{groupName}`, it actually returns `404` instead of `403` for `Group not found or user is not a member`

Might want to double-check that the implementation matches the Swagger

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -135,41 +135,49 @@ final class TrialService
...
+              Future.successful(StatusUpdate.toName(StatusUpdate.NoChangeRequired))
             }
-          } else {
-            logger.warn(s"The user '${userInfo.userEmail}' is already in the trial state of '$newStatus.newState'. No further action will be taken.")
-            Future.successful(StatusUpdate.toName(StatusUpdate.NoChangeRequired))
+          } catch {
+            case ex: FireCloudException =>
+              Future.successful(StatusUpdate.toName(StatusUpdate.Failure(ex.getMessage)))
+            case t: Throwable =>
+              StatusUpdate.ServerError(t.getMessage)
```
davidangb: extraneous line?

davidangb (approved):
:+1: looks good, modulo Adam's comment. I also have one tiny change inline. I did not test.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
ahaessly: yep - thanks

ahaessly:
jenkins retest

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -144,41 +144,48 @@ final class TrialService
...
-    samDao.adminGetUserByEmail(RawlsUserEmail(userEmail)).map(_.userInfo.userSubjectId)
-  }
-
   private def executeStateTransitions(managerInfo: UserInfo, userEmails: Seq[String],
                                       statusTransition: (WorkbenchUserInfo, Option[UserTrialStatus]) => UserTrialStatus,
                                       transitionPostProcessing: (Attempt, Option[UserTrialStatus], UserTrialStatus) => Unit): Future[PerRequestMessage] = {
     var results: Seq[(String, String)] = Seq()
     userEmails.foreach { userEmail =>
-      val finalStatus = getUserSubjectId(userEmail) flatMap { subId =>
+      val finalStatus = samDao.adminGetUserByEmail(RawlsUserEmail(userEmail)) flatMap { regInfo =>
```
kyuksel, line 162: Do we have negative tests for this Sam call? We may get an exception or `404`. Not sure if there are others.

davidangb (approved):
still :+1:

aednichols (approved):
👍

PR #573, opened 2017-12-12
Enhance free tier unit test coverage [GAWB-2930]

aednichols:

In all cases:

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
@@ -109,6 +109,7 @@ final class TrialApiServiceSpec extends BaseServiceSpec with UserApiService with
     "Failing due to Thurloe error should return a server error to the user" in {
       Put(userAgreementPath) ~> dummyUserIdHeaders(dummy1User) ~> userServiceRoutes ~> check {
         status should equal(InternalServerError)
+        assert(localThurloeDao.agreedUsers == Seq())
       }
     }
 
@@ -116,6 +117,7 @@ final class TrialApiServiceSpec extends BaseServiceSpec with UserApiService with
       s"should reject ${method.toString} method" in {
         new RequestBuilder(method)(userAgreementPath) ~> dummyUserIdHeaders(enabledUser) ~> userServiceRoutes ~> check {
           assert(!handled)
+          assert(localThurloeDao.agreedUsers == Seq())
         }
       }
     }
@@ -124,6 +126,7 @@ final class TrialApiServiceSpec extends BaseServiceSpec with UserApiService with
       s"$user should not be able to agree to terms" in {
         Put(userAgreementPath) ~> dummyUserIdHeaders(user) ~> userServiceRoutes ~> check {
           status should equal(Forbidden)
+          assert(localThurloeDao.agreedUsers == Seq())
         }
       }
     }
@@ -131,6 +134,7 @@ final class TrialApiServiceSpec extends BaseServiceSpec with UserApiService with
     "attempting to agree to terms as enabled user" in {
       Put(userAgreementPath) ~> dummyUserIdHeaders(enabledUser) ~> userServiceRoutes ~> check {
         status should equal(NoContent)
+        assert(localThurloeDao.agreedUsers == Seq("enabled-user"))
       }
     }
   }
@@ -418,7 +422,15 @@ final class TrialApiServiceSpec extends BaseServiceSpec with UserApiService with
 
   /** Used by positive and negative tests where `saveTrialStatus` is called */
   final class TrialApiServiceSpecThurloeDAO extends HttpThurloeDAO {
+
+    var agreedUsers = Seq[String]()
+
     override def saveTrialStatus(forUserId: String, callerToken: WithAccessToken, trialStatus: UserTrialStatus) = {
+
+      if (trialStatus.userAgreed) {
+        agreedUsers = agreedUsers :+ forUserId
+      }
+
       // Note: because HttpThurloeDAO catches exceptions, the assertions here will
       // result in InternalServerErrors instead of appearing nicely in unit test output.
       forUserId match {
--- a/src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
@@ -262,11 +262,9 @@ final class TrialApiServiceSpec extends BaseServiceSpec with UserApiService with
           val enableResponse = responseAs[Map[String, Set[String]]].map(_.swap)
 
           assert(enableResponse(Set(enabledUser, dummy2User)) === StatusUpdate.NoChangeRequired.toString)
-          assert(enableResponse(Set(disabledUser)) === StatusUpdate.Success.toString)
+          assert(enableResponse(Set(disabledUser, registeredUser)) === StatusUpdate.Success.toString)
           assert(enableResponse(Set(enrolledUser)).contains("Failure: Cannot transition"))
           assert(enableResponse(Set(terminatedUser)).contains("Failure: Cannot transition"))
-          assert(enableResponse(Set(registeredUser))
-            .contains("ServerError: Should only be updating enabled, disabled or enrolled users"))
           assert(enableResponse(Set(dummy1User))
             .contains("ServerError: ErrorReport(Thurloe,Unable to get user trial status,Some(500 Internal Server Error)"))
 
@@ -537,7 +535,16 @@ final class TrialApiServiceSpec extends BaseServiceSpec with UserApiService with
 
           Future.successful(Success(()))
         }
-        case user @ `dummy2User` => { // Mocking Thurloe status saving failures
+        case `registeredUser` => {
+          assertResult(Some(TrialStates.Enabled)) { trialStatus.state }
+          assert(trialStatus.enabledDate.toEpochMilli > 0)
+          assert(trialStatus.enrolledDate.toEpochMilli === 0)
+          assert(trialStatus.terminatedDate.toEpochMilli === 0)
+          assert(trialStatus.expirationDate.toEpochMilli === 0)
+
+          Future.successful(Success(()))
+        }
+        case `dummy2User` => { // Mocking Thurloe status saving failures
           Future.failed(new FireCloudExceptionWithErrorReport(ErrorReport.apply(StatusCodes.InternalServerError, new FireCloudException(s"Unable to update user profile"))))
         }
         case _ => {
--- a/src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
@@ -10,6 +10,7 @@ import org.broadinstitute.dsde.firecloud.dataaccess.{HttpSamDAO, HttpThurloeDAO,
 import org.broadinstitute.dsde.firecloud.mock.{MockGoogleServicesDAO, MockUtils}
 import org.broadinstitute.dsde.firecloud.mock.MockUtils.thurloeServerPort
 import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol.impProfileWrapper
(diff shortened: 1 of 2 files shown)


src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
```diff
@@ -110,13 +111,15 @@ final class TrialApiServiceSpec extends BaseServiceSpec with UserApiService with
     "Failing due to Thurloe error should return a server error to the user" in {
       Put(userAgreementPath) ~> dummyUserIdHeaders(dummy1User) ~> userServiceRoutes ~> check {
         status should equal(InternalServerError)
+        assert(localThurloeDao.agreedUsers == Seq())
```
davidangb: `assert(localThurloeDao.agreedUsers.isEmpty)`

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
```diff
@@ -247,10 +253,19 @@ final class TrialApiServiceSpec extends BaseServiceSpec with UserApiService with
       }
     }
 
-    "Attempting to enable a previously disabled user should return success" in {
-      Post(enablePath, Seq(disabledUser, enabledUser)) ~> dummyUserIdHeaders(manager) ~> trialApiServiceRoutes ~> check {
-        val enableResponse = responseAs[Map[String, Seq[String]]]
-        assertResult(Map("Success"->Seq(disabledUser), "NoChangeRequired"->Seq(enabledUser))) { enableResponse }
+    "Attempting to enable multiple users in various states should return a map of status lists" in {
+      val assortmentOfUserEmails = Seq(disabledUser, enabledUser, enrolledUser, terminatedUser, dummy1User, dummy2User)
```
davidangb: could you add a user who has no previous state? i.e. a user that returns `TrialApiServiceSpec.noTrialProps`?

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
```diff
@@ -430,7 +445,15 @@ final class TrialApiServiceSpec extends BaseServiceSpec with UserApiService with
 
   /** Used by positive and negative tests where `saveTrialStatus` is called */
   final class TrialApiServiceSpecThurloeDAO extends HttpThurloeDAO {
+
+    var agreedUsers = Seq[String]()
```
davidangb: I prefer `Seq.empty[String]` but it's mostly style and preference

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
```diff
@@ -492,6 +515,9 @@ final class TrialApiServiceSpec extends BaseServiceSpec with UserApiService with
 
   /** Used to ensure that manager endpoints only serve managers */
   final class TrialApiServiceSpecRawlsDAO extends MockRawlsDAO {
+
+    var billingProjectAdds = Map[String, String]()
```
davidangb: same here: `Map.empty[String, String]`

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
```diff
@@ -247,10 +253,19 @@ final class TrialApiServiceSpec extends BaseServiceSpec with UserApiService with
...
+      Post(enablePath, assortmentOfUserEmails) ~> dummyUserIdHeaders(manager) ~> trialApiServiceRoutes ~> check {
+        val enableResponse = responseAs[Map[String, Seq[String]]].map(_.swap)
+
+        assert(enableResponse(List(enabledUser, dummy2User)) === StatusUpdate.NoChangeRequired.toString)
+        assert(enableResponse(List(disabledUser)) === StatusUpdate.Success.toString)
+        assert(enableResponse(List(enrolledUser)).contains("Failure: Cannot transition"))
+        assert(enableResponse(List(terminatedUser)).contains("Failure: Cannot transition"))
+        assert(enableResponse(List(dummy1User))
+          .contains("ServerError: ErrorReport(Thurloe,Unable to get user trial status,Some(500 Internal Server Error)"))
+
```
davidangb: I'd love to see the same type of multi-user test for disable and terminate!

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
```diff
@@ -247,10 +253,19 @@ final class TrialApiServiceSpec extends BaseServiceSpec with UserApiService with
...
-      Post(enablePath, Seq(disabledUser, enabledUser)) ~> dummyUserIdHeaders(manager) ~> trialApiServiceRoutes ~> check {
-        val enableResponse = responseAs[Map[String, Seq[String]]]
-        assertResult(Map("Success"->Seq(disabledUser), "NoChangeRequired"->Seq(enabledUser))) { enableResponse }
+    "Attempting to enable multiple users in various states should return a map of status lists" in {
+      val assortmentOfUserEmails = Seq(disabledUser, enabledUser, enrolledUser, terminatedUser, dummy1User, dummy2User)
+
+      Post(enablePath, assortmentOfUserEmails) ~> dummyUserIdHeaders(manager) ~> trialApiServiceRoutes ~> check {
+        val enableResponse = responseAs[Map[String, Seq[String]]].map(_.swap)
+
+        assert(enableResponse(List(enabledUser, dummy2User)) === StatusUpdate.NoChangeRequired.toString)
```
davidangb: this test - which relies on a `List` with multiple elements - is fragile in that it is dependent on the order of the users passed in to the api, as well as those users being processed in order. If the internals of the api processed futures in parallel and didn't guarantee order, the result could be `List(enabledUser, dummy2User)` or `List(dummy2User, enabledUser)`.

It's working now and seems stable, just pointing this out as potential fragility down the line. In cases like this I tend to convert everything to a `Set` and use those.

davidangb (commented):
good stuff, comments inline, nothing critical.

rushtong (approved):
👍 LGTM. I restarted the travis ci build that was failing - hopefully it passes.

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
```diff
@@ -110,13 +111,15 @@ final class TrialApiServiceSpec extends BaseServiceSpec with UserApiService with
...
         status should equal(InternalServerError)
+        assert(localThurloeDao.agreedUsers.isEmpty)
       }
     }
 
     allHttpMethodsExcept(PUT) foreach { method =>
       s"should reject ${method.toString} method" in {
         new RequestBuilder(method)(userAgreementPath) ~> dummyUserIdHeaders(enabledUser) ~> userServiceRoutes ~> check {
           assert(!handled)
+          assert(localThurloeDao.agreedUsers == Seq())
```
davidangb: there are still a bunch of places where you could check `isEmpty` instead of `== Seq())`

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
```diff
@@ -247,19 +253,71 @@ final class TrialApiServiceSpec extends BaseServiceSpec with UserApiService with
...
+
+        Post(enablePath, assortmentOfUserEmails) ~> dummyUserIdHeaders(manager) ~> trialApiServiceRoutes ~> check {
+          val enableResponse = responseAs[Map[String, Set[String]]].map(_.swap)
+
+          assert(enableResponse(Set(enabledUser, dummy2User)) === StatusUpdate.NoChangeRequired.toString)
+          assert(enableResponse(Set(disabledUser)) === StatusUpdate.Success.toString)
+          assert(enableResponse(Set(enrolledUser)).contains("Failure: Cannot transition"))
+          assert(enableResponse(Set(terminatedUser)).contains("Failure: Cannot transition"))
+          assert(enableResponse(Set(registeredUser))
+            .contains("ServerError: Should only be updating enabled, disabled or enrolled users"))
```
davidangb: this assert passes ... but the underlying logic is incorrect, I believe. We *should* be able to enable a user who has no previous state.

Update: I think this is actually a problem in TrialApiServiceSpecThurloeDAO. The ``case `disabledUser` =>`` should be ``case `disabledUser` | `registeredUser` =>``

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
kyuksel: Hopefully got them all this time 🤞🏼

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
kyuksel: good catch!

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
```diff
@@ -97,6 +98,17 @@ final class TrialApiServiceSpec extends BaseServiceSpec with UserApiService with
...
+    // For testing Thurloe returning no KVPs
+    localThurloeServer
+      .when(request()
+        .withMethod("GET")
+        .withHeader(fireCloudHeader.name, fireCloudHeader.value)
+        .withPath(UserApiService.remoteGetAllPath.format(userWithNoKVPs)))
+      .respond(
+        org.mockserver.model.HttpResponse.response()
+          .withHeaders(MockUtils.header).withStatusCode(OK.intValue)
+          .withBody(profile(userWithNoKVPs, Map.empty[String, String]).toJson.compactPrint))
```
kyuksel: @davidangb Could you please double-check if I got this mocking right? I wanted to mimic the case where a user has no KVPs (inspired by the discussion on @ahaessly's PR#580). See also the tests added accordingly (look for `userWithNoKVPs`).

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
```diff
@@ -537,7 +535,16 @@ final class TrialApiServiceSpec extends BaseServiceSpec with UserApiService with
...
-        case user @ `dummy2User` => { // Mocking Thurloe status saving failures
+        case `registeredUser` => {
+          assertResult(Some(TrialStates.Enabled)) { trialStatus.state }
+          assert(trialStatus.enabledDate.toEpochMilli > 0)
+          assert(trialStatus.enrolledDate.toEpochMilli === 0)
+          assert(trialStatus.terminatedDate.toEpochMilli === 0)
+          assert(trialStatus.expirationDate.toEpochMilli === 0)
+
+          Future.successful(Success(()))
+        }
```
davidangb: style issue (therefore not a blocker for the PR), but these are the exact same lines as ``case `disabledUser`` ... combine them?

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
davidangb: This is functionally correct (footnote below), but I think it could be coded differently.

In the `allUsersAndProps` at line 70, `registeredUser` maps to `noTrialProps`, and `noTrialProps` is already a `Map.empty[String,String]` ... so you don't need this new "// For testing Thurloe returning no KVPs" mock, and you don't need the `userWithNoKVPs`, unless I'm missing something?

footnote: for a registered user who has never been enabled or disabled, Thurloe won't actually return an empty map - it will return KVPs that populate the user's `Profile`: see https://github.com/broadinstitute/firecloud-orchestration/blob/develop/src/main/scala/org/broadinstitute/dsde/firecloud/model/Profile.scala#L82. But it won't return any keys that would populate any part of `UserTrialStatus`. So the tests in here are valid, but to make them the most real, you could append a set of `Profile` KVPs to each response. I don't think you need to do this in order to merge this PR - the comment is for your edification.

davidangb (approved):
:+1: all good!

PR #577, opened 2017-12-13
gawb2947: additional exception handling fix

ahaessly:

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
@@ -168,19 +168,18 @@ final class TrialService
   private def executeStateTransitions(managerInfo: UserInfo, userEmails: Seq[String],
                                       statusTransition: (WorkbenchUserInfo, Option[UserTrialStatus]) => UserTrialStatus,
                                       transitionPostProcessing: (Attempt, Option[UserTrialStatus], UserTrialStatus) => Unit): Future[PerRequestMessage] = {
-    var results: Seq[(String, String)] = Seq()
-    userEmails.foreach { userEmail =>
+    val results: Seq[(String, String)] = userEmails map { userEmail =>
       val finalStatus = executeStateTransition(managerInfo, userEmail, statusTransition, transitionPostProcessing)
       // Use await here so that multiple Futures don't have a conflict when claiming a billing project
 
       val result =
         try {
-          Await.result(finalStatus, scala.concurrent.duration.Duration.Inf)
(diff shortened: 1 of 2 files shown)


src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
```diff
@@ -260,6 +263,13 @@ final class TrialApiServiceSpec extends BaseServiceSpec with UserApiService with
       }
     }
 
+    "Attempting to enable a non registered user should fail" in {
+      Post(enablePath, Seq(unregisteredUser)) ~> dummyUserIdHeaders(manager) ~> trialApiServiceRoutes ~> check {
```
davidangb: use `nonRegisteredUserEmails` instead of `Seq(unregisteredUser)`?

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -170,7 +170,27 @@ final class TrialService
                                       transitionPostProcessing: (Attempt, Option[UserTrialStatus], UserTrialStatus) => Unit): Future[PerRequestMessage] = {
     var results: Seq[(String, String)] = Seq()
     userEmails.foreach { userEmail =>
```
davidangb: eliminate the use of `var` here:
```
val results: Seq[(String, String)] = userEmails.map { userEmail =>
    ...
    ...
    (result, userEmail)
}
```

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -170,7 +170,27 @@ final class TrialService
...
+      val finalStatus = executeStateTransition(managerInfo, userEmail, statusTransition, transitionPostProcessing)
+      // Use await here so that multiple Futures don't have a conflict when claiming a billing project
+
+      val result =
+        try {
+          Await.result(finalStatus, scala.concurrent.duration.Duration.Inf)
+        } catch {
+          case ex: FireCloudException =>
+            StatusUpdate.toName(StatusUpdate.ServerError(ex.getMessage))
+        }
```
davidangb: yeah, I really don't like the use of `Await` here, especially with a `Duration.Inf`. Step 1 is to reduce the duration to something that won't hang the server in an extreme case, e.g. `1.minute` ... step 2 will be to eliminate the blocking code entirely. I can take a shot at refactoring this too.

kyuksel (approved):

PR #580, opened 2017-12-14
Ah gawb 2976: remove await, clean up 

ahaessly:
clean up exceptions
remove await

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
@@ -84,15 +84,14 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
     * @param callerToken the OAuth token of the person making the API call
     * @return the trial status for the specified user, or None if trial status could not be determined.
     */
-  override def getTrialStatus(forUserId: String, callerToken: WithAccessToken): Future[Option[UserTrialStatus]] = {
+  override def getTrialStatus(forUserId: String, callerToken: WithAccessToken): Future[UserTrialStatus] = {
     wrapExceptions {
       userAuthedRequest(Get(UserApiService.remoteGetAllURL.format(forUserId)), false, true)(callerToken) map { response =>
         response.status match {
           case StatusCodes.OK =>
             val status = UserTrialStatus(unmarshal[ProfileWrapper].apply(response))
             assert(forUserId == status.userId, "status id does not match!")
-            Some(status)
-          case StatusCodes.NotFound => None
+            status
           case _ => throw new FireCloudException("Unable to get user trial status")
         }
       }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
@@ -82,7 +82,7 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
     *
     * @param forUserId the subjectid of the user whose trial status to get
     * @param callerToken the OAuth token of the person making the API call
-    * @return the trial status for the specified user, or None if trial status could not be determined.
+    * @return the trial status for the specified user, or a default UserTrialStatus object
     */
   override def getTrialStatus(forUserId: String, callerToken: WithAccessToken): Future[UserTrialStatus] = {
     wrapExceptions {
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ThurloeDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ThurloeDAO.scala
@@ -49,7 +49,7 @@ trait ThurloeDAO extends LazyLogging with ReportsSubsystemStatus {
     * @param callerToken the OAuth token of the person making the API call
     * @return the trial status for the specified user, or None if trial status could not be determined.
     */
-  def getTrialStatus(forUserId: String, callerToken: WithAccessToken): Future[Option[UserTrialStatus]]
+  def getTrialStatus(forUserId: String, callerToken: WithAccessToken): Future[UserTrialStatus]
 
   /**
     * set the UserTrialStatus for a specific user
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ThurloeDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ThurloeDAO.scala
@@ -43,11 +43,12 @@ trait ThurloeDAO extends LazyLogging with ReportsSubsystemStatus {
 
   // methods to work with free-trial objects
   /**
-    * get the UserTrialStatus associated with a specific user.
+    * get the UserTrialStatus associated with a specific user or a default UserTrialStatus object.
     *
     * @param forUserId the subjectid of the user whose trial status to get
     * @param callerToken the OAuth token of the person making the API call
-    * @return the trial status for the specified user, or None if trial status could not be determined.
+    * @return the trial status for the specified user, or a default UserTrialStatus object
+
     */
   def getTrialStatus(forUserId: String, callerToken: WithAccessToken): Future[UserTrialStatus]
 
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
@@ -172,7 +172,14 @@ final class TrialService
     userEmails.foreach { userEmail =>
       val finalStatus = executeStateTransition(managerInfo, userEmail, statusTransition, transitionPostProcessing)
       // Use await here so that multiple Futures don't have a conflict when claiming a billing project
-      val result = Await.result(finalStatus, scala.concurrent.duration.Duration.Inf)
+
+      val result =
+        try {
+          Await.result(finalStatus, scala.concurrent.duration.Duration.Inf)
+        } catch {
+          case ex: FireCloudException =>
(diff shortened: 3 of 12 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
```diff
@@ -84,15 +84,14 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
...
+  override def getTrialStatus(forUserId: String, callerToken: WithAccessToken): Future[UserTrialStatus] = {
     wrapExceptions {
       userAuthedRequest(Get(UserApiService.remoteGetAllURL.format(forUserId)), false, true)(callerToken) map { response =>
         response.status match {
           case StatusCodes.OK =>
             val status = UserTrialStatus(unmarshal[ProfileWrapper].apply(response))
             assert(forUserId == status.userId, "status id does not match!")
-            Some(status)
-          case StatusCodes.NotFound => None
+            status
```
kyuksel, line 94: Why are we throwing an exception when a user has no trial status given this will be the case for every user initially?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
```diff
@@ -84,15 +84,14 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
     * @param callerToken the OAuth token of the person making the API call
     * @return the trial status for the specified user, or None if trial status could not be determined.
```
kyuksel: `return` value description doesn't reflect the change

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ThurloeDAO.scala
```diff
@@ -49,7 +49,7 @@ trait ThurloeDAO extends LazyLogging with ReportsSubsystemStatus {
     * @param callerToken the OAuth token of the person making the API call
     * @return the trial status for the specified user, or None if trial status could not be determined.
```
kyuksel: `return` value description doesn't match the new value

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -150,73 +151,71 @@ final class TrialService
...
-        throw new FireCloudException(s"Cannot transition from ${innerState} to $newState.state.get")
+  private def requiresStateTransition(currentState: UserTrialStatus, newState: UserTrialStatus): Boolean = {
+    if (currentState.state.isEmpty || currentState.state != newState.state) {
+      if (newState.state.isEmpty || !newState.state.get.isAllowedFrom(currentState.state))
+        throw new FireCloudException(s"Cannot transition from ${currentState.state} to $newState.state.get")
       true
     } else {
       false
     }
   }
```
kyuksel: I fear that this logic is complicated and may result in bugs not easy to spot. A few thoughts:

1. If `(currentState.state.isEmpty || currentState.state != newState.state)` is `true`, `newState.state.isEmpty` should be `false` so the latter could be removed from the second `if` clause.

2. Separating the logic that checks if the transition is a valid one should simplify things and make them more modular.

3. If `requiresStateTransition` needs to remain as is, it'd be good to have a unit test for it.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -150,73 +151,71 @@ final class TrialService
...
-        } catch {
-          case ex: FireCloudException =>
-            StatusUpdate.toName(StatusUpdate.ServerError(ex.getMessage))
-        }
-      (result, userEmail)
+  private def throwableToStatus(t: Throwable): String = {
+    t match {
+      case exr: FireCloudExceptionWithErrorReport =>
+        if (exr.errorReport.statusCode.contains(StatusCodes.NotFound))
+          StatusUpdate.toName(StatusUpdate.Failure("User not registered"))
```
kyuksel, line 170: How do we know that the `NotFound` status we get is due to the user not being registered?

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -150,73 +151,71 @@ final class TrialService
...
+        else StatusUpdate.toName(StatusUpdate.Failure(exr.errorReport.message))
+      case ex: FireCloudException =>
+        StatusUpdate.toName(StatusUpdate.Failure(ex.getMessage))
+      case _: Throwable =>
```
kyuksel: This can be `case _ =>`.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
davidangb: The `userAuthedRequest` will only return an error if the user has no KVPs in Thurloe at all ... which we *should* consider an error, since we expect all free-trial users to already be registered and have filled out their profile.

As long as Thurloe returns at least one KVP, the `response.status` will be `OK`.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
davidangb: also please realize that `TrialState.isAllowedFrom` can - and *SHOULD* handle the logic of transitioning from `None` to some state. So, how about:

```
  private def requiresStateTransition(currentState: UserTrialStatus, newState: UserTrialStatus): Boolean = {
    if (newState.state.isEmpty || !newState.state.get.isAllowedFrom(currentState.state))
      throw new FireCloudException(s"Cannot transition from ${currentState.state} to ${newState.state}")

    currentState.state != newState.state
  }
```
(untested, coding in the void)

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
ahaessly: actually thurloe will return an empty key value list even if the user doesn't exist in thurloe - for example:
{
  "userId": "asdf",
  "keyValuePairs": []
}

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
ahaessly: excellent point. i'm really not happy with the complexity of error handling i'm having to go through here. too many different places and checks. i'll think about this one. open to suggestions for simplification

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialServiceSupport.scala
```diff
@@ -34,12 +36,13 @@ trait TrialServiceSupport extends LazyLogging {
...
-      else
-        Future[Option[UserTrialStatus]](None)
-    })
+    val trialStatuses = Future.sequence(
+      projects.foldLeft(Seq[Future[UserTrialStatus]]())((result, p) => {
+        if (p.user.isDefined)
+          result ++ Seq(thurloeDAO.getTrialStatus(p.user.get.userSubjectId, managerInfo))
+        else
+          result
+      }))
```
davidangb: mostly style, but I generally find `fold`s to be hard to read and avoid unless they're really needed. Your call; would either of these be better?
```
    val trialStatuses = Future.sequence(
      projects.filter(p => p.user.isDefined)
        .map(p => thurloeDAO.getTrialStatus(p.user.get.userSubjectId, managerInfo)))
```
or
```
    val trialStatuses = Future.sequence(
      projects.map {
        case p if p.user.isDefined => thurloeDAO.getTrialStatus(p.user.get.userSubjectId, managerInfo))
      })
```
(again, untested)

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockThurloeDAO.scala
```diff
@@ -134,10 +134,10 @@ class MockThurloeDAO extends ThurloeDAO {
     Future.successful(
       forUserId match {
         case "enabled-user" =>
-          Some(UserTrialStatus(forUserId, Some(TrialStates.Enabled), true, 0, 0, 0, 0, Some("date")))
-        case "disabled-user" => None  // TODO return an empty UserTrialStatus - more accurate with actual thurloe call
+          UserTrialStatus(forUserId, Some(TrialStates.Enabled), true, 0, 0, 0, 0, Some("date"))
+        case "disabled-user" => UserTrialStatus(ProfileWrapper(forUserId, List.empty))
```
davidangb, line 138: nice change.

davidangb (commented):
a few code comments inline. I didn't run this, but it does pass tests locally.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
kyuksel: Part of the complexity is in inherent in the number of and nestedness of backend calls. I fiddled a bit with `executeStateTransitions` in hopes of simplifying it further but didn't get far. 

At the least, I'd change `"User not registered"` to something like `"Not Found"` which is all this method could know about.

davidangb (approved):
👍 for the code changes.

#573 has important tests for this. Let's see if we can coordinate these - hopefully get #573 in first, then rebase this PR and ensure everything still passes.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
kyuksel: Seeing the same Thurloe response as @ahaessly described. Doesn't this conflict with your reasoning @davidangb?

davidangb (approved):
:+1:

kyuksel (approved):

PR #581, opened 2017-12-14
GAWB-2986: use caller token for reindexing, add logging

davidangb:
In Library's reindex-all operation, we [require](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala#L73) the caller to be a FireCloud admin. But in the call from orch to rawls, we generate a new access token, using an admin service account. Since the original caller is already an admin, we could/should just use the caller's token.

I've also added some extra logging to reindex-all failures to help out in the future.

I tested by:
* performing a successful reindex as admin
* attempting reindex as non-admin and being rejected
* manually modifying the `rawlsAdminWorkspaces` url in `HttpRawlsDAO` to force errors when running locally, and seeing that my new logging statements are properly logged

In all cases:


ahaessly (approved):
👍

PR #585, opened 2018-01-05
Smarter update strategy for Onix report [GAWB-2963]

aednichols:
~Needs unit tests, probably fix up some of the clunky naming. Added~ There are still two requirements questions to the JIRA ticket, so may need updates for those.

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
@@ -354,7 +354,7 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
   }
 
   private def updatePreservingOrder(newContent: ValueRange, existingContent: ValueRange): List[java.util.List[AnyRef]] = {
-    val header: java.util.List[AnyRef] = existingContent.getValues.get(0)
+    val header: java.util.List[AnyRef] = existingContent.getValues.head
     val existingRecords: List[java.util.List[AnyRef]] = existingContent.getValues.drop(1).toList
 
     val newRecords: List[java.util.List[AnyRef]] = newContent.getValues.drop(1).toList
@@ -362,7 +362,7 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
     // Go through existing records and update them in place
     val existingRecordsUpdated = existingRecords.map { existingRecord =>
       val matchingNewRecord = newRecords.find { newRecordCandidate =>
-        newRecordCandidate.get(0) == existingRecord.get(0)
+        newRecordCandidate.head == existingRecord.head
       }
 
       matchingNewRecord match {
@@ -374,7 +374,7 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
     // Create new records for newly-added projects
     val recordsToAppend = newRecords.filter { newRecordCandidate =>
       !existingRecords.exists { existingRecordCandidate =>
-        existingRecordCandidate.get(0) == newRecordCandidate.get(0)
+        existingRecordCandidate.head == newRecordCandidate.head
       }
     }
 
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
@@ -354,9 +354,9 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
   }
 
   private def updatePreservingOrder(newContent: ValueRange, existingContent: ValueRange): List[java.util.List[AnyRef]] = {
-    val header: java.util.List[AnyRef] = existingContent.getValues.head
     val existingRecords: List[java.util.List[AnyRef]] = existingContent.getValues.drop(1).toList
 
+    val header: java.util.List[AnyRef] = newContent.getValues.head
     val newRecords: List[java.util.List[AnyRef]] = newContent.getValues.drop(1).toList
 
     // Go through existing records and update them in place
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
@@ -354,11 +354,12 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
   }
 
   private def updatePreservingOrder(newContent: ValueRange, existingContent: ValueRange): List[java.util.List[AnyRef]] = {
-    val existingRecords: List[java.util.List[AnyRef]] =
-      existingContent.getValues match {
-        case rows: Any => rows.drop(1).toList
-        case null => List()
-      }
+    val existingRecords =
+      // getValues may come through as an instantiated list of type null with zero entries due to Scala <> Java stuff
+      if (Try(existingContent.getValues.size()).toOption.getOrElse(0) > 0)
+        existingContent.getValues.tail.toList
+      else
+        List()
 
     val header: java.util.List[AnyRef] = newContent.getValues.head
     val newRecords: List[java.util.List[AnyRef]] = newContent.getValues.drop(1).toList
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
@@ -1,5 +1,7 @@
 package org.broadinstitute.dsde.firecloud.dataaccess
 
+import java.util
+
 import akka.actor.ActorRefFactory
 import com.google.api.client.auth.oauth2.Credential
 import com.google.api.client.googleapis.auth.oauth2.GoogleCredential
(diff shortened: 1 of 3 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -334,16 +334,53 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
...
+  private def updatePreservingOrder(newContent: ValueRange, existingContent: ValueRange): List[java.util.List[AnyRef]] = {
+    val header: java.util.List[AnyRef] = existingContent.getValues.get(0)
+    val existingRecords: List[java.util.List[AnyRef]] = existingContent.getValues.drop(1).toList
+
+    val newRecords: List[java.util.List[AnyRef]] = newContent.getValues.drop(1).toList
+
+    // Go through existing records and update them in place
+    val existingRecordsUpdated = existingRecords.map { existingRecord =>
+      val matchingNewRecord = newRecords.find { newRecordCandidate =>
+        newRecordCandidate.get(0) == existingRecord.get(0)
```
aednichols: I don't really like using `.get(0)` everywhere to access the project ID.

I was thinking of refactoring `SpreadsheetRow` to include it, and adopting that class here, but I'm not sure it's worth it.

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAOSpec.scala
```diff
@@ -63,4 +65,32 @@ class HttpGoogleServicesDAOSpec extends FlatSpec with Matchers {
     objectMetadata.name should equal("test-composite-object")
     objectMetadata.md5Hash should equal(None)
   }
+
+  {
+    // https://stackoverflow.com/a/24375762/818054
+    val updatePreservingOrder = PrivateMethod('updatePreservingOrder)
```
breilly2: Does `updatePreservingOrder` have to be private? Or can you extract the thing that you want to test into a separate object that this can logically include in its public API?

aednichols: Good question. The public API is `updateSpreadsheet`, above, which interacts directly with Google - so we can't conveniently test that.

I'd entertain making `updatePreservingOrder` `public` to make tests simpler, but it doesn't naturally belong to the public API of `HttpGoogleServicesDAO`.

This is the first time I've encountered this specific problem in Scala, and decided to explore what's possible in `ScalaTest`.

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAOSpec.scala
```diff
@@ -63,4 +65,86 @@ class HttpGoogleServicesDAOSpec extends FlatSpec with Matchers {
...
+      check(
+        newContent = List(headers, row2, row3, row1),
+        existingContent = List(headers, row1, row2, row3),
+        expectedOutput = List(headers, row1, row2, row3)
+      )
+    }
+
+    def check(newContent: List[java.util.List[AnyRef]], existingContent: List[java.util.List[AnyRef]], expectedOutput: List[java.util.List[AnyRef]]): Unit = {
+      // https://stackoverflow.com/a/24375762/818054
+      val updatePreservingOrder = PrivateMethod('updatePreservingOrder)
```
aednichols, line 193: If people don't like this (and I can see why they wouldn't), happy to just make `updatePreservingOrder` `public`

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAOSpec.scala
```diff
@@ -63,4 +65,102 @@ class HttpGoogleServicesDAOSpec extends FlatSpec with Matchers {
...
+        expectedOutput = List(headers, row1, row2, row3)
+      )
+    }
+
+    it should "output the union of the rows (disjoint sets)" in {
+      check(
+        newContent = List(headers, row3),
+        existingContent = List(headers, row1, row2),
+        expectedOutput = List(headers, row1, row2, row3)
+      )
```
davidangb, line 148: I think these last few are order-dependent, aren't they? Might be worth mucking with the order a bit ... 

```
 check(
        newContent = List(headers, row1),
        existingContent = List(headers, row2, row3),
        expectedOutput = List(headers, row2, row3, row1)
      )
```

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -334,16 +334,53 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
...
+    // Create new records for newly-added projects
+    val recordsToAppend = newRecords.filter { newRecordCandidate =>
+      !existingRecords.exists { existingRecordCandidate =>
+        existingRecordCandidate.get(0) == newRecordCandidate.get(0)
+      }
+    }
+
+    List(header) ++ existingRecordsUpdated ++ recordsToAppend
+  }
+
```
davidangb, line 388: I would have transformed both `newRecords` and `existingRecords` into maps keyed by the project id: `Map[String, List[AnyRef]]`. Then, inside your loops, instead of doing a `find` or `exists`, you can do a more-efficient map lookup.

At the scale we're talking about - getting up to a thousand-ish rows - this may be premature optimization. So don't consider it a requirement for this PR; I'm mentioning it for the sake of talking about code :)

davidangb (approved):
:+1:, optional comments inline

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
rushtong: How about something more scala-y, but the same thing ... `newRecordCandidate.head == existingRecord.head` ?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -334,16 +334,53 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
...
+    // Smart merge existing with new
+    val rows = updatePreservingOrder(newContent, existingContent)
+
+    // Send update
+    val response = service.spreadsheets().values().update(spreadsheetId, newContent.getRange, newContent.setValues(rows)).setValueInputOption("RAW").execute()
     response.toString.parseJson.asJsObject
   }
 
+  private def updatePreservingOrder(newContent: ValueRange, existingContent: ValueRange): List[java.util.List[AnyRef]] = {
+    val header: java.util.List[AnyRef] = existingContent.getValues.get(0)
```
rushtong: Consider `existingContent.getValues.head` if you keep this approach

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
rushtong: +1 to this idea.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -334,16 +334,53 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
...
+    val newRecords: List[java.util.List[AnyRef]] = newContent.getValues.drop(1).toList
+
+    // Go through existing records and update them in place
+    val existingRecordsUpdated = existingRecords.map { existingRecord =>
+      val matchingNewRecord = newRecords.find { newRecordCandidate =>
+        newRecordCandidate.get(0) == existingRecord.get(0)
+      }
+
+      matchingNewRecord match {
+        case Some(newRecord) => newRecord // Overwrite the entire row in place, adequate unless we want to preserve user-added columns
```
rushtong: I do think we should consider preserving user-added columns, not necessarily in this PR. One of the reasons for keeping a data row in the same order would be this exact case - the user has added a new column with a comment that applies to the current row. If we are not trying to preserve user-comments in additional columns, then there's no real point in preserving the order of the data rows. 

That being said, I think this might still work (will test functionally). A new `ValueRange`, when pushed to the spreadsheet, *only* overwrites values *inside* the range and leaves values (and formats, etc.) outside the range intact.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -334,16 +334,53 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
...
+        case _ => existingRecord // Don't delete existing records that aren't in the new export
```
rushtong, line 375: I am conflicted about this. I agree that we should not try to delete rows. However, I think we do need to indicate to the user that FireCloud no longer knows about this particular row. How we do that is an open question ... perhaps a separate story/PR.

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAOSpec.scala
rushtong: Cool! I didn't know this was possible.

rushtong (approved):
👍 For code. Will do some functional testing as well.

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAOSpec.scala
rushtong: Another thought is to pull the method into a supporting trait that would allow for easier testing.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
rushtong: In functional testing, `existingContent.getValues` can be null for brand new reports. This method needs to handle both cases, new + empty, existing + filled.

rushtong: Minor issue. In functional testing, headers are maintained and not changed. So, if I have an older spreadsheet that is missing a newer column, then update, the new column data is written without a header.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
rushtong: Update. I believe this is functioning as it should: 
* New records do not overwrite data in user-added columns and instead are appended as new rows. 
* Existing records are overwritten in place without changing user-added columns outside of the updated data. The caveat to this is that if new columns are added to the output beyond what the existing spreadsheet has, they *will* overwrite the user-added columns that exist in the same position.

rushtong (changes requested):
Need to handle the empty data case in `updatePreservingOrder`

andy7i (commented):
Only reviewed test cases:
Overall, this is awesome. I would recommend the following:
1. test for new columns as greg found.
2. test for a data cell being deleted (emptied).
3. test for cases where rows are updated (similar to your tests for sets) with modified rows (row1a, row2a, row3a).

@andy7i Adam edit:
1. `"use the header from the new data in the update"`
2. modified `row2` and `row3` to have a `null` and a `""`, other tests use the data in ways that cover this case
3. `"update data in a row"` and `"locate a row in existing that moved based on its first column, and replace its remaining columns with new"`

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
aednichols: 1. Done
2. Fixed + added test
3. Fixed + added test

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
aednichols: This is true, I was lazy and optimized for simplicity on the first pass.

The above would need things like `record.head.asInstanceOf[String]` to wrangle row data from `java.util.List[java.util.List[Object]]` into the right types, and I kinda felt like leaving them alone

davidangb (approved):
:+1: did not test locally, looks good!

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAOSpec.scala
```diff
@@ -63,4 +65,140 @@ class HttpGoogleServicesDAOSpec extends FlatSpec with Matchers {
...
+
+    it should "not explode if there is no data (only headers)" in {
+      check(
+        newContent = List(headers),
+        existingContent = List(headers),
+        expectedOutput = List(headers)
+      )
+    }
+
+    // This happens when calling updateBillingReport the first time from within createBillingReport
```
andy7i, line 94: This is great and makes me think of another edge case: 
What would happen if `updateBillingReport` or `createBillingReport` were called when there is no data yet. If this is a valid scenario, it should handle when newContent==null and existingContent==null (or is an empty list)

davidangb: agree this is a valid edge case, but wonder if it's worth expending effort on. Remember that very few people - only campaign managers, and then a subset of those - will call this, so we can be pretty tolerant of unexpected errors.

aednichols: It's possible, but I'm comfortable leaving the behavior undefined, a.k.a. "don't do that"

andy7i (approved):
👍 to test plan

rushtong (approved):
👍 Functionally beautiful!

PR #586, opened 2018-01-09
GAWB-2934: Added passthrough for sam's proxyGroup endpoint.

breilly2:
Ended up just adding the passthrough because, in sam, the proxy group is currently an implementation detail in the Google cloud extension support and, therefore, seemed incorrect to add to `UserInfo`. I think we should either decide to include cloud extension info in `UserInfo` or make an overall cloud extension user info endpoint that the proxy group can be included in. This passthrough will work until we're ready to make that decision.

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
@@ -55,6 +55,7 @@ object UserApiService {
   def rawlsGroupRequestAccessPath(group: String) = rawlsGroupPath(group) + "/requestAccess"
   def rawlsGroupRequestAccessUrl(group: String) = FireCloudConfig.Rawls.baseUrl + rawlsGroupRequestAccessPath(group)
 
+  def samUserProxyGroupURL(email: String) = FireCloudConfig.Sam.baseUrl + s"/api/google/user/proxyGroup/$email"
 }
 
 // TODO: this should use UserInfoDirectives, not StandardUserInfoDirectives. That would require a refactoring
@@ -180,6 +181,11 @@ trait UserApiService extends HttpService with PerRequestCreator with FireCloudRe
               passthrough(UserApiService.rawlsGroupMemberUrl(groupName, role, email), HttpMethods.DELETE)
             }
           }
+        }
+      }
+      pathPrefix("proxyGroup") {
+        path(Segment) { email =>
+          passthrough(UserApiService.samUserProxyGroupURL(email), HttpMethods.GET)
         }
       }
     } ~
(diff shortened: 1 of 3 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
```diff
@@ -182,6 +183,11 @@ trait UserApiService extends HttpService with PerRequestCreator with FireCloudRe
           }
         }
       }
+      pathPrefix("proxyGroup") {
+        path(Segment) { email =>
+          passthrough(UserApiService.samUserProxyGroupURL(email), HttpMethods.GET)
```
rushtong: Should this route segment be wrapped in a `get {...}` block?

breilly2: Good question. Should it? I've seen passthroughs wrapped before but that looked to me like a holdover from the implementations previously not being passthroughs. I was thinking that specifying `HttpMethods.GET` in the call to `passthrough` was enough to only apply to that method, but I'm glad you asked.

Can anyone confirm how this should be?

rushtong: What I think will happen here is that all methods (`PUT`, `POST`, etc.) will be handled and channeled to a GET. It will work, but doesn't exactly follow the swagger spec.

jmthibault79: My understanding is that this would already reject anything that's not a `GET`.

Since we have some confusion on this, would you mind testing this assumption and adding a comment to the `passthrough()` method itself based on what you find?

breilly2: Ha! Tests! Yeah, I should add some!

I will also clarify the behavior of `passthrough` as you suggested. Thanks.

MatthewBemis (approved):
:+1: pending clarification of whether or not the `get { ... }` is needed around the passthrough

rushtong:
Pretty sure Joel is right and my concern is not valid. This test passes without the `get { ... }` block. 
```
      "non-GET methods should be rejected" in {
        allHttpMethodsExcept(GET) foreach { method =>
          new RequestBuilder(method)("/api/proxyGroup/test@test.test") ~> dummyAuthHeaders ~> sealRoute(userServiceRoutes) ~> check {
            status should be(MethodNotAllowed)
          }
        }
      }
```

breilly2:
@rushtong Yes, although does that test verify that it's orch saying method not allowed rather than sam saying it and orch passing it through?

I ran an experiment where I did a `curl` PUT to the orch endpoint and nothing showed up in the sam proxy log so it looks like it is orch rejecting the request as desired.

rushtong:
@breilly2 - Yes, that's correct, the non-`GET` calls are being denied by Orch. I misunderstood how the routing would work without the method specified in the DSL, but it is working correctly.

dvoet (approved):

PR #589, opened 2018-01-19
GAWB-3062: Ability to finalize free trial user status

kyuksel:
- It enables the UI to hide the free trial banner permanently and addresses [GAWB-3062](https://broadinstitute.atlassian.net/browse/GAWB-3062).

- The user-facing free trial API is enhanced to allow a `Terminated` user's status to be updated as `Finalized`. It's also lenient in that if an already `Finalized` user is attempted to be `Finalized` again, no error is thrown.

- The former URL (`/api/profile/trial`) maps to `enroll` for backwards compatibility and regression concerns. 

- No new timestamp was added to Thurloe KVPs since keeping track of when users are finalized didn't feel useful.

- The tests for user-initiated requests (`/api/profile/trial*`) were grouped together for maintainability.

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/model/Trial.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/model/Trial.scala
@@ -40,7 +40,7 @@ object Trial {
       def isAllowedFromNone: Boolean = false
     }
 
-    val allStates = Seq(Disabled, Enabled, Enrolled, Terminated)
+    val allStates = Seq(Disabled, Enabled, Enrolled, Terminated, Finalized)
 
     def withName(name: String): TrialState = {
       name match {
@@ -58,6 +58,7 @@ object Trial {
         case Enabled => "Enabled"
         case Enrolled => "Enrolled"
         case Terminated => "Terminated"
+        case Finalized => "Finalized"
         case _ => throw new FireCloudException(s"invalid TrialState [${state.getClass.getName}]")
       }
     }
@@ -74,6 +75,9 @@ object Trial {
     }
     case object Terminated extends TrialState {
       override def isAllowedFromState(previous: TrialState): Boolean = previous == Enrolled
+    }
+    case object Finalized extends TrialState {
+      override def isAllowedFromState(previous: TrialState): Boolean = previous == Terminated
     }
   }
 
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/model/Trial.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/model/Trial.scala
@@ -40,8 +40,6 @@ object Trial {
       def isAllowedFromNone: Boolean = false
     }
 
-    val allStates = Seq(Disabled, Enabled, Enrolled, Terminated, Finalized)
-
     def withName(name: String): TrialState = {
       name match {
         case "Disabled" => Disabled
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/model/Trial.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/model/Trial.scala
@@ -78,7 +78,7 @@ object Trial {
       override def isAllowedFromState(previous: TrialState): Boolean = previous == Enrolled
     }
     case object Finalized extends TrialState {
-      override def isAllowedFromState(previous: TrialState): Boolean = previous == Terminated
+      override def isAllowedFromState(previous: TrialState): Boolean = previous == Terminated || previous == Finalized
     }
   }
 
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/model/Trial.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/model/Trial.scala
@@ -48,6 +48,7 @@ object Trial {
         case "Enabled" => Enabled
         case "Enrolled" => Enrolled
         case "Terminated" => Terminated
+        case "Finalized" => Finalized
         case _ => throw new FireCloudException(s"invalid TrialState [$name]")
       }
     }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
@@ -332,11 +332,14 @@ final class TrialService
     // Get user's trial status, check and update the current state if it's a valid transition
     // NB: We are being lenient and are not complaining when a user was already 'finalized' previously
     thurloeDao.getTrialStatus(userInfo.id, userInfo) flatMap { status =>
-      if (Finalized.isAllowedFrom(status.state)) {
-        thurloeDao.saveTrialStatus (userInfo.id, userInfo, status.copy (state = Some (Finalized) ) ) flatMap {
-          case Success (_) => Future (RequestComplete (NoContent) )
-          case Failure (ex) => Future (RequestComplete (InternalServerError, ex.getMessage) )
-        }
+      val state = status.state
+      if (Finalized.isAllowedFrom(state)) { 
+        if (state.contains(Terminated)) {
+          thurloeDao.saveTrialStatus(userInfo.id, userInfo, status.copy(state = Some(Finalized))) flatMap {
+            case Success(_) => Future(RequestComplete(NoContent))
+            case Failure(ex) => Future(RequestComplete(InternalServerError, ex.getMessage))
+          }
+        } else Future(RequestComplete(NoContent))
       } else {
         val errMsg = "Your free trial should have been terminated first."
         Future(RequestCompleteWithErrorReport(BadRequest, errMsg))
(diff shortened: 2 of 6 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -322,6 +326,27 @@ final class TrialService
...
+        case Some(Terminated) =>
+          thurloeDao.saveTrialStatus(userInfo.id, userInfo, status.copy(state = Some(Finalized))) flatMap {
+            case Success(_) => Future(RequestComplete(NoContent))
+            case Failure(ex) => Future(RequestComplete(InternalServerError, ex.getMessage))
+          }
+        // user in some other state; can't finalize
+        case _ =>
+          val errMsg = "Your free trial should have been terminated first."
+          Future(RequestCompleteWithErrorReport(BadRequest, errMsg))
+      }
```
davidangb: it feels like this `match` duplicates the logic in `isAllowedFrom`. Could this block be:
```
if (Finalized.isAllowedFrom(status.state)) ... else ...
```
I'm also wondering if we should throw an error if the user is already Finalized. Consider the likely use case where a Terminated user has multiple tabs open, each of which displays the banner. If the user finalizes in one tab, then switches to the next and attempts to finalize again ... perhaps we should be lenient.
```
if (Finalized.isAllowedFrom(status.state)) {
  ... // persist Finalized state
} else {
  status.state match {
    case Some(Finalized) => Future(RequestComplete(NoContent)) // lenient noop
    case _ =>
      val errMsg = "Your free trial should have been terminated first."
      Future(RequestCompleteWithErrorReport(BadRequest, errMsg))
  }
}
```

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
```diff
@@ -105,99 +105,149 @@ final class TrialApiServiceSpec extends BaseServiceSpec with UserApiService with
...
-  "Free Trial Enrollment" - {
-    val enrollPath = "/api/profile/trial"
+    val trialPathBase = "/api/profile/trial"
 
-    "User-initiated enrollment endpoint" - {
-      allHttpMethodsExcept(POST) foreach { method =>
-        s"should reject ${method.toString} method" in {
-          new RequestBuilder(method)(enrollPath) ~> dummyUserIdHeaders(enabledUser) ~> userServiceRoutes ~> check {
-            assert(!handled)
+    val enrollPaths = Seq(trialPathBase, s"$trialPathBase?operation=enroll")
```
davidangb, line 148: it's worth adding a test without the `operation` param to check the default behavior.

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
```diff
@@ -105,99 +105,149 @@ final class TrialApiServiceSpec extends BaseServiceSpec with UserApiService with
...
-          status should equal(BadRequest)
+      s"enrollment as a disabled user via $enrollPath" - {
+        "should be a BadRequest" in {
+          Post(enrollPath) ~> dummyUserIdHeaders(disabledUser) ~> userServiceRoutes ~> check {
+            status should equal(BadRequest)
+          }
+        }
+      }
+
+      s"enrollment as an enabled user that agreed to terms via $enrollPath" - {
```
davidangb: nitpick, pure language clarity, here and in the following test, this should be:
s"enrollment via $enrollPath as an enabled user that agreed to terms"
( the user enrolls via $enrollPath, the previous language makes it sound like the user agrees to terms via $enrollPath )

davidangb (commented):
a few comments inline!

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
kyuksel: By 'without the `operation` param', do you mean `/api/profile/trial`? If so, `trialPathBase` covers that.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
kyuksel: > perhaps we should be lenient

That's a good point. I agree that the system should be lenient. I'd personally make that happen in the UI since it's a user experience feature and to keep the backend consistent with the other status updates...but I don't feel strongly so went ahead and did it.

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
kyuksel: Thanks; I like clarity :)

kyuksel:
@davidangb Does the tech doc, the LucidCharts diagram and/or any other doc need to be updated to incorporate the `Finalized` state?

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -322,6 +326,24 @@ final class TrialService
     )
   }
 
+  private def finalizeUser(userInfo: UserInfo): Future[PerRequestMessage] = {
+    import TrialStates._
+
+    // Get user's trial status, check and update the current state if it's a valid transition
+    // NB: We are being lenient and are not complaining when a user was already terminated previously
```
aednichols: I do not understand how this is lenient - isn't that the expected use case? Do you mean finalized?

src/main/scala/org/broadinstitute/dsde/firecloud/model/Trial.scala
```diff
@@ -40,14 +40,15 @@ object Trial {
       def isAllowedFromNone: Boolean = false
     }
 
-    val allStates = Seq(Disabled, Enabled, Enrolled, Terminated)
+    val allStates = Seq(Disabled, Enabled, Enrolled, Terminated, Finalized)
```
aednichols: This is only used in `TrialSpec.scala` - should it live there?

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialApiServiceSpec.scala
davidangb: yes, you're right. I misread, my bad!

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -322,6 +326,24 @@ final class TrialService
...
+    thurloeDao.getTrialStatus(userInfo.id, userInfo) flatMap { status =>
+      if (Finalized.isAllowedFrom(status.state)) {
+        thurloeDao.saveTrialStatus (userInfo.id, userInfo, status.copy (state = Some (Finalized) ) ) flatMap {
+          case Success (_) => Future (RequestComplete (NoContent) )
+          case Failure (ex) => Future (RequestComplete (InternalServerError, ex.getMessage) )
+        }
```
davidangb: In the case where a finalized user attempts to re-finalize, this code sends a write to Thurloe. We don't need to contact Thurloe at all, and shouldn't write when we don't have to ... if a finalized user attempts to re-finalize, it should be a complete noop.

davidangb (commented):
one noop comment inline!

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
kyuksel: Yes, thanks for the catch!

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
kyuksel: Totally! To fix it, I ended up introducing a nested if block, which looks ugly. If you have suggestions on how to make it more elegant and find it worthwhile, please let me know.

davidangb (approved):
👍 once tests go green!

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
aednichols: Can you do
```
(Finalized.isAllowedFrom(state), state.contains(Terminated)) match {
  case (true, true) => ...
...
}
```

aednichols (approved):
I left a suggestion for the nested block, but totally optional

👍

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
kyuksel: Done. Thanks!

aednichols:
Still 👍

davidangb (approved):
Still 👍 (modulo tests)

aednichols:
jenkins retest

aednichols:
Failed at `DataSpec` `Writer and reader should see new columns` `with no defaults or local preferences when analysis run that creates new columns`

https://fc-jenkins.dsp-techops.broadinstitute.org/job/fiab-test/2117/consoleFull

aednichols:
jenkins retest

kyuksel:
It passed the Jenkins Automation Tests (see [this PR](https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/1101/) for details) but then I inadvertently kicked off another run, which I killed later. That's why there is an X next to Jenkins Automation Tests.

kyuksel:
jenkins retest

PR #591, opened 2018-01-30
GAWB-2937: reduce queries to ontology for parent labels during library reindex

davidangb:
As of DataBiosphere/consent#247, merged in Nov 2017, the ontology index contains parent labels. Orchestration was still making additional queries to populate those labels, even though it didn't have to. I've removed those extra queries.

In all cases:

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/OntologyAutocompleteSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/OntologyAutocompleteSpec.scala
@@ -30,25 +30,14 @@ class OntologyAutocompleteSpec extends FreeSpec {
       val expected = Set(
         // matches in label:
         "acute hemorrhagic leukoencephalitis",
-        "COL4A1-related familial vascular leukoencephalopathy",
         "hypomyelinating leukoencephalopathy",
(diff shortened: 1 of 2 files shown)


src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/OntologyAutocompleteSpec.scala
```diff
@@ -30,25 +30,14 @@ class OntologyAutocompleteSpec extends FreeSpec {
...
-        "CADASIL",
-        "hypomyelinating leukodystrophy 7 with or without oligodontia and-or hypogonadotropic hypogonadism", // note label is "leuko" without the "e"
         "Krabbe disease",
         "myelophthisic anemia",
-        "Nasu-Hakola disease",
         "subacute sclerosing panencephalitis"
-
-        // following should not match; consent-ontology service matches in their definitions; orch does not
-        // "hypomyelinating leukodystrophy 4",
-        // "mitochondrial complex I deficiency",
```
davidangb, line 51: after syncing the `ontology-unittest` index so that it is equivalent to the `ontology` index, this integration test was failing. We made changes a ways back but hadn't synced the two indexes.

rushtong (approved):
👍 Love the cleanup

PR #592, opened 2018-01-30
GAWB-3059: Show Data Use from DUOS

rushtong:
## Addresses 
* https://broadinstitute.atlassian.net/browse/GAWB-3059

## Changes
* Added the DUOS data use object to consent. 

## Testing
* I spun this up in a FiaB, added an example consent in the fiab duos ui and voted on it so it would be visible to consent. Orchestration was then able to successfully see the data use object in real time as I made updates to it in the fiab duos. 

screen shot 2018-01-30 at 6 35 51 pm

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/model/DUOS.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/model/DUOS.scala
@@ -1,8 +1,93 @@
 package org.broadinstitute.dsde.firecloud.model
 
-import spray.json.JsObject
+import spray.json.{JsObject, JsValue, _}
+import DefaultJsonProtocol._
 
 object DUOS {
+
+  case class DuosDataUse(
+    generalUse: Option[Boolean] = None,
+    hmbResearch: Option[Boolean] = None,
+    diseaseRestrictions: Option[Seq[String]] = None,
+    populationOriginsAncestry: Option[Boolean] = None,
+    populationStructure: Option[Boolean] = None,
+    commercialUse: Option[Boolean] = None,
+    methodsResearch: Option[Boolean] = None,
+    aggregateResearch: Option[String] = None,
+    controlSetOption: Option[String] = None,
+    gender: Option[String] = None,
+    pediatric: Option[Boolean] = None,
+    populationRestrictions: Option[Seq[String]] = None,
+    dateRestriction: Option[String] = None,
+    recontactingDataSubjects: Option[Boolean] = None,
+    recontactMay: Option[String] = None,
+    recontactMust: Option[String] = None,
+    genomicPhenotypicData: Option[String] = None,
+    otherRestrictions: Option[Boolean] = None,
+    cloudStorage: Option[String] = None,
+    ethicsApprovalRequired: Option[Boolean] = None,
+    geographicalRestrictions: Option[String] = None,
+    other: Option[String] = None,
+    illegalBehavior: Option[Boolean] = None,
+    addiction: Option[Boolean] = None,
+    sexualDiseases: Option[Boolean] = None,
+    stigmatizeDiseases: Option[Boolean] = None,
+    vulnerablePopulations: Option[Boolean] = None,
+    psychologicalTraits: Option[Boolean] = None,
+    nonBiomedical: Option[Boolean] = None
+  ) {
+    def apply(jsValues: Map[String, JsValue]): DuosDataUse = {
+      def getBoolean(f: String): Option[Boolean] = {
+        jsValues.get(f) match {
+          case x if x.isDefined => Some(jsValues(f).convertTo[Boolean])
+          case _ => None
+        }
+      }
+      def getSeqString(f: String): Option[Seq[String]] = {
+        jsValues.get(f) match {
+          case x if x.isDefined => Some(jsValues(f).convertTo[JsArray].elements.map(_.toString()))
+          case _ => None
+        }
+      }
+      def getString(f: String): Option[String] = {
+        jsValues.get(f) match {
+          case x if x.isDefined => Some(jsValues(f).convertTo[String])
+          case _ => None
+        }
+      }
+      DuosDataUse(
+        generalUse = getBoolean("generalUse"),
+        hmbResearch = getBoolean("hmbResearch"),
+        diseaseRestrictions = getSeqString("diseaseRestrictions"),
+        populationOriginsAncestry = getBoolean("populationOriginsAncestry"),
+        populationStructure = getBoolean("populationStructure"),
+        commercialUse = getBoolean("commercialUse"),
+        methodsResearch = getBoolean("methodsResearch"),
+        aggregateResearch = getString("aggregateResearch"),
+        controlSetOption = getString("controlSetOption"),
+        gender = getString("gender"),
+        pediatric = getBoolean("pediatric"),
+        populationRestrictions = getSeqString("populationRestrictions"),
+        dateRestriction = getString("dateRestriction"),
+        recontactingDataSubjects = getBoolean("recontactingDataSubjects"),
+        recontactMay = getString("recontactMay"),
+        recontactMust = getString("recontactMust"),
+        genomicPhenotypicData = getString("genomicPhenotypicData"),
+        otherRestrictions = getBoolean("otherRestrictions"),
+        cloudStorage = getString("cloudStorage"),
+        ethicsApprovalRequired = getBoolean("ethicsApprovalRequired"),
+        geographicalRestrictions = getString("geographicalRestrictions"),
+        other = getString("other"),
+        illegalBehavior = getBoolean("illegalBehavior"),
+        addiction = getBoolean("addiction"),
+        sexualDiseases = getBoolean("sexualDiseases"),
+        stigmatizeDiseases = getBoolean("stigmatizeDiseases"),
+        vulnerablePopulations = getBoolean("vulnerablePopulations"),
+        psychologicalTraits = getBoolean("psychologicalTraits"),
+        nonBiomedical = getBoolean("nonBiomedical")
+      )
+    }
+  }
 
   case class Consent(
     consentId: String,
@@ -14,7 +99,8 @@ object DUOS {
     dataUseLetter: Option[String] = None,
     useRestriction: Option[JsObject] = None,
     dulName: Option[String] = None,
-    translatedUseRestriction: Option[String] = None
+    translatedUseRestriction: Option[String] = None,
+    dataUse: Option[DuosDataUse] = None
   )
 
   case class ConsentError(
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/model/DUOS.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/model/DUOS.scala
@@ -3,6 +3,8 @@ package org.broadinstitute.dsde.firecloud.model
 import spray.json.{JsObject, JsValue, _}
 import DefaultJsonProtocol._
 
+import scala.util.{Success, Try}
+
 object DUOS {
 
   case class DuosDataUse(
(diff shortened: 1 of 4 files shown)


rushtong:
jenkins retest

rushtong:
jenkins retest

rushtong:
jenkins retest

src/main/scala/org/broadinstitute/dsde/firecloud/model/DUOS.scala
```diff
@@ -1,9 +1,97 @@
...
+    vulnerablePopulations: Option[Boolean] = None,
+    psychologicalTraits: Option[Boolean] = None,
+    nonBiomedical: Option[Boolean] = None
+  )
+
+  object DuosDataUse {
+    def apply(jsValues: Map[String, JsValue]): DuosDataUse = {
+      def getBoolean(f: String): Option[Boolean] = {
+        jsValues.get(f) match {
+          case x if x.isDefined => Some(jsValues(f).convertTo[Boolean])
```
rushtong: I need some better handling around the `convertTo`-s. If someone sends `"generalUse":"FOO"` instead of `"generalUse":true`, this will fail.

src/main/scala/org/broadinstitute/dsde/firecloud/model/DUOS.scala
```diff
@@ -1,9 +1,110 @@
...
+    vulnerablePopulations: Option[Boolean] = None,
+    psychologicalTraits: Option[Boolean] = None,
+    nonBiomedical: Option[Boolean] = None
+  )
+
+  object DuosDataUse {
+    def apply(jsValues: Map[String, JsValue]): DuosDataUse = {
+      def getBoolean(f: String): Option[Boolean] = {
+        jsValues.get(f) match {
+          case x if x.isDefined =>
```
kyuksel: More idiomatic way to write that line may be
```scala
case Some(_) =>
```

src/main/scala/org/broadinstitute/dsde/firecloud/model/DUOS.scala
```diff
@@ -1,9 +1,110 @@
...
+            Try(jsValues(f).convertTo[Boolean]) match {
+              case Success(booleanVal) => Some(booleanVal)
+              case _ => None
+            }
+          case _ => None
```
kyuksel: Lines 47-50 could probably reduce to
```scala
Try(jsValues(f).convertTo[Boolean]).toOption
```

rushtong:
jenkins retest

src/main/scala/org/broadinstitute/dsde/firecloud/model/DUOS.scala
```diff
@@ -1,9 +1,110 @@
...
+        }
+      }
```
kyuksel: Actually, couldn't `getBoolean()` be reduced to
```scala
def getBoolean(f: String): Option[Boolean] = 
  Try(jsValues(f).convertTo[Boolean]).toOption
```
and similarly for the other `get*()`?

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala
```diff
@@ -236,7 +236,28 @@ object ModelJsonProtocol extends WorkspaceJsonSupport {
   implicit val impLibrarySearchResponse = jsonFormat4(LibrarySearchResponse)
   implicit val impLibraryBulkIndexResponse = jsonFormat3(LibraryBulkIndexResponse)
 
-  implicit val impDuosConsent = jsonFormat10(Consent)
+  implicit object impDuosDataUse extends RootJsonFormat[DuosDataUse] {
+    override def write(ddu: DuosDataUse): JsValue = {
+      val existingProps: Seq[(String, JsValue)] = ddu.getClass.getDeclaredFields map { f =>
```
kyuksel: How would an exception that may be thrown by `getDeclaredFields` and/or `f.get()` be handled?

rushtong: In this case, it's not handled at all. Any exception would propagate up to the caller, so it should result in a 500 error when orch tries to return it in json format.

src/main/scala/org/broadinstitute/dsde/firecloud/model/DUOS.scala
```diff
@@ -1,9 +1,110 @@
 package org.broadinstitute.dsde.firecloud.model
 
-import spray.json.JsObject
+import spray.json.{JsObject, JsValue, _}
```
davidangb: just use `import spray.json._`

src/main/scala/org/broadinstitute/dsde/firecloud/model/DUOS.scala
```diff
@@ -1,9 +1,110 @@
...
+      def getString(f: String): Option[String] = {
+        jsValues.get(f) match {
+          case x if x.isDefined =>
+            Try(jsValues(f).convertTo[String]) match {
+              case Success(stringVal) => Some(stringVal)
+              case _ => None
+            }
+          case _ => None
+        }
+      }
```
davidangb: I'd look to use `JsObject`'s internal type system to handle all of these. Something like:
```
      def getBoolean(f: String): Option[Boolean] = {
        jsValues.get(f) match {
          case Some(b:JsBoolean) => Some(b.value)
          case _ => None
        }
      }
      def getSeqString(f: String): Option[Seq[String]] = {
        jsValues.get(f) match {
          case Some(a:JsArray) => Some(a.elements.collect {
            case s:JsString => s.value
          })
          case _ => None
        }
      }
      def getString(f: String): Option[String] = {
        jsValues.get(f) match {
          case Some(s:JsString) => Some(s.value)
          case _ => None
        }
      }
```

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala
```diff
@@ -9,7 +9,7 @@ import org.broadinstitute.dsde.firecloud.model.MethodRepository._
 import org.broadinstitute.dsde.firecloud.model.Ontology.{ESTermParent, TermParent, TermResource}
 import org.broadinstitute.dsde.firecloud.model.Trial.ProjectRoles.ProjectRole
 import org.broadinstitute.dsde.firecloud.model.Trial._
-import spray.json._
+import spray.json.{JsNull, _}
```
davidangb: just `import spray.json._`

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala
```diff
@@ -236,7 +236,28 @@ object ModelJsonProtocol extends WorkspaceJsonSupport {
...
+      val existingProps: Seq[(String, JsValue)] = ddu.getClass.getDeclaredFields map { f =>
+        f.setAccessible(true)
+        f.get(ddu) match {
+          case Some(x: Boolean) => f.getName -> x.toJson
+          case Some(y: String) => f.getName -> y.toJson
+          case Some(ls: Seq[_]) =>
+            ls.head match {
+              case (a: String) => f.getName -> (a +: ls.tail.map(_.asInstanceOf[String])).toJson
+              case _ => f.getName -> JsNull
+            }
```
davidangb: I think you can one-line the `Seq` case, with even more type-checking that avoids `asInstanceOf`:
```
case Some((h:String) :: t) => f.getName -> (h +: t.collect{case z:String => z}).toJson
```

src/test/scala/org/broadinstitute/dsde/firecloud/model/DuosModelSpec.scala
```diff
@@ -0,0 +1,266 @@
...
+      }
+      "diseaseRestrictions: [DOID_1, DOID_2]" in {
+        val diseases = JsArray(JsString("DOID_1"), JsString("DOID_2"))
+        val jsValues: Map[String, JsValue] = Map("diseaseRestrictions" -> diseases)
+        val duosDataUse: DuosDataUse = DuosDataUse.apply(jsValues)
+        val duosDiseases: Seq[String] = duosDataUse.diseaseRestrictions.getOrElse(Seq.empty[String])
+        duosDiseases should not be empty
+        duosDiseases should contain ("DOID_1")
+        duosDiseases should contain ("DOID_2")
+        duosDiseases.size shouldBe 2
```
davidangb: here - and in multiple other tests - can you simplify to:
```
duosDiseases should contain theSameElementsInOrderAs Seq("DOID_1", "DOID_2")
```

src/test/scala/org/broadinstitute/dsde/firecloud/model/DuosModelSpec.scala
```diff
@@ -0,0 +1,266 @@
...
+    duosDataUse.geographicalRestrictions.isDefined shouldBe false
+    duosDataUse.other.isDefined shouldBe false
+    duosDataUse.illegalBehavior.isDefined shouldBe false
+    duosDataUse.addiction.isDefined shouldBe false
+    duosDataUse.sexualDiseases.isDefined shouldBe false
+    duosDataUse.stigmatizeDiseases.isDefined shouldBe false
+    duosDataUse.vulnerablePopulations.isDefined shouldBe false
+    duosDataUse.psychologicalTraits.isDefined shouldBe false
+    duosDataUse.nonBiomedical.isDefined shouldBe false
+  }
```
davidangb: can you use the same reflection pattern here that you used in the json deserialization?
```
  private def assertIsUndefined(duosDataUse: DuosDataUse): Unit = {
    duosDataUse.getClass.getDeclaredFields map { f =>
      f.setAccessible(true)
      f.get(duosDataUse) match {
        case Some(x) => fail(s"field ${f.getName} should be None; it is $x")
        case _ => // noop
      }
    }
}
```

davidangb (commented):
code comments inline!

src/main/scala/org/broadinstitute/dsde/firecloud/model/DUOS.scala
rushtong: 👍  I think I need a combination of this and Kaan's approach.

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala
kyuksel: Wouldn't it be better to log something useful for debug purposes and/or return something friendly to the user? I'm not familiar with the use case so I can't tell.

rushtong: Yes, it would be nice to handle this better. The current `jsonFormatX` all throw `deserializationError` exceptions along with `serializationError` exceptions - I'll add that.

src/main/scala/org/broadinstitute/dsde/firecloud/model/DUOS.scala
```diff
@@ -43,31 +43,19 @@ object DUOS {
     def apply(jsValues: Map[String, JsValue]): DuosDataUse = {
       def getBoolean(f: String): Option[Boolean] = {
         jsValues.get(f) match {
-          case x if x.isDefined =>
-            Try(jsValues(f).convertTo[Boolean]) match {
-              case Success(booleanVal) => Some(booleanVal)
-              case _ => None
-            }
+          case Some(b: JsBoolean) => Try(b.convertTo[Boolean]).toOption
```
davidangb: I really don't think you need the `Try` here. If it matches as a `JsBoolean`, it will contain a Boolean inside `b.value`.

src/main/scala/org/broadinstitute/dsde/firecloud/model/DUOS.scala
```diff
@@ -43,31 +43,19 @@ object DUOS {
...
         }
       }
       def getString(f: String): Option[String] = {
         jsValues.get(f) match {
-          case x if x.isDefined =>
-            Try(jsValues(f).convertTo[String]) match {
-              case Success(stringVal) => Some(stringVal)
-              case _ => None
-            }
+          case Some(s: JsString) => Try(s.convertTo[String]).toOption
```
davidangb: same as the JsBoolean - if it matches as a JsString, it will contain a String, you don't need the Try.

src/main/scala/org/broadinstitute/dsde/firecloud/model/DUOS.scala
```diff
@@ -43,31 +43,19 @@ object DUOS {
...
         }
       }
       def getSeqString(f: String): Option[Seq[String]] = {
         jsValues.get(f) match {
-          case x if x.isDefined =>
-            Try(jsValues(f).convertTo[JsArray].elements.map(_.convertTo[String])) match {
-              case Success(listVal) => Some(listVal)
-              case _ => None
-            }
+          case Some(l: JsArray) => Try(l.elements.map(_.convertTo[String])).toOption
```
davidangb: I still like using match/case/collect here to look for Strings and skip other types, instead of Try-ing to convert to a string and using exceptions to handle non-strings. But I don't feel too strongly about this and won't hold the PR for that.

davidangb (commented):
few more comments in the same spot ... message me if you want to chat about them!

src/main/scala/org/broadinstitute/dsde/firecloud/model/DUOS.scala
rushtong: Ah - we are doing that in the serialization, I didn't port that over to the deserialization.

davidangb (approved):
👍 assuming tests pass!

kyuksel (approved):

rushtong:
jenkins retest

rushtong:
jenkins retest

rushtong:
jenkins retest

rushtong:
jenkins retest

PR #593, opened 2018-02-01
GAWB-3137: Fix StartupChecksSpec

rushtong:
## Addresses:
* https://broadinstitute.atlassian.net/browse/GAWB-3137

## Changes:
* Moved internal state of the mock dao to an actor. When inside the class, it did not return consistent results. Inside of an actor, the results are consistent.
* Updated logic for the test that threw an error.
* Updated `-` -> `in` so the tests would be run more consistently.

---

In all cases:


rushtong:
jenkins retest

aednichols (approved):
👍

I can intuitively believe that there's a concurrency bug here, but I can't put my finger on where. But since this fixes it, the results speak for themselves...

I wonder if it would be good practice to do `system.shutdown` in a cleanup method.

davidangb (approved):
:+1: good job on this, including fixing the semantics of that "should pass when all SAs need to be registered, and succeed" test!

PR #595, opened 2018-02-05
GAWB-3144: Update tests that require mutable internal state

rushtong:
## Addresses
* https://broadinstitute.atlassian.net/browse/GAWB-3144

## Notes
Certain unit tests depended upon changing internal state of application DAOs. Due to tests being running in parallel, this introduces concurrency problems. This PR moves those dependent tests to a different structure where a new app is spun up for each test instead of spinning up the app once and running all tests inside of the single instance.

This PR makes the most minimal set of changes to avoid the pattern of relying on internal mutable DAO state. I am open to the idea of expanding this to a larger suite of tests.

---

In all cases:


project/plugins.sbt
```diff
@@ -6,3 +6,5 @@ addSbtPlugin("org.scoverage" % "sbt-coveralls" % "1.0.0")
 
 // latest coveralls (1.0.0) is *only* compatible with scoverage 1.0.4 or less: https://github.com/scoverage/sbt-coveralls/issues/49
 addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.0.4")
+
+addSbtPlugin("com.github.tkawachi" % "sbt-repeat" % "0.1.0")
```
rushtong, line 10: Let's you run multiple commands N times:
```
sbt "repeat 10 test"
sbt "repeat 10 test-only *LibraryApiServiceSpec"
```

project/Testing.scala
```diff
@@ -18,7 +18,7 @@ object Testing {
     javaOptions in Test += "-Des.set.netty.runtime.available.processors=false",
 
     fork in Test := true,
-    parallelExecution in Test := false
+    parallelExecution in Test := true
```
aednichols, line 21: Is this going in or is it just for testing? It seems like we're signing up for "hard mode" when we're already a bit swamped... but if it speeds things up a lot, probably worth it

rushtong: I intended this to go in ... but am open to other opinions. My intention here was to fail faster due to parallelism problems. We are already failing this way in jenkins - this makes it more clear during development. It might also be worth adding some kind of dev note about running tests multiple times. Maybe it's overkill?

aednichols: OK yeah, this seems right then. Jenkins is the system of record, so it makes sense for local to replicate it as closely as possible.

src/test/scala/org/broadinstitute/dsde/firecloud/service/LibraryApiServiceSpec.scala
```diff
@@ -370,3 +299,96 @@ class LibraryApiServiceSpec extends BaseServiceSpec with LibraryApiService with
...
+/**
+  * This class represents a version of the application with Library APIs that can run a single request.
+  */
+class MockLibraryApiServiceApp(val request: HttpRequest) extends MockApplication with LibraryApiService {
+  val libraryServiceConstructor: (UserInfo) => LibraryService = LibraryService.constructor(app)
+  val ontologyServiceConstructor: () => OntologyService = OntologyService.constructor(app)
+  def checkRequest: (HttpResponse, StatusCode) = {
+    request ~> dummyUserIdHeaders("1234") ~> sealRoute(libraryRoutes) ~> check {(response, status)}
+  }
+}
```
davidangb, line 312: each `MockLibraryApiServiceApp` reuses the same `app`, which means you're reusing the daos associated with that app. Is that really what you intended to do? It seems that you intended to create a new app/dao for each test.

src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
```diff
@@ -1031,3 +968,93 @@ class WorkspaceApiServiceSpec extends BaseServiceSpec with WorkspaceApiService w
...
+/**
+  * This class represents a version of the application with Workspace APIs that can run a single request.
+  */
+class MockWorkspaceApiServiceApp(val request: HttpRequest) extends MockApplication with WorkspaceApiService {
+  val workspaceServiceConstructor: (WithAccessToken) => WorkspaceService = WorkspaceService.constructor(app)
+  val permissionReportServiceConstructor: (UserInfo) => PermissionReportService = PermissionReportService.constructor(app)
+  def checkRequest: (HttpResponse, StatusCode) = {
+    request ~> dummyUserIdHeaders("1234") ~> sealRoute(workspaceRoutes) ~> check {(response, status)}
+  }
+}
```
davidangb, line 981: same here, I don't think you're creating new daos for each test.

davidangb (commented):
concerns about reusing daos inline.

src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
```diff
@@ -1031,3 +968,93 @@ class WorkspaceApiServiceSpec extends BaseServiceSpec with WorkspaceApiService w
...
+        val request = Patch(updateAttributesPath,
+          HttpEntity(MediaTypes.`application/json`, """[
+                                                      |  {
+                                                      |    "op": "AddUpdateAttribute",
+                                                      |    "attributeName": "library:dataCategory",
+                                                      |    "addUpdateAttribute": "test-attribute-value"
+                                                      |  }
+                                                      |]""".stripMargin))
+        val app = new MockWorkspaceApiServiceApp(request)
+        val (response, status) = app.checkRequest
```
aednichols, line 1006: Highly optional:

You can eliminate the unused val annotation by doing
```
val (_, status) = app.checkRequest
```
(Or make a `checkOK`.)

src/test/scala/org/broadinstitute/dsde/firecloud/service/LibraryApiServiceSpec.scala
davidangb: Update: I was wrong! Ignore this.

src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
davidangb: Update: I was wrong! Ignore this.

src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
rushtong: Yes, I was thinking about different things to return here and kept the two implementations the same for no really good reason. Some tests want to know about the response, some might need the status, and others still might need just the entity, even though I didn't add that in here. I'm leaning toward leaving the details of `checkRequest` up to the individual mock.

rushtong:
jenkins retest

rushtong:
@davidangb , @aednichols - The latest jenkins run has a failure in one of my stateful tests, so I am now sure that this approach doesn't really solve the problem. Closing this PR and will re-open if needed. 

https://fc-jenkins.dsp-techops.broadinstitute.org/view/Build/job/firecloud-orchestration-build/528/consoleFull

PR #596, opened 2018-02-05
GAWB-3060: index data use restriction pulled from Consent (from ORSP)

davidangb:
I still need to discuss testing strategy with QA (Andrew), but possibly this coverage is enough.

A bunch of lines of code changed in minimal ways, because I needed to pass the consent DAO around.

To test in the dev env, you can use ORSP-590 or ORSP-640.

You probably don't want to review commit-by-commit because the first 1 or 2 commits were heavily refactored.

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
@@ -119,7 +119,21 @@ trait DataUseRestrictionSupport extends LazyLogging {
       case _ => Map.empty[AttributeName,Attribute]
     }
 
-    val result = Map.empty[AttributeName, Attribute] ++ gru ++ hmb ++ ncu ++ nmds ++ rspd ++ nagr ++ nctrl ++ ds ++ rspop // ++ npu ++ irb
+    // gender restrictions: RS-G, RS-FM, RS-M
+    val rsg = duosDataUse.gender match {
+      case Some(f:String) if "female".equals(f.toLowerCase) => Map(
+        AttributeName.withLibraryNS("RS-G") -> AttributeBoolean(true),
+        AttributeName.withLibraryNS("RS-FM") -> AttributeBoolean(true)
+      )
+      case Some(m:String) if "male".equals(m.toLowerCase) => Map(
+        AttributeName.withLibraryNS("RS-G") -> AttributeBoolean(true),
+        AttributeName.withLibraryNS("RS-M") -> AttributeBoolean(true)
+      )
+      case _ => Map.empty[AttributeName,Attribute]
+    }
+
+    val result = Map.empty[AttributeName, Attribute] ++ gru ++ hmb ++ ncu ++ nmds ++
+      rspd ++ nagr ++ nctrl ++ ds ++ rspop ++ rsg // ++ npu ++ irb
 
     logger.debug("inbound DuosDataUse: " + duosDataUse)
     logger.debug("outbound attrs: " + result)
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
@@ -94,6 +94,7 @@ trait DataUseRestrictionSupport extends LazyLogging {
     // DUOS has no concept of "IRB review required"
     // val irb = None
 
+    // disease restrictions: DS, DS_URL
     val ds = duosDataUse.diseaseRestrictions match {
       case Some(Seq()) => Map.empty[AttributeName,Attribute]
       case Some(nodeList) =>
@@ -108,7 +109,17 @@ trait DataUseRestrictionSupport extends LazyLogging {
       case _ => Map.empty[AttributeName,Attribute]
     }
 
-    val result = Map.empty[AttributeName, Attribute] ++ gru ++ hmb ++ ncu ++ nmds ++ rspd ++ nagr ++ nctrl ++ ds // ++ npu ++ irb
+    // population restrictions: RS-POP
+    val rspop = duosDataUse.populationRestrictions match {
+      case Some(Seq()) => Map.empty[AttributeName,Attribute]
+      case Some(populations) =>
+        Map(
+          AttributeName.withLibraryNS("RS-PD") -> AttributeValueList(populations.map(AttributeString))
+        )
+      case _ => Map.empty[AttributeName,Attribute]
+    }
+
+    val result = Map.empty[AttributeName, Attribute] ++ gru ++ hmb ++ ncu ++ nmds ++ rspd ++ nagr ++ nctrl ++ ds ++ rspop // ++ npu ++ irb
 
     logger.debug("inbound DuosDataUse: " + duosDataUse)
     logger.debug("outbound attrs: " + result)
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
@@ -148,10 +148,10 @@ trait DataUseRestrictionSupport extends LazyLogging {
     (existing -
       structuredUseRestrictionAttributeName -
       consentCodesAttributeName --
-      allDurFieldNames.map(x => AttributeName.withLibraryNS(x))) ++ preferred
+      allDurFieldNames.map(AttributeName.withLibraryNS)) ++ preferred
   }
 
-  // TODO: namespaces on allDurFieldNames??!?!?
+  // TODO: this method needs to respect attribute namespaces: see GAWB-3173
   private def getDataUseAttributes(workspace: Workspace): Map[AttributeName, Attribute] = {
 
     // Find all library attributes that contribute to data use restrictions
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
@@ -1,7 +1,10 @@
 package org.broadinstitute.dsde.firecloud.service
 
 import com.typesafe.scalalogging.LazyLogging
+import org.broadinstitute.dsde.firecloud.dataaccess.OntologyDAO
+import org.broadinstitute.dsde.firecloud.model.DUOS.DuosDataUse
 import org.broadinstitute.dsde.firecloud.model.DataUse.DiseaseOntologyNodeId
+import org.broadinstitute.dsde.rawls.model.Attributable.AttributeMap
 import org.broadinstitute.dsde.rawls.model.WorkspaceJsonSupport.AttributeNameFormat
 import org.broadinstitute.dsde.rawls.model._
 import spray.json.DefaultJsonProtocol._
@@ -59,6 +62,61 @@ trait DataUseRestrictionSupport extends LazyLogging {
   }
 
   /**
+    * Looks at a DuosDataUse object (likely previously retrieved from ORSP) and translates it into a
+    * set of structured Data Use Restriction attributes.
+    *
+    * @param duosDataUse
+    * @return
+    */
+  def generateStructuredUseRestrictionAttribute(duosDataUse: DuosDataUse, ontologyDAO: OntologyDAO): Map[AttributeName, Attribute] = {
+    // these are straightforward mappings
+    val gru = duosDataUse.generalUse.map ( AttributeName.withLibraryNS("GRU") -> AttributeBoolean(_) )
+    val hmb = duosDataUse.hmbResearch.map ( AttributeName.withLibraryNS("HMB") -> AttributeBoolean(_) )
+    val rspd = duosDataUse.vulnerablePopulations.map ( AttributeName.withLibraryNS("RS-PD") -> AttributeBoolean(_) )
+    val ncu = duosDataUse.commercialUse.map ( AttributeName.withLibraryNS("NCU") -> AttributeBoolean(_) )
+    val nmds = duosDataUse.methodsResearch.map ( AttributeName.withLibraryNS("NMDS") -> AttributeBoolean(_) )
+
+    // FC supports both NCU and NPU; DUOS only supports one (NCU)
+    // val npu = duosDataUse.commercialUse.map ( AttributeName.withLibraryNS("NPU") -> AttributeBoolean(_) )
+
+    // DUOS represents NAGR and NCTRL as strings ("yes"|"no")
+    val nagr = duosDataUse.aggregateResearch match {
+      case Some(s) if "yes".equals(s.toLowerCase) => Some(AttributeName.withLibraryNS("NAGR") -> AttributeBoolean(true))
+      case Some(s) if "no".equals(s.toLowerCase) => Some(AttributeName.withLibraryNS("NAGR") -> AttributeBoolean(false))
+      case _ => None
+    }
+    val nctrl = duosDataUse.controlSetOption match {
+      case Some(s) if "yes".equals(s.toLowerCase) => Some(AttributeName.withLibraryNS("NCTRL") -> AttributeBoolean(true))
+      case Some(s) if "no".equals(s.toLowerCase) => Some(AttributeName.withLibraryNS("NCTRL") -> AttributeBoolean(false))
+      case _ => None
+    }
+
+    // DUOS has no concept of "IRB review required"
+    // val irb = None
+
+    val ds = duosDataUse.diseaseRestrictions match {
+      case Some(Seq()) => Map.empty[AttributeName,Attribute]
+      case Some(nodeList) =>
+        Map(
+          AttributeName.withLibraryNS("DS_URL") -> AttributeValueList(nodeList.map( AttributeString )),
+          AttributeName.withLibraryNS("DS") -> AttributeValueList(nodeList.map{ nodeid =>
+            ontologyDAO.search(nodeid) match {
+              case termResource :: Nil => AttributeString(termResource.label)
+              case _ => AttributeString(nodeid)
+            }})
+        )
+      case _ => Map.empty[AttributeName,Attribute]
+    }
+
+    val result = Map.empty[AttributeName, Attribute] ++ gru ++ hmb ++ ncu ++ nmds ++ rspd ++ nagr ++ nctrl ++ ds // ++ npu ++ irb
+
+    logger.debug("inbound DuosDataUse: " + duosDataUse)
+    logger.debug("outbound attrs: " + result)
+
+    result
+  }
+
+  /**
     * Create a display-friendly version of the structured data use restriction in the form of a
     * list of code strings.
     *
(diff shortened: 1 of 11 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -46,20 +47,22 @@ object LibraryService {
...
-                      val ontologyDAO: OntologyDAO)
+                      val ontologyDAO: OntologyDAO,
+                      val consentDAO: ConsentDAO)
                      (implicit protected val executionContext: ExecutionContext) extends Actor
   with LibraryServiceSupport with AttributeSupport with PermissionsSupport with SprayJsonSupport with LazyLogging with WorkspacePublishingSupport {
 
   lazy val log = LoggerFactory.getLogger(getClass)
 
-  implicit val userInfo = argUserInfo
+  implicit val userToken = argUserInfo
```
davidangb, line 64: I changed this so LibraryService, LibraryServiceSupport, WorkspaceService, WorkspacePublishingSupport all use the same name.

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -58,6 +61,61 @@ trait DataUseRestrictionSupport extends LazyLogging {
...
+    val ds = duosDataUse.diseaseRestrictions match {
+      case Some(Seq()) => Map.empty[AttributeName,Attribute]
+      case Some(nodeList) =>
+        Map(
+          AttributeName.withLibraryNS("DS_URL") -> AttributeValueList(nodeList.map( AttributeString )),
+          AttributeName.withLibraryNS("DS") -> AttributeValueList(nodeList.map{ nodeid =>
+            ontologyDAO.search(nodeid) match {
+              case termResource :: Nil => AttributeString(termResource.label)
+              case _ => AttributeString(nodeid)
+            }})
```
davidangb, line 107: this call to `ontologyDAO.search` could be optimized - in a full Library reindex, if there are lots of datasets whose ORSP-based restrictions use the same ontology nodes, we are making duplicate requests. But that's more refactoring that I want to do in this PR.

davidangb:
automation had 1 failure:
```
[info] PublishSpec:
[info] As a user with no TCGA permissions
[info]   a published workspace
[info]     with TCGA control access
[info]     - should see a information message about getting TCGA access *** FAILED ***
[info]       org.broadinstitute.dsde.workbench.service.RestException: {
[info]   "statusCode": 403,
[info]   "source": "rawls",
[info]   "timestamp": 1517871970947,
[info]   "causes": [],
[info]   "stackTrace": [],
[info]   "message": "In order to use a group in an Authorization Domain, you must be a member of that group."
[info] }
```
https://fc-jenkins.dsp-techops.broadinstitute.org/job/fiab-test/2531/consoleFull

davidangb:
jenkins retest

davidangb:
Second run also had one failure, but different:
```
[info] MethodConfigSpec:
[info] For a method config that references a redacted method
[info] - should have warning icon in config list *** FAILED ***
[info]   org.openqa.selenium.TimeoutException: Expected condition failed: waiting for org.broadinstitute.dsde.workbench.service.test.WebBrowserUtil$await$$anon$1@53f8ee (tried for 30 second(s) with 500 MILLISECONDS interval)
```
https://fc-jenkins.dsp-techops.broadinstitute.org/job/fiab-test/2543/consoleFull

davidangb:
( permission from Nicole to merge, despite test failures )

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
```diff
@@ -26,35 +28,56 @@ import scala.util.{Failure, Success, Try}
...
 
+    // identify the unique ORSP codes in this list of workspaces
+    val orspIds = uniqueStrings(workspaces, orspIdAttribute)
+
+    // set up an exception-resilient Future to get the ORSP restrictions for those codes
+    val futureRestrictions:Future[Set[(String, Option[DuosDataUse])]] = Future.sequence(orspIds map {orspId =>
+      consentDAO.getRestriction(orspId) map { restriction =>
+        orspId -> restriction
+      } recover {
+        case e:Exception =>
```
aednichols, line 55: (Optional) Maybe a `Try` would be more Scala-idiomatic here?

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
```diff
@@ -96,6 +119,21 @@ trait LibraryServiceSupport extends DataUseRestrictionSupport with LazyLogging {
     }
   }
 
+  def uniqueStrings(workspaces: Seq[Workspace], attributeName: AttributeName): Set[String] = {
```
aednichols: Optional, maybe this function signature could be more specific in name and types? First attempt,
```
def uniqueValuesForAttributeInWorkspaces(attributeName: AttributeName, workspaces: Seq[Workspace]): Set[AttributeString]
```

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -85,6 +143,15 @@ trait DataUseRestrictionSupport extends LazyLogging {
       Map.empty[AttributeName, Attribute]
   }
 
+  def replaceDataUseAttributes(existing: AttributeMap, preferred: AttributeMap): AttributeMap = {
+    // delete pre-existing DU codes, then add the DU codes from ORSP
+    (existing -
+      structuredUseRestrictionAttributeName -
+      consentCodesAttributeName --
+      allDurFieldNames.map(x => AttributeName.withLibraryNS(x))) ++ preferred
```
aednichols: (Optional) This can be
```
AttributeName.withLibraryNS(_)
```
or even just
```
AttributeName.withLibraryNS
```
Concision vs. clarity tradeoff up to you.

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -58,6 +61,61 @@ trait DataUseRestrictionSupport extends LazyLogging {
...
+    // val irb = None
+
+    val ds = duosDataUse.diseaseRestrictions match {
+      case Some(Seq()) => Map.empty[AttributeName,Attribute]
+      case Some(nodeList) =>
+        Map(
+          AttributeName.withLibraryNS("DS_URL") -> AttributeValueList(nodeList.map( AttributeString )),
+          AttributeName.withLibraryNS("DS") -> AttributeValueList(nodeList.map{ nodeid =>
+            ontologyDAO.search(nodeid) match {
+              case termResource :: Nil => AttributeString(termResource.label)
```
aednichols, line 105: Not specific to this PR, but I'm confused about why `search` returns a `List`.

Looking at the ES DAO, it can only be an empty list or a list with exactly one item.

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -85,6 +143,15 @@ trait DataUseRestrictionSupport extends LazyLogging {
...
 
+  def replaceDataUseAttributes(existing: AttributeMap, preferred: AttributeMap): AttributeMap = {
+    // delete pre-existing DU codes, then add the DU codes from ORSP
+    (existing -
+      structuredUseRestrictionAttributeName -
+      consentCodesAttributeName --
+      allDurFieldNames.map(x => AttributeName.withLibraryNS(x))) ++ preferred
+  }
+
+  // TODO: namespaces on allDurFieldNames??!?!?
```
aednichols: Address or remove TODO?

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -58,6 +61,61 @@ trait DataUseRestrictionSupport extends LazyLogging {
...
+    val rspd = duosDataUse.vulnerablePopulations.map ( AttributeName.withLibraryNS("RS-PD") -> AttributeBoolean(_) )
+    val ncu = duosDataUse.commercialUse.map ( AttributeName.withLibraryNS("NCU") -> AttributeBoolean(_) )
+    val nmds = duosDataUse.methodsResearch.map ( AttributeName.withLibraryNS("NMDS") -> AttributeBoolean(_) )
+
+    // FC supports both NCU and NPU; DUOS only supports one (NCU)
+    // val npu = duosDataUse.commercialUse.map ( AttributeName.withLibraryNS("NPU") -> AttributeBoolean(_) )
+
+    // DUOS represents NAGR and NCTRL as strings ("yes"|"no")
+    val nagr = duosDataUse.aggregateResearch match {
+      case Some(s) if "yes".equals(s.toLowerCase) => Some(AttributeName.withLibraryNS("NAGR") -> AttributeBoolean(true))
```
aednichols, line 84: (Optional) It might be nice to make the string literals in this function into constants, grouped into objects for FC and DUOS.

Motivating example: I feel like we could at some point make a mistake between `"yes"`/`"no"` (DUOS) and `"Yes"`/`"No"` (FC)

src/test/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupportSpec.scala
```diff
@@ -182,6 +184,275 @@ class DataUseRestrictionSupportSpec extends FreeSpec with Matchers with DataUseR
...
+          )
+          assertResult(expected) {attrs}
+        }
+        "should handle empty seq of disease ontology nodes" in {
+          val attrs = generateStructuredUseRestrictionAttribute(new DuosDataUse(
+            diseaseRestrictions = Some(Seq.empty[String])
+          ), ontologyDAO)
+          assert(attrs.isEmpty)
+        }
+        "should ignore the DUOS keys that FireCloud doesn't implement" in {
```
aednichols: Good test, also a handy reference!

aednichols (approved):
Looks OK to me, code seems to do the right thing. I reviewed at the Scala level, not so much subject matter expert (you're up, Greg!)

Love the tests, made reasoning about the functionality easier.

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -58,6 +61,61 @@ trait DataUseRestrictionSupport extends LazyLogging {
...
+  def generateStructuredUseRestrictionAttribute(duosDataUse: DuosDataUse, ontologyDAO: OntologyDAO): Map[AttributeName, Attribute] = {
+    // these are straightforward mappings
+    val gru = duosDataUse.generalUse.map ( AttributeName.withLibraryNS("GRU") -> AttributeBoolean(_) )
+    val hmb = duosDataUse.hmbResearch.map ( AttributeName.withLibraryNS("HMB") -> AttributeBoolean(_) )
+    val rspd = duosDataUse.vulnerablePopulations.map ( AttributeName.withLibraryNS("RS-PD") -> AttributeBoolean(_) )
+    val ncu = duosDataUse.commercialUse.map ( AttributeName.withLibraryNS("NCU") -> AttributeBoolean(_) )
+    val nmds = duosDataUse.methodsResearch.map ( AttributeName.withLibraryNS("NMDS") -> AttributeBoolean(_) )
+
+    // FC supports both NCU and NPU; DUOS only supports one (NCU)
+    // val npu = duosDataUse.commercialUse.map ( AttributeName.withLibraryNS("NPU") -> AttributeBoolean(_) )
```
rushtong, line 80: Support for both, as well as the "IRB review required" issue below, was added to FC, but not in ORSP/DUOS. We should probably create a story for adding support across the products for the newer ones.

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -58,6 +61,61 @@ trait DataUseRestrictionSupport extends LazyLogging {
...
+    * set of structured Data Use Restriction attributes.
+    *
+    * @param duosDataUse
+    * @return
+    */
+  def generateStructuredUseRestrictionAttribute(duosDataUse: DuosDataUse, ontologyDAO: OntologyDAO): Map[AttributeName, Attribute] = {
+    // these are straightforward mappings
+    val gru = duosDataUse.generalUse.map ( AttributeName.withLibraryNS("GRU") -> AttributeBoolean(_) )
+    val hmb = duosDataUse.hmbResearch.map ( AttributeName.withLibraryNS("HMB") -> AttributeBoolean(_) )
+    val rspd = duosDataUse.vulnerablePopulations.map ( AttributeName.withLibraryNS("RS-PD") -> AttributeBoolean(_) )
```
rushtong: I believe this should be looking at the pediatric value:
```
val rspd = duosDataUse.pediatric.map ( AttributeName.withLibraryNS("RS-PD") -> AttributeBoolean(_) )
```

The `vulnerablePopulations` field is something that is asked as a researcher in DUOS: 
```
3.1.8 Does the study target a vulnerable population as defined in 456 CFR (children, prisoners, pregnant women, mentally disabled persons, or [SIGNIFICANTLY] economically or educationally disadvantaged persons)?
```
So, I don't think that needs to be translated here.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
rushtong: I think using recover around a future is more actor-idiomatic, but I might be misunderstanding exactly how you're envisioning the `Try`.

src/test/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupportSpec.scala
```diff
@@ -182,6 +184,275 @@ class DataUseRestrictionSupportSpec extends FreeSpec with Matchers with DataUseR
...
+        "should return no attributes if DUOS is empty" in {
+          val attrs = generateStructuredUseRestrictionAttribute(new DuosDataUse, ontologyDAO)
+          assert(attrs.isEmpty)
+        }
+        "should translate boolean TRUEs" in {
+          val attrs = generateStructuredUseRestrictionAttribute(new DuosDataUse(
+            generalUse = Some(true),
+            hmbResearch = Some(true),
+            commercialUse = Some(true),
+            vulnerablePopulations = Some(true),
```
rushtong: See comments above about this field - I don't think we should be using it to model the pediatric case. Similar to other usages in this test.

src/test/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupportSpec.scala
```diff
@@ -182,6 +184,275 @@ class DataUseRestrictionSupportSpec extends FreeSpec with Matchers with DataUseR
...
+          // The properties that are commented out are the ones that FireCloud implements.
+          val attrs = generateStructuredUseRestrictionAttribute(new DuosDataUse(
+//            generalUse = None,
+//            hmbResearch = None,
+//            diseaseRestrictions = None,
+//            commercialUse = None,
+//            methodsResearch = None,
+//            aggregateResearch = None,
+//            controlSetOption = None,
+//            vulnerablePopulations = None,
```
rushtong: See above :-)

src/test/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupportSpec.scala
```diff
@@ -182,6 +184,275 @@ class DataUseRestrictionSupportSpec extends FreeSpec with Matchers with DataUseR
...
+            populationOriginsAncestry = Some(true),
+            populationStructure = Some(true),
+            gender = Some("Female"),
```
rushtong: We have an `RS-G` library attribute that this should populate.

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -58,6 +61,61 @@ trait DataUseRestrictionSupport extends LazyLogging {
...
+          AttributeName.withLibraryNS("DS") -> AttributeValueList(nodeList.map{ nodeid =>
+            ontologyDAO.search(nodeid) match {
+              case termResource :: Nil => AttributeString(termResource.label)
+              case _ => AttributeString(nodeid)
+            }})
+        )
+      case _ => Map.empty[AttributeName,Attribute]
+    }
+
+    val result = Map.empty[AttributeName, Attribute] ++ gru ++ hmb ++ ncu ++ nmds ++ rspd ++ nagr ++ nctrl ++ ds // ++ npu ++ irb
```
rushtong: I think we need to capture gender, `RS-G` as one of the things we capture from the `DuoDataUse` object.
Additionally, we might need to capture populations as `RS-POP`.

src/test/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupportSpec.scala
```diff
@@ -182,6 +184,275 @@ class DataUseRestrictionSupportSpec extends FreeSpec with Matchers with DataUseR
...
+            pediatric = Some(true),
+            populationRestrictions = Some(Seq("one","two")),
```
rushtong: I think this is what should trigger the `RS-POP` library attribute.

rushtong (changes requested):
Looks good, but I think we need to change how we're translating gender, and if we need to translate population restrictions.

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
davidangb: Good call - and the reason is just legacy. In the past, the implementation *could* return a list, and callers typically just used the head of it, sometimes threw an error if it returned >1. We changed the underlying implementation, but never changed the method signatures. That would be a good cleanup (outside of this PR, I propose): GAWB-3176

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
davidangb: Ah yes, I forgot about that! This is something I noticed in pre-existing code; I will file a ticket: GAWB-3173.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
aednichols: Um... I tried to come up with a working example just now but couldn't - never mind.

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
davidangb: filed GAWB-3177

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -58,6 +61,86 @@ trait DataUseRestrictionSupport extends LazyLogging {
...
+        AttributeName.withLibraryNS("RS-M") -> AttributeBoolean(true)
+      )
+      case _ => Map.empty[AttributeName,Attribute]
+    }
+
+    val result = Map.empty[AttributeName, Attribute] ++ gru ++ hmb ++ ncu ++ nmds ++
+      rspd ++ nagr ++ nctrl ++ ds ++ rspop ++ rsg // ++ npu ++ irb
+
+    logger.debug("inbound DuosDataUse: " + duosDataUse)
+    logger.debug("outbound attrs: " + result)
```
rushtong, line 139: Do you still want these (here and elsewhere)? I'm fine if we keep them around for some period of time (they will be super helpful for debugging), but maybe a TODO to eliminate in X months?

rushtong (approved):
👍  Looks great!

aednichols (approved):
`(constantly 👍)`

PR #607, opened 2018-02-26
Add allRepos support to methodconfigs route [GAWB-3100]

aednichols:
~Needs tests, Swagger, workspace model update.~

~Not necessary for https://github.com/broadinstitute/rawls/pull/854 to merge.~

~Existing tests pass, currently trying to find a reasonable way to test that query parameters are forwarded for a particular route (we have had bugs with this in the past so it seems worthwhile)~

Existing tests pass, added test for the new behavior.

----

In all cases:


project/Dependencies.scala
```diff
@@ -21,20 +21,21 @@ object Dependencies {
...
-    "io.spray"                      %% "spray-client"        % sprayV,
-    "io.spray"                      %% "spray-testkit"       % sprayV    % "test",
-    "com.typesafe.akka"             %% "akka-actor"          % akkaV,
-    "com.typesafe.akka"             %% "akka-testkit"        % akkaV     % "test",
-    "com.typesafe.akka"             %% "akka-slf4j"          % akkaV,
-    "com.typesafe.akka"             %% "akka-stream"         % akkaV,
+    "io.spray"                      %% "spray-can"                 % sprayV,
+    "io.spray"                      %% "spray-json"                % "1.3.3",
+    "io.spray"                      %% "spray-client"              % sprayV,
+    "io.spray"                      %% "spray-routing-shapeless23" % sprayV,
```
aednichols, line 33: What went down here was, the new Rawls model introduced libraries that rely on Shapeless 2.3.2, so I upgraded us from `spray-routing` to `spray-routing-shapeless23` to match.

At that point, `vault-common` still pulled in `spray-routing` and by extension Shapeless 1.2.4, so I just threw an exclusion on that.

src/main/scala/org/broadinstitute/dsde/firecloud/model/MethodRepository.scala
```diff
@@ -50,10 +50,15 @@ object MethodRepository {
     def toShortString: String = s"Method($namespace,$name,$snapshotId)"
   }
 
+  case class AgoraConfigurationShort(
```
aednichols, line 53: This file is full of stuff that's vaguely copy-pasta from Rawls

I'm not thrilled about adding to it, but I found that replacing usages of `MethodConfigurationShort` with `AgoraConfigurationShort` is a lot cleaner than doing

```
config.methodRepoMethod match {
  case agoraMethod: AgoraMethod => [...]
  case _ => throw Exception(...)
}
```
in a bunch of places. I'm not married to this solution though, if someone can find a nicer way to encode this information in the type system I'm glad to hear it.

src/main/scala/org/broadinstitute/dsde/firecloud/model/MethodRepository.scala
```diff
@@ -50,10 +50,15 @@ object MethodRepository {
...
+    name: String,
+    rootEntityType: String,
+    methodRepoMethod: AgoraMethod,
+    namespace: String)
+
   object Method {
```
aednichols, line 59: This thing should probably be called `AgoraMethod` now, or be replaced wholesale by `AgoraEntity` from Rawls... along with its sibling `Configuration`

I'm going for maintenance-adequate with this PR though, not an opportunistic refactoring

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/RawlsDAO.scala
```diff
@@ -89,7 +90,7 @@ trait RawlsDAO extends LazyLogging with ReportsSubsystemStatus {
 
   def patchCatalog(workspaceNamespace: String, workspaceName: String, updates: Seq[WorkspaceCatalog])(implicit userToken: WithAccessToken): Future[WorkspaceCatalogUpdateResponseList]
 
-  def getMethodConfigs(workspaceNamespace: String, workspaceName: String)(implicit userToken: WithAccessToken): Future[Seq[MethodConfigurationShort]]
+  def getAgoraMethodConfigs(workspaceNamespace: String, workspaceName: String)(implicit userToken: WithAccessToken): Future[Seq[AgoraConfigurationShort]]
```
aednichols, line 93: I chose `Future[Seq[AgoraConfigurationShort]]` as the return type based on the following:

If we ask Rawls for Agora configs, everything we get back should look like an Agora config. If that's not the case, it's better for the code to blow up immediately during JSON unmarshalling instead of potentially letting instances of the wrong thing float around in memory (and trying to put checks in a bunch of places instead of one)

src/main/resources/swagger/api-docs.yaml
```diff
@@ -2840,17 +2840,48 @@ paths:
...
+          type: boolean
+          default: false
       responses:
         200:
           description: List of Method Configurations
         404:
           description: Workspace does not exist
         500:
           description: Internal Server Error
+      desciption: |
```
davidangb: typo: 'desciption' is missing the letter 'r' ... and makes this entire text disappear from swagger ui.

project/Dependencies.scala
davidangb: another option: we only use `ServerInitializer` and `ImplicitMagnet` from vault-common. We could pretty easily copy those into the orch codebase and remove the dependency on vault-common entirely. Definitely not required for this PR and probably more than you wanted to deal with - mentioning it for the sake of discussion.

davidangb (approved):
one swagger typo that must get fixed; otherwise :+1:!

rushtong:
FYI, the yaml validates against both validators (as used in DSPTO-246):
* https://codebeautify.org/yaml-validator
* https://editor.swagger.io/
Nice!

src/main/scala/org/broadinstitute/dsde/firecloud/model/MethodRepository.scala
rushtong: 👍

src/main/scala/org/broadinstitute/dsde/firecloud/model/MethodRepository.scala
```diff
@@ -81,7 +86,7 @@ object MethodRepository {
 
   case class EntityAccessControlAgora(entity: Method, acls: Seq[AgoraPermission], message: Option[String] = None)
 
-  case class MethodAclPair(method:MethodRepoMethod, acls: Seq[FireCloudPermission], message: Option[String] = None)
+  case class MethodAclPair(method: AgoraMethod, acls: Seq[FireCloudPermission], message: Option[String] = None)
```
rushtong, line 89: I've always hated the name `MethodRepoMethod`. 👍

rushtong (approved):
👍

aednichols:
jenkins retest

aednichols:
jenkins retest

jacmrob:
jenkins retest

PR #609, opened 2018-03-08
made startup checks part of health checks

dvoet:

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/HealthChecks.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/HealthChecks.scala
@@ -2,7 +2,7 @@ package org.broadinstitute.dsde.firecloud
 
 import akka.actor.ActorSystem
 import com.typesafe.scalalogging.LazyLogging
-import org.broadinstitute.dsde.firecloud.model.{AccessToken, RegistrationInfo}
+import org.broadinstitute.dsde.firecloud.model.{AccessToken, RegistrationInfo, WorkbenchEnabled}
 import org.broadinstitute.dsde.workbench.util.health.{SubsystemStatus, Subsystems}
 import org.broadinstitute.dsde.workbench.util.health.Subsystems._
 import spray.http.StatusCodes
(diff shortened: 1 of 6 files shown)


davidangb (approved):
:+1: did not test

src/main/scala/org/broadinstitute/dsde/firecloud/HealthChecks.scala
```diff
@@ -82,5 +62,20 @@ class StartupChecks(app: Application, registerSAs: Boolean = true)
...
+    Map(
+      Agora -> app.agoraDAO.status,
+      Consent -> app.consentDAO.status,
+      GoogleBuckets -> app.googleServicesDAO.status,
+      LibraryIndex -> app.searchDAO.status,
+      OntologyIndex -> app.ontologyDAO.status,
+      Rawls -> app.rawlsDAO.status,
+      Sam -> app.samDAO.status,
+      Thurloe -> app.thurloeDAO.status,
+      adminSaRegistered -> isAdminSARegistered.map(SubsystemStatus(_, None)),
```
breilly2: Since `isServiceAccountRegistered` is potentially side-effecty, I don't think it's right to move it into the health checks as-is. I think we need to divorce the part that creates the SAs (with retries if necessary) and keep that as a startup activity and have `isServiceAccountRegistered` simply return the status.

Is there an invocation of the health checks at startup that I missed? I didn't see anything.

breilly2 (changes requested):

src/main/scala/org/broadinstitute/dsde/firecloud/HealthChecks.scala
dvoet: I removed the initial checks at startup because this code does it, it was already side effecty.

I would argue to leave it as is because it a) is no worse than it was before and b) is lower complexity because there is not some other async process to deal with. This is also more self healing.

davidangb: The invocation of health checks at startup is FireCloudServiceActor.scala lines 66-67.

If the registration stays in the health check, maybe turn `logger.info(s"$name is properly registered.")` down to debug? Otherwise it logs once a minute, at least when I run locally.

breilly2 (approved):
Ahh, right... "at startup" being at scheduling the health monitor... so delayed by 3 seconds 🤔 but I guess that's fine.

Lower complexity, I guess. Self-healing... :man_shrugging:.

I slightly disagree with "no worse" just because it breaks my expectations by there being side effects of the health check... which feels wrong to me and is not how it used to be. But... we have bigger problems at the moment. ☹️ I agree with David's comment about logging, but otherwise 👍.

PR #610, opened 2018-03-14
GAWB-3160: Automate free credit whitelist for all new users

davidangb:
Objective: all newly-registered users are enabled for free credits.

In the current state, users are manually enabled for free credits by a campaign manager. At enable-time, users claim a billing project. Once the user starts (enrolls in) their trial, the user is added to that billing project. If the user never enrolls, the billing project remains claimed and can't be used by anyone else.

We currently have a 36.89% conversion rate for enrollment. This means that 63.11% of the claimed projects are going unused.

In this PR, I have changed the workflow such that:
* when a new user registers, the user gets `trialState: Enabled` in their profile, but does NOT claim a project.
* when a user enrolls, the enrollment code uses a pre-claimed project if one exists (campaign manager path), otherwise claims a project (newly registered user path)
* if, during registration, the user encounters an error writing `trialState: Enabled` to their profile (very unlikely - they would have seen errors before reaching this point), we complete the registration anyway but send a warning message in the response. As of this PR, I have *not* updated any UI to actually display that error message.

-----

requires broadinstitute/firecloud-develop#1026

-----

Commit history is a mess; I suggest you review as a whole.

-----

In all cases:


src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -60,7 +60,7 @@ object TrialService {
 
 final class TrialService
   (val samDao: SamDAO, val thurloeDao: ThurloeDAO, val rawlsDAO: RawlsDAO,
-   val trialDAO: TrialDAO, val googleDAO: GoogleServicesDAO, projectManager: ActorRef)
+   val trialDao: TrialDAO, val googleDAO: GoogleServicesDAO, projectManager: ActorRef)
```
davidangb, line 63: needed to rename `trialDAO` to `trialDao` due to naming conflicts between traits - unfortunately this affects many lines of code in the diff.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -259,58 +259,78 @@ final class TrialService
...
-            case Failure(profileUpdateFail) => {
-              // We couldn't update trial status, so clean up
-              rawlsDAO.removeUserFromBillingProject(projectId, ProjectRoles.Owner, userInfo.userEmail)(userToken = saToken) map { _ =>
-                logger.warn(s"Enrolling user ${userInfo.userEmail} failed at profile update: ${profileUpdateFail.getMessage}. User has been backed out of billing project $projectId.")
-                RequestCompleteWithErrorReport(InternalServerError, s"We could not process your enrollment. Please try again later. (Error 70)")
-              } recover {
-                case bpFail: Throwable => {
-                  logger.warn(s"Enrolling user ${userInfo.userEmail} failed at profile update: ${profileUpdateFail.getMessage}. User is still in billing project $projectId due to cleanup failure: ${bpFail.getMessage}.")
-                  RequestCompleteWithErrorReport(InternalServerError, s"We could not process your enrollment. Please contact support. (Error 80)")
+    statusFuture flatMap { status =>
```
davidangb, line 268: this looks like a big diff but is mostly whitespace change - I wrapped this block of code in a `statusFuture flatMap {...` and indented it. Try ?w=1 in the github url to ignore whitespace changes.

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialRegistrationSpec.scala
```diff
@@ -0,0 +1,140 @@
...
+  )
+
+  "When updating the profile for a pre-existing, registered user" - {
+    "should not read or write trial status" in {
+      Post(registerPath, fullProfile) ~> dummyUserIdHeaders(NORMAL_USER, email = NORMAL_USER) ~> sealRoute(registerRoutes) ~> check {
+        status should equal(OK)
+        val regInfo = responseAs[RegistrationInfo]
+        // if NORMAL_USER triggers a call to either getTrialStatus or saveTrialStatus, UserRegistrationThurloeDAO will
+        // fail and RegistrationInfo will have a message below.
+        assert(regInfo.message.isEmpty, "- if this reports a BoxedError it is likely the assert inside UserRegistrationThurloeDAO.getTrialStatus or saveTrialStatus")
```
davidangb: I find the "if this reports a BoxedError" distasteful - but it's the result of an `assert` failing and throwing an exception inside a Future - the Future is in the mock DAOs at the bottom of this file. I didn't want to rewrite the runtime Future-handling and recover-handling - so I declared this good enough for the tests.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialServiceSupport.scala
```diff
@@ -148,37 +151,64 @@ trait TrialServiceSupport extends LazyLogging {
     if (date.after(zeroDate)) spreadsheetFormat.format(date) else ""
   }
 
+  /**
+    * Enables the current user for free credits. Should only be used during the registration process.
+    * @param userInfo the current user
+    * @return true if enabling succeeded; will throw an exception if failed.
```
kyuksel: Should it be
```scala
@return [[UserTrialStatus]]
```
instead?

src/main/scala/org/broadinstitute/dsde/firecloud/service/RegisterService.scala
```diff
@@ -56,7 +57,15 @@ class RegisterService(val rawlsDao: RawlsDAO, val samDao: SamDAO, val thurloeDao
...
+        } yield {
+          val message = freeCredits match {
+            case Left(ex) => Some(s"Error enabling free credits during registration. Underlying error: ${ex.getMessage}")
+            case Right(_) => None
+          }
+          registrationInfo.copy(message = message)
+        }
       } else Future.successful(isRegistered)
     } yield {
       RequestComplete(StatusCodes.OK, userStatus)
```
kyuksel: Moving 
```scala
      isRegistered <- samDao.getRegistrationStatus(userInfo) recover {
        case e: FireCloudExceptionWithErrorReport if e.errorReport.statusCode == Option(StatusCodes.NotFound) =>
          RegistrationInfo(WorkbenchUserInfo(userInfo.id, userInfo.userEmail), WorkbenchEnabled(false, false, false))
      }
```
and
```scala
      userStatus <- if (!isRegistered.enabled.google || !isRegistered.enabled.ldap) {
        for {
          registrationInfo <- samDao.registerUser(userInfo)
          _ <- rawlsDao.registerUser(userInfo) //This call to rawls handles leftover registration pieces (welcome email and pending workspace access)
          freeCredits:Either[Exception,UserTrialStatus] <- enableSelfForFreeCredits(userInfo)
            .map(Right(_)) recover { case e: Exception => Left(e) }
        } yield {
          val message = freeCredits match {
            case Left(ex) => Some(s"Error enabling free credits during registration. Underlying error: ${ex.getMessage}")
            case Right(_) => None
          }
          registrationInfo.copy(message = message)
        }
      } else Future.successful(isRegistered)
```
into their own `def`s may make it easier to follow what's going on.

src/main/scala/org/broadinstitute/dsde/firecloud/service/RegisterService.scala
```diff
@@ -56,7 +57,15 @@ class RegisterService(val rawlsDao: RawlsDAO, val samDao: SamDAO, val thurloeDao
         for {
           registrationInfo <- samDao.registerUser(userInfo)
           _ <- rawlsDao.registerUser(userInfo) //This call to rawls handles leftover registration pieces (welcome email and pending workspace access)
-        } yield registrationInfo
+          freeCredits:Either[Exception,UserTrialStatus] <- enableSelfForFreeCredits(userInfo)
```
kyuksel: Does `Either` add value? I wonder if pattern matching directly on `Future[UserTrialStatus]` (the output of `enableSelfForFreeCredits`) would be more concise?

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -259,58 +259,78 @@ final class TrialService
     }
   }
 
-  private def enrollUserInternal(userInfo: UserInfo, status: UserTrialStatus): Future[PerRequestMessage] = {
+  private def enrollUserInternal(userInfo: UserInfo, enabledStatus: UserTrialStatus): Future[PerRequestMessage] = {
 
-    val projectId = status.billingProjectName match {
-      case Some(name: String) => name
+    val statusFuture = enabledStatus.billingProjectName match {
+      case Some(name: String) => Future.successful(enabledStatus)
```
kyuksel: Consider using `_` instead of unused `name`

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
kyuksel: It may help to factor out the unchanged code into its own `def`, especially given how complicated `enrollUserInternal` had already become.

aednichols (commented):
Looks like this needs a firecloud-develop PR for `projectBufferSize`

davidangb:
@aednichols ah sorry! I made the branch but not the PR. broadinstitute/firecloud-develop#1026

src/main/scala/org/broadinstitute/dsde/firecloud/service/RegisterService.scala
```diff
@@ -56,7 +57,15 @@ class RegisterService(val rawlsDao: RawlsDAO, val samDao: SamDAO, val thurloeDao
...
+            .map(Right(_)) recover { case e: Exception => Left(e) }
+        } yield {
+          val message = freeCredits match {
+            case Left(ex) => Some(s"Error enabling free credits during registration. Underlying error: ${ex.getMessage}")
+            case Right(_) => None
+          }
+          registrationInfo.copy(message = message)
```
aednichols: Perhaps we should append the free credits `message` to the existing message, instead of replacing it.

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialRegistrationSpec.scala
```diff
@@ -0,0 +1,140 @@
...
+        status should equal(OK)
+        val regInfo = responseAs[RegistrationInfo]
+        assert(regInfo.message.isEmpty, "- if this reports a BoxedError it is likely the assert inside UserRegistrationThurloeDAO.saveTrialStatus")
+      }
+
+    }
+  }
+
+  "When registering a brand new user encounters an error saving Enabled trial status for free credits" - {
+    "should complete registration with a message in responses" in {
```
aednichols: Nice, seems like the right call to let registration carry on.

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialRegistrationSpec.scala
```diff
@@ -0,0 +1,140 @@
...
+        val regInfo = responseAs[RegistrationInfo]
+        // if NORMAL_USER triggers a call to either getTrialStatus or saveTrialStatus, UserRegistrationThurloeDAO will
+        // fail and RegistrationInfo will have a message below.
+        assert(regInfo.message.isEmpty, "- if this reports a BoxedError it is likely the assert inside UserRegistrationThurloeDAO.getTrialStatus or saveTrialStatus")
+      }
+    }
+  }
+
+  "When registering a brand new user in sam" - {
+    "should enable for free credits" in {
```
kyuksel: Should the above be 
```scala
"the user should be enabled for free credits" in {
```
? 

Similarly for the following ones...

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -259,58 +259,78 @@ final class TrialService
     }
   }
 
-  private def enrollUserInternal(userInfo: UserInfo, status: UserTrialStatus): Future[PerRequestMessage] = {
+  private def enrollUserInternal(userInfo: UserInfo, enabledStatus: UserTrialStatus): Future[PerRequestMessage] = {
```
aednichols, line 262: This function was already a bit of a monster before this PR, might be nice to break out the block that creates `statusFuture`

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/TrialRegistrationSpec.scala
```diff
@@ -0,0 +1,140 @@
...
+import org.broadinstitute.dsde.firecloud.service.{BaseServiceSpec, FireCloudRequestBuilding, RegisterService}
+import org.broadinstitute.dsde.rawls.model.RawlsUserEmail
+import spray.http.HttpResponse
+import spray.http.StatusCodes.{NotFound, OK}
+import spray.httpx.SprayJsonSupport
+import spray.json.DefaultJsonProtocol
+
+import scala.concurrent.Future
+import scala.util.Try
+
```
kyuksel, line 21: great to see this test suite!

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
```diff
@@ -259,58 +259,78 @@ final class TrialService
...
+    val statusFuture = enabledStatus.billingProjectName match {
+      case Some(name: String) => Future.successful(enabledStatus)
       case None => {
-        logger.warn(s"User ${userInfo.userEmail} attempted to enroll in trial but no billing project in profile.")
-        throw new FireCloudException("We could not process your enrollment. Please contact support. (Error 56)")
+        Try(claimProjectWithRetries(WorkbenchUserInfo(userInfo.id, userInfo.userEmail))) match {
+          case Success(claimed) =>
+            // persist claimed project name to user's profile
+            val updatedStatus = enabledStatus.copy(billingProjectName = Some(claimed.name.value))
+            thurloeDao.saveTrialStatus(userInfo.id, userInfo, updatedStatus) map {
```
aednichols: Maybe we are not doing this in the most efficient order. Would it make sense to check e.g. `getProjectMembers` _before_ assigning the project to the user, so that we can try with a different project, instead of failing later?

kyuksel (commented):
Good work! Didn't do manual testing.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialServiceSupport.scala
```diff
@@ -148,37 +151,64 @@ trait TrialServiceSupport extends LazyLogging {
...
+    *
+    * @param userInfo user for whom to claim project
+    * @param numAttempts how many times should we try to find an available project? Retries here help when multiple users are enabled at once
+    * @return claimed project
+    */
+  def claimProjectWithRetries(userInfo: WorkbenchUserInfo, numAttempts: Int = 50): TrialProject = {
+    // log a warning, which should be seen by dev team, if the project pool is running low.
+    val numAvailable:Long = trialDao.countProjects.getOrElse("available", 0L)
+    if (numAvailable < FireCloudConfig.Trial.projectBufferSize)
+      logger.error(s"There are only $numAvailable free trial projects available; create more as soon as possible!")
```
aednichols, line 196: Can we just... do this programmatically from here? It won't help the user that's currently trying to claim because the request will time out before the project creates, but it would help subsequent users

aednichols (approved):
👍 Looks reasonable, just a few thoughts

kyuksel:
Two more questions:

1. Does the Free Trial documentation need updating?
2. Does the Selenium test need changes?

src/main/scala/org/broadinstitute/dsde/firecloud/service/RegisterService.scala
rushtong: Agreed for the `userStatus` case at least.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialServiceSupport.scala
```diff
@@ -148,37 +151,64 @@ trait TrialServiceSupport extends LazyLogging {
...
 
+  /**
+    * Enables the current user for free credits. Should only be used during the registration process.
+    * @param userInfo the current user
+    * @return true if enabling succeeded; will throw an exception if failed.
+\   */
+  def enableSelfForFreeCredits(userInfo: UserInfo): Future[UserTrialStatus] = {
+    thurloeDao.getTrialStatus(userInfo.id, userInfo) flatMap { userTrialStatus =>
+      val doTransition = Enabled.isAllowedFrom(userTrialStatus.state)
+      val newStatus = userTrialStatus.copy(state = Some(Enabled))
```
rushtong: minor nit: this val could be inside the `if (doTransition)` block.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialServiceSupport.scala
rushtong: Alternatively, we could spin off an async request that requests `n` projects so the pool never gets too low - that would not interfere with the current user's claim process. However, I lean towards keeping this "manual-intervention-required" for the sake of keeping tight control over trial project creation.

rushtong (approved):
👍 Minor comments inline, but looks good, modulo others' feedback.

src/main/scala/org/broadinstitute/dsde/firecloud/service/RegisterService.scala
davidangb: I think it adds value; I'm open to suggestions here. My thinking was:

1. I need a way for `enableSelfForFreeCredits` to be tolerant to exceptions, while not changing the exception behavior of `samDao.registerUser(userInfo)` and `rawlsDao.registerUser(userInfo)`
2. If `enableSelfForFreeCredits` fails, I need a way to capture its exception message and pass it into the `yield` clause

src/main/scala/org/broadinstitute/dsde/firecloud/service/RegisterService.scala
davidangb: Good call; I'll change. I know this is safe for now because I just added the `message` field and therefore know nothing else writes to it ... but it's worth changing for the future.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialServiceSupport.scala
davidangb: the problem with triggering creation of new projects here is that:
* project creation via `ProjectManager` is intentionally throttled to avoid hitting quotas, especially because end users may be creating projects at the same time
* we have two instances of orch in production
* there's currently no coordination between orch instances on project creation

So, if a user hits this code on instance 1 and triggers creation of N projects, then a second user hits instance 2 and triggers another creation of N projects, we have both instances spinning and we're creating at 2x our throttle. I'd love to fix this for other reasons but I think it's out of scope of this PR.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
davidangb: This sounds good but is a much larger change than I was hoping to tackle in this PR. Do you feel strongly about this?

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
davidangb: done!

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
davidangb: in the spirit of keeping changes small, I'm choosing not to touch the unchanged code ... do you feel strongly about this?

davidangb:
@kyuksel I'll update the doc at some point; I don't think anyone is using that now, but I'll get to it. 

regarding selenium tests: this PR passed existing tests, so no changes are necessary afaik for compatibility. I should probably add some new tests ... but given the state of testing right now I might wait until after things settle down. Opinions?

davidangb:
@kyuksel @rushtong @aednichols I believe I have either responded to or addressed all the comments! Back to you, please and thank you.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialServiceSupport.scala
aednichols: Agree that it is out of scope!

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialService.scala
aednichols: Out of scope, I just pointed it out so that you might have small opportunities to pull the codebase in that direction while making other changes

aednichols:
👍 LGTM

rushtong (approved):
👍

src/main/scala/org/broadinstitute/dsde/firecloud/service/RegisterService.scala
kyuksel: Ah...I see that what I had in mind wouldn't work...not as trivially as I thought anyway.

kyuksel (approved):
Looks good 👍

PR #612, opened 2018-03-28
GAWB-3272: logit metrics

davidangb:
requires broadinstitute/firecloud-develop#1056
requires broadinstitute/rawls#883

This PR adds a new actor and a new DAO to communicate with Logit. Every 6 hours (configurable), orch queries Rawls to count the number of samples in a given set of workspaces, then sends that metric to Logit.

I'm building this in orch specifically, because in a future story (GAWB-3265) we'll also query Library for stats, and send those stats to Logit as well.

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
@@ -3,6 +3,7 @@ package org.broadinstitute.dsde.firecloud
 import akka.stream.ActorMaterializer
 import org.broadinstitute.dsde.firecloud.dataaccess._
 import org.broadinstitute.dsde.firecloud.elastic.ElasticUtils
+import org.broadinstitute.dsde.firecloud.metrics.MetricsActor
 import org.broadinstitute.dsde.firecloud.model.{UserInfo, WithAccessToken}
 import org.slf4j.LoggerFactory
 import spray.http.StatusCodes._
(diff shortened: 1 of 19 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
```diff
@@ -68,6 +75,16 @@ class FireCloudServiceActor extends HttpServiceActor with FireCloudDirectives
 
   val trialProjectManager = system.actorOf(ProjectManager.props(app.rawlsDAO, app.trialDAO, app.googleServicesDAO), "trial-project-manager")
 
+  if (logitMetricsEnabled) {
+    logger.info("Logit metrics are enabled.")
+    val metricsActor = system.actorOf(MetricsActor.props(app), "metrics-actor")
+    // use a randomized startup delay to avoid multiple instances of this app executing on the same cycle
```
rushtong, line 81: If I understand correctly, each instance would be querying across a time range of one second on a 6-hour schedule, so we would be getting two data points every 6 hours under the current configuration.

rushtong (approved):
👍 Looks good. I was able to test locally and see the traffic go by ... very cool stuff!

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
davidangb: that's correct. I have a random delay on each schedule so they don't hit rawls at the same time ... but I left the schedule on both prod instances because:
1. coordinating across instances is a lot of work
2. having both instances send metrics keeps us safe in the case where we need to take one instance down for maintenance/upgrades/etc
3. recording 8 metrics per day vs. 4 per day is trivial given the size of other logs we send to kibana

PR #615, opened 2018-03-30
Enhanced error messages for /me [GAWB-3116]

aednichols:
Requires https://github.com/broadinstitute/firecloud-ui/pull/1255

---

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
@@ -83,43 +83,43 @@ trait UserApiService extends HttpService with PerRequestCreator with FireCloudRe
         authorizationHeader match {
           // no Authorization header; the user must be unauthorized
           case None =>
-            respondWithErrorReport(Unauthorized, Unauthorized.defaultMessage, requestContext)
-          // browser sent Authorization header; try to query rawls for user status
+            respondWithErrorReport(Unauthorized, "No authorization header in request.", requestContext)
+          // browser sent Authorization header; try to query Sam for user status
           case Some(c) =>
             val pipeline = authHeaders(requestContext) ~> sendReceive
-            val extReq = Get(UserApiService.samRegisterUserURL)
-            pipeline(extReq) onComplete {
+            val samRequest = Get(UserApiService.samRegisterUserURL)
+            pipeline(samRequest) onComplete {
               case Success(response) =>
                 response.status match {
-                  // rawls rejected our request. User is either invalid or their token timed out; this is truly unauthorized
-                  case Unauthorized => respondWithErrorReport(Unauthorized, Unauthorized.defaultMessage, requestContext)
+                  // Sam rejected our request. User is either invalid or their token timed out; this is truly unauthorized
+                  case Unauthorized => respondWithErrorReport(Unauthorized, "Request rejected by identity service - invalid user or expired token.", requestContext)
                   // rawls 404 means the user is not registered with FireCloud
-                  case NotFound => respondWithErrorReport(NotFound, "FireCloud user registration not found", requestContext)
+                  case NotFound => respondWithErrorReport(NotFound, "FireCloud user registration not found.", requestContext)
                   // rawls error? boo. All we can do is respond with an error.
-                  case InternalServerError => respondWithErrorReport(InternalServerError, InternalServerError.defaultMessage, requestContext)
+                  case InternalServerError => respondWithErrorReport(InternalServerError, "Identity service encountered an unknown error, please try again. (Error 10)", requestContext)
                   // rawls found the user; we'll try to parse the response and inspect it
                   case OK =>
                     val respJson = response.entity.as[RegistrationInfo]
                     respJson match {
                       case Right(regInfo) =>
                         if (regInfo.enabled.google && regInfo.enabled.ldap && regInfo.enabled.allUsersGroup) {
-                          // rawls says the user is fully registered and activated!
+                          // Sam says the user is fully registered and activated!
                           requestContext.complete(OK, regInfo)
                         } else {
-                          // rawls knows about the user, but the user isn't activated
-                          respondWithErrorReport(Forbidden, "FireCloud user not activated", requestContext)
+                          // Sam knows about the user, but the user isn't activated
+                          respondWithErrorReport(Forbidden, "FireCloud user not activated.", requestContext)
                         }
-                      // we couldn't parse the rawls response. Respond with an error.
+                      // we couldn't parse the Sam response. Respond with an error.
                       case Left(error) =>
-                        respondWithErrorReport(InternalServerError, InternalServerError.defaultMessage, requestContext)
+                        respondWithErrorReport(InternalServerError, "Identity service encountered an unknown error, please try again. (Error 20)", requestContext)
                     }
                   case x =>
-                    // if we get any other error from rawls, pass that error on
+                    // if we get any other error from Sam, pass that error on
                     respondWithErrorReport(x.intValue, "Unexpected response validating registration: " + x.toString, requestContext)
                 }
-              // we couldn't reach rawls (within timeout period). Respond with a Service Unavailable error.
+              // we couldn't reach Sam (within timeout period). Respond with a Service Unavailable error.
               case Failure(error) =>
-                respondWithErrorReport(ServiceUnavailable, ServiceUnavailable.defaultMessage, requestContext)
+                respondWithErrorReport(ServiceUnavailable, "Identity service did not produce a timely response, please try again later.", requestContext)
             }
         }
       }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
@@ -93,11 +93,11 @@ trait UserApiService extends HttpService with PerRequestCreator with FireCloudRe
                 response.status match {
                   // Sam rejected our request. User is either invalid or their token timed out; this is truly unauthorized
                   case Unauthorized => respondWithErrorReport(Unauthorized, "Request rejected by identity service - invalid user or expired token.", requestContext)
-                  // rawls 404 means the user is not registered with FireCloud
+                  // Sam 404 means the user is not registered with FireCloud
                   case NotFound => respondWithErrorReport(NotFound, "FireCloud user registration not found.", requestContext)
-                  // rawls error? boo. All we can do is respond with an error.
+                  // Sam error? boo. All we can do is respond with an error.
                   case InternalServerError => respondWithErrorReport(InternalServerError, "Identity service encountered an unknown error, please try again. (Error 10)", requestContext)
-                  // rawls found the user; we'll try to parse the response and inspect it
+                  // Sam found the user; we'll try to parse the response and inspect it
                   case OK =>
                     val respJson = response.entity.as[RegistrationInfo]
                     respJson match {
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
@@ -114,6 +114,7 @@ trait UserApiService extends HttpService with PerRequestCreator with FireCloudRe
                         }
                       // we couldn't parse the Sam response. Respond with an error.
                       case Left(_) =>
+                        // TODO: no obvious way to include the JSON parsing error in the ErrorReport. We define `message` below and it does not belong there.
                         respondWithErrorReport(InternalServerError, "Received unparseable response from identity service.", requestContext)
                     }
                   case x =>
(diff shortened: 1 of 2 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
```diff
@@ -83,43 +83,43 @@ trait UserApiService extends HttpService with PerRequestCreator with FireCloudRe
         authorizationHeader match {
           // no Authorization header; the user must be unauthorized
           case None =>
-            respondWithErrorReport(Unauthorized, Unauthorized.defaultMessage, requestContext)
-          // browser sent Authorization header; try to query rawls for user status
+            respondWithErrorReport(Unauthorized, "No authorization header in request.", requestContext)
+          // browser sent Authorization header; try to query Sam for user status
```
aednichols: My basis for the Rawls -> Sam renames is that the request is a `Get(UserApiService.samRegisterUserURL)`

rushtong: Commenting here for my own learning .... this is true. Following the URL, it is actually making a call to `Sam` and not `Rawls`. It looks like we still have a bit of a mix between where user info lives wrt Sam vs Rawls.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
```diff
@@ -83,43 +83,43 @@ trait UserApiService extends HttpService with PerRequestCreator with FireCloudRe
...
                   case x =>
-                    // if we get any other error from rawls, pass that error on
+                    // if we get any other error from Sam, pass that error on
                     respondWithErrorReport(x.intValue, "Unexpected response validating registration: " + x.toString, requestContext)
                 }
-              // we couldn't reach rawls (within timeout period). Respond with a Service Unavailable error.
+              // we couldn't reach Sam (within timeout period). Respond with a Service Unavailable error.
               case Failure(error) =>
-                respondWithErrorReport(ServiceUnavailable, ServiceUnavailable.defaultMessage, requestContext)
+                respondWithErrorReport(ServiceUnavailable, "Identity service did not produce a timely response, please try again later.", requestContext)
```
rushtong: Also here for my own clarity - it's not necessary to be consistent with an error code here due to the fact that the messages are different and can be clearly followed from log reporting.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
```diff
@@ -83,43 +83,43 @@ trait UserApiService extends HttpService with PerRequestCreator with FireCloudRe
...
-            pipeline(extReq) onComplete {
+            val samRequest = Get(UserApiService.samRegisterUserURL)
+            pipeline(samRequest) onComplete {
               case Success(response) =>
                 response.status match {
-                  // rawls rejected our request. User is either invalid or their token timed out; this is truly unauthorized
-                  case Unauthorized => respondWithErrorReport(Unauthorized, Unauthorized.defaultMessage, requestContext)
+                  // Sam rejected our request. User is either invalid or their token timed out; this is truly unauthorized
+                  case Unauthorized => respondWithErrorReport(Unauthorized, "Request rejected by identity service - invalid user or expired token.", requestContext)
                   // rawls 404 means the user is not registered with FireCloud
```
kyuksel: You may want to rename the other `rawls`s as well.

rushtong (approved):
👍

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
```diff
@@ -83,43 +83,43 @@ trait UserApiService extends HttpService with PerRequestCreator with FireCloudRe
...
                       case Right(regInfo) =>
                         if (regInfo.enabled.google && regInfo.enabled.ldap && regInfo.enabled.allUsersGroup) {
-                          // rawls says the user is fully registered and activated!
+                          // Sam says the user is fully registered and activated!
                           requestContext.complete(OK, regInfo)
                         } else {
-                          // rawls knows about the user, but the user isn't activated
-                          respondWithErrorReport(Forbidden, "FireCloud user not activated", requestContext)
+                          // Sam knows about the user, but the user isn't activated
+                          respondWithErrorReport(Forbidden, "FireCloud user not activated.", requestContext)
```
kyuksel, line 113: Would `enabled` be easier to understand than `activated` since that seems to be term we've been using (with Google, LDAP, etc.)?

kyuksel (commented):
Good improvement. Comments inline.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
aednichols: Yeah, I only put an error code in the two cases that shared the same message. Could standardize across all, but wasn’t sure it added value

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
aednichols: I don't want to rewrite existing terminology that people are familiar with, short of correcting factual errors

aednichols:
One thing I could add on top of this is taking the `response` from Sam and adding it to the string we send back - but that would mess with neatly displaying the error string in the UI (unless we used e.g. a carriage return to identify that the second part of the string should be displayed in the error report viewer)

kyuksel:
> One thing I could add on top of this is taking the response from Sam and adding it to the string we send back - but that would mess with neatly displaying the error string in the UI (unless we used e.g. a carriage return to identify that the second part of the string should be displayed in the error report viewer)

I think someone (David?) had floated the idea of including more detailed debug info in an expandable fashion that defaults to `not displayed`.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
kyuksel: sensible

aednichols:
Yes, that is an implementation suggestion in the story description

I definitely thought about it (as you can see) but my sense was that it might be overkill. Obviously subject to change based on reviewer comments

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
```diff
@@ -83,43 +83,43 @@ trait UserApiService extends HttpService with PerRequestCreator with FireCloudRe
...
-                  case NotFound => respondWithErrorReport(NotFound, "FireCloud user registration not found", requestContext)
-                  // rawls error? boo. All we can do is respond with an error.
-                  case InternalServerError => respondWithErrorReport(InternalServerError, InternalServerError.defaultMessage, requestContext)
-                  // rawls found the user; we'll try to parse the response and inspect it
+                  // Sam rejected our request. User is either invalid or their token timed out; this is truly unauthorized
+                  case Unauthorized => respondWithErrorReport(Unauthorized, "Request rejected by identity service - invalid user or expired token.", requestContext)
+                  // Sam 404 means the user is not registered with FireCloud
+                  case NotFound => respondWithErrorReport(NotFound, "FireCloud user registration not found.", requestContext)
+                  // Sam error? boo. All we can do is respond with an error.
+                  case InternalServerError => respondWithErrorReport(InternalServerError, "Identity service encountered an unknown error, please try again. (Error 10)", requestContext)
```
davidangb: for this unexpected error, suggest you pass the root cause back up to the UI? Something like:
```
case ise:InternalServerError => respondWithErrorReport(InternalServerError, "Identity service encountered an unknown error, please try again. (Error 10)", ise, requestContext)
```
of course, that signature for `respondWithErrorReport` - a signature that accepts a `Throwable` and uses the throwable when creating its `ErrorReport` - doesn't exist, so you'll have to create it.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
```diff
@@ -83,43 +83,43 @@ trait UserApiService extends HttpService with PerRequestCreator with FireCloudRe
...
-                          // rawls knows about the user, but the user isn't activated
-                          respondWithErrorReport(Forbidden, "FireCloud user not activated", requestContext)
+                          // Sam knows about the user, but the user isn't activated
+                          respondWithErrorReport(Forbidden, "FireCloud user not activated.", requestContext)
                         }
-                      // we couldn't parse the rawls response. Respond with an error.
+                      // we couldn't parse the Sam response. Respond with an error.
                       case Left(error) =>
-                        respondWithErrorReport(InternalServerError, InternalServerError.defaultMessage, requestContext)
+                        respondWithErrorReport(InternalServerError, "Identity service encountered an unknown error, please try again. (Error 20)", requestContext)
```
davidangb: same here - pass the `error` back up to the UI in the `ErrorReport`

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
```diff
@@ -83,43 +83,43 @@ trait UserApiService extends HttpService with PerRequestCreator with FireCloudRe
...
                   case x =>
-                    // if we get any other error from rawls, pass that error on
+                    // if we get any other error from Sam, pass that error on
                     respondWithErrorReport(x.intValue, "Unexpected response validating registration: " + x.toString, requestContext)
                 }
-              // we couldn't reach rawls (within timeout period). Respond with a Service Unavailable error.
+              // we couldn't reach Sam (within timeout period). Respond with a Service Unavailable error.
               case Failure(error) =>
-                respondWithErrorReport(ServiceUnavailable, ServiceUnavailable.defaultMessage, requestContext)
+                respondWithErrorReport(ServiceUnavailable, "Identity service did not produce a timely response, please try again later.", requestContext)
```
davidangb: same here - pass the `error` back up to the UI in the `ErrorReport`

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
```diff
@@ -83,43 +83,43 @@ trait UserApiService extends HttpService with PerRequestCreator with FireCloudRe
...
-                      // we couldn't parse the rawls response. Respond with an error.
+                      // we couldn't parse the Sam response. Respond with an error.
                       case Left(error) =>
-                        respondWithErrorReport(InternalServerError, InternalServerError.defaultMessage, requestContext)
+                        respondWithErrorReport(InternalServerError, "Identity service encountered an unknown error, please try again. (Error 20)", requestContext)
                     }
                   case x =>
-                    // if we get any other error from rawls, pass that error on
+                    // if we get any other error from Sam, pass that error on
                     respondWithErrorReport(x.intValue, "Unexpected response validating registration: " + x.toString, requestContext)
```
davidangb, line 122: might want to tweak this one too, so we pass the full error object up to the UI, instead of just the `toString` of it

davidangb (commented):
Love these changes, have suggestions inline.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
aednichols: `InternalServerError` is a `val` on `spray.http.StatusCodes`, not the error itself

I see what you're getting at though, I think it's possible

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
davidangb: it used to be Rawls, and when it moved to Sam nobody updated the code comments.

kyuksel (approved):
👍

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
```diff
@@ -83,43 +83,47 @@ trait UserApiService extends HttpService with PerRequestCreator with FireCloudRe
...
-                          respondWithErrorReport(Forbidden, "FireCloud user not activated", requestContext)
+                          // Sam knows about the user, but the user isn't activated
+                          respondWithErrorReport(Forbidden, "FireCloud user not activated.", requestContext)
                         }
-                      // we couldn't parse the rawls response. Respond with an error.
-                      case Left(error) =>
-                        respondWithErrorReport(InternalServerError, InternalServerError.defaultMessage, requestContext)
+                      // we couldn't parse the Sam response. Respond with an error.
+                      case Left(_) =>
+                        // TODO: no obvious way to include the JSON parsing error in the ErrorReport. We define `message` below and it does not belong there.
```
aednichols, line 117: I think our best bet is to add a `description` field to `ErrorReport` and add support for it to UI's `ErrorViewer`

davidangb: It's pretty ugly, but you could do something like:
```
                      // we couldn't parse the Sam response. Respond with an error.
                      case Left(e) =>
                        val er:ErrorReport = e match {
                          case MalformedContent(errorMessage: String, None) => ErrorReport(errorMessage)
                          case MalformedContent(errorMessage: String, Some(cause:Throwable)) => ErrorReport(InternalServerError, errorMessage, cause)
                          case UnsupportedContentType(errorMessage: String) => ErrorReport(errorMessage)
                          case _ => ErrorReport("Deserialization error.")
                        }
                        respondWithErrorReport(InternalServerError, s"Received unparseable response from identity service: ${er.message}", er, requestContext)

...
...

private def respondWithErrorReport(statusCode: StatusCode, message: String, cause: ErrorReport, requestContext: RequestContext): Unit = {
    requestContext.complete(statusCode, ErrorReport(statusCode = statusCode, message = message, cause = cause))
  }

```

davidangb (approved):
:+1:

aednichols:
jenkins retest

PR #621, opened 2018-04-25
GAWB-2350: TCGA and TARGET groups owned by Orch SA

MatthewBemis:

In all cases:


src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockGroupSupport.scala
```diff
@@ -0,0 +1,15 @@
+package org.broadinstitute.dsde.firecloud.dataaccess
+
+import org.broadinstitute.dsde.workbench.model.{WorkbenchEmail, WorkbenchGroupName}
+
+/**
+  * Created by mbemis on 4/25/18.
+  */
+trait MockGroupSupport {
```
MatthewBemis, line 8: this only exists because the endpoints are sticking around in rawls for now but there's still one endpoint related to groups that we need in sam. this lets the tests share a common group map

MatthewBemis:
Notes on the two automated test failures:

1) The free trial failure could be legitimate, I did modify that code a little bit (isGroupMember stuff). But it's probably because this test run doesn't use the rawls piece that talks to sam for managed groups.

2) The TCGA error- same reason as above. Once this PR is tested with the rawls piece, it should be OK.

gpolumbo-broad (approved):

PR #622, opened 2018-05-03
GAWB-3556: DUAS Questionnaire Ingest API

ansingh7115:
THIS ISN'T TOTALLY DONE YET, I HAVE QUESTIONS

This API is built on fiab-ansingh-liked-ewe (104.198.197.63)

DUAS API Tech Doc - https://docs.google.com/document/d/1gOtzZCQxnpc3vrMzVJX2z6ehJBM_PjhKDfBpfsh1Tgo

This is basically taking a whole bunch of questionnaire results like this:

> { "generalResearchUse": false,
>   "healthMedicalBiomedicalUseOnly": false,
>   "diseaseUseOnly": [9351,1287],
>   "commercialUseProhibited": true,
>   "forProfitUseProhibited": true,
>   "methodsResearchProhibited": true,
>   "aggregateLevelDataProhibited": false,
>   "controlsUseProhibited": false,
>   "genderUseOnly": "male",
>   "pediatricResearchOnly": false,
>   "IRB": true,
>   “prefix”: “library:”
>   }

and then turning it into this: 

> {“library:consentCodes”:[
>   "NPU",
>   "NCU",
>   "NCTRL",
>   "NMDS",
>   "RS-G",
>   "RS-M",
>   "IRB",
>   "DS:diabetes mellitus",
>   "DS:cardiovascular system disease"
> ],
>    "library:dulvn":1.0,
> “library:structuredUseRestriction”: {
>   "GRU": false,
>   "HMB": false,
>   "DS": [9351,1287],
>   "NCU": true,
>   "NPU": true,
>   "NMDS": true,
>   "NAGR": false,
>   "NCTRL": false,
>   "RS-G": true,
>   "RS-FM": false,
>   "RS-M": true,
>   "RS-PD": false,
>   }
> }
> 

I,  @the developer opening this PR, do solemnly pinky swear that:

In all cases:

### DIFF
--- a/src/main/resources/library/attribute-definitions.json
+++ b/src/main/resources/library/attribute-definitions.json
@@ -325,12 +325,6 @@
       "indexable": false,
       "consentCode": "RS-PD"
     },
-    "library:futureUseDate": {
-      "title": "Future use is limited to data generated from samples collected after the following consent form date",
-      "type": "string",
-      "indexable": false,
-      "renderHint": {"datatype": "date"}
-    },
     "library:IRB": {
       "title": "Requester must provide documentation of local IRB/REC approval",
       "type": "boolean",
@@ -410,7 +404,7 @@
         "questionnaire": {
           "enumerate": true,
           "questions": [{"requireGroup": "library:primaryLimitation", "items": ["library:GRU", "library:HMB", "library:DS"]}, "library:DS_URL", "library:NCU", "library:NPU", "library:NMDS",
-            "library:NAGR", "library:NCTRL", "library:RS-G", "library:RS-PD", "library:futureUseDate", "library:IRB"]
+            "library:NAGR", "library:NCTRL", "library:RS-G", "library:RS-PD", "library:IRB"]
         },
         "orsp": {
           "questions": ["library:orsp"]
--- a/src/main/resources/library/attribute-definitions.json
+++ b/src/main/resources/library/attribute-definitions.json
@@ -359,7 +359,6 @@
     "library:dulvn": {
       "title": "Structured Data Use Limitations  Version Number",
       "description": "Structured Data Use Limitations Version Number",
-      "default": 3,
       "indexable": false,
       "type": "integer"
     },
--- a/src/main/resources/library/attribute-definitions.json
+++ b/src/main/resources/library/attribute-definitions.json
@@ -325,13 +325,6 @@
       "indexable": false,
       "consentCode": "RS-PD"
     },
-    "library:RS-POP": {
-      "title": "Future use is limited to research involving a specific population",
-      "type": "array",
-      "items": { "type": "string" },
-      "indexable": false,
-      "consentCode": "RS-POP"
-    },
     "library:futureUseDate": {
       "title": "Future use is limited to data generated from samples collected after the following consent form date",
       "type": "string",
@@ -382,8 +375,7 @@
     "NAGR": "Use of the data includes aggregate level analysis to address research questions (e.g., determine variant frequencies across the general population) ONLY within the bounds of other specified data use limitations.",
     "NCTRL": "Data can be used as a control set ONLY within the bounds of other specified data use limitations (e.g. only for cancer studies).",
     "RS-G": "Use of the data is limited to studies of particular gender.",
-    "RS-PD": "Use of the data is limited to pediatric research.",
-    "RS-POP": "Use of the data is limited to studies of particular population."
+    "RS-PD": "Use of the data is limited to pediatric research."
   },
   "display": {
     "primary": ["library:diseaseOntologyLabel", "library:numSubjects", "library:dataCategory", "library:datatype", "library:dataUseRestriction"],
@@ -392,7 +384,7 @@
       "library:projectName", "library:reference", "library:dataFileFormats", "library:technology", "library:profilingProtocol",
       "library:coverage", "library:studyDesign", "library:cellType", "library:ethnicity", "library:cohortCountry", "library:requiresExternalApproval",
       "library:lmsvn", "library:dulvn"],
-    "consentCodes": ["GRU", "HMB", "DS", "NCU", "NPU", "NMDS", "NAGR", "NCTRL", "RS-G", "RS-PD", "RS-POP"]
+    "consentCodes": ["GRU", "HMB", "DS", "NCU", "NPU", "NMDS", "NAGR", "NCTRL", "RS-G", "RS-PD"]
   },
   "wizard":
   [
@@ -418,7 +410,7 @@
         "questionnaire": {
           "enumerate": true,
           "questions": [{"requireGroup": "library:primaryLimitation", "items": ["library:GRU", "library:HMB", "library:DS"]}, "library:DS_URL", "library:NCU", "library:NPU", "library:NMDS",
-            "library:NAGR", "library:NCTRL", "library:RS-G", "library:RS-PD", "library:RS-POP", "library:futureUseDate", "library:IRB"]
+            "library:NAGR", "library:NCTRL", "library:RS-G", "library:RS-PD", "library:futureUseDate", "library:IRB"]
         },
         "orsp": {
           "questions": ["library:orsp"]
@@ -427,7 +419,6 @@
     }
   ],
   "versions": [
-    "library:dulvn",
     "library:lmsvn"
   ],
   "requireGroups": {
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -4307,41 +4307,41 @@ definitions:
     properties:
       generalResearchUse:
         type: boolean
-        description: something something
+        description: Is the data available for future general research use?
       healthMedicalUseOnly:
         type: boolean
-        description: something something
+        description: Is future use limited for health/medical/biomedical research?
       diseaseUseOnly:
         items:
           type: integer
-        description: something something
+        description: Future use is limited to research involving the following disease area(s)
       commercialUseProhibited:
         type: boolean
-        description: something something
+        description: Is future commercial use prohibited?
       forProfitUseProhibited:
         type: boolean
-        description: something something
+        description: Is future use by for-profit entities prohibited?
       methodsResearchProhibited:
         type: boolean
-        description: something something
+        description: Is future use for methods research (analytic/software/technology development) outside the bounds of the other specified restrictions prohibited?
       aggregateLevelDataProhibited:
         type: boolean
-        description: something something
+        description: Is future use of aggregate-level data for general research purposes prohibited?
       controlsUseProhibited:
         type: boolean
-        description: something something
+        description: Is future use as a control set for diseases other than those specified prohibited?
       genderUseOnly:
         type: string
-        description: something something
+        description: If future use is limited to research involving a particular gender, specify "male" or "female". Any other value will be read as false.
       pediatricResearchOnly:
         type: boolean
-        description: something something
+        description: Is future use limited to pediatric research?
       IRB:
         type: boolean
-        description: something something
+        description: Should the requester provide documentation of local IRB/REC approval?
       prefix:
         type: string
-        description: something something
+        description: A prefix that is prepended to the field names of the responses. This is purely for the user's personalization.
 
 
   CopyConfigurationIngest:
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -4303,6 +4303,17 @@ definitions:
         description: The translated use restriction of the consent.
 
   structuredDataRequest:
+    required:
+        - generalResearchUse
+        - healthMedicalUseOnly
+        - commercialUseProhibited
+        - forProfitUseProhibited
+        - methodsResearchProhibited
+        - aggregateLevelDataProhibited
+        - controlsUseProhibited
+        - genderUseOnly
+        - pediatricResearchOnly
+        - irbRequired
     type: object
     properties:
       generalResearchUse:
@@ -4336,7 +4347,7 @@ definitions:
       pediatricResearchOnly:
         type: boolean
         description: Is future use limited to pediatric research?
-      IRB:
+      irbRequired:
         type: boolean
         description: Should the requester provide documentation of local IRB/REC approval?
       prefix:
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -4305,24 +4305,25 @@ definitions:
   structuredDataRequest:
     required:
         - generalResearchUse
-        - healthMedicalUseOnly
+        - healthMedicalUseRequired
+        - diseaseUseRequired
         - commercialUseProhibited
         - forProfitUseProhibited
         - methodsResearchProhibited
         - aggregateLevelDataProhibited
         - controlsUseProhibited
-        - genderUseOnly
-        - pediatricResearchOnly
+        - genderUseRequired
+        - pediatricResearchRequired
         - irbRequired
     type: object
     properties:
       generalResearchUse:
         type: boolean
         description: Is the data available for future general research use?
-      healthMedicalUseOnly:
+      healthMedicalUseRequired:
         type: boolean
         description: Is future use limited for health/medical/biomedical research?
-      diseaseUseOnly:
+      diseaseUseRequired:
         items:
           type: string
         description: Future use is limited to research involving the following disease area(s). These values must be urls such as "http://purl.obolibrary.org/obo/DOID_12345".
@@ -4341,10 +4342,10 @@ definitions:
       controlsUseProhibited:
         type: boolean
         description: Is future use as a control set for diseases other than those specified prohibited?
-      genderUseOnly:
+      genderUseRequired:
         type: string
         description: If future use is limited to research involving a particular gender, specify "male" or "female". Any other value will be read as false. This field is case-insensitive.
-      pediatricResearchOnly:
+      pediatricResearchRequired:
         type: boolean
         description: Is future use limited to pediatric research?
       irbRequired:
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -3481,6 +3481,26 @@ paths:
         200:
           description: JSON of ES filter clause
 
+  /duos/structuredData:
+    post:
+      tags:
+      - Library
+      operationId: dataIngest
+      summary: Data Use Search ingest
+      description: The ingest API will take the answers to consent questions and return the elasticsearch structure to persist with the data reference.
+      parameters:
+        - in: body
+          description: Structured Data Request
+          name: body
+          required: true
+          schema:
+            $ref: '#/definitions/structuredDataRequest'
+      security:
+        - googleoauth:
+          - openid
+          - email
+          - profile
+
   /api/duos/consent/orsp/{orspId}:
     get:
       tags:
@@ -3506,26 +3526,6 @@ paths:
           description: The consent associated with the provided name could not be found.
         500:
           description: Internal Server Error
-
-  /duos/structuredData:
-    post:
-      tags:
-      - Library # Is this tag right or should it be DataUse?
-      operationId: dataIngest
-      summary: Data Use Search ingest
-      description: The ingest API will take the answers to consent questions and return the elasticsearch structure to persist with the data reference.
-      parameters:
-        - in: body
-          description: create project request
-          name: createProjectRequest
-          required: true
-          schema:
-            $ref: '#/definitions/structuredDataRequest'
-      security:
-        - googleoauth:
-          - openid
-          - email
-          - profile
 
   /error:
     get:
@@ -4343,7 +4343,7 @@ definitions:
         description: Is future use as a control set for diseases other than those specified prohibited?
       genderUseOnly:
         type: string
-        description: If future use is limited to research involving a particular gender, specify "male" or "female". Any other value will be read as false. This value is case-insensitive.
+        description: If future use is limited to research involving a particular gender, specify "male" or "female". Any other value will be read as false. This field is case-insensitive.
       pediatricResearchOnly:
         type: boolean
         description: Is future use limited to pediatric research?
@@ -4352,7 +4352,7 @@ definitions:
         description: Should the requester provide documentation of local IRB/REC approval?
       prefix:
         type: string
-        description: A prefix that is prepended to the field names of the responses. This is purely for the user's personalization.
+        description: Optional. A prefix that is prepended to the field names of the responses. This is purely for personalization.
 
 
   CopyConfigurationIngest:
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -3507,6 +3507,26 @@ paths:
         500:
           description: Internal Server Error
 
+  /duos/structuredData:
+    post:
+      tags:
+      - Library # Is this tag right or should it be DataUse?
+      operationId: dataIngest
+      summary: Data Use Search ingest
+      description: The ingest API will take the answers to consent questions and return the elasticsearch structure to persist with the data reference.
+      parameters:
+        - in: body
+          description: create project request
+          name: createProjectRequest
+          required: true
+          schema:
+            $ref: '#/definitions/structuredDataRequest'
+      security:
+        - googleoauth:
+          - openid
+          - email
+          - profile
+
   /error:
     get:
       tags:
@@ -4281,6 +4301,48 @@ definitions:
       translatedUseRestriction:
         type: string
         description: The translated use restriction of the consent.
+
+  structuredDataRequest:
+    type: object
+    properties:
+      generalResearchUse:
+        type: boolean
+        description:
+      healthMedicalUseOnly:
+        type: boolean
+        description:
+      diseaseUseOnly:
+        items:
+          type: integer  # or is it string?
+          description:
+      commercialUseProhibited:
+        type: boolean
+        description:
+      forProfitUseProhibited:
+        type: boolean
+        description:
+      methodsResearchProhibited:
+        type: boolean
+        description:
+      aggregateLevelDataProhibited:
+        type: boolean
+        description:
+      controlsUseProhibited:
+        type: boolean
+        description:
+      genderUseOnly:
+        type: string
+        description:
+      pediatricResearchOnly:
+        type: boolean
+        description:
+      IRB:
+        type: boolean
+        description:
+      prefix:
+        type: string
+        description:
+
 
   CopyConfigurationIngest:
     type: object
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -4324,8 +4324,8 @@ definitions:
         description: Is future use limited for health/medical/biomedical research?
       diseaseUseOnly:
         items:
-          type: integer
-        description: Future use is limited to research involving the following disease area(s)
+          type: string
+        description: Future use is limited to research involving the following disease area(s). These values must be urls such as "http://purl.obolibrary.org/obo/DOID_12345".
       commercialUseProhibited:
         type: boolean
         description: Is future commercial use prohibited?
@@ -4343,7 +4343,7 @@ definitions:
         description: Is future use as a control set for diseases other than those specified prohibited?
       genderUseOnly:
         type: string
-        description: If future use is limited to research involving a particular gender, specify "male" or "female". Any other value will be read as false.
+        description: If future use is limited to research involving a particular gender, specify "male" or "female". Any other value will be read as false. This value is case-insensitive.
       pediatricResearchOnly:
         type: boolean
         description: Is future use limited to pediatric research?
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -4307,41 +4307,41 @@ definitions:
     properties:
       generalResearchUse:
         type: boolean
-        description:
+        description: something something
       healthMedicalUseOnly:
         type: boolean
-        description:
+        description: something something
       diseaseUseOnly:
         items:
-          type: integer  # or is it string?
-          description:
+          type: integer
+        description: something something
       commercialUseProhibited:
         type: boolean
-        description:
+        description: something something
       forProfitUseProhibited:
         type: boolean
-        description:
+        description: something something
       methodsResearchProhibited:
         type: boolean
-        description:
+        description: something something
       aggregateLevelDataProhibited:
         type: boolean
-        description:
+        description: something something
       controlsUseProhibited:
         type: boolean
-        description:
+        description: something something
       genderUseOnly:
         type: string
-        description:
+        description: something something
       pediatricResearchOnly:
         type: boolean
-        description:
+        description: something something
       IRB:
         type: boolean
-        description:
+        description: something something
       prefix:
         type: string
-        description:
+        description: something something
 
 
   CopyConfigurationIngest:
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/model/DUOS.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/model/DUOS.scala
@@ -1,49 +1,45 @@
 package org.broadinstitute.dsde.firecloud.model
 
 import org.broadinstitute.dsde.rawls.model._
-import spray.json.{JsObject, JsValue}
 import org.broadinstitute.dsde.rawls.model.{AttributeFormat, PlainArrayAttributeListSerializer}
 import org.broadinstitute.dsde.rawls.model.WorkspaceJsonSupport.AttributeNameFormat
-import org.broadinstitute.dsde.rawls.model.Attributable.AttributeMap
 import spray.json._
 import spray.json.DefaultJsonProtocol._
 import spray.json.{JsObject, JsValue}
 
-import scala.util.Try
-
 object DUOS {
 
   case class DuosDataUse(
-                          generalUse: Option[Boolean] = None,
-                          hmbResearch: Option[Boolean] = None,
-                          diseaseRestrictions: Option[Seq[String]] = None,
-                          populationOriginsAncestry: Option[Boolean] = None,
-                          populationStructure: Option[Boolean] = None,
-                          commercialUse: Option[Boolean] = None,
-                          methodsResearch: Option[Boolean] = None,
-                          aggregateResearch: Option[String] = None,
-                          controlSetOption: Option[String] = None,
-                          gender: Option[String] = None,
-                          pediatric: Option[Boolean] = None,
-                          populationRestrictions: Option[Seq[String]] = None,
-                          dateRestriction: Option[String] = None,
-                          recontactingDataSubjects: Option[Boolean] = None,
-                          recontactMay: Option[String] = None,
-                          recontactMust: Option[String] = None,
-                          genomicPhenotypicData: Option[String] = None,
-                          otherRestrictions: Option[Boolean] = None,
-                          cloudStorage: Option[String] = None,
-                          ethicsApprovalRequired: Option[Boolean] = None,
-                          geographicalRestrictions: Option[String] = None,
-                          other: Option[String] = None,
-                          illegalBehavior: Option[Boolean] = None,
-                          addiction: Option[Boolean] = None,
-                          sexualDiseases: Option[Boolean] = None,
-                          stigmatizeDiseases: Option[Boolean] = None,
-                          vulnerablePopulations: Option[Boolean] = None,
-                          psychologicalTraits: Option[Boolean] = None,
-                          nonBiomedical: Option[Boolean] = None
-                        )
+    generalUse: Option[Boolean] = None,
+    hmbResearch: Option[Boolean] = None,
+    diseaseRestrictions: Option[Seq[String]] = None,
+    populationOriginsAncestry: Option[Boolean] = None,
+    populationStructure: Option[Boolean] = None,
+    commercialUse: Option[Boolean] = None,
+    methodsResearch: Option[Boolean] = None,
+    aggregateResearch: Option[String] = None,
+    controlSetOption: Option[String] = None,
+    gender: Option[String] = None,
+    pediatric: Option[Boolean] = None,
+    populationRestrictions: Option[Seq[String]] = None,
+    dateRestriction: Option[String] = None,
+    recontactingDataSubjects: Option[Boolean] = None,
+    recontactMay: Option[String] = None,
+    recontactMust: Option[String] = None,
+    genomicPhenotypicData: Option[String] = None,
+    otherRestrictions: Option[Boolean] = None,
+    cloudStorage: Option[String] = None,
+    ethicsApprovalRequired: Option[Boolean] = None,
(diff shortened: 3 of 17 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -4265,6 +4283,48 @@ definitions:
...
+        description: If future use is limited to research involving a particular gender, specify "male" or "female". Any other value will be read as false.
+      pediatricResearchOnly:
+        type: boolean
+        description: Is future use limited to pediatric research?
+      IRB:
+        type: boolean
+        description: Should the requester provide documentation of local IRB/REC approval?
+      prefix:
+        type: string
+        description: A prefix that is prepended to the field names of the responses. This is purely for the user's personalization.
```
ansingh7115: Currently a prefix is required - should it be optional instead?

src/main/scala/org/broadinstitute/dsde/firecloud/model/DUOS.scala
```diff
@@ -93,6 +95,32 @@ object DUOS {
...
+
+  case class StructuredDataResponse(consentCodes: Array[String],
+                                    dulvn: Double, //what the heck is this?
+                                    prefix: String,
+                                    structuredUseRestriction: Map[AttributeName, Attribute]) {
+    def formatWithPrefix(): Map[String, JsValue] = {
+      implicit val impAttributeFormat = new AttributeFormat with PlainArrayAttributeListSerializer
+      Map(prefix + "consentCodes" -> consentCodes.toJson,
+        prefix + "dulvn" -> dulvn.toJson,
+        prefix + "structuredUseRestriction" -> structuredUseRestriction.toJson)
```
ansingh7115: Currently, this is the simplest implementation of adding a prefix to each field. There are a couple of other options though: 

(a) Don't use a case class - instead of creating a `StructuredDataResponse`, just turn it into a Map[String, JsValue] directly

(b) Write function that takes in any object (Product) and a prefix (String) and turns it into a Map[String, JsValue]. The problem is that .toJson get weird when you apply it to something of type Any. I sort of started going down that path - (see comment below in ModelJsonProtocol.scala) but the turning Any to json got a bit hairy.

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -12,6 +12,30 @@ import spray.json._
...
+  val DSURL = "DS_URL"
+  val RSPOP = "RS-POP"
+  val DS = "DS"
+
+  val booleanCodes = Seq(GRU, HMB, NCU, NPU, NMDS, NAGR, NCTRL, RSPD, IRB)
+  val genderCodes = Seq(RSG, RSFM, RSM)
+  val duRestrictionFieldNames = booleanCodes ++ genderCodes ++ Seq(DSURL, RSPOP)
+  val allDurFieldNames = duRestrictionFieldNames ++ Seq(DS)
+  val diseaseLabelsAttributeName: AttributeName = AttributeName.withLibraryNS(DS)
+}
```
ansingh7115: Shoved everything into an enum. If we're good with this, I can replace other stuff in this file with the vals in here.

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -141,6 +164,30 @@ trait DataUseRestrictionSupport extends LazyLogging {
...
+//        case JsString(s) => s
+//        case JsTrue => true
+//        case JsFalse => false
+//      }
+//    }
+//
+//    val values = product.productIterator
+//    product.getClass.getDeclaredFields.map( prefix + _.getName -> values.next.toJson ).toMap
+//  }
+
```
ansingh7115: This is what I tried to do to generalize adding a prefix to the field names of an object. 

The problem is values.next is of type `Any`. It's all fun and games until your value isn't atomic. So if I have an Any that might be an Array[Any], I'm unsure how to define the json write function to handle that. 

Ideas and suggestions are welcome

We can also just scratch this and go with the options in my comment above in DUOS.scala

src/main/scala/org/broadinstitute/dsde/firecloud/model/DUOS.scala
```diff
@@ -93,6 +95,32 @@ object DUOS {
     }
   }
 
+
```
ansingh7115: I shoved this in DUOS.scala because that's what made sense to me - but I can move it elsewhere.

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -154,6 +202,7 @@ trait DataUseRestrictionSupport extends LazyLogging {
       case (attr: AttributeName, AttributeBoolean(true)) => attr.name
     }.toSeq
 
+    //this isn't actually calling ontologyDAO?
     val dsLabels:Seq[String] = (workspace.attributes.get(diseaseLabelsAttributeName) collect {
       case value: AttributeValueList => value.list.collect {
         case a: AttributeString => "DS:" + a.value
```
ansingh7115: `generateUseRestrictionDisplayAttribute` is supposed to get the data use attributes out of the attributes of the workspace and then package into something like `Map("NPU", "HMB", "RS-G", "DS:some cancer")`. However, this part which appends the disease to "DS:" doesn't call an ontologyDAO, meaning the attributes already have the disease them in them? This is a bit different from the way our DS attributes are populated for the display field (we're calling ontologyDAO to get the disease names). Which makes me wonder if we should keep these two sections separate instead of refactoring them together?

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -119,6 +142,7 @@ trait DataUseRestrictionSupport extends LazyLogging {
       case _ => Map.empty[AttributeName,Attribute]
     }
 
+    // this is inconsistent with our API
```
ansingh7115: The below section `val rsg = ...` is consistent with how we're treating gender codes for our data use api.

For ours, if RS-G is, for example, "female" the codes would be Map( "RS-G" -> true, "RS-FM" -> true, "RS-M" -> false). HERE, however, the result excludes RS-M rather than setting it as false. Which do we want to go with? If it's gender-specified, do we include the other gender that's false?

rushtong: It is the case that no value means not specifying it at all and not specifying something should mean that it does **NOT** show up as a consent code. I think the end result of the current code is correct ... you should never get a `library:consentCodes:RS-M` tag in a result set in both of the cases - no value and a false value. So, setting it to false might be nice for clarity, I don't think it's necessary for the correct results.

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
rushtong: It would be nice to somehow reuse this enum, and/or colocate it, with the similar fields in `DataUse.ResearchPurpose`. The `DataUse` object was intended to be the home for exactly this kind of information.

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
rushtong: When this method is called, the disease name attributes have already been pulled from ontology otherwise they wouldn't exist. IIRC, all data-use specific attributes would have been populated during the publishing process so this method would only be looking at attributes that exist post-publish.

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -176,6 +225,60 @@ trait DataUseRestrictionSupport extends LazyLogging {
...
+
+    // convert to array of consent codes
+    val consentCodes = consentMap.filter(_._2.value).map(_._1.name).toArray ++ diseaseCodesArray
+
+    //formatWithPrefix(request.prefix, StructuredDataResponse(consentCodes, 1.0, consentMap ++ Map(ConsentCodes.DS -> AttributeValueList(request.diseaseUseOnly.map(AttributeNumber(_))))))
+    StructuredDataResponse(consentCodes, 1.0, request.prefix, consentMap ++ Map(AttributeName.withDefaultNS(ConsentCodes.DS) -> AttributeValueList(request.diseaseUseOnly.map(AttributeNumber(_))))).formatWithPrefix
+  }
+
+  private def getDiseaseNames(diseaseCodes: Array[Int], ontologyDAO: OntologyDAO): Array[String] = {
+    getDiseaseNames(diseaseCodes.map("http://purl.obolibrary.org/obo/DOID_" + _.toString), ontologyDAO)
```
rushtong: I'm concerned about this. For example, if we have something tagged with Pharynx Cancer, http://purl.obolibrary.org/obo/DOID_0060119, the `int` part of that DOID is `60119`, not `0060119` and http://purl.obolibrary.org/obo/DOID_60119 is not a valid DOID url.

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -176,6 +225,60 @@ trait DataUseRestrictionSupport extends LazyLogging {
...
+  }
+
+  private def getDiseaseNames(diseaseCodes: Array[String], ontologyDAO: OntologyDAO): Array[String] = {
+   diseaseCodes.map { nodeid =>
+      ontologyDAO.search(nodeid) match {
+        case termResource :: Nil => termResource.label
+        case _ =>  "" // return an error if the int does not correlate to a disease?
```
rushtong: I think you want to log a warning here so we can address it from the sentry logs. It is not uncommon for nodes to be deprecated, so we need to inform the catalog administrators (i.e. Moran and Jonathan) that something is wrong in this particular case. It might be nice to log the specific request for this case as well, but you'd have to re-work the method chain a little bit to make that happen.

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -176,6 +225,60 @@ trait DataUseRestrictionSupport extends LazyLogging {
...
+    }
+  }
+
+
+  private def getGenderCodeMap(rsg: String): Map[AttributeName, AttributeBoolean] = {
+    rsg.toLowerCase match {
+      case "female" =>
+        Map(AttributeName.withDefaultNS(ConsentCodes.RSG) -> AttributeBoolean(true),
+          AttributeName.withDefaultNS(ConsentCodes.RSFM) -> AttributeBoolean(true),
+          AttributeName.withDefaultNS(ConsentCodes.RSM) -> AttributeBoolean(false))
```
rushtong, line 240: Setting a false value goes back to an earlier comment - what is the effect of a consent code having a false value vs no value at all? Do we use false values in any way? If so, then this makes sense. If not, then it's just extraneous data that we don't need to store.

src/main/scala/org/broadinstitute/dsde/firecloud/model/DUOS.scala
```diff
@@ -93,6 +95,32 @@ object DUOS {
...
+                                   healthMedicalBiomedicalUseOnly: Boolean,
+                                   diseaseUseOnly: Array[Int],
+                                   commercialUseProhibited: Boolean,
+                                   forProfitUseProhibited: Boolean,
+                                   methodsResearchProhibited: Boolean,
+                                   aggregateLevelDataProhibited: Boolean,
+                                   controlsUseProhibited: Boolean,
+                                   genderUseOnly: String,
+                                   pediatricResearchOnly: Boolean,
+                                   IRB: Boolean,
```
rushtong: Nit: I would case this similarly to the other codes and maybe even name it better, `irbRequired` for example.

rushtong (commented):
Mostly some logical comments inline that may be out of date with respect to recent requirements. Feel free to let me know if you want me to re-review in the future.

src/main/scala/org/broadinstitute/dsde/firecloud/model/DUOS.scala
breilly2: I put mine in `DataUse.scala`, mostly because it was so closely related to something already in there. I'm also not sure if that's where it belongs.

src/main/scala/org/broadinstitute/dsde/firecloud/model/DUOS.scala
breilly2: I tend to agree. Worth noting that this affects the request payload so it's part of the service API. As it is now adheres to the tech doc so changing it here would necessitate changing it there too.

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -154,6 +202,7 @@ trait DataUseRestrictionSupport extends LazyLogging {
       case (attr: AttributeName, AttributeBoolean(true)) => attr.name
     }.toSeq
 
+    //this isn't actually calling ontologyDAO?
```
breilly2: For what reason? It looks like everything needed here is in the workspace attribute values.

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
breilly2: So we need to use `String` instead of `Int` to represent these things. I wonder if there's any reason to use just the `0060119` part or if we should always use the full `http://purl.obolibrary.org/obo/DOID_0060119`.

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
breilly2: I'm very interested in learning how we handle deprecation. This feels like a potentially big problem that we (and anyone else using this API) will have to deal with.

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
aednichols: One reason to use the whole URI is namespacing - it positively identifies which ontology you're referencing. This is similar to the use of URI namespaces in XML.

Besides, it's just darn convenient when debugging 😄

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -2,16 +2,40 @@ package org.broadinstitute.dsde.firecloud.service
...
 import org.broadinstitute.dsde.rawls.model.Attributable.AttributeMap
 import org.broadinstitute.dsde.rawls.model.WorkspaceJsonSupport.AttributeNameFormat
-import org.broadinstitute.dsde.rawls.model._
+import org.broadinstitute.dsde.rawls.model.{Attribute, _}
 import spray.json.DefaultJsonProtocol._
 import spray.json._
 
 import scala.util.Try
 
+object ConsentCodes extends Enumeration {
```
breilly2: You might be able to make better use of `Enumeration`. If you define the values like:
```
val GRU = Value("GRU")
```
and add a function here:
```
  def withNameIgnoreCase(s: String): Value =
    values.find(_.toString.equalsIgnoreCase(s)).getOrElse(throw new NoSuchElementException(s"No value found for '$s'"))
```
then you can match against the enum values:
```
ConsentCodes.withNameIgnoreCase(attr.name) match {
  case ConsentCodes.NAGR => ...
```
The downside is that you have to use `toString` everywhere to get the string value out.

Alternately, you can just remove the `extends Enumeration` here and leave it as a bag of string constants.

src/main/scala/org/broadinstitute/dsde/firecloud/model/DUOS.scala
```diff
@@ -93,6 +95,32 @@ object DUOS {
...
+                                   prefix: String)
+
+  case class StructuredDataResponse(consentCodes: Array[String],
+                                    dulvn: Double, //what the heck is this?
```
breilly2: According to `attribute-definitions.json`:

> Structured Data Use Limitations Version Number

Though that describes it as an integer. Being a version number, we should use an integer (or a string if we're using \<major\>.\<minor\>) rather than a double.

src/test/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupportSpec.scala
```diff
@@ -17,6 +18,117 @@ class DataUseRestrictionSupportSpec extends FreeSpec with Matchers with DataUseR
...
+            forProfitUseProhibited = true,
+            methodsResearchProhibited = true,
+            aggregateLevelDataProhibited = true,
+            controlsUseProhibited = true,
+            genderUseOnly = "female",
+            pediatricResearchOnly = true,
+            IRB = true,
+            prefix = "blah")
+
+          val expected = Map("blahconsentCodes" -> Array("NPU","RS-G","NCU","HMB","RS-FM","NCTRL","RS-PD","IRB","NAGR","GRU","NMDS","DS:Ebola hemorrhagic fever","DS:hematologic cancer").toJson,
```
breilly2: Maybe more of a requirements question but I assumed that the prefix should have some sort of delimiter such as `:`, so `blahconsentCodes` should be `blah:consentCodes`.

ansingh7115: The example in the API Doc indicated that there wasn't an in-built delimiter. (the example prefix was "library:" in the request) I assume this was so that users could choose their own delimiters.

src/main/resources/swagger/api-docs.yaml
breilly2: Optional would make sense to me. I'm sure we'll see people passing in `""` if we make it required.

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
breilly2: From what I can tell, we _only_ use a duo of codes when calculating data use from DUOS. We use a trio of codes when building attributes for storing in ES. Furthermore, there is code that fills in `false` for missing attributes when calculating data use restriction field to index in ES... but then doesn't fill in `false` when calculating the _display_ data use restriction field for ES.

Maybe this was all intentional, but it's really difficult to be certain that there isn't a subtle bug here just from reading the code.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/LibraryApiService.scala
```diff
@@ -54,6 +55,19 @@ trait LibraryApiService extends HttpService with FireCloudRequestBuilding
             externalHttpPerRequest(requestContext, extReq)
           }
         } ~
+        pathPrefix("dataUse") {
```
breilly2: We need to settle on what the prefix will be. It was originally `dataUse` but the autocomplete was `duos`. I suggested putting autocomplete under `dataUse` which was done, but now the tech doc says `duos` for all 3 endpoints.

src/main/resources/library/attribute-definitions.json
```diff
@@ -325,13 +325,6 @@
       "indexable": false,
       "consentCode": "RS-PD"
     },
-    "library:RS-POP": {
```
breilly2, line 328: Why is this going away?

src/main/resources/swagger/api-docs.yaml
```diff
@@ -3490,6 +3490,24 @@ paths:
         500:
           description: Internal Server Error
 
+  /api/dataUse/structuredData:
```
breilly2: This shouldn't be under `/api`, right?
Also, are we going with `dataUse` or `duos`? @ahaessly

src/main/scala/org/broadinstitute/dsde/firecloud/model/DataUse.scala
```diff
@@ -1,10 +1,11 @@
 package org.broadinstitute.dsde.firecloud.model
 
+import org.broadinstitute.dsde.rawls.model.AttributeName
 import spray.http.Uri
 
 object DataUse {
 
-  private final val doid_prefix = "http://purl.obolibrary.org/obo/DOID_"
+  final val doid_prefix = "http://purl.obolibrary.org/obo/DOID_"
```
breilly2: Interesting. Where do you need this outside of `DataUse`? I'm inclined to say it should stay private.

src/main/scala/org/broadinstitute/dsde/firecloud/model/DataUse.scala
```diff
@@ -30,3 +31,26 @@ object DataUse {
...
+  val RSFM = "RS-FM"
+  val RSM = "RS-M"
+  val DSURL = "DS_URL"
+  val DS = "DS"
+
+  val booleanCodes = Seq(GRU, HMB, NCU, NPU, NMDS, NAGR, NCTRL, RSPD, IRB)
+  val genderCodes = Seq(RSG, RSFM, RSM)
+  val duRestrictionFieldNames = booleanCodes ++ genderCodes ++ Seq(DSURL/*, RSPOP*/)
+  val allDurFieldNames = duRestrictionFieldNames ++ Seq(DS)
+  val diseaseLabelsAttributeName: AttributeName = AttributeName.withLibraryNS(DS)
```
breilly2: Like the other one we were working with this afternoon, if this is used in only one place (or one class/file), it might be better living there. I'm thinking that based on the fact that this is an `AttributeName` which makes it opinionated about the namespace which probably isn't appropriate here.

src/main/scala/org/broadinstitute/dsde/firecloud/model/DataUse.scala
```diff
@@ -30,3 +31,26 @@ object DataUse {
...
+  val IRB = "IRB"
+  val RSG = "RS-G"
+  val RSFM = "RS-FM"
+  val RSM = "RS-M"
+  val DSURL = "DS_URL"
+  val DS = "DS"
+
+  val booleanCodes = Seq(GRU, HMB, NCU, NPU, NMDS, NAGR, NCTRL, RSPD, IRB)
+  val genderCodes = Seq(RSG, RSFM, RSM)
+  val duRestrictionFieldNames = booleanCodes ++ genderCodes ++ Seq(DSURL/*, RSPOP*/)
```
breilly2: Did you mean to leave `RSPOP` commented out?

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -110,15 +156,16 @@ trait DataUseRestrictionSupport extends LazyLogging {
...
-    val rspop = duosDataUse.populationRestrictions match {
-      case Some(Seq()) => Map.empty[AttributeName,Attribute]
-      case Some(populations) =>
-        Map(
-          AttributeName.withLibraryNS("RS-PD") -> AttributeValueList(populations.map(AttributeString))
-        )
-      case _ => Map.empty[AttributeName,Attribute]
-    }
-
+//    val rspop = duosDataUse.populationRestrictions match {
```
breilly2: Should this all be commented out?

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -173,65 +223,94 @@ trait DataUseRestrictionSupport extends LazyLogging {
...
-          if (diseaseNumericIdValues.nonEmpty)
-            Map(AttributeName.withDefaultNS("DS") -> AttributeValueList(diseaseNumericIdValues.map { n => AttributeNumber(n) }))
-          else
-            Map.empty[AttributeName, Attribute]
-        case (attr: AttributeName, value: AttributeValueList) if attr.name.equals("RS-POP") => Map(AttributeName.withDefaultNS(attr.name) -> value)
-        case unmatched =>
-          logger.warn(s"Unexpected library data use attribute type: (workspace-id: ${workspace.workspaceId}, attribute name: ${unmatched._1.name}, attribute value: ${unmatched._2.toString})")
-          Map.empty[AttributeName, Attribute]
+          }.flatten.toArray.map(_.toString)
+//          if (diseaseNumericIdValues.nonEmpty)
```
breilly2: Should this be commented out?

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
```diff
@@ -102,9 +102,9 @@ trait LibraryServiceSupport extends DataUseRestrictionSupport with LazyLogging {
       case None => Map()
     }
 
-    val durAttributeNames = allDurFieldNames.map(AttributeName.withLibraryNS)
-    val dur: Map[AttributeName, Attribute] = generateStructuredUseRestrictionAttribute(workspace)
-    val displayDur: Map[AttributeName, Attribute] = generateUseRestrictionDisplayAttribute(workspace)
+    val durAttributeNames = ConsentCodes.allDurFieldNames.map(AttributeName.withLibraryNS)
+    val structuredAndDisplayAttributes = generateStructuredAndDisplayAttributes(workspace, ontologyDAO)
+    val (dur, displayDur) = (structuredAndDisplayAttributes.structured, structuredAndDisplayAttributes.display)
```
breilly2: Nice!

src/test/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupportSpec.scala
```diff
@@ -17,102 +20,214 @@ class DataUseRestrictionSupportSpec extends FreeSpec with Matchers with DataUseR
...
-          val durs: Map[String, DataUseRestriction] = diseaseDatasets.flatMap { ds =>
-            Map(ds.name -> makeDurFromWorkspace(ds))
-          }.toMap
-
-          Seq("DS").foreach { code =>
-            val dur: DataUseRestriction = durs(code)
-            checkDiseaseValues(dur, code)
-          }
-        }
+//        "dataset should have a fully populated data use restriction attribute" in {
```
breilly2: Lots of test cases commented out here that I'm thinking shouldn't be:
* dataset should have a fully populated data use restriction attribute
* dur should have appropriate gender codes populated
* dur should have appropriate NAGR code populated
* dataset should have a true value for the consent code for which it was specified
* dataset should have the correct list values for the consent code for which it was specified
* dataset should have the correct disease values for the consent code for which it was specified
* dataset should not have any data use restriction for empty attributes
* dataset should not have any data use restriction for non-library attributes
* valid datasets should have some form of data use display attribute
* datasets with single boolean code should have that single display code

src/test/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupportSpec.scala
```diff
@@ -122,32 +237,35 @@ class DataUseRestrictionSupportSpec extends FreeSpec with Matchers with DataUseR
...
+//            // Boolean datasets are named with the same code value
+//            codes should contain theSameElementsAs Seq(ds.name)
+//          }
+//        }
 
         "'EVERYTHING' dataset should have the right codes" in {
-          val attrs: Map[AttributeName, Attribute] = generateUseRestrictionDisplayAttribute(everythingDataset.head)
+          val ontologyDAO = new MockOntologyDAO
+          val attrs = generateStructuredAndDisplayAttributes(everythingDataset.head, ontologyDAO).display
+          //attrs should be (1)
```
breilly2: Remove commented-out debugging code.

src/test/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupportSpec.scala
```diff
@@ -325,16 +446,10 @@ class DataUseRestrictionSupportSpec extends FreeSpec with Matchers with DataUseR
         }
         "should handle populationRestrictions string lists" in {
           val attrs = generateStructuredUseRestrictionAttribute(new DuosDataUse(
-            populationRestrictions = Some(Seq(
-              "population 1",
-              "population 2"
-            ))
+            pediatric = Some(true)
```
breilly2, line 437: This looks like a very different test now. Why?

src/test/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSpec.scala
```diff
@@ -439,7 +439,7 @@ class LibraryServiceSpec extends BaseServiceSpec with FreeSpecLike with LibraryS
         "NCTRL" -> JsBoolean(false),
         "GRU" -> JsBoolean(false),
         "HMB" -> JsBoolean(false),
-        "RS-POP" -> JsArray(),
+        //"RS-POP" -> JsArray(),
```
breilly2: Okay... I finally asked and got the story about RS-POP. Probably worth removing the code instead of commenting out?

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -209,7 +208,7 @@ trait DataUseRestrictionSupport extends LazyLogging {
     }.toSeq
 
 
-    val dsLabels:Seq[String] = (attributes.get(AttributeName.withDefaultNS(DS)) collect {
+    val dsLabels:Seq[String] = (attributes.get(AttributeName.withDefaultNS(ConsentCodes.DSURL)) collect {
```
breilly2: This is probably the right thing to do, but I think it needs a corresponding change to `getDiseaseNames` below to not prepend `DataUse.doid_prefix` when calling `OntologyDao.search`.

src/main/resources/library/attribute-definitions.json
breilly2: Follow-up: I got an answer. It wasn't really being used so we want to clean it up now rather than expose it in a public API.

src/main/scala/org/broadinstitute/dsde/firecloud/model/DataUse.scala
breilly2: Answer: yes, it is being removed.

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
breilly2: Answer: RS-POP is going away so it should be safe to remove this.

breilly2 (commented):
Mostly a nit about `diseaseLabelsAttributeName` but also some commented out code in `DataUseRestrictionSupport` that doesn't appear to be related to RS-POP and possibly suspicious changes to a test that I'd like to understand before giving thumbs.

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -146,22 +198,25 @@ trait DataUseRestrictionSupport extends LazyLogging {
...
+  def transformUseRestrictionDisplayAttribute(attributes: Map[AttributeName, Attribute], ontologyDAO: OntologyDAO): Map[AttributeName, Attribute] = {
 
-    val booleanCodes:Seq[String] = getDataUseAttributes(workspace).collect {
+    val booleanCodes:Seq[String] = attributes.collect {
       case (attr: AttributeName, AttributeBoolean(true)) => attr.name
     }.toSeq
 
-    val dsLabels:Seq[String] = (workspace.attributes.get(diseaseLabelsAttributeName) collect {
+
+    val dsLabels:Seq[String] = (attributes.get(AttributeName.withDefaultNS(ConsentCodes.DSURL)) collect {
```
breilly2, line 202: This is probably the right thing to do, but I think it needs a corresponding change to `getDiseaseNames` below to not prepend `DataUse.doid_prefix` when calling `OntologyDao.search`.

(GitHub is ridiculous. I originally posted this comment while viewing a subset of commits so it wasn't showing up here. 😡

src/main/scala/org/broadinstitute/dsde/firecloud/model/DataUse.scala
```diff
@@ -26,6 +27,10 @@ object DataUse {
   case class DiseaseOntologyNodeId(uri: Uri, numericId: Int)
   object DiseaseOntologyNodeId {
     def apply(stringid:String) = {
+      println("id " + stringid)
+      println("pre " + doid_prefix)
+      println("result " + stringid.startsWith(doid_prefix))
```
breilly2: `println`

src/main/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupport.scala
```diff
@@ -33,35 +37,87 @@ trait DataUseRestrictionSupport extends LazyLogging {
...
         val allAttrs = existingAttrs ++ booleanAttrs ++ listAttrs
         Map(structuredUseRestrictionAttributeName -> AttributeValueRawJson.apply(allAttrs.toJson.compactPrint))
     }
+  }
+
+  def generateStructuredAndDisplayAttributes(workspace: Workspace, ontologyDAO: OntologyDAO): UseRestriction = {
+    getDataUseAttributes(workspace)  match {
+      case None => UseRestriction(Map.empty[AttributeName, Attribute],Map.empty[AttributeName, Attribute])
+      case Some(request) => {
+        println("request" + request)
```
breilly2: `println`

src/main/scala/org/broadinstitute/dsde/firecloud/model/DataUse.scala
ansingh7115: Still doing some work on this, will let you know when it's back in reviewable state!

src/test/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupportSpec.scala
ansingh7115: There was a (potential) bug that was translating RS-POP to RS-PD codes. @ahaessly is this truly a bug? If so, should I change the test to check pediatric codes (as I did here) or simply delete the test?

ansingh7115: Confirmed that it was a real bug.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -4282,6 +4302,59 @@ definitions:
...
+        - generalResearchUse
+        - healthMedicalUseOnly
+        - commercialUseProhibited
+        - forProfitUseProhibited
+        - methodsResearchProhibited
+        - aggregateLevelDataProhibited
+        - controlsUseProhibited
+        - genderUseOnly
+        - pediatricResearchOnly
+        - irbRequired
```
ahaessly: change the name of the 3 attributes that end in Only to end it Required

src/main/scala/org/broadinstitute/dsde/firecloud/model/DataUse.scala
```diff
@@ -59,4 +64,55 @@ object DataUse {
...
+                                   commercialUseProhibited: Boolean,
+                                   forProfitUseProhibited: Boolean,
+                                   methodsResearchProhibited: Boolean,
+                                   aggregateLevelDataProhibited: Boolean,
+                                   controlsUseProhibited: Boolean,
+                                   genderUseOnly: String,
+                                   pediatricResearchOnly: Boolean,
+                                   irbRequired: Boolean,
+                                   prefix: Option[String])
+
```
ahaessly: also in the code, change the names that end in Only to end in Required

src/test/resources/reference.conf
```diff
@@ -74,6 +74,7 @@ elasticsearch {
 duos {
   baseConsentUrl = "http://localhost:8992"
   baseOntologyUrl = "http://localhost:8993"
+  dulvn = 3
```
ahaessly: i think this also need to be 4

breilly2: This is really interesting. Since this is in a test resource, we should be considering whether we need to maintain tests for multiple versions. Unless someone can think of a problem that the current changes would cause, I propose that we don't worry about it right now but keep it in mind for the future.

ahaessly (approved):
👍 woohoo!

src/test/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionSupportSpec.scala
breilly2: Wow. Good catch!

src/test/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionTestFixtures.scala
```diff
@@ -37,7 +36,7 @@ object DataUseRestrictionTestFixtures {
     mkWorkspace(attributes, code, s"{${code.replace("-","")}}-unique")
   }
 
-  val listCodes: Seq[String] = Seq("RS-POP")
+  val listCodes: Seq[String] = Seq()
```
breilly2: There are quite a few tests that are driven by `listCodes` and `listDatasets`. Now that it's empty, those tests do little to nothing. We need to understand if those tests need to be modified or if they are now irrelevant. I haven't looked too closely yet but my gut tells me there are enough of them that they can't all be irrelevant (but I could be wrong).

breilly2 (changes requested):
I hate to do this because I know we want to get this merged, but we need to make sure the tests we have are right.

src/test/resources/reference.conf
ansingh7115: Oh, good catch :) 

I think @ahaessly said we don't need to worry about previous versions? (it was an earlier conversation today)

ahaessly: yes, that is correct. moran has said we do not yet need to worry about maintaining compatibility with previous versions.

src/test/scala/org/broadinstitute/dsde/firecloud/service/DataUseRestrictionTestFixtures.scala
ahaessly: because we removed RS-POP. we don't have any generic lists to test. (we do test disease lists with a different, more specific fixture). if we re-introduce a code that contains a list, we will probably want these tests again. but currently we don't need them.

ahaessly:
jenkins retest

ansingh7115:
jenkins retest

ansingh7115:
jenkins retest

ansingh7115:
jenkins retest

breilly2 (approved):
👍

PR #624, opened 2018-05-09
GAWB-3557: Added /duos/researchPurposeQuery endpoint.

breilly2:
There's one bit left to do... making use of the prefix from the request... which I'll work on after reviewing Anu's PR.

Update 5/14: Prefix support now in.
Update 5/15: Switched to full DOIDs in request rather than just the integer part.

In all cases:

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -32,6 +32,7 @@ object Dependencies {
     "io.spray"                      %% "spray-client"              % sprayV,
     "io.spray"                      %% "spray-routing-shapeless23" % sprayV,
     "io.spray"                      %% "spray-testkit"             % sprayV   % "test",
+    "net.virtual-void"              %% "json-lenses"               % "0.6.2"  % "test",
     "com.typesafe.akka"             %% "akka-actor"                % akkaV,
     "com.typesafe.akka"             %% "akka-testkit"              % akkaV    % "test",
     "com.typesafe.akka"             %% "akka-slf4j"                % akkaV,
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -3464,6 +3464,23 @@ paths:
           description: Internal Server Error
       security: []
 
+  /duos/researchPurposeQuery:
+    post:
+      tags:
+      - Library
+      operationId: duosResearchPurposeQuery
+      summary: Generates an ElasticSearch clause for filtering by research purpose
+      parameters:
+        - in: body
+          name: body
+          description: Research Purpose Request
+          required: true
+          schema:
+            $ref: '#/definitions/ResearchPurposeRequest'
+      responses:
+        200:
+          description: JSON of ES filter clause
+
   /api/duos/consent/orsp/{orspId}:
     get:
       tags:
@@ -5267,6 +5284,33 @@ definitions:
         type: boolean
         default: false
         description: Commercial purpose/by a commercial entity
+
+  ResearchPurposeRequest:
+    type: object
+    properties:
+      DS:
+        type: array
+        default: []
+        items:
+          type: integer
+      NMDS:
+        type: boolean
+        default: false
+      NCTRL:
+        type: boolean
+        default: false
+      NAGR:
+        type: boolean
+        default: false
+      POA:
+        type: boolean
+        default: false
+      NCU:
+        type: boolean
+        default: false
+      prefix:
+        type: string
+        default: ""
 
   SearchTermRef:
     type: object
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -5292,23 +5292,34 @@ definitions:
         type: array
         default: []
         items:
-          type: integer
+          type: string
+          example: "http://purl.obolibrary.org/obo/DOID_4"
+        description: >
+          Disease focused research. Values are DOID ontology node ids in the form
+          http://purl.obolibrary.org/obo/DOID_NNN. Specify an empty array if this
+          is not disease focused research.
       NMDS:
         type: boolean
         default: false
+        description: Methods development/Validation study
       NCTRL:
         type: boolean
         default: false
+        description: Control set
       NAGR:
         type: boolean
         default: false
+        description: Aggregate analysis to understand variation in the general population
       POA:
         type: boolean
         default: false
+        description: Study population origins or ancestry
       NCU:
         type: boolean
         default: false
+        description: Commercial purpose/by a commercial entity
       prefix:
+        description: Prefix for query properties. Must match prefix used when indexing.
         type: string
         default: ""
 
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/Application.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/Application.scala
@@ -17,6 +17,7 @@ case class Application(agoraDAO: AgoraDAO,
                        rawlsDAO: RawlsDAO,
                        samDAO: SamDAO,
                        searchDAO: SearchDAO,
+                       researchPurposeDAO: ResearchPurposeDAO,
                        thurloeDAO: ThurloeDAO,
                        trialDAO: TrialDAO,
                        logitDAO: LogitDAO)
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/Application.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/Application.scala
@@ -17,7 +17,7 @@ case class Application(agoraDAO: AgoraDAO,
                        rawlsDAO: RawlsDAO,
                        samDAO: SamDAO,
                        searchDAO: SearchDAO,
-                       researchPurposeDAO: ResearchPurposeDAO,
+                       researchPurposeSupport: ResearchPurposeSupport,
                        thurloeDAO: ThurloeDAO,
                        trialDAO: TrialDAO,
                        logitDAO: LogitDAO)
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ESResearchPurposeDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ESResearchPurposeDAO.scala
@@ -1,0 +1,11 @@
+package org.broadinstitute.dsde.firecloud.dataaccess
+
+import org.broadinstitute.dsde.firecloud.model.DataUse.ResearchPurpose
+import org.elasticsearch.index.query.BoolQueryBuilder
+
+import scala.concurrent.ExecutionContext.Implicits.global
+
+class ESResearchPurposeDAO(ontologyDAO: OntologyDAO) extends ResearchPurposeDAO with ElasticSearchDAOResearchPurposeSupport {
+  def researchPurposeFilters(researchPurpose: ResearchPurpose): BoolQueryBuilder =
+    researchPurposeFilters(researchPurpose, ontologyDAO)
+}
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
@@ -28,7 +28,7 @@ import scala.concurrent.{ExecutionContext, Future}
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.util.{Failure, Success, Try}
 
-class ElasticSearchDAO(client: TransportClient, indexName: String, researchPurposeDAO: ResearchPurposeDAO) extends SearchDAO with ElasticSearchDAOSupport with ElasticSearchDAOQuerySupport {
+class ElasticSearchDAO(client: TransportClient, indexName: String, researchPurposeSupport: ResearchPurposeSupport) extends SearchDAO with ElasticSearchDAOSupport with ElasticSearchDAOQuerySupport {
 
   private final val datatype = "dataset"
 
@@ -119,11 +119,11 @@ class ElasticSearchDAO(client: TransportClient, indexName: String, researchPurpo
   }
 
   override def findDocuments(criteria: LibrarySearchParams, groups: Seq[String]): Future[LibrarySearchResponse] = {
-    findDocumentsWithAggregateInfo(client, indexName, criteria, groups, researchPurposeDAO)
+    findDocumentsWithAggregateInfo(client, indexName, criteria, groups, researchPurposeSupport)
   }
 
   override def suggestionsFromAll(criteria: LibrarySearchParams, groups: Seq[String]): Future[LibrarySearchResponse] = {
-    autocompleteSuggestions(client, indexName, criteria, groups, researchPurposeDAO)
+    autocompleteSuggestions(client, indexName, criteria, groups, researchPurposeSupport)
   }
 
   override def suggestionsForFieldPopulate(field: String, text: String): Future[Seq[String]] = {
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
@@ -28,7 +28,7 @@ import scala.concurrent.{ExecutionContext, Future}
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.util.{Failure, Success, Try}
 
-class ElasticSearchDAO(client: TransportClient, indexName: String, ontologyDAO: OntologyDAO) extends SearchDAO with ElasticSearchDAOSupport with ElasticSearchDAOQuerySupport {
+class ElasticSearchDAO(client: TransportClient, indexName: String, researchPurposeDAO: ResearchPurposeDAO) extends SearchDAO with ElasticSearchDAOSupport with ElasticSearchDAOQuerySupport {
 
   private final val datatype = "dataset"
 
@@ -119,11 +119,11 @@ class ElasticSearchDAO(client: TransportClient, indexName: String, ontologyDAO: 
   }
 
   override def findDocuments(criteria: LibrarySearchParams, groups: Seq[String]): Future[LibrarySearchResponse] = {
-    findDocumentsWithAggregateInfo(client, indexName, criteria, groups, ontologyDAO)
+    findDocumentsWithAggregateInfo(client, indexName, criteria, groups, researchPurposeDAO)
   }
 
   override def suggestionsFromAll(criteria: LibrarySearchParams, groups: Seq[String]): Future[LibrarySearchResponse] = {
-    autocompleteSuggestions(client, indexName, criteria, groups, ontologyDAO)
+    autocompleteSuggestions(client, indexName, criteria, groups, researchPurposeDAO)
   }
 
   override def suggestionsForFieldPopulate(field: String, text: String): Future[Seq[String]] = {
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
@@ -66,7 +66,7 @@ trait ElasticSearchDAOQuerySupport extends ElasticSearchDAOSupport {
     */
 
 
-  def createQuery(criteria: LibrarySearchParams, groups: Seq[String], researchPurposeDAO: ResearchPurposeDAO, searchField: String = fieldAll, phrase: Boolean = false): QueryBuilder = {
+  def createQuery(criteria: LibrarySearchParams, groups: Seq[String], researchPurposeSupport: ResearchPurposeSupport, searchField: String = fieldAll, phrase: Boolean = false): QueryBuilder = {
     val query: BoolQueryBuilder = boolQuery // outer query, all subqueries should be added to the must list
     query.must(criteria.searchString match {
       case None => matchAllQuery
@@ -94,7 +94,7 @@ trait ElasticSearchDAOQuerySupport extends ElasticSearchDAOSupport {
       query.must(fieldQuery)
     }
     criteria.researchPurpose map { rp =>
-      query.must(researchPurposeDAO.researchPurposeFilters(rp))
+      query.must(researchPurposeSupport.researchPurposeFilters(rp))
     }
 
     query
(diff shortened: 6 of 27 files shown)


project/Dependencies.scala
```diff
@@ -32,6 +32,7 @@ object Dependencies {
     "io.spray"                      %% "spray-client"              % sprayV,
     "io.spray"                      %% "spray-routing-shapeless23" % sprayV,
     "io.spray"                      %% "spray-testkit"             % sprayV   % "test",
+    "net.virtual-void"              %% "json-lenses"               % "0.6.2"  % "test",
```
breilly2, line 35: This is used in `LibraryApiServiceSpec` to look into the JSON response. Pretty nice! I limited it to test scope (at least for now).

src/main/scala/org/broadinstitute/dsde/firecloud/Application.scala
```diff
@@ -17,6 +17,7 @@ case class Application(agoraDAO: AgoraDAO,
                        rawlsDAO: RawlsDAO,
                        samDAO: SamDAO,
                        searchDAO: SearchDAO,
+                       researchPurposeDAO: ResearchPurposeDAO,
```
breilly2: The code that this feature reuses was originally being pulled in to `ElasticSearchDAO` by extension. This made trying to reuse it (and the tests) feel dirty.

This change is the first of many in this PR that introduces a `ResearchPurposeDAO` and does all the necessary plumbing to make it available in the right places. Now the code has more of a compositional structure which makes the reuse feel cleaner to me.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
```diff
@@ -28,7 +28,7 @@ import scala.concurrent.{ExecutionContext, Future}
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.util.{Failure, Success, Try}
 
-class ElasticSearchDAO(client: TransportClient, indexName: String, ontologyDAO: OntologyDAO) extends SearchDAO with ElasticSearchDAOSupport with ElasticSearchDAOQuerySupport {
+class ElasticSearchDAO(client: TransportClient, indexName: String, researchPurposeDAO: ResearchPurposeDAO) extends SearchDAO with ElasticSearchDAOSupport with ElasticSearchDAOQuerySupport {
```
breilly2: Turned out `ElasticSearchDAO` didn't need `OntologyDAO` for anything other than research purpose queries so it's no longer a direct dependency.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
```diff
@@ -22,7 +22,7 @@ import scala.util.{Failure, Success, Try}
 import scala.util.matching.Regex
 
 
-trait ElasticSearchDAOQuerySupport extends ElasticSearchDAOSupport with ElasticSearchDAOResearchPurposeSupport {
```
breilly2, line 25: No need for ElasticSearchDAOQuerySupport to extend `ElasticSearchDAOResearchPurposeSupport` here anymore.

src/main/scala/org/broadinstitute/dsde/firecloud/model/DataUse.scala
```diff
@@ -29,4 +33,30 @@ object DataUse {
     }
   }
 
+  case class ResearchPurposeRequest(
```
breilly2, line 36: `Option`s allow for defaults if not in the request JSON. Also this adds the `prefix` attribute that is in the tech doc and `ResearchPurpose` didn't already have.

src/main/scala/org/broadinstitute/dsde/firecloud/service/OntologyService.scala
```diff
@@ -4,40 +4,44 @@ import akka.actor.{Actor, Props}
...
+    case ResearchPurposeQuery(researchPurposeRequest: ResearchPurposeRequest) => buildResearchPurposeQuery(researchPurposeRequest) pipeTo sender
   }
 
   def autocompleteOntology(term: String): Future[PerRequestMessage] = {
     Future(RequestComplete(ontologyDAO.autocomplete(term)))
   }
 
-
+  def buildResearchPurposeQuery(request: ResearchPurposeRequest): Future[PerRequestMessage] = {
+    // TODO: Don't ignore request.prefix!
```
breilly2: Yeah, this could be interesting. I missed this until today. I'm going to see if there's anything in Anu's PR that I can borrow.

src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/ResearchPurposeSearchSpec.scala
```diff
@@ -32,16 +32,19 @@ class ResearchPurposeSearchSpec extends FreeSpec with SearchResultValidation wit
...
 
     "Research purpose for aggregate analysis (NAGR)" - {
-      "should return any dataset where NAGR is false and is (GRU or HMB)" in {
-        val researchPurpose = ResearchPurpose.default.copy(NAGR = true)
-        val searchResponse = searchWithPurpose(researchPurpose)
-        validateResultNames(
-          Set("one", "two", "six", "seven", "eleven", "twelve"),
-          searchResponse
-        )
+      val researchPurpose = ResearchPurpose.default.copy(NAGR = true)
```
breilly2, line 35: I extracted the research purpose input and the expected search output in order to run the same tests against the original monolithic search and a simulated external search using the `researchPurposeQuery` function. Ultimately, we may only want one set of these along with some supporting tests to make sure the monolithic search uses the filter code... but maybe not and these tests as they are provide good confidence that both behave the same way.

src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/SearchResultValidation.scala
```diff
@@ -46,6 +49,20 @@ trait SearchResultValidation {
     Await.result(searchDAO.suggestionsFromAll(criteria.copy(size=100), Seq.empty[String]), dur)
   }
 
+  /**
+    * Mimics a 3rd party searching against an ElasticSearch instance using a research purpose filter
+    * from us. Our SearchDAO is used to create and execute the search request, but the research
+    * purpose is fetched identically to how a 3rd party would.
+    */
+  def searchWithResearchPurposeQuery(researchPurpose: ResearchPurpose): SearchResponse = {
```
breilly2: This is what all of the new test cases in `ResearchPurposeSearchSpec` use to pretend to be a 3rd party using the research purpose query endpoint.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
```diff
@@ -119,11 +119,11 @@ class ElasticSearchDAO(client: TransportClient, indexName: String, ontologyDAO:
   }
 
   override def findDocuments(criteria: LibrarySearchParams, groups: Seq[String]): Future[LibrarySearchResponse] = {
-    findDocumentsWithAggregateInfo(client, indexName, criteria, groups, ontologyDAO)
+    findDocumentsWithAggregateInfo(client, indexName, criteria, groups, researchPurposeDAO)
```
breilly2: More plumbing. You'll see a lot of swapping out `ontologyDAO` for `researchPurposeDAO` in this PR. I wish we had an easier way to inject dependencies instead of manually passing them around everywhere.

src/test/scala/org/broadinstitute/dsde/firecloud/service/ElasticSearchDAOQuerySupportSpec.scala
```diff
@@ -28,21 +28,21 @@ class ElasticSearchDAOQuerySupportSpec extends FreeSpec with ElasticSearchDAOQue
   val client: TransportClient = ElasticUtils.buildClient(FireCloudConfig.ElasticSearch.servers, FireCloudConfig.ElasticSearch.clusterName)
 
   // create a mock ontology dao
-  val ontologyDao: OntologyDAO = new MockOntologyDAO
+  val researchPurposeDAO: ResearchPurposeDAO = new MockResearchPurposeDAO
```
breilly2: Nothing but more boring plumbing changes in this file.

src/test/scala/org/broadinstitute/dsde/firecloud/service/LibraryApiServiceSpec.scala
```diff
@@ -368,5 +371,23 @@ class LibraryApiServiceSpec extends BaseServiceSpec with LibraryApiService with
...
+        new RequestBuilder(HttpMethods.POST)(duosResearchPurposeQuery, request) ~> sealRoute(libraryRoutes) ~> check {
+          status should equal(OK)
+        }
+      }
+
+      "POST with disease focus should return OK" in {
+        val request = ResearchPurposeRequest.empty.copy(DS = Some(Seq(1234, 5678)))
+        new RequestBuilder(HttpMethods.POST)(duosResearchPurposeQuery, request) ~> sealRoute(libraryRoutes) ~> check {
+          status should equal(OK)
+          val diseaseIds = responseAs[JsObject].extract[Int]('bool / 'should / * / 'term / "structuredUseRestriction.DS" / 'value)
```
breilly2: It's almost like using something like XPath but for reaching into JSON.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ESResearchPurposeDAO.scala
```diff
@@ -0,0 +1,11 @@
+package org.broadinstitute.dsde.firecloud.dataaccess
+
+import org.broadinstitute.dsde.firecloud.model.DataUse.ResearchPurpose
+import org.elasticsearch.index.query.BoolQueryBuilder
+
+import scala.concurrent.ExecutionContext.Implicits.global
+
+class ESResearchPurposeDAO(ontologyDAO: OntologyDAO) extends ResearchPurposeDAO with ElasticSearchDAOResearchPurposeSupport {
```
breilly2: Yes, this extends `ElasticSearchDAOResearchPurposeSupport`. That code could have moved into here... or `ElasticSearchDAOResearchPurposeSupport` could extend the `ResearchPurposeDAO` trait directly. I chose this path for now in part to both make it clear that the function being reused isn’t changing.

Also, I recognize that this isn’t exactly a “DAO” in the same way our other DAOs abstract external data access. However, I couldn’t identify any other naming convention for a thing instantiated by the application that gets injected into things and has other things injected into it. Alternate naming suggestions welcome.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -5268,6 +5285,33 @@ definitions:
         default: false
         description: Commercial purpose/by a commercial entity
 
+  ResearchPurposeRequest:
+    type: object
+    properties:
+      DS:
+        type: array
+        default: []
```
ansingh7115, line 5293: I should use this for the structuredData api!

src/main/resources/swagger/api-docs.yaml
```diff
@@ -5268,6 +5285,33 @@ definitions:
         default: false
         description: Commercial purpose/by a commercial entity
 
+  ResearchPurposeRequest:
+    type: object
+    properties:
+      DS:
```
breilly2, line 5291: I need to add descriptions to all of these.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ESResearchPurposeDAO.scala
ansingh7115: Maybe something like ESResearchPurposeSupport? But I'm fine with this, the name isn't all that important to me.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
ansingh7115: Could we could pass them as implicits?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ESResearchPurposeDAO.scala
helgridly: -DAO +something like "support" or "helper". prefer the DAO moniker for objects that actually access data. ofc we're risking the weeds of ComponentHelperSupportManager so don't overthink it

breilly2: I don't like either of those but can't think of anything better. Will go with `ResearchPurposeSupport` because it fits with existing names.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
breilly2: Maybe. I think there are DI frameworks for scala out there... some probably use implicits. I mainly made that statement to see whether or not other people agreed. I really enjoyed using java's CDI in the past so I'm sad to be back to doing it all manually.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ESResearchPurposeDAO.scala
rushtong: 👍 to `ResearchPurposeSupport` for consistency. For devs working in the other similarly named areas, this will make more sense than having to dig around for something else.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAO.scala
rushtong: 👎 to using implicits unless necessary. Agreed that DI is a little more painful here.

src/main/scala/org/broadinstitute/dsde/firecloud/model/DataUse.scala
rushtong: 👍

src/test/scala/org/broadinstitute/dsde/firecloud/service/LibraryApiServiceSpec.scala
rushtong: This is really cool - I haven not seen this before.

rushtong (approved):
👍 Have not tested locally, but the general refactoring looks good to me.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -5292,23 +5292,34 @@ definitions:
         type: array
         default: []
         items:
-          type: integer
+          type: string
```
breilly2: Still need to finalize whether this is a string or an integer.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
```diff
@@ -94,7 +95,7 @@ trait ElasticSearchDAOQuerySupport extends ElasticSearchDAOSupport {
       query.must(fieldQuery)
     }
     criteria.researchPurpose map { rp =>
-      query.must(researchPurposeDAO.researchPurposeFilters(rp))
+      query.must(researchPurposeSupport.researchPurposeFilters(rp, Some(s"${AttributeName.libraryNamespace}${AttributeName.delimiter}")))
```
breilly2: Not thrilled that this prefix building logic is outside of `AttributeName`. May still work on improving this.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOResearchPurposeSupport.scala
```diff
@@ -5,16 +5,62 @@ import org.broadinstitute.dsde.firecloud.model.DataUse.{DiseaseOntologyNodeId, R
...
 
 trait ElasticSearchDAOResearchPurposeSupport extends DataUseRestrictionSupport with LazyLogging {
 
-  val durRoot = AttributeName.toDelimitedName(structuredUseRestrictionAttributeName)
+  def researchPurposeFilters(rp: ResearchPurpose, ontologyDAO: OntologyDAO, attributePrefix: Option[String])(implicit ec: ExecutionContext): BoolQueryBuilder = {
 
-  def researchPurposeFilters(rp: ResearchPurpose, ontologyDAO: OntologyDAO)(implicit ec: ExecutionContext): BoolQueryBuilder = {
+    val durRoot: String = attributePrefix.getOrElse("") + structuredUseRestrictionName
+
+    def generateDiseaseMatchLogic(rp: ResearchPurpose, ontologyDAO: OntologyDAO)(implicit ec: ExecutionContext): Option[BoolQueryBuilder] = {
```
breilly2: All of these functions, along with `durRoot`, moved inside `researchPurposeFilters` unchanged so they could have access to the `attributePrefix` parameter without having to pass it around.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOResearchPurposeSupport.scala
```diff
@@ -5,16 +5,62 @@ import org.broadinstitute.dsde.firecloud.model.DataUse.{DiseaseOntologyNodeId, R
...
+          val parentsToAugment:Seq[DiseaseOntologyNodeId] = (allTermResults collect {
+            case termWithParents => termWithParents.parents.getOrElse(List.empty[TermParent]).map(parent => DiseaseOntologyNodeId(parent.id))
+          }).flatten
+          // append the parent node info to the original research purpose
+          nodeids ++ parentsToAugment
+        }
+      }
+    }
+
+    def encode[T](code: String, value: T) = termQuery(s"$durRoot.$code", value)
```
breilly2, line 62: I lied. I renamed this from `code` to `encode` but that's actually left over from a previous refactoring. I think I like it being more verb-like but I can undo the rename if anyone cares.

project/Dependencies.scala
ahaessly: this is great info! we might want to use that instead of the spray json stuff when we implement the entity service

src/main/resources/swagger/api-docs.yaml
```diff
@@ -5268,6 +5285,44 @@ definitions:
...
+          example: "http://purl.obolibrary.org/obo/DOID_4"
+        description: >
+          Disease focused research. Values are DOID ontology node ids in the form
+          http://purl.obolibrary.org/obo/DOID_NNN. Specify an empty array if this
+          is not disease focused research.
+      NMDS:
+        type: boolean
+        default: false
+        description: Methods development/Validation study
```
ahaessly: in these descriptions, i'd like to include whether the boolean value is indicating something is prohibited or required. For this one: Whether Methods development/Validation study is prohibited.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
```diff
@@ -93,8 +94,9 @@ trait ElasticSearchDAOQuerySupport extends ElasticSearchDAOSupport with ElasticS
       values foreach { value:String => fieldQuery.should(termQuery(field+".keyword", value))}
       query.must(fieldQuery)
     }
-    criteria.researchPurpose map { rp =>
-      query.must(researchPurposeFilters(rp, ontologyDAO))
+    criteria.researchPurpose map {
+      def toLibraryAttributeName(name: String): String = AttributeName.toDelimitedName(AttributeName.withLibraryNS(name))
+      rp => query.must(researchPurposeSupport.researchPurposeFilters(rp, toLibraryAttributeName))
```
ahaessly: i don't think we need the library prefix here

src/main/scala/org/broadinstitute/dsde/firecloud/model/DataUse.scala
ahaessly: In the tech doc all the fields except prefix are required. I'm not sure how important this is. I will check with Moran

src/main/scala/org/broadinstitute/dsde/firecloud/service/OntologyService.scala
```diff
@@ -4,40 +4,44 @@ import akka.actor.{Actor, Props}
...
 import spray.httpx.SprayJsonSupport
-import spray.json._
 import spray.json.DefaultJsonProtocol._
 
 import scala.concurrent.{ExecutionContext, Future}
 
 object OntologyService {
   sealed trait OntologyServiceMessage
   case class AutocompleteOntology(term: String) extends OntologyServiceMessage
+  case class ResearchPurposeQuery(researchPurposeRequest: ResearchPurposeRequest) extends OntologyServiceMessage
```
ahaessly, line 21: I'm not sure I would have put the ResearchPurposeQuery in OntologyService. is this because they are under the same /duos endpoint? If we want to have a separate service for the Data Use Aware search API, I think it would be separate from the Autocomplete service (autocomplete needs an ES backend but our 2 APIs don't)

breilly2: That is why I put it here, but I wasn't 100% confident with that decision. My thought was to keep related functionality together, which should happily correspond with the organization of the web service endpoints. @ahaessly, do you think we should have a new service for the data use endpoints?

src/main/scala/org/broadinstitute/dsde/firecloud/model/DataUse.scala
breilly2: Right, I made them optional out of convenience before noticing that. My thinking here is that it makes sense for all to be required for the ingest endpoint to guarantee we have deliberate responses to all of the questions. The search gesture is different enough that I think they maybe shouldn't be required, but I could go either way.

src/main/scala/org/broadinstitute/dsde/firecloud/service/OntologyService.scala
ahaessly: actually, i think this is making more sense to me now. i forgot that our research purpose request needed to query ontology for parents. i think if we did a separate deploy we actually would consider deploying our apis with the consent-ontology service

ahaessly (approved):
looks good. i'm just going to check with Moran about allowing research purpose values to be defaulted.

breilly2:
@helgridly gave me a verbal 👍 just now.

helgridly (approved):

PR #637, opened 2018-05-23
GAWB-3591 unpack TSVs from a bagit

helgridly:
Note that the implementation here just looks at every file in the zip to see if it's called `samples.tsv` or `participants.tsv` -- it doesn't validate that the zip confirms to the bagit spec because that's more work for zero gain.

Progress checklist:

- [x] Swagger definition
- [x] Route specification passing through to `EntityClient` and ultimately to Rawls
- [x] Decompressing a zip file
- [x] Check that iterating over all zip entries looks in subdirectories automagically (thanks Andrea)
- [x] Finding out if our collabs want a .gz or a .tar.gz or what
- [ ] Receiving an example one just to be sure
- [ ] Decompressing that instead
- [x] Unit testing zip decompression
- [x] Route tests using a mock server that serves up hand-rolled zips
- [x] Use better temp file creation methods (thanks Brian)
- [x] Check for multiple participant or sample TSVs in the zip and complain if so
- [x] Exception handling from downloading the URL (e.g. 404s)
- [x] Make sure downloaded zip and extracted TSVs are deleted (even if unzipping or the Rawls call throws an exception)

Security hardening:
- [x] Reject anything that's not https
- [ ] Stat zip file size before downloading and reject if "too big" (how big?)

In all cases:


src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -50,6 +59,35 @@ object EntityClient {
...
+  //returns (contents of participants.tsv, contents of samples.tsv)
+  def unzipTSVs(zipFile: ZipFile): (Option[String], Option[String]) = {
+    val zipEntries = zipFile.entries.asScala
+
+    val rand = java.util.UUID.randomUUID.toString.take(8)
+    val participantsFn = s"/tmp/$rand-participants.tsv"
+    val samplesFn = s"/tmp/$rand-samples.tsv"
+
+    val unzippedFiles = zipEntries.foldLeft((None: Option[String], None: Option[String])){ (acc: (Option[String], Option[String]), ent: ZipEntry) =>
+      if(!ent.isDirectory && ent.getName.contains("participants.tsv")) {
```
ahaessly: does this code only look one directory down? it could be nested several directories deep

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -50,6 +59,35 @@ object EntityClient {
     }
   }
 
+  //returns (contents of participants.tsv, contents of samples.tsv)
+  def unzipTSVs(zipFile: ZipFile): (Option[String], Option[String]) = {
+    val zipEntries = zipFile.entries.asScala
+
+    val rand = java.util.UUID.randomUUID.toString.take(8)
+    val participantsFn = s"/tmp/$rand-participants.tsv"
+    val samplesFn = s"/tmp/$rand-samples.tsv"
```
breilly2: Should use `System.getProperty("java.io.tmpdir")` in place of `/tmp`.

Another option is to use `Files.createTempFile`: https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#createTempFile(java.lang.String,%20java.lang.String,%20java.nio.file.attribute.FileAttribute...)

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
helgridly: as far as i can tell from the docs it's flat, not recursive, but i'll doublecheck

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
helgridly: nice, thanks

ahaessly (commented):
one comment. looking good!

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
breilly2: If there are multiple matching files in different directories, this will pick the "last" one based on whatever order is returned from the `ZipFile`. I'd rather we either explicitly search only the root (or wherever we agree they will live) or fail if there isn't a standard location and we find more than one match for either filename.

(I skipped over this when looking at the preview because of the `//TODO: eww` comment.)

helgridly: fair point. i'll accumulate all of them and if there's multiples i'll throw an annoyed error

ahaessly: in this doc https://docs.google.com/document/d/1iepsRql4EmuUSHsF1nguIADbovlav6NTbTR3F7gEHIc/edit#heading=h.wjhyckqmf2oj
the data dir has 2 sub dirs before the actual files and it says the data dir can have any format you want.

src/test/scala/org/broadinstitute/dsde/firecloud/EntityClientSpec.scala
```diff
@@ -0,0 +1,32 @@
...
+class EntityClientSpec extends BaseServiceSpec with BeforeAndAfterEach {
+  override def beforeEach(): Unit = {
+    searchDao.reset
+  }
+
+  override def afterEach(): Unit = {
+    searchDao.reset
+  }
+
+  "EntityClient should extract TSV files out of bagit zips" - {
```
breilly2, line 21: The only other thing I can think of is a test that participants are imported before samples. Not sure how to verify that, though... except maybe an integration test that has references that would fail if done in the wrong order.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -209,4 +250,42 @@ class EntityClient (requestContext: RequestContext)(implicit protected val execu
...
+      //Java URL handles http, https, ftp, file, and jar protocols.
+      //We're only OK with https to avoid MITM attacks.
+      val bagitURL = new URL(bagitRq.bagitURL)
+      val acceptableProtocols = Seq("https") //for when we inevitably change our mind and need to support others
+      if (!acceptableProtocols.contains(bagitURL.getProtocol)) {
+        Future.successful(RequestCompleteWithErrorReport(StatusCodes.BadRequest, "Invalid bagitURL protocol: must be https only"))
+      } else {
+
+        val rand = java.util.UUID.randomUUID.toString.take(8)
+        val localZipPath = s"/tmp/$rand-bagit.zip"
```
breilly2: Temp file changes here too.

breilly2 (changes requested):
Just a couple of small things then it should be good.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
helgridly: sorry, i meant that the list of entries (i.e. the variable `zipEntries` here) is a list, not some kind of tree structure. so walking that list traverses all files in the zip, including in subdirectories, because the directory structure has been flattened

ahaessly: great - thanks!

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -210,4 +266,62 @@ class EntityClient (requestContext: RequestContext)(implicit protected val execu
...
+        val bagItFile = File.createTempFile(s"$rand-samples", ".tsv")
+
+        try {
+          val conn = bagitURL.openConnection()
+          val length = conn.getContentLength
+          logger.info("BAGIT FILE LENGTH: " + length.toString)
+
+          //how big is this supposed to be
+          if (length > 1000000) {
+            Future.successful(RequestCompleteWithErrorReport(StatusCodes.BadRequest, s"BDBAg size is too large."))
```
helgridly: nit: capitalisation

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -210,4 +266,61 @@ class EntityClient (requestContext: RequestContext)(implicit protected val execu
...
+      } else {
+
+        val rand = java.util.UUID.randomUUID.toString.take(8)
+        val bagItFile = File.createTempFile(s"$rand-samples", ".tsv")
+
+        try {
+          val conn = bagitURL.openConnection()
+          val length = conn.getContentLength
+
+          if (length > 1000000) {
```
ansingh7115: This value is not set in stone yet - talking with Bernick on what it should be. Probs should put it in config?

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -51,6 +60,50 @@ object EntityClient {
...
+    val zipEntries = zipFile.entries.asScala
+
+    val rand = java.util.UUID.randomUUID.toString.take(8)
+    val participantsTmp = File.createTempFile(s"$rand-participants", ".tsv")
+    val samplesTmp = File.createTempFile(s"$rand-samples", ".tsv")
+
+    val unzippedFiles = zipEntries.foldLeft((None: Option[String], None: Option[String])){ (acc: (Option[String], Option[String]), ent: ZipEntry) =>
+      if(!ent.isDirectory && (ent.getName.contains("/participants.tsv") || ent.getName.equals("participants.tsv"))) {
+        acc._1 match {
+          case Some(x) => throw new FireCloudException(s"More than one participants.tsv file found in BDBag $bagName")
```
helgridly: nit: bagit, not bdbag

helgridly: also: i assume this exception turns into a 400 somewhere upstream?

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
helgridly: +1 put it in config

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -210,4 +266,61 @@ class EntityClient (requestContext: RequestContext)(implicit protected val execu
...
+      val acceptableProtocols = Seq("https") //for when we inevitably change our mind and need to support others
+      if (!acceptableProtocols.contains(bagitURL.getProtocol)) {
+        Future.successful(RequestCompleteWithErrorReport(StatusCodes.BadRequest, "Invalid bagitURL protocol: must be https only"))
+      } else {
+
+        val rand = java.util.UUID.randomUUID.toString.take(8)
+        val bagItFile = File.createTempFile(s"$rand-samples", ".tsv")
+
+        try {
+          val conn = bagitURL.openConnection()
```
helgridly, line 287: does this connection need to be closed?

helgridly (commented):
GH won't let me thumb this because I opened the PR :) minor comments but thumb once they're addressed

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -210,4 +266,61 @@ class EntityClient (requestContext: RequestContext)(implicit protected val execu
...
+
+            //make two big strings containing the participants and samples TSVs
+            //if i could turn back time this would use streams to save memory, but hopefully this will all go away when entity service comes along
+            unzipTSVs(bagitRq.bagitURL, new ZipFile(bagItFile.getAbsolutePath)) { (participantsStr, samplesStr) =>
+              (participantsStr, samplesStr) match {
+                case (None, None) =>
+                  Future.successful(RequestCompleteWithErrorReport(StatusCodes.BadRequest, "You must have either (or both) participants.tsv and samples.tsv in the zip file"))
+                case _ =>
+                  for {
+                    // This should toss back errors from rawls.
```
helgridly: minor, but the word "toss" has connotations in British English that make this comment extremely stick out for me. i will politely defer to [Stack Exchange](https://english.stackexchange.com/questions/409677/how-popular-is-the-slang-usage-of-toss-in-british-english) for a discussion of the vulgar details

ansingh7115: Alright, will change to "vomit"

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
ansingh7115: URLConnection doesn't seem to have a disconnect or close method. Reading on the internet, it appears you only have to close it if you open a stream? I _think_ it's okay like this (closes on its own).

rushtong: One thing you could do is cast it as a `HttpURLConnection` and use that class' disconnect. Take this with a grain of salt as I'm not a huge fan of casting.

ahaessly (approved):
👍

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
ansingh7115: no it does not, I should fix that

ansingh7115:
jenkins retest

PR #648, opened 2018-06-05
encode workflow cost uri

MatthewBemis:
this fixes a bug when viewing workflows for a workspace that has spaces in its name.

it essentially came about as a race condition where this new feature was implemented at the same time as spaces in names were supported, so the new endpoint URI was never wrapped in `encodeUri`

i would like to make it so `encodeUri` cannot be forgotten for any URI that has a workspace name in it, but that would be a larger refactor and i'd like to at least get a fix out in the next release. 

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/SubmissionService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/SubmissionService.scala
@@ -58,7 +58,7 @@ trait SubmissionService extends HttpService with PerRequestCreator with FireClou
             } ~
             pathPrefix("cost") {
               get {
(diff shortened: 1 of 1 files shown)


MatthewBemis:
jenkins retest because free trial spec is being flakey tonight

dvoet (approved):

jmthibault79 (approved):
Looks ok.

Can we apply `encodeURL` to all endpoints?  Make a new `passthroughAndEncode` or something?  I dunno.

MatthewBemis:
@jmthibault79 we can but there is still a hole where this can be forgotten, which is for requests to rawls using RawlsDAO. i will try to implement something that's less prone to being left out in both places... (and maybe that will be a follow-up PR)

gpolumbo-broad (approved):

MatthewBemis:
verbal QA approval to merge despite TSV related errors

PR #650, opened 2018-06-11
GAWB-3647: Add default tsv types

ansingh7115:

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
@@ -248,21 +248,24 @@ class EntityClient (requestContext: RequestContext)(implicit protected val execu
       // 1. The type of the tsv import file (membership, entity, or update)
       // 2. The entity type we're trying to import or update, which is magically verified in the withFoo functions below.
       tsv.firstColumnHeader.split(":") match {
-        case Array( tsvType, entityHeader ) =>
-          val entityType = entityHeader.stripSuffix("_id")
-          if( entityType == entityHeader ) {
+        case h: Array[String] if h.length == 1 || h.length == 2 =>
+          val entityType = h.last.stripSuffix("_id")
+          if (entityType == h.last) {
             Future(RequestCompleteWithErrorReport(BadRequest, "Invalid first column header, entity type should end in _id"))
           } else {
             val strippedTsv = backwardsCompatStripIdSuffixes(tsv, entityType)
-            Try(TsvTypes.withName(tsvType)) match {
-              case Success(TsvTypes.MEMBERSHIP) => importMembershipTSV (pipeline, workspaceNamespace, workspaceName, strippedTsv, entityType)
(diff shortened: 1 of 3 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -248,21 +248,24 @@ class EntityClient (requestContext: RequestContext)(implicit protected val execu
       // 1. The type of the tsv import file (membership, entity, or update)
       // 2. The entity type we're trying to import or update, which is magically verified in the withFoo functions below.
       tsv.firstColumnHeader.split(":") match {
-        case Array( tsvType, entityHeader ) =>
-          val entityType = entityHeader.stripSuffix("_id")
-          if( entityType == entityHeader ) {
+        case h: Array[String] if h.length == 1 || h.length == 2 =>
```
helgridly: suggestion, since `h.head` and `h.last` are a little difficult to read:

```
case entityType :: Nil => //use the default TSV type
case tsvType :: entityType :: Nil => //use the TSV type given
case _ => //error
```

The last line is equivalent to `case tsvType :: theRest`, where `theRest` is an array of at least two, and implies at least one of a) someone's using a colon in an entity name b) they are very confused

helgridly: i'd be inclined to make the result of this `match` block be a `TsvType`. then this function becomes

1. determine the type of the TSV
2. import it

which i'd argue is a touch more readable

helgridly (approved):
looks good, a couple of style nits

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -238,32 +239,42 @@ class EntityClient (requestContext: RequestContext)(implicit protected val execu
...
-              case Success(TsvTypes.MEMBERSHIP) => importMembershipTSV (pipeline, workspaceNamespace, workspaceName, strippedTsv, entityType)
-              case Success(TsvTypes.ENTITY) => importEntityTSV (pipeline, workspaceNamespace, workspaceName, strippedTsv, entityType)
-              case Success(TsvTypes.UPDATE) => importUpdateTSV (pipeline, workspaceNamespace, workspaceName, strippedTsv, entityType)
-              case Failure(err) => Future(RequestCompleteWithErrorReport(BadRequest, err.toString))
-            }
+      val (tsvType, entityType) = tsv.firstColumnHeader.split(":") match {
+        case Array(entityTypeString) => (TsvTypes.ENTITY, stripEntityType(entityTypeString))
+        case Array(tsvTypeString, entityTypeString) => {
+          val tsvType = Try(TsvTypes.withName(tsvTypeString)) match {
+            case Success(t) => t
```
ansingh7115, line 269: I'm not sure if they're a less verbose way to do a Try() match when you're just passing up the results of a success and then throwing an error that uses the error message if it fails.

helgridly: nah i think this is it. you might be able to do `getOrElse(throw exc)` but that seems _way_ crazy

ahaessly (approved):
👍

PR #661, opened 2018-06-20
 Restore GET URL for cookie-authed entity TSV download for fc-app-80

jmthibault79:
This endpoint was changed from GET to POST in #337.  GET is now restored and comments are added indicating when each is useful.

Also moved some test-only code out of FireCloudRequestBuilding

In all cases:

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -2408,6 +2408,10 @@ paths:
             type: file
         404:
           description: Workspace or entity type does not exist
+        414:
+          description: |
+           URI length exceeds the configured limit of 2048 characters.
+           Please use the POST endpoint when it's necessary to supply a large number of attribute names.
         500:
           description: Internal Server Error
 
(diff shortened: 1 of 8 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/FireCloudRequestBuilding.scala
```diff
@@ -37,18 +37,6 @@ trait FireCloudRequestBuilding extends spray.httpx.RequestBuilding {
     authHeaders(Some(accessToken.accessToken))
   }
 
-  def dummyAuthHeaders = {
```
jmthibault79, line 40: moved to TestRequestBuilding

jmthibault79:
2 failures, both related to test cleanup.  Worth looking into, but they're not caused by this PR.

jenkins retest

src/main/resources/swagger/api-docs.yaml
```diff
@@ -3534,6 +3534,36 @@ paths:
...
+      responses:
+        200:
+          description: Workspace entities of specified type in TSV format
+          schema:
+            type: file
+        404:
+          description: Workspace or entity type does not exist
+        500:
+          description: Internal Server Error
+      security: []
```
davidangb: if the caller exceeds URL length by specifying too many `attributeNames`, what do we return? Is that a 400? Please add this potential response here in the swagger doc. (I don't think this error case needs a unit test)

davidangb (commented):
one small comment re: swagger, otherwise :+1:

PR #663, opened 2018-07-11
API for logging shares in ElasticSearch

ghost:
The backend work for logging shares in ElasticSearch. 

There is a companion PR in firecloud-develop: [#1263](https://github.com/broadinstitute/firecloud-develop/pull/1263)

In all cases:

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1713,17 +1713,6 @@ paths:
         Get a list of users with whom current user has shared a Workspace.
       description: |
         Get a list of users with whom current user has shared a Workspace.
-      parameters:
-        - name: shareType
-          required: false
-          in: query
-          description: |
-           The type of share - Workspace, Group, or Method. If left empty, all types are returned.
-          type: string
-          enum:
-            - workspace
-            - group
-            - method
       responses:
         200:
           description: Success
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1719,13 +1719,15 @@ paths:
           description: |
            The type of share - Workspace, Group, or Method. If left empty, all types are returned.
           type: string
+          enum:
+            - workspace
+            - group
+            - method
       responses:
         200:
           description: Success
           schema:
             $ref: '#/definitions/StringArray'
-        404:
-          description: Not Found
         500:
           description: Internal Error
 
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1710,9 +1710,9 @@ paths:
       - ShareLog
       operationId: getSharees
       summary: |
-        Get a list of users with whom current user has shared a Workspace, Group, or Method.
+        Get a list of users with whom current user has shared a Workspace.
       description: |
-        Get a list of users with whom current user has shared a Workspace, Group, or Method.
+        Get a list of users with whom current user has shared a Workspace.
       parameters:
         - name: shareType
           required: false
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1704,6 +1704,31 @@ paths:
               requiresRefresh:
                 type: boolean
 
+  /api/sharelog/sharees:
+    get:
+      tags:
+      - ShareLog
+      operationId: getSharees
+      summary: |
+        Get a list of users with whom current user has shared a Workspace, Group, or Method.
+      description: |
+        Get a list of users with whom current user has shared a Workspace, Group, or Method.
+      parameters:
+        - name: shareType
+          required: false
+          description: |
+           The type of share - Workspace, Group, or Method. If left empty, all types are returned.
+          type: string
+      responses:
+        200:
+          description: Success
+          schema:
+            $ref: '#/definitions/StringArray'
+        404:
+          description: Not Found
+        500:
+          description: Internal Error
+
   /api/storage/{bucket}/{object}:
     get:
       tags:
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1716,6 +1716,7 @@ paths:
       parameters:
         - name: shareType
           required: false
+          in: query
           description: |
            The type of share - Workspace, Group, or Method. If left empty, all types are returned.
           type: string
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
@@ -39,6 +39,7 @@ class FireCloudServiceActor extends HttpServiceActor with FireCloudDirectives
   with Ga4ghApiService
   with UserApiService
   with TrialApiService
+  with ShareLogApiService
 {
 
   override lazy val log = LoggerFactory.getLogger(getClass)
@@ -103,6 +104,7 @@ class FireCloudServiceActor extends HttpServiceActor with FireCloudDirectives
   val permissionReportServiceConstructor: (UserInfo) => PermissionReportService = PermissionReportService.constructor(app)
   val trialServiceConstructor: () => TrialService = TrialService.constructor(app, trialProjectManager)
   val userServiceConstructor: (WithAccessToken) => UserService = UserService.constructor(app)
+  val shareLogServiceConstructor: () => ShareLogService = ShareLogService.constructor(app)
 
   if (FireCloudConfig.Trial.spreadsheetId.nonEmpty && FireCloudConfig.Trial.spreadsheetUpdateFrequencyMinutes > 0) {
     val freq = FireCloudConfig.Trial.spreadsheetUpdateFrequencyMinutes
@@ -121,7 +123,7 @@ class FireCloudServiceActor extends HttpServiceActor with FireCloudDirectives
   val billingService = new BillingService with ActorRefFactoryContext
   val apiRoutes = methodsApiServiceRoutes ~ profileRoutes ~
     methodConfigurationService.routes ~ submissionsService.routes ~
-    nihRoutes ~ billingService.routes ~ trialApiServiceRoutes
+    nihRoutes ~ billingService.routes ~ trialApiServiceRoutes ~ shareLogServiceRoutes
 
   val healthService = new HealthService with ActorRefFactoryContext
 
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
@@ -66,8 +66,9 @@ class FireCloudServiceActor extends HttpServiceActor with FireCloudDirectives
       new HttpLogitDAO(FireCloudConfig.Metrics.logitUrl, FireCloudConfig.Metrics.logitApiKey.get)
     else
       new NoopLogitDAO
+  val shareLogDAO:ShareLogDAO = new ElasticSearchShareLogDAO(elasticSearchClient, FireCloudConfig.ElasticSearch.shareLogIndexName)
 
-  val app:Application = new Application(agoraDAO, googleServicesDAO, ontologyDAO, consentDAO, rawlsDAO, samDAO, searchDAO, researchPurposeSupport, thurloeDAO, trialDAO, logitDAO)
+  val app:Application = new Application(agoraDAO, googleServicesDAO, ontologyDAO, consentDAO, rawlsDAO, samDAO, searchDAO, researchPurposeSupport, thurloeDAO, trialDAO, logitDAO, shareLogDAO)
   val materializer: ActorMaterializer = ActorMaterializer()
 
   private val healthChecks = new HealthChecks(app)
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
@@ -14,7 +14,6 @@ import org.broadinstitute.dsde.firecloud.service._
 import org.broadinstitute.dsde.firecloud.trial.ProjectManager
 import org.broadinstitute.dsde.firecloud.webservice._
 import org.broadinstitute.dsde.workbench.util.health.HealthMonitor
-import org.elasticsearch.action.support.WriteRequest.RefreshPolicy
 import org.elasticsearch.client.transport.TransportClient
 
 import scala.concurrent.duration._
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
@@ -14,6 +14,7 @@ import org.broadinstitute.dsde.firecloud.service._
 import org.broadinstitute.dsde.firecloud.trial.ProjectManager
 import org.broadinstitute.dsde.firecloud.webservice._
 import org.broadinstitute.dsde.workbench.util.health.HealthMonitor
+import org.elasticsearch.action.support.WriteRequest.RefreshPolicy
 import org.elasticsearch.client.transport.TransportClient
 
 import scala.concurrent.duration._
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchShareLogDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchShareLogDAO.scala
@@ -3,8 +3,8 @@ package org.broadinstitute.dsde.firecloud.dataaccess
 import java.time.Instant
 
 import org.broadinstitute.dsde.firecloud.FireCloudException
-import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol.ShareFormat
-import org.broadinstitute.dsde.firecloud.model.ShareLog.Share
+import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol.impShareFormat
+import org.broadinstitute.dsde.firecloud.model.ShareLog.{Share, ShareType}
 import org.broadinstitute.dsde.workbench.util.health.SubsystemStatus
 import org.elasticsearch.action.admin.indices.create.{CreateIndexRequest, CreateIndexRequestBuilder, CreateIndexResponse}
 import org.elasticsearch.action.admin.indices.exists.indices.{IndicesExistsRequest, IndicesExistsRequestBuilder, IndicesExistsResponse}
@@ -26,14 +26,15 @@ import scala.util.{Failure, Success, Try}
 
 trait ShareQueries {
   def userShares(userId: String): QueryBuilder = termQuery("userId", userId)
-  def userSharesOfType(userId: String, shareType: Option[String]) = {
-    if (shareType.isDefined)
-      boolQuery().filter(
-        boolQuery()
-          .must(userShares(userId))
-          .must(termQuery("shareType", shareType.get)))
-    else
-      userShares(userId)
+  def userSharesOfType(userId: String, shareType: Option[ShareType.Value]): QueryBuilder = {
+    shareType match {
+      case Some(typeOfShare) =>
+        boolQuery().filter(
+          boolQuery()
+            .must(userShares(userId))
+            .must(termQuery("shareType", typeOfShare.toString)))
+      case _ => userShares(userId)
+    }
   }
 
 }
@@ -59,9 +60,9 @@ class ElasticSearchShareLogDAO(client: TransportClient, indexName: String, refre
     * @param shareType  The type (workspace, group, or method) see `ShareLog`
     * @return           The record of the share
     */
-  override def logShare(userId: String, sharee: String, shareType: String): Share = {
+  override def logShare(userId: String, sharee: String, shareType: ShareType.Value): Share = {
     val share = Share(userId, sharee, shareType, Some(Instant.now))
-    val id = MurmurHash3.stringHash(userId + sharee + shareType).toString
+    val id = generateId(share)
     val insert = client
       .prepareIndex(indexName, datatype, id)
       .setSource(share.toJson.compactPrint, XContentType.JSON)
@@ -73,14 +74,14 @@ class ElasticSearchShareLogDAO(client: TransportClient, indexName: String, refre
 
   /**
     * Logs records of a user sharing a workspace, group, or method with users.
+    * todo could be a batch query
     *
     * @param userId The workbench user id
     * @param sharees The emails of the users being shared with
     * @param shareType The type (workspace, group, or method) see `ShareLog`
     * @return The records of the shares - see `ShareLog.Share`
     */
-  override def logShares(userId: String, sharees: Seq[String], shareType: String): Seq[Share] = {
-    // todo how do i make this a batch query?
+  override def logShares(userId: String, sharees: Seq[String], shareType: ShareType.Value): Seq[Share] = {
     sharees map { sharee => logShare(userId, sharee, shareType) }
   }
 
@@ -108,7 +109,7 @@ class ElasticSearchShareLogDAO(client: TransportClient, indexName: String, refre
     * @param shareType  The type (workspace, group, or method) - if left blank returns all shares
     * @return A list of `ShareLog.Share`s
     */
-  override def getShares(userId: String, shareType: Option[String] = None): Seq[Share] = {
+  override def getShares(userId: String, shareType: Option[ShareType.Value] = None): Seq[Share] = {
     val getSharesRequest = client
       .prepareSearch(indexName)
       .setQuery(userSharesOfType(userId, shareType))
@@ -122,7 +123,12 @@ class ElasticSearchShareLogDAO(client: TransportClient, indexName: String, refre
       getSharesResponse.getHits.getHits.toList map (_.getSourceAsString.parseJson.convertTo[Share])
   }
 
-//  todo
+//  todo Leveraging ElasticSearch's autocomplete functionality will likely be useful
+//       as we iterate and the number of shares increases to such a size that the list of
+//       shares returned by `getShares` is too large for HTML autocomplete to efficiently handle.
+//       Rather than limit the number of values returned, it ought to be beneficial to let
+//       ElasticSearch handle the autocompletion.
+//
 //  override def autocomplete(userId: String, term: String): List[String] = ???
 
   private def indexExists: Boolean = {
@@ -151,7 +157,7 @@ class ElasticSearchShareLogDAO(client: TransportClient, indexName: String, refre
     * @return the hash
     */
   def generateId(share: Share): String = {
-    val rawId = Seq(share.userId, share.sharee, share.shareType).mkString
+    val rawId = Seq(share.userId, share.sharee, share.shareType).mkString("|")
     MurmurHash3.stringHash(rawId).toString
   }
 }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchShareLogDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchShareLogDAO.scala
@@ -11,7 +11,6 @@ import org.elasticsearch.action.admin.indices.exists.indices.{IndicesExistsReque
 import org.elasticsearch.action.get.{GetRequest, GetRequestBuilder, GetResponse}
 import org.elasticsearch.action.index.{IndexRequest, IndexRequestBuilder, IndexResponse}
 import org.elasticsearch.action.search.{SearchRequest, SearchRequestBuilder, SearchResponse}
-import org.elasticsearch.action.support.WriteRequest.RefreshPolicy
 import org.elasticsearch.client.transport.TransportClient
 import org.elasticsearch.common.xcontent.XContentType
 import org.elasticsearch.index.query.QueryBuilder
@@ -27,8 +26,9 @@ import scala.util.{Failure, Success, Try}
 trait ShareQueries {
   def userShares(userId: String): QueryBuilder = termQuery("userId", userId)
   def userSharesOfType(userId: String, shareType: String) = boolQuery()
-    .must(termQuery("userId", userId))
-    .must(termQuery("shareType", shareType))
+    .filter(boolQuery()
+      .must(userShares(userId))
+      .must(termQuery("shareType", shareType)))
 }
 
 /**
@@ -36,11 +36,8 @@ trait ShareQueries {
   *
   * @param client      The ElasticSearch client
   * @param indexName   The name of the target share log index in ElasticSearch
-  * @param refreshMode Using IMMEDIATE - the default - is a performance hit but ensures transactionality
-  *                    of updates across multiple users.
-  *                    TODO would NONE be a better option here?
   */
-class ElasticSearchShareLogDAO(client: TransportClient, indexName: String, refreshMode: RefreshPolicy = RefreshPolicy.IMMEDIATE)
+class ElasticSearchShareLogDAO(client: TransportClient, indexName: String)
   extends ShareLogDAO with ElasticSearchDAOSupport with ShareQueries {
 
   lazy private final val datatype = "sharelog"
@@ -56,13 +53,11 @@ class ElasticSearchShareLogDAO(client: TransportClient, indexName: String, refre
     * @return           The record of the share
     */
   override def logShare(userId: String, sharee: String, shareType: String): Share = {
-    val share = Share(userId, sharee, shareType, Instant.now)
+    val share = Share(userId, sharee, shareType, Some(Instant.now))
     val id = MurmurHash3.stringHash(userId + sharee + shareType).toString
     val insert = client
       .prepareIndex(indexName, datatype, id)
       .setSource(share.toJson.compactPrint, XContentType.JSON)
-      .setCreate(true) // fail the request if the logged share already exists
-      .setRefreshPolicy(refreshMode)
 
     executeESRequest[IndexRequest, IndexResponse, IndexRequestBuilder](insert)
     share
@@ -71,15 +66,16 @@ class ElasticSearchShareLogDAO(client: TransportClient, indexName: String, refre
   /**
     * Gets a share by the ID, a `MurmurHash3` of `userId` + `sharee` + `shareType`
     *
-    * @param id The ID of the share
+    * @param share The share to get
     * @return A record of the share
     */
-  override def getShare(id: String): Share = {
+  override def getShare(share: Share): Share = {
+    val id = generateId(share)
     val getSharesQuery = client.prepareGet(indexName, datatype, id)
     Try(executeESRequest[GetRequest, GetResponse, GetRequestBuilder](getSharesQuery)) match {
       case Success(get) if get.isExists => get.getSourceAsString.parseJson.convertTo[Share]
-      case Success(_) => throw new FireCloudException(s"share for $id not found")
-      case Failure(f) => throw new FireCloudException(s"error receiving share for $id: ${f.getMessage}")
+      case Success(_) => throw new FireCloudException(s"share not found")
+      case Failure(f) => throw new FireCloudException(s"error getting share for $share: ${f.getMessage}")
     }
   }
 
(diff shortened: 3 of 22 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -1704,6 +1704,31 @@ paths:
...
 
+  /api/sharelog/sharees:
+    get:
+      tags:
+      - ShareLog
+      operationId: getSharees
+      summary: |
+        Get a list of users with whom current user has shared a Workspace, Group, or Method.
+      description: |
+        Get a list of users with whom current user has shared a Workspace, Group, or Method.
```
davidangb: "Workspace, Group, or Method" <- is this scope creep? Original ticket only specifies workspace shares.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1704,6 +1704,31 @@ paths:
...
+          description: |
+           The type of share - Workspace, Group, or Method. If left empty, all types are returned.
+          type: string
+      responses:
+        200:
+          description: Success
+          schema:
+            $ref: '#/definitions/StringArray'
+        404:
+          description: Not Found
```
davidangb: what conditions cause the 404? i.e. what do we return if the current user has no entries in the log? I suggest this should return a 200 OK with a payload of `[]` instead of a 404.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1704,6 +1704,31 @@ paths:
...
+      summary: |
+        Get a list of users with whom current user has shared a Workspace, Group, or Method.
+      description: |
+        Get a list of users with whom current user has shared a Workspace, Group, or Method.
+      parameters:
+        - name: shareType
+          required: false
+          description: |
+           The type of share - Workspace, Group, or Method. If left empty, all types are returned.
+          type: string
```
davidangb: this should probably be a type `enum`, showing the legal values.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchShareLogDAO.scala
```diff
@@ -0,0 +1,133 @@
...
+import scala.concurrent.Future
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.util.hashing.MurmurHash3
+import scala.util.{Failure, Success, Try}
+
+trait ShareQueries {
+  def userShares(userId: String): QueryBuilder = termQuery("userId", userId)
+  def userSharesOfType(userId: String, shareType: String) = boolQuery()
+    .must(termQuery("userId", userId))
+    .must(termQuery("shareType", shareType))
```
davidangb: I think you can reuse `userShares` here:
```
def userSharesOfType(userId: String, shareType: String) = boolQuery()
    .must(userShares(userId))
    .must(termQuery("shareType", shareType))
```

also, a bit of a query optimization - we probably want to use filter context here, not query context.
```
    boolQuery().filter(boolQuery()
    .must(termQuery("userId", userId))
    .must(termQuery("shareType", shareType)))
```
see https://www.elastic.co/guide/en/elasticsearch/reference/5.4/term-level-queries.html for an explanation of this - in short, filter context turns off a bunch of relevance calculations and other search features that we don't need, because these are exact-match queries.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchShareLogDAO.scala
```diff
@@ -0,0 +1,133 @@
...
+}
+
+/**
+  * DAO that uses ElasticSearch to log and get records of shares.
+  *
+  * @param client      The ElasticSearch client
+  * @param indexName   The name of the target share log index in ElasticSearch
+  * @param refreshMode Using IMMEDIATE - the default - is a performance hit but ensures transactionality
+  *                    of updates across multiple users.
+  *                    TODO would NONE be a better option here?
```
davidangb: yeah, I'd go for NONE here. I don't see any reason this needs to be atomic/immediately available.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchShareLogDAO.scala
```diff
@@ -0,0 +1,133 @@
...
+  override def getShares(userId: String, shareType: Option[String] = None): Seq[Share] = {
+    val getSharesRequest = client
+      .prepareSearch(indexName)
+      .setQuery(
+        if (shareType.isDefined)
+          userSharesOfType(userId, shareType.get)
+        else
+          userShares(userId)
+      )
+      // possibly sort and limit size
```
davidangb: if you don't set a page size, you get the default of 10 - which means that after an end user shares a bunch of times, we won't be returning all sharees.

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala
```diff
@@ -325,4 +327,35 @@ object ModelJsonProtocol extends WorkspaceJsonSupport {
...
+          case _ => throw new DeserializationException("could not deserialize Share")
+        }
+        val sharee: String = jso.get("sharee") match {
+          case Some(JsString(s)) => s
+          case _ => throw new DeserializationException("could not deserialize Share")
+        }
+        val shareType: String = jso.get("shareType") match {
+          case Some(JsString(s)) => s
+          case _ => throw new DeserializationException("could not deserialize Share")
+        }
```
davidangb: I know I gave you this code :) There's a lot of cut-and-paste here, maybe extract a common method for these three?

src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/ElasticSearchShareLogDAOSpec.scala
```diff
@@ -0,0 +1,65 @@
...
+    }
+  }
+}
+
+object ElasticSearchShareLogDAOFixtures {
+  val fixtureShares: Seq[Share] = Seq(
+    Share("fake1", "fake2@gmail.com", ShareLog.WORKSPACE, Instant.now),
+    Share("fake1", "fake3@gmail.com", ShareLog.WORKSPACE, Instant.now),
+    Share("fake1", "fake4@gmail.com", ShareLog.WORKSPACE, Instant.now)
+  )
```
davidangb: probably worth setting up > 10 entries so you can test pagination/query size.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ShareLogService.scala
```diff
@@ -0,0 +1,57 @@
...
+
+  override def receive = {
+    case LogShare(userId: String, sharee: String, shareType: String) => logShare(userId, sharee, shareType) pipeTo sender
+    case GetShares(userId: String, shareType: Option[String]) => getShares(userId) pipeTo sender
+    case GetSharees(userId: String, shareType: Option[String]) => getSharees(userId, shareType) pipeTo sender
+//    todo
+//    case Autocomplete(userId: String, term: String) => autocomplete(userId, term) pipeTo sender
+  }
+
+  def logShare(userId: String, sharee: String, shareType: String) = Future(RequestComplete(shareDAO.logShare(userId, sharee, shareType)))
```
davidangb: this will throw an error when attempting to log a duplicate entry. Is that desired?

As I type this, I wonder (scope creep warning) ... should we log the number of times a specific share (sharer, sharee) occurs? We could use that quantity to drive some kind of relevance - if Alice shared with Bob 20 times but only shared with Charlie twice, we could show Bob first.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchShareLogDAO.scala
```diff
@@ -0,0 +1,133 @@
...
+    val id = MurmurHash3.stringHash(userId + sharee + shareType).toString
+    val insert = client
+      .prepareIndex(indexName, datatype, id)
+      .setSource(share.toJson.compactPrint, XContentType.JSON)
+      .setCreate(true) // fail the request if the logged share already exists
+      .setRefreshPolicy(refreshMode)
+
+    executeESRequest[IndexRequest, IndexResponse, IndexRequestBuilder](insert)
+    share
+  }
```
davidangb, line 75: might wanna break out the id generation into a reusable method, e.g. 
```
  def generateId(share: Share) = {
    val rawString = Seq(share.userId, share.sharee, share.shareType).mkString("|")
    MurmurHash3.stringHash(rawString).toString
  }
```
I can see needing to use this from other parts of the code sometime.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchShareLogDAO.scala
```diff
@@ -0,0 +1,133 @@
...
+    * @return A record of the share
+    */
+  override def getShare(id: String): Share = {
+    val getSharesQuery = client.prepareGet(indexName, datatype, id)
+    Try(executeESRequest[GetRequest, GetResponse, GetRequestBuilder](getSharesQuery)) match {
+      case Success(get) if get.isExists => get.getSourceAsString.parseJson.convertTo[Share]
+      case Success(_) => throw new FireCloudException(s"share for $id not found")
+      case Failure(f) => throw new FireCloudException(s"error receiving share for $id: ${f.getMessage}")
+    }
+  }
```
davidangb: should this accept the string id as argument, or a `Share` object?

davidangb (commented):
this is looking good overall! I have a bunch of comments inline as a first pass. I haven't tried to run this or anything yet.

src/main/resources/swagger/api-docs.yaml
ghost: Good point!

src/main/scala/org/broadinstitute/dsde/firecloud/service/ShareLogService.scala
ghost: My initial thought was yes to throwing an error but now it doesn't seem practical to me, and I like your suggestion. I think ElasticSearch has functionality for this, I'll do some digging.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchShareLogDAO.scala
ghost: just wondering - why delimit with `"|"`?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchShareLogDAO.scala
ghost: Cool, do you think these relevance calculations will come in handy when keeping track of times shared?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchShareLogDAO.scala
```diff
@@ -0,0 +1,157 @@
...
+
+import scala.concurrent.Future
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.util.hashing.MurmurHash3
+import scala.util.{Failure, Success, Try}
+
+trait ShareQueries {
+  def userShares(userId: String): QueryBuilder = termQuery("userId", userId)
+  def userSharesOfType(userId: String, shareType: Option[String]) = {
+    if (shareType.isDefined)
```
jmthibault79: use `shareType match {` instead

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchShareLogDAO.scala
jmthibault79: Not sure how relevant this will be, but the delimiter allows us to distinguish between (`foo.bar`, `baz`) and (`foo.barb`, `az`)

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchShareLogDAO.scala
```diff
@@ -0,0 +1,157 @@
...
+    * Logs a record of a user sharing a workspace, group, or method with a user.
+    *
+    * @param userId     The workbench user id
+    * @param sharee     The email of the user being shared with
+    * @param shareType  The type (workspace, group, or method) see `ShareLog`
+    * @return           The record of the share
+    */
+  override def logShare(userId: String, sharee: String, shareType: String): Share = {
+    val share = Share(userId, sharee, shareType, Some(Instant.now))
+    val id = MurmurHash3.stringHash(userId + sharee + shareType).toString
```
jmthibault79: change to `generateId`

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchShareLogDAO.scala
```diff
@@ -0,0 +1,157 @@
...
+    val getSharesResponse = executeESRequest[SearchRequest, SearchResponse, SearchRequestBuilder](getSharesRequest)
+
+    if (getSharesResponse.getHits.totalHits == 0)
+      Seq.empty[Share]
+    else
+      getSharesResponse.getHits.getHits.toList map (_.getSourceAsString.parseJson.convertTo[Share])
+  }
+
+//  todo
+//  override def autocomplete(userId: String, term: String): List[String] = ???
```
jmthibault79: can you say more about this?

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala
```diff
@@ -325,4 +327,36 @@ object ModelJsonProtocol extends WorkspaceJsonSupport {
   implicit val impCreateRawlsBillingProjectFullRequestFormat = jsonFormat2(CreateRawlsBillingProjectFullRequest)
   implicit val impSpreadsheetResponse = jsonFormat1(SpreadsheetResponse)
 
+  implicit object ShareFormat extends RootJsonFormat[Share] {
```
jmthibault79: This could simply be `jsonFormat4(Share)`

You may need to `import org.broadinstitute.dsde.workbench.model.google.GoogleModelJsonSupport.InstantFormat` to enable this.

src/main/scala/org/broadinstitute/dsde/firecloud/model/ShareLog.scala
```diff
@@ -0,0 +1,12 @@
+package org.broadinstitute.dsde.firecloud.model
+
+import java.time.Instant
+
+object ShareLog {
+  final val WORKSPACE = "workspace"
+  final val GROUP = "group"
+  final val METHOD = "method"
+  case class Share(userId: String, sharee: String, shareType: String, timestamp: Option[Instant] = None)
```
jmthibault79: `shareType` should be an enum of some kind

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockShareLogDAO.scala
```diff
@@ -0,0 +1,36 @@
...
+import org.broadinstitute.dsde.firecloud.model.ShareLog.Share
+import org.broadinstitute.dsde.workbench.util.health.SubsystemStatus
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+
+class MockShareLogDAO extends ShareLogDAO {
+
+  val exception = "unit test exception: override in a class specific to test"
+
+  override def logShare(userId: String, sharee: String, shareType: String): Share = throw new Exception("unit test exception: override in a class specific to test")
```
jmthibault79: might as well make these use the `val` here

jmthibault79 (approved):

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchShareLogDAO.scala
```diff
@@ -0,0 +1,163 @@
...
+  def userSharesOfType(userId: String, shareType: Option[ShareType.Value]): QueryBuilder = {
+    shareType match {
+      case Some(typeOfShare) =>
+        boolQuery().filter(
+          boolQuery()
+            .must(userShares(userId))
+            .must(termQuery("shareType", typeOfShare.toString)))
+      case _ => userShares(userId)
+    }
+  }
```
davidangb: if you set a default of `None`:
```
def userSharesOfType(userId: String, shareType: Option[ShareType.Value] = None)
```
then `userShares` becomes obsolete, and you only need one def!

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchShareLogDAO.scala
davidangb: sorry, missed this question! Exactly what Joel said.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchShareLogDAO.scala
```diff
@@ -0,0 +1,163 @@
...
+    val getSharesQuery = client.prepareGet(indexName, datatype, id)
+    Try(executeESRequest[GetRequest, GetResponse, GetRequestBuilder](getSharesQuery)) match {
+      case Success(get) if get.isExists => get.getSourceAsString.parseJson.convertTo[Share]
+      case Success(_) => throw new FireCloudException(s"share not found")
+      case Failure(f) => throw new FireCloudException(s"error getting share for $share: ${f.getMessage}")
+    }
+  }
+
+  /**
+    * Gets all shares that have been logged for a workbench user which fall under the
```
davidangb: this will get "all" up to a limit of 100. We won't hit that too soon, but it may cause a problem in the long run. Mention that in the comment!

src/main/scala/org/broadinstitute/dsde/firecloud/service/ShareLogService.scala
```diff
@@ -0,0 +1,52 @@
...
+
+  def logShare(userId: String, sharee: String, shareType: ShareType.Value) = Future(RequestComplete(shareDAO.logShare(userId, sharee, shareType)))
+
+  def getShares(userId: String, shareType: Option[ShareType.Value] = None) = {
+    Future(RequestComplete(shareDAO.getShares(userId, shareType)))
+  }
+
+  def getSharees(userId: String, shareType: Option[ShareType.Value] = None) = {
+    Future(RequestComplete(shareDAO.getShares(userId, shareType).flatMap(_.sharee)))
+  }
```
davidangb: nit - these three methods should have similar formatting - decide whether or not to have on a single line or multiple, decide if you want the braces around the method body.

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -133,10 +134,21 @@ class WorkspaceService(protected val argUserToken: WithAccessToken, val rawlsDAO
...
+      shareLogDAO.logShares(originId, sharees, ShareType.WORKSPACE)
+    }
+
     val aclUpdate = rawlsDAO.patchWorkspaceACL(workspaceNamespace, workspaceName, aclUpdates, inviteUsersNotFound)
+
     aclUpdate map { actualUpdates =>
-      RequestComplete(actualUpdates)
+      val shares = logShares(actualUpdates)
+      // todo swagger 2.0 doesn't support arrays of mixed types. will need to revisit
+      RequestComplete((actualUpdates, shares))
```
davidangb: why does this need to change the response payload at all? Is there a reason to return `shares` in the response?

src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/ElasticSearchShareLogDAOSpec.scala
```diff
@@ -0,0 +1,107 @@
...
+        val share = Share("roger", "syd@gmail.com", ShareType.WORKSPACE)
+        val loggedShare = shareLogDAO.logShare(share.userId, share.sharee, share.shareType)
+        val check = shareLogDAO.getShare(share)
+        assertResult(loggedShare) { check }
+      }
+      "should successfully log a duplicate share" in {
+        val loggedShare = shareLogDAO.logShare("fake4", "fake3@gmail.com", ShareType.WORKSPACE)
+        val check = Try(shareLogDAO.logShare(loggedShare.userId, loggedShare.sharee, loggedShare.shareType))
+        assert(check.isSuccess)
+      }
```
davidangb: how is this a duplicate?

src/main/scala/org/broadinstitute/dsde/firecloud/service/ShareLogService.scala
```diff
@@ -0,0 +1,52 @@
...
+import spray.httpx.SprayJsonSupport
+import spray.json.DefaultJsonProtocol._
+
+import scala.concurrent.{ExecutionContext, Future}
+
+object ShareLogService {
+  sealed trait ShareLogMessage
+  case class LogShare(userId: String, sharee: String, shareType: ShareType.Value) extends ShareLogMessage
+  case class GetShares(userId: String, shareType: Option[ShareType.Value] = None) extends ShareLogMessage
+  case class GetSharees(userId: String, shareType: Option[ShareType.Value] = None) extends ShareLogMessage
```
davidangb: it looks like the `GetShares` case class, and the `getShares` method, are never used. Clean them up?

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/ShareLogApiService.scala
```diff
@@ -0,0 +1,41 @@
...
+              shareType match {
+                case Some(typeOfShare) =>
+                  perRequest(requestContext,
+                    ShareLogService.props(shareLogServiceConstructor),
+                    ShareLogService.GetSharees(userInfo.id, Some(ShareType.withName(typeOfShare))))
+                case None =>
+                  perRequest(requestContext,
+                    ShareLogService.props(shareLogServiceConstructor),
+                    ShareLogService.GetSharees(userInfo.id))
+              }
```
davidangb: could this entire match block be reduced to:
```
              perRequest(requestContext,
                ShareLogService.props(shareLogServiceConstructor),
                ShareLogService.GetSharees(userInfo.id, shareType.map(ShareType.withName)))
```

src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/ElasticSearchShareLogDAOSpec.scala
```diff
@@ -0,0 +1,107 @@
...
+        val userId = "fake2"
+        val shareType = ShareType.GROUP
+        val expected = ElasticSearchShareLogDAOSpecFixtures.fixtureShares.filter(_.userId == userId).filter(_.shareType == shareType)
+        val sharees = expected.map(_.sharee)
+        val check = shareLogDAO.logShares(userId, sharees, ShareType.GROUP)
+        assertResult(expected.map(s => (s.userId, s.sharee, s.shareType))) {
+          check.map(s => (s.userId, s.sharee, s.shareType))
+        }
+      }
+    }
```
davidangb: since everything in `ElasticSearchShareLogDAOSpecFixtures` is inserted into the index during `init`, this test is really only checking that `init` did its thing and that attempting to insert those rows again does not cause an error. Am I reading that right?

davidangb:
Okay, done with this round of review, comments are inline. One notable task that remains is to confer with the Saturn team and see if this is a feature they want to take advantage of - that could change our architecture significantly.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchShareLogDAO.scala
ghost: Would it be worth adding a `log.debug` when 100 items are returned?

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
ghost: It was strictly so that I could make a test assertion in `WorkspaceApiServiceSpec`. Is that assertion worth having in your opinion?

src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/ElasticSearchShareLogDAOSpec.scala
ghost: Bad wording! Changing to `"should successfully log a record of a user sharing a workspace with the same user twice"`

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
davidangb: I do like being able to test it. BUT - since changing the response payload affects end users, and this is a breaking API change, I don't think it's worth it. We should refactor code to find a way to test `logShares` without affecting the end user.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchShareLogDAO.scala
davidangb: our default log level is `info` so we'd never see it! https://github.com/broadinstitute/firecloud-orchestration/blob/develop/src/main/resources/logback.xml#L22

I think a `warn` is good here. That'll trigger Sentry - but we should be bugged about it, because if we reach that point we need to update code/rethink approach.

src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/ElasticSearchShareLogDAOSpec.scala
ghost: that is right... seems a bit redundant now. i'll remove the `logShares` call and change the wording/logic to assert that `init` was successful.

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
ghost: technically `ElasticSearchShareLogDAO.logShares` is unit tested, and this is really just testing that the MockShareLogDAO endpoint is called. Do we need more test coverage than that?

ghost:
jenkins retest

ghost:
automation passed, integration failed fast but passed locally. jenkins retest

ghost:
testing infrastructure needed a fix - jenkins retest

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
```diff
@@ -14,6 +14,7 @@ import org.broadinstitute.dsde.firecloud.service._
 import org.broadinstitute.dsde.firecloud.trial.ProjectManager
 import org.broadinstitute.dsde.firecloud.webservice._
 import org.broadinstitute.dsde.workbench.util.health.HealthMonitor
+import org.elasticsearch.action.support.WriteRequest.RefreshPolicy
```
davidangb: unused import?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchShareLogDAO.scala
```diff
@@ -0,0 +1,166 @@
...
+  def userShares(userId: String, shareType: Option[ShareType.Value] = None): QueryBuilder = {
+    shareType match {
+      case Some(typeOfShare) =>
+        boolQuery().filter(
+          boolQuery()
+            .must(termQuery("userId", userId))
+            .must(termQuery("shareType", typeOfShare.toString)))
+      case _ => termQuery("userId", userId)
+    }
+  }
```
davidangb: these are technically not parallel queries, since the first has the extra `boolQuery.filter` nesting but the second doesn't. I don't *think* this would cause any difference in results, but if it did ... it could be a difficult bug to track down. I suggest making the two queries exactly the same, except for the `shareType` criteria.

I might write it like this - your call on whether or not to refactor like this:
```
  def userShares(userId: String, shareType: Option[ShareType.Value] = None): QueryBuilder = {
    // always include the sharer in query criteria
    val userIdQuery = boolQuery().must(termQuery("userId", userId))
    // if a shareType was specified, include that also
    shareType.map{ typeOfShare => userIdQuery.must(termQuery("shareType", typeOfShare.toString)) }
    boolQuery.filter(userIdQuery)
  }
```
alternately, skip the `filter`:
```
  def userShares(userId: String, shareType: Option[ShareType.Value] = None): QueryBuilder = {
    // always include the sharer in query criteria
    val userIdQuery = boolQuery().must(termQuery("userId", userId))
    // if a shareType was specified, include that also
    shareType.map{ typeOfShare => userIdQuery.must(termQuery("shareType", typeOfShare.toString)) }
    userIdQuery
  }
```

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/ShareLogApiServiceSpec.scala
```diff
@@ -0,0 +1,39 @@
...
+  val localShareLogDao = new ShareLogApiServiceSpecShareLogDAO
+
+  override val shareLogServiceConstructor: () => ShareLogService = ShareLogService.constructor(app.copy(shareLogDAO = localShareLogDao))
+
+  override def actorRefFactory: ActorRefFactory = system
+
+  "ShareLogApiService" - {
+    "when getting all sharees" in {
+      Get(getShareesPath)  ~> getUserHeaders("fake1", "fake1@gmail.com") ~> sealRoute(shareLogServiceRoutes) ~> check {
+        assertResult(OK) { status }
```
davidangb, line 31: check the response payload for correctness, too - should be the values from `ElasticSearchShareLogDAOSpecFixtures.fixtureShares`

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/ShareLogApiServiceSpec.scala
```diff
@@ -0,0 +1,39 @@
...
+      }
+    }
+    "when getting workspace sharees" in {
+      Get(makeGetShareesPath(ShareType.WORKSPACE)) ~> getUserHeaders("fake1", "fake1@gmail.com") ~> sealRoute(shareLogServiceRoutes) ~> check {
+        assertResult(OK) { status }
```
davidangb, line 37: check the response payload here too.

src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
```diff
@@ -595,15 +601,25 @@ class WorkspaceApiServiceSpec extends BaseServiceSpec with WorkspaceApiService w
...
+        status should equal(OK)
+      }
+    }
+
+    "OK status and logged shares returned from PATCH on /workspaces/%s/%s/acl" in {
+      val aclUpdates = ElasticSearchShareLogDAOSpecFixtures.fixtureShares map { share =>
+        WorkspaceACLUpdate(share.sharee, WorkspaceAccessLevels.Read, Some(false))
+      }
+
+      Patch(aclPath, aclUpdates) ~> dummyUserIdHeaders(dummyUserId) ~> sealRoute(workspaceRoutes) ~> check {
```
davidangb: is this test now irrelevant?

src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/ElasticSearchShareLogDAOSpec.scala
```diff
@@ -0,0 +1,97 @@
...
+    // using the recreate from search dao because we don't have recreate in sharelog dao
+    searchDAO.recreateIndex()
+    ElasticSearchShareLogDAOSpecFixtures.fixtureShares map { share =>
+      shareLogDAO.logShare(share.userId, share.sharee, share.shareType)
+    }
+  }
+
+  override def afterAll = {
+    // using the delete from search dao because we don't have recreate in sharelog dao
+    searchDAO.deleteIndex()
```
davidangb, line 21: It would be cleaner to move `recreateIndex` and `deleteIndex` implementations into a trait or base class, and allow multiple DAOs to reuse them. That refactor could be done in a follow-on PR.

src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/ElasticSearchShareLogDAOSpec.scala
```diff
@@ -0,0 +1,97 @@
...
+        val expected = ElasticSearchShareLogDAOSpecFixtures.fixtureShares
+          .filter(s => s.userId.equals("fake1"))
+          .filter(s => s.shareType.equals(ShareType.GROUP))
+          .sortBy(_.sharee)
+        val check = shareLogDAO.getShares("fake1", Some(ShareType.GROUP)).sortBy(_.sharee)
+
+        assertResult(expected.size) { check.size }
+        assertResult(expected.map(s => (s.userId, s.sharee, s.shareType))) { check.map(s => (s.userId, s.sharee, s.shareType)) }
+      }
+    }
```
davidangb: there's a lot of `foo.map(s => (s.userId, s.sharee, s.shareType))` in this test. I suggest making a helper method to do this, and rely on `copy`:
```
private def scrubShares(in: Seq[Share]) = in.map(_.copy(timestamp = None))
```

also I believe (untested) you can use `theSameElementsAs` to avoid all the sorting. instead of having to match sizes and sort both actual and expected, you could do something like:
```
scrubShares(check) should contain theSameElementsAs expected
```

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchShareLogDAO.scala
```diff
@@ -0,0 +1,166 @@
...
+      executeESRequest[CreateIndexRequest, CreateIndexResponse, CreateIndexRequestBuilder](
+      client.admin.indices.prepareCreate(indexName))
+      // Try one more time and fail if index creation fails
+      if (!indexExists)
+      throw new FireCloudException(s"index $indexName does not exist!")
+    }
+  }
+
+  /**
+    * Uses MurmerHash3 for quick hashing -
```
davidangb: nit: typo :)

davidangb (commented):
comments inline!

src/main/resources/swagger/api-docs.yaml
```diff
@@ -1704,6 +1704,34 @@ paths:
...
+      summary: |
+        Get a list of users with whom current user has shared a Workspace.
+      description: |
+        Get a list of users with whom current user has shared a Workspace.
+      parameters:
+        - name: shareType
+          required: false
+          in: query
+          description: |
+           The type of share - Workspace, Group, or Method. If left empty, all types are returned.
```
ghost: @davidangb would you suggest hiding the possibility of  `Group` and `Method` for the time being?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchShareLogDAO.scala
ghost: Really good point! Thank you

src/test/scala/org/broadinstitute/dsde/firecloud/integrationtest/ElasticSearchShareLogDAOSpec.scala
ghost: Very nice!

src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
ghost: yes

src/main/resources/swagger/api-docs.yaml
davidangb: how about remove the entire parameter? End users can't do anything useful with `shareType` right now.

davidangb (approved):
:+1: after removing `shareType` param in swagger-ui, or otherwise making it clearer to end users

ghost:
swat tests passed prior to recent change which did not touch code. jenkins retest

PR #670, opened 2018-08-13
mtalbott add me param

marctalbott:
Change /me endpoint to accept a parameter which utilizes the register/user/v2/self/info endpoint instead of the register/user/v1 endpoint to cutdown on expensive calls to populate the google field

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
@@ -77,7 +77,7 @@ trait UserApiService extends HttpService with PerRequestCreator with FireCloudRe
 
   val userServiceRoutes =
     path("me") {
-      parameter("V2".?) { V2 =>
+      parameter("userDetailsOnly".?) { userDetailsOnly =>
         get { requestContext =>
 
           // inspect headers for a pre-existing Authorization: header
@@ -92,7 +92,7 @@ trait UserApiService extends HttpService with PerRequestCreator with FireCloudRe
             // browser sent Authorization header; try to query Sam for user status
             case Some(_) =>
               val pipeline = authHeaders(requestContext) ~> sendReceive
-              val version2 = V2.exists(_.equalsIgnoreCase("true"))
+              val version2 = userDetailsOnly.exists(_.equalsIgnoreCase("true"))
               val samRequest = if (version2) Get(UserApiService.samRegisterUserV2URL) else Get(UserApiService.samRegisterUserURL)
               pipeline(samRequest) onComplete {
                 case Success(response: HttpResponse) =>
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
@@ -279,10 +279,10 @@ trait UserApiService extends HttpService with PerRequestCreator with FireCloudRe
           case InternalServerError =>
             respondWithErrorReport(InternalServerError, "Identity service encountered an unknown error, please try again.", requestContext)
           case OK =>
-            response.entity.as[WorkbenchEnabled] match {
+            response.entity.as[WorkbenchEnabledV2] match {
               case Right(diagnostics) =>
-                if (diagnostics.allUsersGroup && diagnostics.google) {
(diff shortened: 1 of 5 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
```diff
@@ -74,58 +77,75 @@ trait UserApiService extends HttpService with PerRequestCreator with FireCloudRe
...
-                    val respJson = response.entity.as[RegistrationInfo]
-                    respJson match {
-                      case Right(regInfo) =>
-                        if (regInfo.enabled.google && regInfo.enabled.ldap && regInfo.enabled.allUsersGroup) {
-                          // Sam says the user is fully registered and activated!
-                          requestContext.complete(OK, regInfo)
-                        } else {
-                          // Sam knows about the user, but the user isn't activated
-                          respondWithErrorReport(Forbidden, "FireCloud user not activated.", requestContext)
+          authorizationHeader match {
```
gpolumbo-broad: I'm trying to figure out what you changed vs. what git _thinks_ you changed.  I think I count at least 4 neseted case statements, on top of the normal akka nesting.  With caution, I would recommend that if you're making changes here, try to start extracting these nested statements into their own methods.  

This is the sort of thing I'd normally reject a PR for, but seeing as it doesn't seem like you are leaving the code in any worse shape than when you started (I think, it's hard to tell, and that's part of the problem) I'm not going to worry too much about it.

src/test/scala/org/broadinstitute/dsde/firecloud/service/UserApiServiceSpec.scala
```diff
@@ -543,6 +543,130 @@ class UserApiServiceSpec extends BaseServiceSpec with RegisterApiService with Us
...
+      "OK response is returned" in {
+
+        samServer.clear(request.withMethod("GET").withPath(UserApiService.samRegisterUserV2Path))
+        samServer
+          .when(request.withMethod("GET").withPath(UserApiService.samRegisterUserV2Path))
+          .respond(
+            org.mockserver.model.HttpResponse.response()
+              .withBody("""{"userSubjectId": "1111111111", "userEmail": "no@nope.org", "enabled": true}""")
+              .withHeaders(MockUtils.header).withStatusCode(OK.intValue)
+          )
```
gpolumbo-broad: There are different style preferences in the tests about copy/pasting mocking setup statements like these between tests.  I prefer to treat your testing code as you would production code and keep it DRY and extract the copy/pasted stuff out into its own method.  Do what you want here.

gpolumbo-broad (approved):

gpolumbo-broad:
@marctalbott @Qi77Qi not sure if you have been granted access to Codacy yet:  https://app.codacy.com/app/BroadInstitute/firecloud-orchestration/pullRequest?prid=2038494 here's the link to the analysis for this branch, which is in good shape based on the rules being enforced at the moment.  Of particular note, the cyclometric complexity rule is disabled.  Anyway, we should make sure you have access.

MatthewBemis (approved):

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
```diff
@@ -74,58 +77,31 @@ trait UserApiService extends HttpService with PerRequestCreator with FireCloudRe
...
-                respondWithErrorReport(ServiceUnavailable, "Identity service did not produce a timely response, please try again later.", error, requestContext)
-            }
+          authorizationHeader match {
+            // no Authorization header; the user must be unauthorized
+            case None =>
+              respondWithErrorReport(Unauthorized, "No authorization header in request.", requestContext)
+            // browser sent Authorization header; try to query Sam for user status
+            case Some(_) =>
+              val pipeline = authHeaders(requestContext) ~> sendReceive
+              val version2 = userDetailsOnly.exists(_.equalsIgnoreCase("true"))
```
dvoet: do we have to support both v1 and v2?

dvoet (changes requested):
assuming we want to continue support of /me v1, the api should be composed of the 2 new sam v2 api calls. The intent is that the sam v1 api is deprecated so we should stop using it. I think that also might simplify things a little: always do the sam info call then add in the sam details call if supporting v1.

dvoet (approved):

PR #672, opened 2018-09-10
upgrade jackson libs [risk:low]

davidangb:
upgrade Jackson libs due to proactive security scan warnings.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -3,8 +3,7 @@ import sbt._
 object Dependencies {
   val akkaV = "2.4.19"
   val sprayV = "1.3.4"
-  val jacksonV = "2.8.10"
-     // note that jackson-databind overrides this below! 2.8.8.1 is not released for core or annotations.
+  val jacksonV = "2.9.6"
 
   def excludeGuavaJDK5(m: ModuleID): ModuleID = m.exclude("com.google.guava", "guava-jdk5")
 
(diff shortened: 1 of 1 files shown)


project/Dependencies.scala
```diff
@@ -3,8 +3,7 @@ import sbt._
 object Dependencies {
   val akkaV = "2.4.19"
   val sprayV = "1.3.4"
-  val jacksonV = "2.8.10"
-     // note that jackson-databind overrides this below! 2.8.8.1 is not released for core or annotations.
```
jmthibault79, line 7: Is this still true?

project/Dependencies.scala
```diff
@@ -49,7 +48,8 @@ object Dependencies {
     excludeGuavaJDK5("com.google.apis"                % "google-api-services-cloudbilling" % "v1-rev14-1.23.0"),
 
     "org.webjars"                    % "swagger-ui"          % "2.2.5",
-    "com.jason-goodwin"             %% "authentikat-jwt"     % "0.4.1",
+    "com.jason-goodwin"             %% "authentikat-jwt"     % "0.4.5"
```
jmthibault79, line 51: Is this because 0.4.1 is no longer compatible with new Jackson?

davidangb: no. I was seeing that sourceclear reported a jackson issue as a transitive dependency from authentikat-jwt. So I'm just being explicit here and making sure that authentikat-jwt doesn't pull in jackson - instead, it'll use the jackson I defined above.

project/Dependencies.scala
davidangb: it is not true for 2.9.6 - annotations, databind, and core are all available at 2.9.6. In fact, this comment became untrue as of #570 but I failed to remove it then.

jmthibault79 (approved):

davidangb:
jenkins retest, now that gpalloc projects are fixed.

davidangb:
two UI test failures, both `group not found RawlsGroupRef(RawlsGroupName(AuthDomain_ ... ))`; will retest those manually.
```
[info] *** 2 TESTS FAILED ***
[error] Failed: Total 122, Failed 2, Errors 0, Passed 120, Ignored 7
[error] Failed tests:
[error] 	org.broadinstitute.dsde.firecloud.test.security.AuthDomainMultiGroupsSpec
[error] (Test / testOnly) sbt.TestsFailedException: Tests unsuccessful
```

Update: passed in a manual run at https://fc-jenkins.dsp-techops.broadinstitute.org/view/Testing/view/Pipelines/job/swatomation-pipeline/9205/.

PR #673, opened 2018-09-12
add git revision endpoint at /version [risk: medium]

davidangb:
I got frustrated always wondering which version of the app was running in which environment.

This PR adds git revision/hash endpoint at /version. Developers/admins/other interested parties can call this endpoint to see which specific commit is in use. 

The `getOrElse` fallback should should make this code resilient to environments where git information isn't available - notably, when running via ./config/docker-rsync-local-ui.sh. In that local environment, we rsync codebases into a docker volume and don't sync the .git directory. But in Jenkins builds, this information should be available; I verified by starting a custom fiab and saw an API response of:

```
{
  "version": "0.1-ecb1bbc-SNAPSHOT"
}
```

I'm very conservatively labeling this [risk:medium] instead of low. Potentially, an attacker could use the revision information to see which bugs/vulnerabilities have been fixed since the commit in use. However, since our github repo is open-source anyway, this doesn't actually give the attacker any new information - it just makes it slightly easier to know which code is in use when.

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -1878,6 +1878,19 @@ paths:
         500:
           description: Internal Server Error
 
+  /version:
+    get:
+      tags:
+        - Version
+      operationId: orchestrationVersion
+      summary: Returns the currently deployed version of this service.
+      responses:
+        200:
+          description: OK
+          schema:
+            $ref: '#/definitions/OrchestrationVersion'
+      security: []
+
   /version/executionEngine:
     get:
       tags:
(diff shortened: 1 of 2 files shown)


davidangb:
jenkins retest (transient build-time failure)

davidangb:
and now a fiab-start failure. jenkins retest, again.

adrazhi (commented):
Hi @davidangb, your logic makes a lot of sense in that even if this introduces an attack surface at some level, it does not add much due to the fact that our repo is open to the world.

Is there anything else associated with this that's likely to change in the future?

adrazhi (approved):

jmthibault79 (approved):
Looks good.  Tested locally with
```
{
  "version": "n/a"
}
```

PR #677, opened 2018-10-02
GAWB-3846: initial impl to support tsv download

ahaessly:

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelSchema.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelSchema.scala
@@ -2,9 +2,7 @@ package org.broadinstitute.dsde.firecloud.model
 
 import scala.io.Source
 import scala.util.{Failure, Success, Try}
-
 import spray.json._
-
 import org.broadinstitute.dsde.firecloud.FireCloudConfig
 
 /**
@@ -15,6 +13,7 @@ import org.broadinstitute.dsde.firecloud.FireCloudConfig
  */
 
 object ModelSchema {
+
 
   def getTypeSchema(entityType: String): Try[EntityMetadata] = {
     EntityTypes.types.get(entityType) match {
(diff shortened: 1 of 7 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelSchema.scala
```diff
@@ -59,3 +58,42 @@ case class EntityMetadata(
 )
 
 case class EntityModel(schema : Map[String, EntityMetadata]) //entity name -> stuff about it
+
+object FlexibleModelSchema {
```
ahaessly, line 63: does it make sense for FlexibleModelSchema to be an object? i think after i implement the upload, we may be able to get rid of ModelSchema.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
```diff
@@ -153,11 +153,13 @@ object TSVFormatter {
...
   def makeEntityRows(entityType: String, entities: Seq[Entity], headers: IndexedSeq[String]): IndexedSeq[IndexedSeq[String]] = {
-    val memberPlural = pluralizeMemberType(memberTypeFromEntityType(entityType))
     // if we have a set entity, we need to filter out the attribute array of the members so that we only
     // have top-level attributes to construct columns from.
-    val filteredEntities = if (isCollectionType(entityType)) {
-      filterAttributeFromEntities(entities, memberPlural)
+    val filteredEntities = if (FlexibleModelSchema.isCollectionType(entityType)) {
+      FlexibleModelSchema.memberTypeFromEntityType(entityType) match {
+        case Success(memberType) => filterAttributeFromEntities(entities, FlexibleModelSchema.pluralizeEntityType(memberType))
+        case Failure(regret) => ??? // should not happen?
```
ahaessly, line 161: i could use some help with this implementation. throw an exception from the failure case?

PR #678, opened 2018-10-02
GAWB-3846, GAWB-3847 tsv download and upload for non-FC model entities

ahaessly:

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
@@ -126,10 +126,7 @@ class EntityClient (requestContext: RequestContext)(implicit protected val execu
    * Returns the plural form of the entity type.
    * Bails with a 400 Bad Request if the entity type is unknown to the schema. */
   private def withPlural(entityType: String)(op: (String => Future[PerRequestMessage])): Future[PerRequestMessage] = {
-    ModelSchema.getPlural(entityType) match {
-      case Failure(regret) => Future(RequestCompleteWithErrorReport(BadRequest, regret.getMessage))
-      case Success(plural) => op(plural)
-    }
+    op(FlexibleModelSchema.pluralizeEntityType(entityType))
   }
 
 
@@ -138,15 +135,12 @@ class EntityClient (requestContext: RequestContext)(implicit protected val execu
    * Bails with a 400 Bad Request if the entity type is unknown or if some attributes are missing.
    * Returns the list of required attributes if all is well. */
   private def withRequiredAttributes(entityType: String, headers: Seq[String])(op: (Map[String, String] => Future[PerRequestMessage])):Future[PerRequestMessage] = {
-    ModelSchema.getRequiredAttributes(entityType) match {
-      case Failure(regret) => Future(RequestCompleteWithErrorReport(BadRequest, regret.getMessage))
-      case Success(requiredAttributes) =>
-        if( !requiredAttributes.keySet.subsetOf(headers.toSet) ) {
-          Future( RequestCompleteWithErrorReport(BadRequest,
-            "TSV is missing required attributes: " + (requiredAttributes.keySet -- headers).mkString(", ")) )
-        } else {
-          op(requiredAttributes)
-        }
+    val requiredAttributes = FlexibleModelSchema.getRequiredAttributes(entityType)
+    if( !requiredAttributes.keySet.subsetOf(headers.toSet) ) {
+      Future( RequestCompleteWithErrorReport(BadRequest,
+        "TSV is missing required attributes: " + (requiredAttributes.keySet -- headers).mkString(", ")) )
+    } else {
+      op(requiredAttributes)
     }
   }
 
@@ -225,15 +219,9 @@ class EntityClient (requestContext: RequestContext)(implicit protected val execu
     checkFirstColumnDistinct(tsv) {
       withMemberCollectionType(entityType) { memberTypeOpt =>
         checkNoCollectionMemberAttribute(tsv, memberTypeOpt) {
-          ModelSchema.getRequiredAttributes(entityType) match {
-            //Required attributes aren't required to be headers in update TSVs - they should already have been
-            //defined when the entity was created. But we still need the type information if the headers do exist.
-            case Failure(regret) => Future(RequestCompleteWithErrorReport(BadRequest, regret.getMessage))
-            case Success(requiredAttributes) =>
-              val colInfo = colNamesToAttributeNames(tsv.headers, requiredAttributes)
-              val rawlsCalls = tsv.tsvData.map(row => setAttributesOnEntity(entityType, memberTypeOpt, row, colInfo))
-              batchCallToRawls(pipeline, workspaceNamespace, workspaceName, entityType, rawlsCalls, "batchUpdate")
-          }
+          val colInfo = colNamesToAttributeNames(tsv.headers, FlexibleModelSchema.getRequiredAttributes(entityType))
+          val rawlsCalls = tsv.tsvData.map(row => setAttributesOnEntity(entityType, memberTypeOpt, row, colInfo))
+          batchCallToRawls(pipeline, workspaceNamespace, workspaceName, entityType, rawlsCalls, "batchUpdate")
         }
       }
     }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelSchema.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelSchema.scala
@@ -2,9 +2,7 @@ package org.broadinstitute.dsde.firecloud.model
 
 import scala.io.Source
 import scala.util.{Failure, Success, Try}
-
 import spray.json._
-
 import org.broadinstitute.dsde.firecloud.FireCloudConfig
 
 /**
@@ -15,6 +13,7 @@ import org.broadinstitute.dsde.firecloud.FireCloudConfig
  */
 
 object ModelSchema {
+
 
   def getTypeSchema(entityType: String): Try[EntityMetadata] = {
     EntityTypes.types.get(entityType) match {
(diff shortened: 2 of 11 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
```diff
@@ -153,11 +153,13 @@ object TSVFormatter {
...
   def makeEntityRows(entityType: String, entities: Seq[Entity], headers: IndexedSeq[String]): IndexedSeq[IndexedSeq[String]] = {
-    val memberPlural = pluralizeMemberType(memberTypeFromEntityType(entityType))
     // if we have a set entity, we need to filter out the attribute array of the members so that we only
     // have top-level attributes to construct columns from.
-    val filteredEntities = if (isCollectionType(entityType)) {
-      filterAttributeFromEntities(entities, memberPlural)
+    val filteredEntities = if (FlexibleModelSchema.isCollectionType(entityType)) {
+      FlexibleModelSchema.memberTypeFromEntityType(entityType) match {
+        case Success(memberType) => filterAttributeFromEntities(entities, FlexibleModelSchema.pluralizeEntityType(memberType))
+        case Failure(regret) => ??? // should not happen?
```
ahaessly, line 161: any suggestions on what to do here? throw exception?

ahaessly:
jenkins retest

ahaessly:
jenkins retest

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
```diff
@@ -95,13 +96,12 @@ object TSVFormatter {
...
     */
   def makeMembershipRows(entityType: String, entities: Seq[Entity]): Seq[IndexedSeq[String]] = {
-    val memberPlural = pluralizeMemberType(memberTypeFromEntityType(entityType))
     entities.filter { _.entityType == entityType }.flatMap {
       entity =>
         entity.attributes.filter {
           // To make the membership file, we need the array of elements that correspond to the set type.
           // All other top-level properties are not necessary and are only used for the data load file.
-          case (attributeName, _) => attributeName.equals(AttributeName.withDefaultNS(memberPlural))
+          case (attributeName, _) => attributeName.equals(AttributeName.withDefaultNS(FlexibleModelSchema.pluralizeEntityType(FlexibleModelSchema.memberTypeFromEntityType(entityType).get)))
```
breilly2, line 104: `.get` here could throw an exception before `pluralizeEntityType` has a chance to append an "s" as a fallback.

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelSchema.scala
```diff
@@ -59,3 +43,41 @@ case class EntityMetadata(
...
+
+object FlexibleModelSchema {
+  def memberTypeFromEntityType(entityType: String): Try[String] = {
+    val memberType = EntityTypes.types.get(entityType) match {
+      case Some(schema) => schema.memberType
+      case None => if (entityType.endsWith("_set"))
+        Some(entityType.replace("_set", "")) else
+        None
+    }
+    Try(memberType.get)
```
breilly2, line 55: With the fixed data model, it would be programmer error for the JSON to contain a set type without a `memberType` property. When the set type name is user supplied, we're enforcing that it must be named "${memberType}_set". Therefore, we need to report an error back to the user if they deviate from that.

I think something like this will do it:
```
Try(memberType.getOrElse(throw new RuntimeException("Entity set type name must end with \"_set\": " + entityType)))
```

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
```diff
@@ -84,7 +85,7 @@ object TSVFormatter {
     * @return IndexedSeq of header Strings
     */
   def makeMembershipHeaders(entityType: String): IndexedSeq[String] = {
-    IndexedSeq[String](s"${TsvTypes.MEMBERSHIP}:${entityType}_id", memberTypeFromEntityType(entityType))
+    IndexedSeq[String](s"${TsvTypes.MEMBERSHIP}:${entityType}_id", FlexibleModelSchema.memberTypeFromEntityType(entityType).getOrElse(""))
```
breilly2, line 88: I think we want just `.get` instead of `.getOrElse`. If getting the membership type fails, we want to raise that exception, not use `""` as a header.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
```diff
@@ -119,7 +119,7 @@ object TSVFormatter {
     * @return Entity name as first column header, followed by matching entity attribute labels
     */
   def makeEntityHeaders(entityType: String, allHeaders: Seq[String], requestedHeaders: Option[IndexedSeq[String]]): IndexedSeq[String] = {
-    val memberPlural = pluralizeMemberType(memberTypeFromEntityType(entityType))
+    val memberPlural = FlexibleModelSchema.pluralizeEntityType(FlexibleModelSchema.memberTypeFromEntityType(entityType).getOrElse(""))
```
breilly2, line 122: Same comment as line 88.

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelSchema.scala
```diff
@@ -22,22 +22,6 @@ object ModelSchema {
       case None => Failure(new RuntimeException("Unknown entity type: " + entityType))
     }
   }
```
ansingh7115, line 24: This `getTypeSchema` is the only thing remaining in `ModelSchema` and is the only used in one other place (`backwardsCompatStripIdSuffixes(...)`) - is it necessary to keep it in a separate object?

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelSchema.scala
```diff
@@ -59,3 +43,41 @@ case class EntityMetadata(
 )
 
 case class EntityModel(schema : Map[String, EntityMetadata]) //entity name -> stuff about it
+
```
ansingh7115, line 46: Maybe some code comments to explain FlexibleModelSchema and ModelSchema and how they relate?

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
ansingh7115: Maybe a 500 error, yeah

breilly2: It can happen if the collection type is not named with a `_set` suffix which, outside of the FC data model, is a user error. Therefore, I think just calling `.get` instead of matching and letting the exception be thrown would be fine.
```
    val filteredEntities = if (FlexibleModelSchema.isCollectionType(entityType)) {
      val memberPlural = FlexibleModelSchema.pluralizeEntityType(FlexibleModelSchema.memberTypeFromEntityType(entityType).get)
      filterAttributeFromEntities(entities, memberPlural)
    } else {
      entities
    }
```

breilly2:
Worth noting that this change introduces a convention that set types are named with a `_set` suffix... and all entity types with a `_set` suffix will be treated as set types... whether or not that's the desired interpretation.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -126,10 +126,7 @@ class EntityClient (requestContext: RequestContext)(implicit protected val execu
    * Returns the plural form of the entity type.
    * Bails with a 400 Bad Request if the entity type is unknown to the schema. */
```
breilly2, line 127: This doc comment and others need to be updated.

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelSchema.scala
```diff
@@ -59,3 +43,41 @@ case class EntityMetadata(
...
+object FlexibleModelSchema {
+  def memberTypeFromEntityType(entityType: String): Try[String] = {
+    val memberType = EntityTypes.types.get(entityType) match {
+      case Some(schema) => schema.memberType
+      case None => if (entityType.endsWith("_set"))
+        Some(entityType.replace("_set", "")) else
```
breilly2, line 52: Nit: instead of `else None`, I think a 3rd `case` would read a little bit better:
```
      case Some(schema) => schema.memberType
      case None if entityType.endsWith("_set") =>
        Some(entityType.replace("_set", ""))
      case _ => None
```

breilly2 (changes requested):
Suggested changes are in comments. Let me know if you have any questions.

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelSchema.scala
ahaessly: ugh. this doesn't work. it needs to return None instead of throwing an error because of how withMemberCollectionType in EntityClient is used. it makes this call regardless of whether is it a set or not. and None is the propper response when it is not a set. more suggestions welcome.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
ahaessly: nice catch. since it is filtering, i'll just do a getOrElse "", which should not pass the filter

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
ahaessly: this is also used in filtering (filterNot), so i think it's actually ok to default to empty string

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelSchema.scala
breilly2: Hmm... thinking out loud, haven't tried... but maybe `withMemberCollectionType`'s `op` parameter should be a `Try[Option[String]] => Future[PerRequestMessage]` so that the caller can decide what to do with the failure. When called from `EntityClient.importMembershipTSV`, throwing an error is right, I think... whereas the other two cases in `EntityClient` more-or-less expect a `None` result... I think... kinda hard to tell because they're passing it to functions that are called from other places that have different expectations about the value.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
breilly2: It looks to me like `getOrElse(entityType)` would most closely mimic the previous behavior:
```
  def memberTypeFromEntityType(entityType: String): String = ModelSchema.getCollectionMemberType(entityType).get.getOrElse(entityType.replace("_set", ""))
```
This previously looked up the entity type in the schema, throwing an exception if missing. Now, we want any entity type to be valid because there's no schema. However, we might still get `None` from `FlexibleModelSchema.memberTypeFromEntityType(entityType)` if the name doesn't end in `_set`. Previously, as long as it was a valid type, we'd trim `_set` from the name if it's in there, otherwise leave the name alone. Therefore `getOrElse(entityType)`.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
breilly2: Good point, though again I think that `getOrElse(entityType)` would be more similar to the previous behavior and feels a little less arbitrary than `""`. However, I haven't fully thought through whether or not this would affect the `filterNot`.

PR #680, opened 2018-10-10
GAWB-3840 use sam for workspace authz and update es query

ahaessly:

In all cases:


dvoet:
@ahaessly is this still a thing?

ahaessly:
@dvoet yes. i'm waiting until the sam changes to orch are merged and then i'll rebase, test and merge this PR.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -225,26 +225,26 @@ class LibraryService (protected val argUserInfo: UserInfo,
...
-
-      (for {
-        docs <- docsFuture
-        ids <- idsFuture
-      } yield RequestComplete(updateAccess(docs, ids))) recover {
-        case ex =>
-          throw new FireCloudExceptionWithErrorReport(ErrorReport(errorMessageFromSearchException(ex)))
-      }
+  def searchFor(criteria: LibrarySearchParams, searchMethod:(LibrarySearchParams, Seq[String], Map[String, String])=>Future[LibrarySearchResponse]): Future[PerRequestMessage] ={
+    val workspaces: Future[Seq[WorkspaceListResponse]] = rawlsDAO.getWorkspaces
```
davidangb: this call -  to list all workspaces for the user - is the bottleneck in Library.  Now that Sam owns workspace authz, could this be replaced with a call to Sam's `/api/resources/v1/workspace`?

Even better would be to query Sam for only those workspaceIds returned from `findDocuments`, or even just the visible page of workspaceIds, but in the current Sam APIs this doesn't look possible.

ahaessly: so the plan now is to also match data sets where the user already has access to the workspace. so i am planning on getting workspace ids and passing them in as filter criteria. i agree that it would be more efficient going straight to sam, but i'm not sure i can get all the info i need (may need auth domain). i'll check it out.

ahaessly:
reopening to merge into develop

PR #681, opened 2018-10-12
GAWB-3847 flexible tsv export/import (better structure)

ahaessly:
This approach uses a separate endpoint for import so we know whether the tsvs are supposed to be in the FC model for flexible model. (We will eventually need to make some changes to the bd bag implementation if/when we want to support the flexible model - but this could be some data in the bd bag itself instead of a separate endpoint).

Take a look at the ModelSchema file first. This implementation is more OO than functional using polymorphic subclasses to do the appropriate validation.

For export, we determine if we should be using the FC model by seeing if the entity type exporting is the same name as in our FC model. This feels fragile. Initially I thought that for export, it wouldn't matter - that we could use the flexible model for export because we are not validating the data. The one area where this broke down was in determining if the export file was of type 'entity' or 'update'. If the user have asked for an export of a subset of columns that do not contain all the required columns, it creates the export file as an update file (this behavior is the same in the other PR as well though - only realized it doing this implementation).

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
@@ -126,10 +126,7 @@ class EntityClient (requestContext: RequestContext)(implicit protected val execu
    * Returns the plural form of the entity type.
    * Bails with a 400 Bad Request if the entity type is unknown to the schema. */
   private def withPlural(entityType: String)(op: (String => Future[PerRequestMessage])): Future[PerRequestMessage] = {
-    ModelSchema.getPlural(entityType) match {
-      case Failure(regret) => Future(RequestCompleteWithErrorReport(BadRequest, regret.getMessage))
-      case Success(plural) => op(plural)
-    }
+    op(FlexibleModelSchema.pluralizeEntityType(entityType))
   }
 
 
@@ -138,15 +135,12 @@ class EntityClient (requestContext: RequestContext)(implicit protected val execu
    * Bails with a 400 Bad Request if the entity type is unknown or if some attributes are missing.
    * Returns the list of required attributes if all is well. */
   private def withRequiredAttributes(entityType: String, headers: Seq[String])(op: (Map[String, String] => Future[PerRequestMessage])):Future[PerRequestMessage] = {
-    ModelSchema.getRequiredAttributes(entityType) match {
-      case Failure(regret) => Future(RequestCompleteWithErrorReport(BadRequest, regret.getMessage))
-      case Success(requiredAttributes) =>
-        if( !requiredAttributes.keySet.subsetOf(headers.toSet) ) {
-          Future( RequestCompleteWithErrorReport(BadRequest,
-            "TSV is missing required attributes: " + (requiredAttributes.keySet -- headers).mkString(", ")) )
-        } else {
-          op(requiredAttributes)
-        }
+    val requiredAttributes = FlexibleModelSchema.getRequiredAttributes(entityType)
+    if( !requiredAttributes.keySet.subsetOf(headers.toSet) ) {
+      Future( RequestCompleteWithErrorReport(BadRequest,
+        "TSV is missing required attributes: " + (requiredAttributes.keySet -- headers).mkString(", ")) )
+    } else {
+      op(requiredAttributes)
     }
   }
 
@@ -225,15 +219,9 @@ class EntityClient (requestContext: RequestContext)(implicit protected val execu
     checkFirstColumnDistinct(tsv) {
       withMemberCollectionType(entityType) { memberTypeOpt =>
         checkNoCollectionMemberAttribute(tsv, memberTypeOpt) {
-          ModelSchema.getRequiredAttributes(entityType) match {
-            //Required attributes aren't required to be headers in update TSVs - they should already have been
-            //defined when the entity was created. But we still need the type information if the headers do exist.
-            case Failure(regret) => Future(RequestCompleteWithErrorReport(BadRequest, regret.getMessage))
-            case Success(requiredAttributes) =>
-              val colInfo = colNamesToAttributeNames(tsv.headers, requiredAttributes)
-              val rawlsCalls = tsv.tsvData.map(row => setAttributesOnEntity(entityType, memberTypeOpt, row, colInfo))
-              batchCallToRawls(pipeline, workspaceNamespace, workspaceName, entityType, rawlsCalls, "batchUpdate")
-          }
+          val colInfo = colNamesToAttributeNames(tsv.headers, FlexibleModelSchema.getRequiredAttributes(entityType))
+          val rawlsCalls = tsv.tsvData.map(row => setAttributesOnEntity(entityType, memberTypeOpt, row, colInfo))
+          batchCallToRawls(pipeline, workspaceNamespace, workspaceName, entityType, rawlsCalls, "batchUpdate")
         }
       }
     }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
@@ -48,7 +48,7 @@ object EntityClient {
 
   def backwardsCompatStripIdSuffixes(tsvLoadFile: TSVLoadFile, entityType: String, modelSchema: ModelSchema): TSVLoadFile = {
     modelSchema.getTypeSchema(entityType) match {
-      case Failure(exception) => tsvLoadFile // the failure will be handled during parsing
+      case Failure(regrets) => tsvLoadFile // the failure will be handled during parsing
       case Success(metaData) =>
         val newHeaders = tsvLoadFile.headers.map { header =>
           val headerSansId = header.stripSuffix("_id")
@@ -122,6 +122,7 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema)(im
       importBagit(pipeline, workspaceNamespace, workspaceName, bagitRq) pipeTo sender
   }
 
+
   /**
    * Returns the plural form of the entity type.
    * Bails with a 400 Bad Request if the entity type is unknown to the schema and we are using firecloud model
@@ -184,7 +185,7 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema)(im
     withMemberCollectionType(entityType, modelSchema) { memberTypeOpt =>
       validateMembershipTSV(tsv, memberTypeOpt) {
         withPlural(memberTypeOpt.get) { memberPlural =>
-          val rawlsCalls = tsv.tsvData groupBy(_ (0)) map { case (entityName, rows) =>
+          val rawlsCalls = tsv.tsvData groupBy(_(0)) map { case (entityName, rows) =>
             val ops = rows map { row =>
               //row(1) is the entity to add as a member of the entity in row.head
               val attrRef = AttributeEntityReference(memberTypeOpt.get,row(1))
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
@@ -48,6 +48,7 @@ object EntityClient {
 
   def backwardsCompatStripIdSuffixes(tsvLoadFile: TSVLoadFile, entityType: String, modelSchema: ModelSchema): TSVLoadFile = {
     modelSchema.getTypeSchema(entityType) match {
+      case Failure(exception) => tsvLoadFile // the failure will be handled during parsing
       case Success(metaData) =>
         val newHeaders = tsvLoadFile.headers.map { header =>
           val headerSansId = header.stripSuffix("_id")
@@ -58,7 +59,6 @@ object EntityClient {
           }
         }
         tsvLoadFile.copy(headers = newHeaders)
-      case Failure(exception) => tsvLoadFile // the failure will be handled during parsing
     }
   }
 
@@ -148,8 +148,8 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema)(im
         } else {
           op(requiredAttributes)
         }
-      }
     }
+  }
 
   def batchCallToRawls(
     pipeline: WithTransformerConcatenation[HttpRequest, Future[HttpResponse]],
@@ -184,13 +184,13 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema)(im
     withMemberCollectionType(entityType, modelSchema) { memberTypeOpt =>
       validateMembershipTSV(tsv, memberTypeOpt) {
         withPlural(memberTypeOpt.get) { memberPlural =>
-          val rawlsCalls = tsv.tsvData groupBy (_ (0)) map { case (entityName, rows) =>
+          val rawlsCalls = tsv.tsvData groupBy(_ (0)) map { case (entityName, rows) =>
             val ops = rows map { row =>
               //row(1) is the entity to add as a member of the entity in row.head
-              val attrRef = AttributeEntityReference(memberTypeOpt.get, row(1))
-              Map(addListMemberOperation, "attributeListName" -> AttributeString(memberPlural), "newMember" -> attrRef)
+              val attrRef = AttributeEntityReference(memberTypeOpt.get,row(1))
+              Map(addListMemberOperation,"attributeListName"->AttributeString(memberPlural),"newMember"->attrRef)
             }
-            EntityUpdateDefinition(entityName, entityType, ops)
+            EntityUpdateDefinition(entityName,entityType,ops)
           }
           batchCallToRawls(pipeline, workspaceNamespace, workspaceName, entityType, rawlsCalls.toSeq, "batchUpsert")
         }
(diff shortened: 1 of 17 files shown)


breilly2:
Focusing mostly on the high-level structure, I think this works nicely. Once I understand `ModelSchema`, I can easily see when each specific schema would be applied and not worry about their details when looking at how they're used. If we choose to keep this in favor of the other, I will review more closely.

@ahaessly, for the problem you mentioned with export, could that also be a separate endpoint that explicitly chooses the flexible schema rather than inferring it?

Also, which code do you prefer?

ahaessly:
yes, i think i will use a separate endpoint for export as well. also, i think i do prefer this code.
making a few changes and will submit for more in depth review

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelSchema.scala
```diff
@@ -14,35 +13,114 @@ import org.broadinstitute.dsde.firecloud.FireCloudConfig
...
+      case Failure(_) => false
+      case _ => true
+    }
+  }
+}
+
+
+
+object SchemaTypes {
+  trait SchemaType
```
ansingh7115: nit: should be `sealed trait`

ahaessly: i wanted this trait to be extendable...
although i'm not sure that just leaving off the sealed really allows it to be extended.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -39,16 +40,17 @@ object EntityClient {
...
 
-  def backwardsCompatStripIdSuffixes(tsvLoadFile: TSVLoadFile, entityType: String): TSVLoadFile = {
-    ModelSchema.getTypeSchema(entityType) match {
-      case Failure(regrets) => tsvLoadFile
-      case Success(metaData) =>
+  def backwardsCompatStripIdSuffixes(tsvLoadFile: TSVLoadFile, entityType: String, modelSchema: ModelSchema): TSVLoadFile = {
+    modelSchema.isInstanceOf[FlexibleModelSchema] match {
+      case true => tsvLoadFile
+      case false =>
+        val metaData = modelSchema.getTypeSchema(entityType)
```
ansingh7115: instead of match true/false, you could do this as just an if/else:

```
if (modelSchema.isInstanceOf[FlexibleModelSchema])
     tsvLoadFile
else {
     val metaData = modelSchema.getTypeSchema(entityType)
     val newHeaders = tsvLoadFile.headers.map { header =>
       val headerSansId = header.stripSuffix("_id")
       if (metaData.requiredAttributes.keySet.contains(headerSansId) || metaData.memberType.contains(headerSansId)) {
         headerSansId
       } else {
         header
       }
     }
     tsvLoadFile.copy(headers = newHeaders)
   }
```

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
```diff
@@ -118,8 +122,11 @@ object TSVFormatter {
     * @param requestedHeaders Which, if any, columns were requested. If none, return allHeaders (subject to sanitization)
     * @return Entity name as first column header, followed by matching entity attribute labels
     */
-  def makeEntityHeaders(entityType: String, allHeaders: Seq[String], requestedHeaders: Option[IndexedSeq[String]]): IndexedSeq[String] = {
-    val memberPlural = pluralizeMemberType(memberTypeFromEntityType(entityType))
+  def makeEntityHeaders(entityType: String, allHeaders: Seq[String], requestedHeaders: Option[IndexedSeq[String]])(implicit modelSchema: ModelSchema): IndexedSeq[String] = {
+    val memberPlural = modelSchema.isCollectionType(entityType) match {
+      case true => Some(modelSchema.getPlural(modelSchema.memberTypeFromEntityType(entityType).getOrElse("")))
+      case false => None
+    }
```
ansingh7115: This match true/false can also be just an if/else:

```
    val memberPlural =
      if (modelSchema.isCollectionType(entityType))
        Some(modelSchema.getPlural(modelSchema.memberTypeFromEntityType(entityType).getOrElse("")))
      else None
```

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelSchema.scala
```diff
@@ -14,35 +13,114 @@ import org.broadinstitute.dsde.firecloud.FireCloudConfig
...
+  }
+}
+
+
+class FlexibleModelSchema extends ModelSchema {
+
+  def memberTypeFromEntityType(entityType: String): Option[String] = {
+    isCollectionType(entityType) match {
+      case true => Some(entityType.replace("_set", ""))
+      case _ => None
```
ansingh7115: This match true/false can  be an if/else 

```
    if (isCollectionType(entityType))
      Some(entityType.replace("_set", ""))
    else None
```

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelSchema.scala
```diff
@@ -14,35 +13,114 @@ import org.broadinstitute.dsde.firecloud.FireCloudConfig
...
+    entityType + "s"
   }
 
-  def getRequiredAttributes(entityType: String): Try[Map[String, String]] = {
-    getTypeSchema(entityType).map(_.requiredAttributes)
+  def getTypeSchema(entityType: String): EntityMetadata = {
+    isCollectionType(entityType) match {
+      case true => EntityMetadata(entityType+"s", Map.empty, Some(entityType+"_members"))
+      case _ => EntityMetadata(entityType+"s", Map.empty, None)
+    }
```
ansingh7115: This match true/false can be turned into an if/else like above

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelSchema.scala
```diff
@@ -14,35 +13,114 @@ import org.broadinstitute.dsde.firecloud.FireCloudConfig
...
-  val types : Map[String, EntityMetadata] = ModelJsonProtocol.impModelSchema.read(
-    Source.fromURL(getClass.getResource(FireCloudConfig.Rawls.model)).mkString.parseJson ).schema
+class FirecloudModelSchema extends ModelSchema {
+
+  object EntityTypes {
+    val types : Map[String, EntityMetadata] = ModelJsonProtocol.impModelSchema.read(
+      Source.fromURL(getClass.getResource(FireCloudConfig.Rawls.model)).mkString.parseJson ).schema
+  }
+
+  def getTypeSchema(entityType: String): EntityMetadata = {
```
breilly2: This used to return a `Failure` instead of throwing an exception. Unless callers are wrapping this call in `try { ... } catch` or `Try( ... )`, I suspect there may be some behavior changes.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
breilly2: This looks like an operation on `ModelSchema`. If you push it down into `ModelSchema` then you don't need this conditional at all.

breilly2: To avoid possibly throwing an exception in this case (the FC schema case), you can put this is a `Try`:
```
        Try(modelSchema.getTypeSchema(entityType)) match {
          case Failure(_) => tsvLoadFile
          case Success(metaData) =>
            val newHeaders = tsvLoadFile.headers.map { header =>
              val headerSansId = header.stripSuffix("_id")
              if (metaData.requiredAttributes.keySet.contains(headerSansId) || metaData.memberType.contains(headerSansId)) {
                headerSansId
              } else {
                header
              }
            }
            tsvLoadFile.copy(headers = newHeaders)
        }
```
This should maintain the behavior of the function for this case.

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelSchema.scala
```diff
@@ -14,35 +13,114 @@ import org.broadinstitute.dsde.firecloud.FireCloudConfig
...
+
+  def memberTypeFromEntityType(entityType: String): Option[String] = {
+    getTypeSchema(entityType).memberType
+  }
+
+  def getRequiredAttributes(entityType: String): Map[String, String] = {
+    getTypeSchema(entityType).requiredAttributes
+  }
+
+  def isCollectionType(entityType: String): Boolean = {
```
breilly2, line 109: This is more permissive than it used to be. For the FC schema, this used to return `Failure` if `entityType` was not defined in the schema. I think `endsWith("_set")` logic belongs only in the flexible model.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -138,15 +141,12 @@ class EntityClient (requestContext: RequestContext)(implicit protected val execu
...
-    ModelSchema.getRequiredAttributes(entityType) match {
-      case Failure(regret) => Future(RequestCompleteWithErrorReport(BadRequest, regret.getMessage))
-      case Success(requiredAttributes) =>
-        if( !requiredAttributes.keySet.subsetOf(headers.toSet) ) {
-          Future( RequestCompleteWithErrorReport(BadRequest,
-            "TSV is missing required attributes: " + (requiredAttributes.keySet -- headers).mkString(", ")) )
-        } else {
-          op(requiredAttributes)
-        }
+    val requiredAttributes = modelSchema.getRequiredAttributes(entityType)
```
breilly2: Since `getRequiredAttributes` now throws an exception (from `getTypeSchema`) instead of returning `Failure`, I think this will result in 500 (internal server error) instead of 400 (bad request) when the entity type doesn't exist in the FC schema.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -121,12 +124,12 @@ class EntityClient (requestContext: RequestContext)(implicit protected val execu
...
 
-
   /**
    * Returns the plural form of the entity type.
-   * Bails with a 400 Bad Request if the entity type is unknown to the schema. */
+   * Bails with a 400 Bad Request if the entity type is unknown to the schema and we are using firecloud model
+   * If using flexible model, just appends an 's' */
   private def withPlural(entityType: String)(op: (String => Future[PerRequestMessage])): Future[PerRequestMessage] = {
-    ModelSchema.getPlural(entityType) match {
+    Try(modelSchema.getPlural(entityType)) match {
```
breilly2: This is the type of translation I would expect to see more of as a result of changing the result type of the `ModelSchema` functions to remove the `Try`.

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelSchema.scala
```diff
@@ -14,35 +13,114 @@ import org.broadinstitute.dsde.firecloud.FireCloudConfig
  * Time: 14:18
  */
 
-object ModelSchema {
+trait ModelSchema {
```
breilly2, line 18: The more I think about it, the more I feel like these functions should continue returning `Try`s rather than throwing exceptions. If we expect the possibility that an implementation might need to throw an exception (and we already have one case of that), we should build that into the API. I would guess that it would also minimize the need to change client code.

ahaessly:
thanks for these comments. i agree and will make the modifications.

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelSchema.scala
```diff
@@ -14,35 +13,106 @@ import org.broadinstitute.dsde.firecloud.FireCloudConfig
...
-    getTypeSchema(entityType).map(_.requiredAttributes)
+    Success(Map.empty)
   }
+
+  def getPlural(entityType: String): Try[String] =  {
+    Success(entityType + "s")
+  }
+
+  def getTypeSchema(entityType: String): Try[EntityMetadata] = {
+    Success(EntityMetadata(entityType, Map.empty, Some(entityType+"_members").filter(_ => isCollectionType(entityType))))
```
breilly2: Not sure if it matters right now but the first property of `EntityMetadata` should be the pluralized entity type. I suggest extracting a function from the current `getPlural`:
```
  private def pluralize(entityType: String) = {
    entityType + "s"
  }
```

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelSchema.scala
```diff
@@ -14,35 +13,106 @@ import org.broadinstitute.dsde.firecloud.FireCloudConfig
...
+  def getModelForEntityType(entityType: String): ModelSchema = {
+    val possibleMatch = schemas.values.collectFirst({case schema if schema.isEntityTypeInSchema(entityType) => schema})
+    possibleMatch.getOrElse(schemas.last._2)
   }
+}
 
-  def getPlural(entityType: String): Try[String] = {
-    getTypeSchema(entityType).map(_.plural)
+
+class FlexibleModelSchema extends ModelSchema {
```
breilly2: This and `FirecloudModelSchema` don't have any properties so they can be objects instead of classes (I'm pretty sure... it at least compiled after also removing all of the `new`s).

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -63,13 +64,19 @@ class ExportEntitiesByTypeActor(rawlsDAO: RawlsDAO,
...
                                 argMaterializer: ActorMaterializer)
                                (implicit protected val executionContext: ExecutionContext) extends Actor with LazyLogging {
 
   implicit val timeout: Timeout = Timeout(1 minute)
   implicit val userInfo: UserInfo = argUserInfo
   implicit val materializer: ActorMaterializer = argMaterializer
 
+  implicit val modelSchema: ModelSchema = model match {
+    case Some(name) => ModelSchemaRegistry.getModelForSchemaType(SchemaTypes.withName(name))
+    case None => ModelSchemaRegistry.getModelForEntityType(entityType)
```
breilly2: This case allows for the flexible model to be chosen if schema is not specified. I think the schema type should be known by this point in the code and should be determined right at the point where we get the optional parameter from the request. That way, if the client doesn't specify a schema, it behaves exactly the same regardless of the entity type.

src/test/scala/org/broadinstitute/dsde/firecloud/mock/MockTSV.scala
```diff
@@ -107,8 +107,8 @@ object MockTSVStrings {
     List("sset_01", "sample_02").tabbed).newlineSeparated
 
   //entity TSVs
-  val entityUnknownFirstColumnHeader = List(
-    List("entity:sampel_id", "bar", "baz").tabbed,
+  val entityNonModelFirstColumnHeader = List(
+    List("entity:bigQueyr_id", "bar", "baz").tabbed,
```
breilly2, line 111: Typo here. Not sure what this means about the validity of the test.

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelSchema.scala
```diff
@@ -14,35 +13,110 @@ import org.broadinstitute.dsde.firecloud.FireCloudConfig
...
+
+  def memberTypeFromEntityType(entityType: String): Try[Option[String]] = {
+    getTypeSchema(entityType).map(_.memberType)
+  }
+
+  def getRequiredAttributes(entityType: String): Try[Map[String, String]] = {
+    getTypeSchema(entityType).map(_.requiredAttributes)
+  }
+
+  def isCollectionType(entityType: String): Boolean = {
```
breilly2, line 109: I'm still uncertain about this not returning a `Try`. This used to use `getTypeSchema` so  `Failure` was a possible result. I can't convince myself that the places that use `isCollectionType` could possibly return an "Unknown entity type" error like they used to. Is there evidence that this is a safe change?

src/test/scala/org/broadinstitute/dsde/firecloud/mock/MockTSV.scala
ahaessly: it's a typo on purpose. it fails with the firecloud model when calling import but will succeed with the flexible model because it doesn't care what the type is

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
ahaessly: for backwards compatibility, no model specified should be the firecloud model. this is so that existing code does not have to start passing a model param to indicate the "firecloud".

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelSchema.scala
ahaessly: before my changes it was used in 2 places in the code and both did a .getOrElse(false)

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
breilly2: Right. So if model isn't specified but the entity type doesn't match then this case will select the flexible model. In prod today, the firecloud model would still be used in that case. Although that usage is unsupported, it does look to me like a change in behavior.

ahaessly: ah. got it.

breilly2 (approved):
👍 Looks good!

ansingh7115 (approved):

PR #683, opened 2018-11-02
GAWB-3983: increase max aggregates size and move to config.

ahaessly:
get max aggregations from config and default to 10000 if missing or invalid

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
@@ -154,7 +154,7 @@ object FireCloudConfig {
     val trialIndexName = elasticsearch.getString("trialIndex")
     val discoverGroupNames = elasticsearch.getStringList("discoverGroupNames")
     val shareLogIndexName: String = elasticsearch.getString("shareLogIndex")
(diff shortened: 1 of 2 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
```diff
@@ -152,6 +154,7 @@ object FireCloudConfig {
     val trialIndexName = elasticsearch.getString("trialIndex")
     val discoverGroupNames = elasticsearch.getStringList("discoverGroupNames")
     val shareLogIndexName: String = elasticsearch.getString("shareLogIndex")
+    val maxAggregations: Int = Try(elasticsearch.getInt("maxAggregations")).getOrElse(10000)
```
davidangb: your call on preference - instead of a `Try ... getOrElse` you could specify `maxAutocomplete = 1000` in reference.conf and leave this at just `elasticsearch.getInt("maxAggregations")`

davidangb (approved):
:+1:

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
ahaessly: someone could still set it to an invalid value (i.e. not an int) in the generated conf. unlikely, but still...

ahaessly:
jenkins retest

PR #686, opened 2018-11-12
better error message on too-large chunked response [risk: low]

davidangb:
DataBiosphere/firecloud-app#83

if orch receives a chunked response from a subsystem, and that response is over the `response-chunk-aggregation-limit` size limit, we previously returned an error message that was confusing to end users: "sendReceive doesn't support chunked responses, try sendTo instead"

This PR attempts to make that error message more understandable.

-----

In all cases:


rtitle (approved):
Not a big fan of parsing error message strings as it's unreliable (and will break if/when we upgrade to akka-http). But I don't really have a better alternative, and I agree the new error message is much better. 👍

davidangb:
Thanks - I agree it's unreliable in the face of upgrades, but the "failure" case in this code is to fall back to the current behavior of `case Failure(error) =>` ... we'd simply stop rewording this particular error.

PR #688, opened 2018-11-16
set email only on user registration, not profile update

dvoet:
when users create a google account with an external email they have the ability to later change that email. If they do that then update their profile the new email address is saved in thurloe but this is then out of sync with the rest of the system. This PR stops the update, email is only set when user first registers.

In all cases:


davidangb (approved):
:+1: changes here look fine. Is there a longer-term story to update the rest of the system so that everything is in sync within Workbench AND in sync with the user's current email address?

dvoet:
@davidangb no, the thought for the moment is that it is too scary to allow a user's email address to change.

PR #689, opened 2018-11-20
add single whitelist sync endpoint

dvoet:
in prep for 60 some whitelists from AnVIL changing the sync process to be an api call per whiltelist

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
@@ -48,10 +48,10 @@ object NihStatus {
 
 object NihService {
   sealed trait ServiceMessage
-  case class GetNihStatus(userInfo: UserInfo) extends ServiceMessage
-  case class UpdateNihLinkAndSyncSelf(userInfo: UserInfo, nihLink: NihLink) extends ServiceMessage
+  final case class GetNihStatus(userInfo: UserInfo) extends ServiceMessage
+  final case class UpdateNihLinkAndSyncSelf(userInfo: UserInfo, nihLink: NihLink) extends ServiceMessage
   case object SyncAllWhitelists extends ServiceMessage
-  case class SyncWhitelist(whitelistName: String) extends ServiceMessage
+  final case class SyncWhitelist(whitelistName: String) extends ServiceMessage
 
   def props(service: () => NihServiceActor): Props = {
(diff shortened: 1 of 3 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
```diff
@@ -103,6 +105,17 @@ trait NihService extends LazyLogging {
     usersList.getLines().toSet
   }
 
+  def syncWhitelistAllUsers(whitelistName: String): Future[PerRequestMessage] = {
+    println(s"foo $nihWhitelists")
```
Qi77Qi: println

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
```diff
@@ -103,6 +105,17 @@ trait NihService extends LazyLogging {
...
+    nihWhitelists.find(_.name.equals(whitelistName)) match {
```
Qi77Qi, line 109: does `if (nihWhitelists.exists(_.name == whitelistName)) ... else ...` suffice? (feels the extra `whitelist` variable is a bit unnecessary)

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
```diff
@@ -50,7 +50,8 @@ object NihService {
   sealed trait ServiceMessage
   case class GetNihStatus(userInfo: UserInfo) extends ServiceMessage
   case class UpdateNihLinkAndSyncSelf(userInfo: UserInfo, nihLink: NihLink) extends ServiceMessage
-  case object SyncWhitelist extends ServiceMessage
+  case object SyncAllWhitelists extends ServiceMessage
+  case class SyncWhitelist(whitelistName: String) extends ServiceMessage
```
Qi77Qi: nice to add `final`

Qi77Qi (changes requested):

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
dvoet: nihWhitelists is a set of objects, not just names. We need to find the right object in the set.

Qi77Qi (approved):

PR #692, opened 2018-12-03
GAWB-4014: TSV download filename should be .tsv

breilly2:
Changes TSV filename to have `.tsv` extension instead of `.txt`. Also changes content-type to `text/tab-separated-values` instead of `text/plain`.

This is a bug fix. It could also be interpreted as a breaking API change. Let me know if you object to this change as-is and if you have any suggestions for making it backward-compatible. One thought I had was to only apply this change if the request has `text/tab-separated-values` in its `Accept` header.

In all cases:


davidangb (approved):
:+1: looks good to me, assuming tests pass.

I also consider this a bugfix, not a breaking change, so in my opinion we don't need to jump through hoops to preserve compatibility. However, I do suggest working with comms to ensure that this change is specifically highlighted in release notes, just in case it breaks anyone's scripts.

dvoet (approved):

PR #699, opened 2019-01-02
rename consentCodes library title to Data Use Limitations [risk: no]

davidangb:
DataBiosphere/firecloud-app#237

renames a Library column from "Consent Codes" to "Data Use Limitations".

This value appears to only be used when displaying the column header in FC UI's Library table. Since it is display-only, I don't feel that the `attribute-definitions.json` file needs a version bump, nor do we need to change anything about the browser's local storage of Library display preferences.

-----

In all cases:


breilly2 (approved):
👍 My only question would be if any UI tests need updating, but those all passed.

PR #700, opened 2019-01-02
gawb-3840: data library tab will show workspaces user already has read access to

ahaessly:
This PR is for both gawb-3840 and gawb-3842

it calls sam to get the workspaces the user has access to and adds this as filter criteria for matching published data sets. it also changes the way we determine whether the user has access to the underlying workspace. 

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
@@ -5,12 +5,14 @@ import org.broadinstitute.dsde.firecloud.FireCloudConfig
 import org.broadinstitute.dsde.firecloud.model._
 import org.broadinstitute.dsde.firecloud.model.ElasticSearch._
 import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
+import org.broadinstitute.dsde.firecloud.model.SamResource.AccessPolicyName
 import org.broadinstitute.dsde.rawls.model.AttributeName
 import org.elasticsearch.search.aggregations.{AggregationBuilders, Aggregations}
 import org.elasticsearch.client.transport.TransportClient
 import org.elasticsearch.action.search.{SearchRequest, SearchRequestBuilder, SearchResponse}
 import org.elasticsearch.index.query.{BoolQueryBuilder, QueryBuilder}
 import org.elasticsearch.index.query.QueryBuilders._
+import org.elasticsearch.search.SearchHit
 import org.elasticsearch.search.aggregations.bucket.terms.{StringTerms, Terms}
 import org.elasticsearch.search.fetch.subphase.highlight.HighlightBuilder
 import org.elasticsearch.search.sort.SortOrder
@@ -38,6 +40,14 @@ trait ElasticSearchDAOQuerySupport extends ElasticSearchDAOSupport {
     * {"query":{
     *    "bool":{
     *      "must":[
+    *        {"bool":
+    *          {"should":[
+    *            {"term":{"library:indication":"n/a"}},
+    *            {"term":{"library:indication":"disease"}},
+    *            {"term":{"library:indication":"lukemia"}}]
+    *          }
+    *        }],
+    *      "filter":[
     *        {"bool" :
     *          {"should" : [
     *            {"bool" :
@@ -46,15 +56,8 @@ trait ElasticSearchDAOQuerySupport extends ElasticSearchDAOSupport {
     *            {"terms" :
     *            {"_discoverableByGroups" : ["37e1e624-84e0-4046-98d6-f89388a8f727-OWNER", "fdae9abd-be33-4be9-9eb2-321df017b8a2-OWNER"]}}]
     *          }
-    *        },
-    *        {"bool":
-    *          {"should":[
-    *            {"term":{"library:indication":"n/a"}},
-    *            {"term":{"library:indication":"disease"}},
-    *            {"term":{"library:indication":"lukemia"}}]
-    *          }
-    *        },
-    *        {"match":{"all_":"broad"}}]}},
+    *        }]
+    *        }},
     *  "aggregations":{
     *    "library:dataUseRestriction":{
     *      "terms":{"field":"library:dataUseRestriction.raw"}},
@@ -68,7 +71,7 @@ trait ElasticSearchDAOQuerySupport extends ElasticSearchDAOSupport {
     */
 
 
-  def createQuery(criteria: LibrarySearchParams, groups: Seq[String], researchPurposeSupport: ResearchPurposeSupport, searchField: String = fieldAll, phrase: Boolean = false): QueryBuilder = {
+  def createQuery(criteria: LibrarySearchParams, groups: Seq[String], workspaceIds: Seq[String], researchPurposeSupport: ResearchPurposeSupport, searchField: String = fieldAll, phrase: Boolean = false): QueryBuilder = {
     val query: BoolQueryBuilder = boolQuery // outer query, all subqueries should be added to the must list
     query.must(criteria.searchString match {
       case None => matchAllQuery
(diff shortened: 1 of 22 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryService.scala
```diff
@@ -225,26 +226,26 @@ class LibraryService (protected val argUserInfo: UserInfo,
...
+    val workspacePolicies: Future[Seq[UserPolicy]] = samDao.listWorkspaceResources
+    workspacePolicies flatMap { policyList =>
+      val workspacePolicyMap = (policyList map { policy =>
+        (policy.resourceId.value, policy)
+      }).toMap
+      getEffectiveDiscoverGroups(samDao) map { userGroups =>
+        // we want docsFuture and ids to be parallelized - so declare them here, outside
+        // of the for-yield.
+        searchMethod(criteria, userGroups, workspacePolicyMap)
+      } map (RequestComplete(_))
```
davidangb: could you parallelize the two calls to Sam? `listWorkspaceResources`  and `getEffectiveDiscoverGroups` are independent of each other, but in this code are serialized.  Maybe something like:
```
  def searchFor(criteria: LibrarySearchParams, searchMethod:(LibrarySearchParams, Seq[String], Map[String, UserPolicy])=>Future[LibrarySearchResponse]): Future[PerRequestMessage] ={
    val workspacePoliciesFuture: Future[Map[String, UserPolicy]] = samDao.listWorkspaceResources map { policyList =>
      (policyList map { policy =>
        (policy.resourceId.value, policy)
      }).toMap
    }
    val userGroupsFuture: Future[Seq[String]] = getEffectiveDiscoverGroups(samDao)

    for {
      workspacePolicyMap <- workspacePoliciesFuture
      userGroups <- userGroupsFuture
      searchResults <- searchMethod(criteria, userGroups, workspacePolicyMap)
    } yield {
      RequestComplete(searchResults)
    }
  }
```

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
```diff
@@ -170,26 +171,7 @@ trait LibraryServiceSupport extends DataUseRestrictionSupport with LazyLogging {
...
-          case _ => document.asJsObject.fields + ("workspaceAccess" -> JsString(WorkspaceAccessLevels.NoAccess.toString))
-        }
-        case _ => document.asJsObject.fields
-      }
-      JsObject(newJson)
-    }
-
-    docs.copy(results = updatedResults)
-  }
+  val accessible = Seq(AccessPolicyName("reader"), AccessPolicyName("writer"), AccessPolicyName("owner"))
```
davidangb: is this used anywhere? If so - and depending on its usage pattern - might want to move it up to an object level so we're not constantly creating a new Seq each time it's used.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
```diff
@@ -100,6 +99,15 @@ trait ElasticSearchDAOQuerySupport extends ElasticSearchDAOSupport {
       rp => query.must(researchPurposeSupport.researchPurposeFilters(rp, toLibraryAttributeName))
     }
 
+    val groupsQuery = boolQuery
+    // https://www.elastic.co/guide/en/elasticsearch/reference/2.4/query-dsl-exists-query.html
+    groupsQuery.should(boolQuery.mustNot(existsQuery(fieldDiscoverableByGroups)))
+    if (groups.nonEmpty) {
+      groupsQuery.should(termsQuery(fieldDiscoverableByGroups, groups.asJavaCollection))
+    }
+    groupsQuery.should(termsQuery(fieldWorkspaceId+".keyword" , workspaceIds.asJavaCollection))
```
davidangb: I'm trying to think of a case in which `workspaceIds` will be empty - and I can't come up with it - but might want to protect against that case.

On  the flip side - have you tested this with arbitrarily large numbers of workspaceIds? Is there *any* limit on size? What if the user has tens of thousands of workspaces?

src/test/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSpec.scala
```diff
@@ -794,32 +794,32 @@ class LibraryServiceSpec extends BaseServiceSpec with FreeSpecLike with LibraryS
...
+//      "set workspaceAccess to workspace access level if workspace is returned from workspace list" in {
+//        val workspaceList = Seq(WorkspaceListResponse(WorkspaceAccessLevels.Owner, testWorkspace, WorkspaceSubmissionStats(None, None, 0), false))
+//        val result = testLibraryMetadataJsObject.copy(testLibraryMetadataJsObject.fields.updated("workspaceId", JsString(testUUID.toString)))
+//        val expectedResult: JsValue = result.copy(result.fields.updated("workspaceAccess", JsString(WorkspaceAccessLevels.Owner.toString)))
+//        val doc = LibrarySearchResponse(params, 1, Seq(result: JsValue), Seq())
+//        assertResult(doc.copy(results=Seq(expectedResult))) {
+//          updateAccess(doc, workspaceList)
+//        }
+//      }
+//    }
```
davidangb: These are being  fixed/updated/replaced in a separate PR, right?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpSamDAO.scala
```diff
@@ -25,6 +26,10 @@ import scala.concurrent.{ExecutionContext, Future}
 class HttpSamDAO( implicit val system: ActorSystem, implicit val executionContext: ExecutionContext )
   extends SamDAO with RestJsonClient {
 
+  override def listWorkspaceResources(implicit userInfo: WithAccessToken): Future[Seq[UserPolicy]] = {
+    authedRequestToObject[Seq[UserPolicy]](Get(samListResources("workspace")))
+  }
```
davidangb, line 31: fyi - and no action necessary for this PR - add a `label` if you want to see performance logging on this. See `registerUser` just  below as an example.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
davidangb: also - for future thought, *definitely* not for this PR ... I think the most likely case is that any given user will have access to multiple workspaces that aren't published into Library. Currently, the corpus of published workspaces changes relatively infrequently (on publish, on unpublish,  on delete, and potentially on reindex). As a brainstorm,  we could maintain an in-memory "whitelist"  of all the published workspace ids, which updates on any of the change  events, and  then prune the user's workspaceId list down to just those that actually exist in Library. This would keep the filter - and the request payload - smaller. Probably only an  issue for some power users that have thousands of workspaces that aren't published into Library.

davidangb (commented):
a few smaller thoughts, one suggestion on query parallelization. Looking good!

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchDAOQuerySupport.scala
ahaessly: i can't remember now. some of the documentation says there is no explicit limit, but performance will degrade. months ago when i first tried out this change in implementation, i did test by adding thousands of ids to the list but i can't remember exactly how many :(

davidangb (commented):
latest code changes look good. Last thing for me is just ensuring this is tested with large numbers of  workspaceIds in the filter. Maybe an integration test?

gpolumbo-broad (approved):

dvoet (approved):
I don't completely grok what is going on but I don't see any red flags

src/test/scala/org/broadinstitute/dsde/firecloud/model/SamResourceSpec.scala
```diff
@@ -0,0 +1,58 @@
...
+    |    "public": false,
+    |    "accessPolicyName": "owner",
+    |    "missingAuthDomainGroups": [],
+    |    "authDomainGroups": []
+    |  }]
+  """.stripMargin
+
+  "UserPolicy JSON" - {
+
+    "should should convert to UserPolicy object" in {
```
davidangb: nit: "should should"

davidangb (approved):
yay new test! one nit inline. 👍 assuming tests pass.

ahaessly:
jenkins retest

PR #706, opened 2019-01-24
Resiliency/debugging in free credit spreadsheet update [risk: low]

davidangb:
As of Friday 18th, FireCloud started getting errors during its scheduled/automatic updates to the free credits tracking spreadsheet. A Sentry issue exists for this error: 556815659.

Unfortunately the pre-existing error logging in this area is insufficient to fully diagnose this problem. My hypothesis is that some user is missing one or more Thurloe keys that we normally require. From code inspection, I think we are vulnerable to this condition.

This PR:
* adds more logs and adds more error details to existing logs
* adds a `Try` around a potential trouble spot, with additional logging in case of error, and the ability to continue processing instead of fully breaking on error
* increases a hardcoded limit that I know we have already overflowed

This may not be the final fix, but I'd like to get this additional logging into place as soon as possible so we CAN get to the final fix.

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
@@ -282,7 +282,7 @@ class ElasticSearchTrialDAO(client: TransportClient, indexName: String, refreshM
       .prepareSearch(indexName)
       .setQuery(Ready)
       .addSort("name.keyword", SortOrder.ASC)
-      .setSize(1000)
+      .setSize(5000) // we should loop and make paginated requests; this will have to suffice for now
 
     val reportProjectResponse = executeESRequest[SearchRequest, SearchResponse, SearchRequestBuilder](reportProjectRequest)
 
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/model/Profile.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/model/Profile.scala
@@ -88,7 +88,7 @@ object Profile {
     }).toMap
 
     requiredKeys foreach {req =>
-      assert(mappedKVPs.contains(req), s"Profile must contain a key-value entry for $req")
+      assert(mappedKVPs.contains(req), s"Profile for user ${wrapper.userId} must contain a key-value entry for $req")
     }
 
     new Profile(
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/model/Profile.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/model/Profile.scala
@@ -79,23 +79,30 @@ object Profile {
   // increment this number every time you make a change to the user-provided profile fields
   val currentVersion:Int = 3
 
-  def apply(wrapper: ProfileWrapper) = {
+  val requiredKeys = List("firstName", "lastName", "title", "institute", "institutionalProgram", "programLocationCity",
+                          "programLocationState", "programLocationCountry", "pi", "nonProfitStatus")
 
-    val mappedKVPs:Map[String,String] = (wrapper.keyValuePairs map {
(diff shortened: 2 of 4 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
```diff
@@ -282,7 +282,7 @@ class ElasticSearchTrialDAO(client: TransportClient, indexName: String, refreshM
       .prepareSearch(indexName)
       .setQuery(Ready)
       .addSort("name.keyword", SortOrder.ASC)
-      .setSize(1000)
+      .setSize(5000) // we should loop and make paginated requests; this will have to suffice for now
```
davidangb, line 285: since the initial launch of free credits, we have grown to ~1300 projects, so we are past this hardcoded limit. While we should move to a paginated query, I want to get the rest of the logging/resilience in this PR out without being blocked on more complex changes.

src/main/scala/org/broadinstitute/dsde/firecloud/model/Profile.scala
```diff
@@ -79,30 +79,38 @@ object Profile {
...
-      programLocationCountry = mappedKVPs.get("programLocationCountry").get,
-      pi = mappedKVPs.get("pi").get,
-      nonProfitStatus = mappedKVPs.get("nonProfitStatus").get,
+      institute = mappedKVPs("institute"),
+      institutionalProgram = mappedKVPs("institutionalProgram"),
+      programLocationCity = mappedKVPs("programLocationCity"),
+      programLocationState = mappedKVPs("programLocationState"),
+      programLocationCountry = mappedKVPs("programLocationCountry"),
+      pi = mappedKVPs("pi"),
+      nonProfitStatus = mappedKVPs("nonProfitStatus"),
```
davidangb, line 105: there are no functional changes in this refactor;  I'm just moving any problems out from the ugly `.get("foo").get` into the earlier `assert` so I have  more control over the error message.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialServiceSupport.scala
```diff
@@ -59,13 +65,19 @@ trait TrialServiceSupport extends LazyLogging {
...
       // loop over all projects (including those that have no defined user) and build spreadsheet rows
       val rows: List[util.List[AnyRef]] = projects.map { trialProject =>
         val rowStrings = trialProject.user match {
-          case Some(user) => getTrialUserInformation(user, userKVPMap).toSpreadsheetValues
+          case Some(user) => Try(getTrialUserInformation(user, userKVPMap).toSpreadsheetValues) match {
+            case Success(s) => s
+            case Failure(x) =>
+              logger.error(s"Failure during getTrialUserInformation for user ${user.userSubjectId} ${user.userEmail} in project ${trialProject.name.value}: ${x.getMessage}", x)
+              errorValues(user.userSubjectId)
+          }
```
davidangb, line 80: this is the most important change in the PR. Previously, if we were looping over hundreds of users, a single user who was missing a single key in Thurloe would cause an error that prevented the entire spreadsheet from being updated. Now, with `Try` in place, if a user is missing a key we'll write "ERROR IN FIRECLOUD" into the spreadsheet for that user's row, but keep processing all other users.

davidangb:
2 failures in UI tests in https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/15528/, I am sure these are unrelated but will re-run:
```
[info] *** 2 TESTS FAILED ***
[error] Failed tests:
[error] 	org.broadinstitute.dsde.firecloud.test.metadata.DataSpec
[error] (Test / testOnly) sbt.TestsFailedException: Tests unsuccessful
```

EDIT: re-run passed in https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/15531/.

gpolumbo-broad (approved):

dvoet (approved):

PR #708, opened 2019-01-30
if free trial user has no profile, still send status fields to spreadsheet [risk: low]

davidangb:
DataBiosphere/firecloud-app#245

in the process that updates our free-credits tracking spreadsheet:
* we originally failed to update the spreadsheet entirely if any user was missing required `Profile`  fields
* as of #706, we successfully update the spreadsheet for fully-populated user, but any user missing fields resulted in an entire row of "ERROR IN FIRECLOUD"  cells.
* as of this PR - now that we understand the use case better - we correctly send the internal fields such as `trialStatus` even if the user is missing `Profile` fields. This means we can track all users through their trials;  some might just be missing firstname/lastname and other cosmetic fields.

The use case for this is Terra - Terra has turned on free trial via their UI,  but don't require a fully-populated `Profile`.

-----

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialServiceSupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialServiceSupport.scala
@@ -32,6 +32,8 @@ trait TrialServiceSupport extends LazyLogging {
     "Enrollment Date", "Terminated Date", "Expiration Date", "User Agreement", "User Agreement Date",
     "First Name", "Last Name", "Organization", "City", "State", "Country").map(_.asInstanceOf[AnyRef]).asJava
 
+  // default value to use if user has no entry for a field in their profile
+  private val defaultProfileField = ""
   // the default set of values to use if a project has no users; one fewer than # of columns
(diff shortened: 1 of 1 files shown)


elizzhao (approved):
couldn't test, but code makes sense to me!

breilly2 (approved):
👍 Interesting technique for setting defaults. I wouldn't have thought of that. I probably would have created a stub profile object with empty string values for all properties. (However I haven't studied this code enough to know whether or not that's possible.) Anyway, while I did not test it myself, I believe that this works and I checked that all of the fields line up positionally.

davidangb:
@breilly2 I wanted to go that path too, but there are `require`ments on the `Profile` object that look for empty strings: https://github.com/broadinstitute/firecloud-orchestration/blob/develop/src/main/scala/org/broadinstitute/dsde/firecloud/model/Profile.scala#L64 ... I didn't want to refactor that because it's used elsewhere in the code.

breilly2:
And that's exactly the kind of thing that I suspected might exist that led to this implementation. :)

PR #709, opened 2019-02-04
Pass through Job Manager requests to CromIAM

aednichols:
Requires https://github.com/broadinstitute/firecloud-develop/pull/1527

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -590,7 +590,7 @@ paths:
           schema:
             type: object
       tags:
-        - Workflows
+        - CromIAM Workflows
       responses:
         '200':
           description: Successful Request
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -530,6 +530,42 @@ paths:
           - email
           - profile
 
+  /api/workflows/{version}/{id}/abort:
+    post:
+      summary: Abort a workflow based on workflow id
+      parameters:
+        - name: version
+          description: API Version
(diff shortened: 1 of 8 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -530,6 +530,427 @@ paths:
           - email
           - profile
 
+  /api/workflows/{version}/{id}/abort:
```
breilly2, line 533: If we plan to remove these in the future, maybe we shouldn't publish them in the swagger interface, or we should mark them as deprecated to discourage anyone else from discovering and using them.

aednichols: I think having them in Swagger is valuable for our own documentation and trying out requests, but I agree we could use a "Job Manager only for now" notice. I'll see what I can do.

cjllanwarne: OOI, what makes you think we'd remove them in the future?

aednichols: I put a comment in the code saying if we ever go through the work to route Job Manager directly to CromIAM, we should clean up these routes

aednichols: I qualified the endpoints with a "(for Job Manager)" comment in Swagger; and after discussing with @cjllanwarne removed my code comment about them going away eventually because I learned we intend this to be the permanent arrangement

breilly2 (approved):
👍 I did not test but changes look fine to me.

rsasch (approved):
did not test, by LGTM!

PR #716, opened 2019-04-16
apis to set terra ui vs legacy view preference [risk: low]

davidangb:
three new APIs in orchestration that get/set the "Terra Preference" - a well-known KVP in Thurloe that Terra UI and FireCloud UI will look at in order to honor the user's choice of which UI they prefer.

each of the three endpoints returns a json a la:
```
{
  "preferTerra": false,
  "preferTerraLastUpdated": 1555519882001
}
```

"preferTerra" is true if the user should be redirected from FCUI to Terra. This is the default if the user has not explicitly set a preference.
"preferTerraLastUpdated" is the epoch timestamp of when the user last set their preference. This defaults to 0 if the user has not explicitly set a preference.

We are intentionally storing this preference in Thurloe, and tracking a timestamp, because Thurloe is connected to our analytics warehouse and thus we can report on these  preferences.

The new endpoints are:
* GET /api/profile/terra - returns the user's current preference.
* POST /api/profile/terra - explicitly sets the user's preference to `true` and returns the same payload as GET, after writing the preference
* DELETE /api/profile/terra - explicitly sets the user's preference to `false` and returns the same payload as GET, after writing the preference

After some discussion, I am choosing to leave the `POST` and  `DELETE` as separate endpoints so they are logged separately and we have easy access to analytics on those request logs.

If clients (UIs) would prefer to read this preference from the pre-existing /register/profile endpoint and consolidate ajax calls, they should:
* look for the `preferTerra` key
* be resilient to `preferTerra` not existing in profile
* default `preferTerra` to true if it is missing or otherwise has a value that cannot be converted to boolean
* expect `preferTerra` to contain a string, and parse that string into a boolean

-----

In all cases:

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -6483,9 +6483,12 @@ definitions:
   TerraPreference:
     description: Current user's choice of UI views.
     properties:
-      PreferTerra:
+      preferTerra:
         type: boolean
         description: When true, prefer Terra UI; when false, prefer Legacy FireCloud UI.
+      preferTerraLastUpdated:
+        type: number
+        description: Epoch timestamp representing when the Terra Preference was last saved.
 
   ToolClass:
     description: Describes a class (type) of tool allowing us to categorize workflows, tools, and maybe even other entities (such as services) separately
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -2100,6 +2100,44 @@ paths:
         500:
           description: Internal Server Error
 
+  /api/profile/terra:
+    get:
+      tags:
+        - Profile
+      operationId: setTerraPreference
+      summary: Returns the current user's preference for Terra UI vs. Legacy view
+      responses:
+        200:
+          description: Success
+          schema:
+            $ref: '#/definitions/TerraPreference'
+        500:
+          description: Internal Server Error
+    post:
+      tags:
+        - Profile
+      operationId: setTerraPreference
+      summary: Sets the current user's preference to use Terra UI, not Legacy view
+      responses:
+        200:
+          description: Success
+          schema:
+            $ref: '#/definitions/TerraPreference'
+        500:
+          description: Internal Server Error
+    delete:
+      tags:
+        - Profile
+      operationId: setTerraPreference
+      summary: Sets the current user's preference to use Legacy view, not Terra UI
+      responses:
+        200:
+          description: Success
+          schema:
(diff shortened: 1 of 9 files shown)


davidangb:
automation and unit tests passed; integration tests failed with what looks like a transient sbt/networking error. Let's try again: jenkins retest.

panentheos:
A couple comments:

Having a separate `DELETE` endpoint seems unusual. If the conceptual model is that we're managing a boolean, I'd expect a single `PUT` that takes e.g. `{'PreferTerra': true}` or `{'PreferTerra': false}`

The 'negative timestamp' is clever, but non-standard. Since the UI is already loading the whole profile on page load anyway, we'd prefer to just read the info from there. What about splitting the data into a boolean and a regular timestamp? That would remove some cognitive overhead on both the UI and analytics sides.

davidangb:
one more time: jenkins retest

davidangb:
@panentheos heard about the separate `DELETE` endpoint.

Regarding reading the preference from the main profile API ... the issue here is that many users will not have a KVP in place for this particular setting - only those users who have explicitly set a preference will have a value. The new APIs in this PR take care of managing all the defaults - they always return a value, even if its the default value in the case of the missing KVP. And they abstract away the timestamp stuff and just return a boolean.  If you (or any UI/client) reads from the main profile API instead of this new one, we'd either have to:

1. add the defaulting logic to the main profile api: distasteful, since it adds custom business logic just for one KVP
2. replicate the defaulting logic in all the UIs/clients

do you prefer one of those approaches over calling this new API?

breilly2:
This is already a special KVP that only 2 UIs/clients should care about, so that makes the special case nature of it a little more palatable to me. That said, this is temporary transition code that we should remove later so I don't have a strong preference one way or the other (with the disclaimer that I haven't looked at how making 2 API calls instead of 1 might affect the UI code flow).

helgridly (approved):
I think the code is fine. But doesn't the UI hit `/profile` as it is already, and thus retrieves this information anyway?

davidangb:
@helgridly yes, UIs currently hit /register/profile and thus can get access to the raw Thurloe KVP. I added the GET endpoint here for a couple reasons:
* it returns a nicer, concise view of the relevant data, without all the other KVPs
* it returns default values in the cases where users don't have a KVP - which I expect will be many of them
* architecturally, it feels right to have a GET that parallels the POST/DELETE

As I described above in the section "If clients (UIs) would prefer to read this preference from the pre-existing /register/profile" ... UIs must do extra work if they bypass the GET api, and that extra work must be re-implemented  in each UI.

PR #720, opened 2019-04-22
add retryable and before

jchanbroad:

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/RegistrationApiSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/RegistrationApiSpec.scala
@@ -9,17 +9,17 @@ import org.broadinstitute.dsde.workbench.service.{Sam, Thurloe}
 import org.scalatest.concurrent.{Eventually, ScalaFutures}
 import org.scalatest.tagobjects.Retryable
 import org.scalatest.time.{Seconds, Span}
-import org.scalatest.{BeforeAndAfter, BeforeAndAfterEach, FreeSpec, Matchers}
+import org.scalatest.{BeforeAndAfterAll, FreeSpec, Matchers}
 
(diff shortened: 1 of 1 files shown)


jchanbroad:
jenkins retest

automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/RegistrationApiSpec.scala
```diff
@@ -7,18 +7,19 @@ import org.broadinstitute.dsde.workbench.config.{Credentials, UserPool}
...
-  with BillingFixtures with BeforeAndAfterEach {
+  with BillingFixtures with BeforeAndAfter {
   implicit override val patienceConfig: PatienceConfig = PatienceConfig(timeout = scaled(Span(5, Seconds)))
 
   val subjectId: String = Users.tempSubjectId
   val adminUser: Credentials = UserPool.chooseAdmin
   implicit val authToken: AuthToken = adminUser.makeAuthToken()
 
-  override protected def beforeEach() = {
+  before{
```
marctalbott: `BeforeAndAfter` has been deprecated (see [scalatest documentation](http://doc.scalatest.org/1.0/org/scalatest/BeforeAndAfter.html)) so you should use either `BeforeAndAfterEach` or `BeforeAndAfterAll`. In this case, it doesn't really matter since there's only one test in this suite so `beforeAll` and `beforeEach` will both be run once before this test, but I think how you had it before should be fine.

marctalbott (changes requested):

automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/RegistrationApiSpec.scala
jchanbroad: I have changed the "before" to "beforeAll".

marctalbott (approved):

PR #726, opened 2019-04-30
NIH link status api should not require full user profile [risk: low]

davidangb:
The NIH link status API - `GET /api/nih/status` - requires a full user profile, including e.g. `firstName`, `institute` keys in Thurloe. This is because:
* the API internally relied on `ThurloeDao.getProfile`
* which relies on `Profile.apply`
* which has assertions on the existence of all the relevant keys

Terra UI, as compared to FireCloud UI, does not enforce that a user must complete a full profile. Therefore, any new users in Terra who have not visited FireCloud UI will not have a profile and will be unable to successfully call `GET /api/nih/status`,  in turn preventing proper linking.

This PR changes this. Instead of relying on `ThurloeDao.getProfile`, we'll rely on `ThurloeDao.getAllKVPs`, which does not have assertions.

Additionally, I've added a `recover` to the API so that in the case of error, we have better/clearer messages.

-----

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
@@ -36,14 +36,6 @@ case class NihDatasetPermission(name: String, authorized: Boolean)
 object NihStatus {
   implicit val impNihDatasetPermission = jsonFormat2(NihDatasetPermission)
   implicit val impNihStatus = jsonFormat3(NihStatus.apply)
-
-  def apply(profile: Profile, whitelistAccess: Set[NihDatasetPermission]): NihStatus = {
-    new NihStatus(
-      profile.linkedNihUsername,
-      whitelistAccess,
-      profile.linkExpireTime
-    )
-  }
 }
 
 object NihService {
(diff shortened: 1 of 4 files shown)


breilly2 (approved):
👍 Looks good to me. I see you were somewhat familiar with this phenomenon after dealing with sparse Terra profiles for free credits. I don't see any other places where this could bite us now. (In fact, there appear to be no more calls to `ThurloeDAO.getProfile`.)

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
```diff
@@ -83,18 +75,21 @@ trait NihService extends LazyLogging {
...
-          case Some(_) =>
+    thurloeDao.getAllKVPs(userInfo.id, userInfo) flatMap {
+      case Some(profileWrapper) =>
+        profileWrapper.keyValuePairs.find(_.key.contains("linkedNihUsername")).flatMap(_.value) match {
+          case Some(linkedNihUsername) =>
             Future.traverse(nihWhitelists) { whitelistDef =>
               samDao.isGroupMember(whitelistDef.groupToSync, userInfo).map(isMember => NihDatasetPermission(whitelistDef.name, isMember))
             }.map { whitelistMembership =>
-              RequestComplete(NihStatus(profile, whitelistMembership))
+              val linkExpireTime = Try(profileWrapper.keyValuePairs.find(_.key.contains("linkExpireTime")).flatMap(_.value.map(_.toLong))).toOption.flatten
```
Qi77Qi: minor, maybe slightly cleaner if we do 
```
val linkExpireTime = profileWrapper.keyValuePairs.collectFirst {
                case fckv if(fckv.key.contains("linkExpireTime")) => fckv.value.flatMap(x => Try(x.toLong).toOption)
              }.flatten
```
Just so that we're not wrapping a whole block inside `Try`

Qi77Qi (approved):
👍 

would be even nicer if we can add a unit test :)

davidangb:
@Qi77Qi I've updated to refactor a bit and add some targeted unit tests. `NihApiServiceSpec` has api-level tests for this code already.

@breilly2 good call on additional cleanup to (potentially) deal with unused methods. I'd like to save that for a  follow-on PR to make sure this fix lands asap ... ok?

breilly2 (approved):
👍 Good changes!

breilly2:
Absolutely can wait. I wasn't so much suggesting that we remove that function right now, if at all. More saying that I used this opportunity to look for other features that could potentially fall victim to the same problem and didn't find anything.

PR #728, opened 2019-05-14
Free credits spreadsheet-update refactor  [risk: low]

davidangb:
The scheduled task to update the free-credits spreadsheet was not performant. For reference: https://docs.google.com/document/d/12pwsl_xT1THr-e2K-Dcx35s3baWk_11QQ02VG3eEg2E/edit?usp=sharing (access allowed to Broad internal only)

This PR changes it such that:
* instead of one big query to Elasticsearch (hardcoded limit: 5000 documents), we make multiple paginated [scroll queries](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-scroll.html) to get all documents.
* instead of N small queries to Thurloe, where N=number of free trial users, we batch the queries to Thurloe into groups of 100 users.
* more resilience to/error-handling of missing data when generating the spreadsheet cells

Some rough analysis of how this would change query load in production:

service | Before  PR | After PR
--- | ---------- | ---------- |
Elasticsearch |  1 query, returning 1649 documents | 7 queries, returning up to 250 documents
Thurloe | 1579 queries, 1 user per query | 40 queries, up to 40 users per query
 Thurloe | all keys per user, roughly 20 | 13 keys per  user

-----
Changes since this PR was approved (8dd899080eeabb704275a1cfc23317579e175e30):
1. Changed the order of `key` and `userid` query params in the request to Thurloe. This is purely cosmetic: since we expect the `key`s to be the same for all requests but the `userid`s to  change, I  wanted the static params to come first.
2. Changed the batch size from 100 to 40. It was necessary to reduce this below  ~50 to address url-length restrictions (I originally miscalculated when choosing 100). From empirical perf testing, 40 seemed the sweet spot.
3. Ran a bunch of perf tests at multiple levels of concurrency and multiple query batch sizes. Landed on 40 as the best batch size. I have this data if anyone wants to review.
4. Stood up this version of orch in perf and watched the spreadsheet-update process repeatedly at a batch size of 40. It performed well and the requests to Thurloe were all fast - most under 100ms with a few outliers.

-----

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
@@ -342,7 +342,7 @@ class ElasticSearchTrialDAO(client: TransportClient, indexName: String, refreshM
     val (scrollId, accum) = startScroll // initial query
     val (_, finalAccum) = nextScroll(scrollId, startCount = 0, accum)
 
-    finalAccum
+    finalAccum.sortBy(_.name.value)
   }
 
   private def indexExists: Boolean = {
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
@@ -11,10 +11,11 @@ import org.elasticsearch.action.admin.indices.create.{CreateIndexRequest, Create
 import org.elasticsearch.action.admin.indices.exists.indices.{IndicesExistsRequest, IndicesExistsRequestBuilder, IndicesExistsResponse}
 import org.elasticsearch.action.get.{GetRequest, GetRequestBuilder, GetResponse}
 import org.elasticsearch.action.index.{IndexRequest, IndexRequestBuilder, IndexResponse}
-import org.elasticsearch.action.search.{SearchRequest, SearchRequestBuilder, SearchResponse}
+import org.elasticsearch.action.search._
 import org.elasticsearch.action.support.WriteRequest.RefreshPolicy
 import org.elasticsearch.action.update.{UpdateRequest, UpdateRequestBuilder, UpdateResponse}
 import org.elasticsearch.client.transport.TransportClient
+import org.elasticsearch.common.unit.TimeValue
 import org.elasticsearch.common.xcontent.XContentType
 import org.elasticsearch.index.query.QueryBuilder
 import org.elasticsearch.index.query.QueryBuilders._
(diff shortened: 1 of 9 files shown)


src/test/scala/org/broadinstitute/dsde/firecloud/service/ElasticSearchDAOQuerySupportSpec.scala
```diff
@@ -373,7 +375,9 @@ class ElasticSearchDAOQuerySupportSpec extends FreeSpec with ElasticSearchDAOQue
             boolMap.fields("should") match {
               case shouldArray:JsArray =>
                 shouldArray
+              case x => throw new Exception(s"unmatched case for  ${x.getClass.getName}: ${x.toString()}")
             }
+          case x => throw new Exception(s"unmatched case for  ${x.getClass.getName}: ${x.toString()}")
```
davidangb, line 380: changes in this file are not in scope for this PR, but they fix compile warnings that were bugging me.

davidangb:
fiab-start failed with a github rate-limit exception. Let's try again: jenkins retest

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialServiceSupport.scala
```diff
@@ -57,13 +62,19 @@ trait TrialServiceSupport extends LazyLogging {
...
-    // TODO: this should switch to using Thurloe's bulk endpoint, so we're not making a large burst of requests.
-    val profileWrappers:Future[Seq[ProfileWrapper]] = Future.sequence(projects.filter(p => p.user.isDefined).map { p =>
-      thurloeDAO.getAllKVPs(p.user.get.userSubjectId, managerInfo) map { kvpOption =>
-        kvpOption.getOrElse(ProfileWrapper(p.user.get.userSubjectId, List.empty[FireCloudKeyValue]))
+    // from the list of projects, get assigned userids
+    val assignedUsers:List[String] = projects.toList.filter(p => p.user.isDefined).flatMap(_.user.map(_.userSubjectId))
+    logger.info(s"makeSpreadsheetValues found ${assignedUsers.length} users assigned to projects ...")
+
+    // split users into chunks of size N for efficient querying to Thurloe
+    val profileQueries = assignedUsers
```
kcibul, line 76: Is this going to potentially spin up a bunch of futures (they're eager right?) all hitting Thurloe at once?

kcibul (approved):

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialServiceSupport.scala
davidangb: yes, that's exactly what it will do. This is the same behavior as before this PR; only the quantity+quality of the queries has changed, but they're still shot off in one big eager batch. Also see the `TODO` a few lines below ...

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
```diff
@@ -273,23 +274,75 @@ class ElasticSearchTrialDAO(client: TransportClient, indexName: String, refreshM
...
+    (scrollId, trialProjects)
+  }
+
+  private def startScroll: (String, List[TrialProject]) = {
+    val startScrollRequest = client
+      .prepareSearch(indexName)
+      .setQuery(Ready)
+      .setScroll(TimeValue.timeValueMinutes(1))
+      .addSort("_doc", SortOrder.ASC)
+      .setSize(250) // tweak pagesize here
```
rtitle, line 304: Should this be configurable?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ElasticSearchTrialDAO.scala
```diff
@@ -273,23 +274,75 @@ class ElasticSearchTrialDAO(client: TransportClient, indexName: String, refreshM
...
+        case Some(s) => continueScroll(s)
+      }
 
-    if (reportProjectResponse.getHits.totalHits == 0)
-      Seq.empty[TrialProject]
-    else
-      reportProjectResponse.getHits.getHits.toSeq map ( _.getSourceAsString.parseJson.convertTo[TrialProject] )
+      // if the query returned results, add them to the accumulator and make another query
+      if (batch.nonEmpty) {
+        nextScroll(Some(newScrollId), accum ++ batch)
```
rtitle, line 344: Scala `List` is optimized for prepending; `++` doesn't have great performance. Though I suspect it's not too much of an issue in this cause, where each batch is a webservice call.

Cats [Chain](https://typelevel.org/cats/datatypes/chain.html) is a nice data structure which has the semantics of `List` but allows constant time prepend and append.

Just FYI, don't need to change this PR I think.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
```diff
@@ -119,6 +118,33 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
...
+  override def bulkUserQuery(userIds: List[String], keySelection: List[String]): Future[List[ProfileWrapper]] = {
+    val userIdParams:List[(String,String)] = userIds.map(("userId", _))
+    val keyParams:List[(String,String)] = keySelection.map(("key", _))
+
+    val allQueryParams = userIdParams ++ keyParams
+
+    val queryUri = Uri(UserApiService.remoteGetQueryURL).withQuery(allQueryParams:_*)
+
+    // default uri length for Spray - which Thurloe uses - is 2048 chars
+    assert(queryUri.toString().length <  2048, s"generated url is too long at ${queryUri.toString().length} chars.")
```
rtitle, line 130: Instead of `assert` which throws an `AssertionError` and tends to crash the ActorSystem in my experience -- should this return a `Future.failed(...)`?

rtitle (approved):
A few comments above, overall looks good.

Do you have any performance test results to show the impact of this change?

davidangb:
thanks @rtitle ! Next step after this is perf testing; currently bottlenecked on access to the perf env. I'm hoping that the only tweak to this code as a result of perf testing is the batch size -  currently 100 - for queries to Thurloe, so I felt safe  putting this PR up for code  review.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TrialServiceSupport.scala
```diff
@@ -57,13 +62,17 @@ trait TrialServiceSupport extends LazyLogging {
...
-        kvpOption.getOrElse(ProfileWrapper(p.user.get.userSubjectId, List.empty[FireCloudKeyValue]))
-      }
-    })
+    // from the list of projects, get assigned userids
+    val assignedUsers:List[String] = projects.toList.filter(p => p.user.isDefined).flatMap(_.user.map(_.userSubjectId))
+    logger.info(s"makeSpreadsheetValues found ${assignedUsers.length} users assigned to projects ...")
+
+    // split users into chunks of size N for efficient querying to Thurloe
+    val profileQueries = assignedUsers
+      .grouped(40) // tweak chunk size here!
```
rtitle, line 77: Since this number is a result of performance tuning I'm wondering if this should be configurable rather than hard-coded. Maybe we'll want to change it in the future...

rtitle:
The latest changes look good to me, one question about whether the magic numbers (250, 40, etc) should live in a config file instead of code. My approval still stands though. 👍

davidangb:
@rtitle thanks for the review! I'm not putting these in config because I'm not sure if we'll ever change them or use environment-specific values, so I don't want to add overhead/confusion. But I did add some explicit code comments in those areas.

PR #729, opened 2019-05-30
Add query param passthrough to CromIAM endpoints [BA-5670]

aednichols:
This functionality should have been here all along, we missed it during the JM rush to production.

Job Manager uses the `POST` version of `/query` rather than `GET`, so there was no impact on JM users - only the rare person who tried using Swagger.

Dropping query params on `/metadata` caused unnecessary metadata to be returned, which affected performance but not correctness.

---

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/CromIamApiService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/CromIamApiService.scala
@@ -1,7 +1,5 @@
 package org.broadinstitute.dsde.firecloud.webservice
 
-
-import akka.http.scaladsl.server.PathMatchers._
 import org.broadinstitute.dsde.firecloud.FireCloudConfig
 import org.broadinstitute.dsde.firecloud.service.{FireCloudDirectives, FireCloudRequestBuilding}
 import org.broadinstitute.dsde.firecloud.utils.StandardUserInfoDirectives
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/CromIamApiService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/CromIamApiService.scala
@@ -5,7 +5,7 @@ import akka.http.scaladsl.server.PathMatchers._
 import org.broadinstitute.dsde.firecloud.FireCloudConfig
 import org.broadinstitute.dsde.firecloud.service.{FireCloudDirectives, FireCloudRequestBuilding}
 import org.broadinstitute.dsde.firecloud.utils.StandardUserInfoDirectives
-import spray.http.HttpMethods
+import spray.http.{HttpMethods, Uri}
 import spray.routing.Route
 import spray.routing.HttpService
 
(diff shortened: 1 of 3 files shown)


src/test/scala/org/broadinstitute/dsde/firecloud/webservice/CromIamApiServiceSpec.scala
```diff
@@ -183,5 +228,5 @@ class CromIamApiServiceSpec extends BaseServiceSpec with CromIamApiService with
 
   }
 
-  override implicit def actorRefFactory: ActorRefFactory = system
```
aednichols, line 186: Seems necessary, not sure why

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/CromIamApiService.scala
```diff
@@ -25,7 +23,9 @@ trait CromIamApiService extends HttpService with FireCloudRequestBuilding with F
       path("query") {
         pathEnd {
           get {
-            passthrough(s"$workflowRoot/query", HttpMethods.GET)
+            extract(_.request.uri.query) { query =>
+              passthrough(Uri(s"$workflowRoot/query").withQuery(query), HttpMethods.GET)
```
cjllanwarne, line 27: My "DRY" senses are tingling... Presumably this is a common enough pattern that it might be worth making a `passthroughWithQueryParams` alternative to `passthrough`?

dvoet (approved):

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/CromIamApiService.scala
dvoet: I don't know why passthrough doesn't just do the right thing... who would think passthrough would not include the query parameters?

cjllanwarne: It was certainly a surprise to all of us....

cjllanwarne: Maybe if `passthrough` is our own custom function (rather than something predefined in the dsl), maybe we can rename it to `passthroughWithoutQueryParams`...?

aednichols: I think the basic flaw in `passthrough()` is that it accepts a `spray.http.Uri` instead of a `spray.routing.RequestContext#request`.

We could add a version that takes the request instead and automatically passes through the query params, but we won't be much safer until we remove the ability to use the old query-dropping version in new code... which entails changing 70+ code locations.

Since Orch is almost EOL I didn't think it was worth it.

aednichols: But now that I've thought through the whole thing in a detailed manner maybe I'll just do it.

cjllanwarne (approved):

PR #735, opened 2019-06-25
AoU RW-2776: Add Calhoun Passthrough Endpoint (Orch)

jmthibault79:
All of Us needs to access the Calhoun service via Firecloud-Orchestration (instead of directly) for security process reasons.

https://precisionmedicineinitiative.atlassian.net/browse/RW-2776

Depends on https://github.com/broadinstitute/firecloud-develop/pull/1704

Have you read [CONTRIBUTING.md](../CONTRIBUTING.md) lately? If not, do that first.

I, the developer opening this PR, do solemnly pinky swear that:

- [x] I've followed [the instructions](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/CONTRIBUTING.md#api-changes) if I've made any changes to the API, _especially_ if they're breaking changes
- [x] I've updated the RC_XXX release ticket with any manual steps required to release this change
- [x] I've updated the [FISMA documentation](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/CONTRIBUTING.md#fisma-documentation-changes) if I've made any security-related changes, including auth, encryption, or auditing

In all cases:

- [ ] Get two thumbsworth of review and PO signoff if necessary
- [ ] Verify all tests go green
- [ ] Squash and merge; you can delete your branch after this **unless it's for a hotfix**. In that case, don't delete it!
- [ ] Test this change deployed correctly and works on dev environment after deployment

PR #737, opened 2019-06-26
Fix for AoU RW-2776: Add Calhoun Passthrough Endpoint (Orch)

jmthibault79:
Followup to #735 after some testing with stricter clients on the AoU side.

Swagger and Spray now correctly report that the content they're passing along from Calhoun is `text/html` rather than JSON.  

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/StaticNotebooksApiService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/StaticNotebooksApiService.scala
@@ -3,8 +3,8 @@ package org.broadinstitute.dsde.firecloud.webservice
 import org.broadinstitute.dsde.firecloud.FireCloudConfig
 import org.broadinstitute.dsde.firecloud.service._
 import org.broadinstitute.dsde.firecloud.utils.StandardUserInfoDirectives
-import spray.http.HttpMethods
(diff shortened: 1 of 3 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/webservice/StaticNotebooksApiService.scala
```diff
@@ -13,12 +13,16 @@ trait StaticNotebooksApiService extends HttpService
   private implicit val executionContext = actorRefFactory.dispatcher
 
   val calhounStaticNotebooksRoot: String = FireCloudConfig.StaticNotebooks.baseUrl
+  val calhounStaticNotebooksURL: String = s"$calhounStaticNotebooksRoot/api/convert"
 
   val staticNotebooksRoutes: Route = {
     path("staticNotebooks" / "convert") {
       requireUserInfo() { _ =>
         post {
-          passthrough(s"$calhounStaticNotebooksRoot/api/convert", HttpMethods.POST)
```
jmthibault79, line 21: `passthrough()` automatically wraps a `respondWithMediaType(`application/json`)` directive, which we do not want.

breilly2 (approved):
:+1:

ericsong (approved):

PR #741, opened 2019-08-08
use jwt-core instead of authentikat-jwt [no ticket; risk: low]

davidangb:
We have a medium-priority recommendation from SourceClear to upgrade or replace the `authentikat-jwt` library. That library looks abandoned - the repo says they are "looking for maintainer" and its readme recommends we use something else.

So, I've replaced `authentikat-jwt` with `jwt-core`. As a bonus, `jwt-core` is better about decoding our non-standard JWTs, so I've refactored/simplified the code a bit.

---

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/RegisterApiService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/RegisterApiService.scala
@@ -1,14 +1,10 @@
 package org.broadinstitute.dsde.firecloud.webservice
 
-import akka.actor.Props
-import authentikat.jwt._
-import org.broadinstitute.dsde.firecloud.FireCloudConfig
-import org.broadinstitute.dsde.firecloud.model._
 import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
+import org.broadinstitute.dsde.firecloud.model._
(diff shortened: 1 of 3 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/webservice/RegisterApiService.scala
```diff
@@ -1,14 +1,10 @@
...
-import authentikat.jwt._
-import org.broadinstitute.dsde.firecloud.FireCloudConfig
-import org.broadinstitute.dsde.firecloud.model._
 import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
+import org.broadinstitute.dsde.firecloud.model._
 import org.broadinstitute.dsde.firecloud.service.{FireCloudDirectives, PerRequestCreator, RegisterService}
-import org.broadinstitute.dsde.firecloud.utils.{DateUtils, StandardUserInfoDirectives}
+import org.broadinstitute.dsde.firecloud.utils.StandardUserInfoDirectives
 import org.slf4j.LoggerFactory
-import spray.http.StatusCodes
```
davidangb, line 11: cleaned up the imports in RegisterApiService.scala. This refactor was necessary because of the (unused) `import authentikat.jwt._`

helgridly (approved):

PR #742, opened 2019-08-09
Remove refresh tokens [no ticket yet; risk: no]

davidangb:
Removes almost everything in Orch that deals with refresh tokens, since we don't use those anymore:
* model classes
* methods in Rawls DAOs to call Rawls refresh token APIs
* business logic for retrieving/validating/storing refresh tokens
* google client library dependencies

I've left the two public-facing APIs (`GET /api/refresh-token-status`, `POST /handle-oauth-code`) in place for backwards compatibility with anyone that might be calling these. But I've gutted them so they don't do anything - they are now hardcoded to return successful responses.

----

In all cases:


davidangb:
transient failure during jenkins build; jenkins retest

project/Dependencies.scala
```diff
@@ -42,9 +42,7 @@ object Dependencies {
 
     "com.google.guava"               % "guava"               % "19.0",
 
-    excludeGuavaJDK5("com.google.api-client"          % "google-api-client"   % "1.23.0"),
```
davidangb, line 45: this lib is not actually removed from the codebase; it's pulled in as a transitive dependency from other google libs, e.g. `google-api-services-cloudbilling`. But, since I removed the code that directly referenced the `google-api-client`, I've removed it as a direct dependency.

project/Dependencies.scala
```diff
@@ -42,9 +42,7 @@ object Dependencies {
...
     excludeGuavaJDK5("com.google.apis"                % "google-api-services-storage" % "v1-rev116-1.23.0"),
-    excludeGuavaJDK5("com.google.apis"                % "google-api-services-compute" % "v1-rev163-1.23.0"),
```
davidangb, line 47: this was only used in `OAuthService.scala`, which is now deleted.

MatthewBemis (approved):
Thanks for doing this!

PR #744, opened 2019-08-19
Add importPFB endpoint

breilly2:
See epic [SATURN-984](https://broadworkbench.atlassian.net/browse/SATURN-984) and/or the project design document for more info.

In all cases:

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -3366,8 +3366,12 @@ paths:
           description: Successful Request
         400:
           description: Bad Request
+        401:
+          description: Unauthorized access to Avro file or Workspace
+        403:
+          description: Forbidden access to Avro file or Workspace
         404:
-          description: Workspace not found
+          description: Avro file or Workspace not found
         500:
           description: Internal Error
 
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -3308,34 +3308,6 @@ paths:
         500:
           description: Internal Error
 
-  /api/workspaces/{workspaceNamespace}/{workspaceName}/importPFB:
-    post:
-      x-passthrough: false
-      tags:
-        - Entities
-      operationId: importPFB
-      summary: Import PFB data from an [Avro](https://avro.apache.org/) file
-      produces:
-        - application/json
-      parameters:
-        - $ref: '#/parameters/workspaceNamespaceParam'
-        - $ref: '#/parameters/workspaceNameParam'
-        - in: body
-          description: JSON object containing PFB URL
-          name: pfbImportRequest
-          required: true
-          schema:
-            $ref: '#/definitions/PFBRequest'
-      responses:
-        200:
-          description: Successful Request
-        400:
-          description: Bad Request
-        404:
-          description: Source Workspace not found
-        500:
-          description: Internal Error
-
   /api/workspaces/{workspaceNamespace}/{workspaceName}/importEntities:
     post:
       x-passthrough: false
@@ -3368,6 +3340,34 @@ paths:
           description: Bad Request
         404:
           description: Source Workspace not found
+        500:
+          description: Internal Error
+
+  /api/workspaces/{workspaceNamespace}/{workspaceName}/importPFB:
+    post:
+      x-passthrough: false
+      tags:
+        - Entities
+      operationId: importPFB
+      summary: Import PFB data from an [Avro](https://avro.apache.org/) file
+      produces:
+        - application/json
+      parameters:
+        - $ref: '#/parameters/workspaceNamespaceParam'
+        - $ref: '#/parameters/workspaceNameParam'
+        - in: body
+          description: JSON object containing PFB URL
+          name: pfbImportRequest
+          required: true
+          schema:
+            $ref: '#/definitions/PFBRequest'
+      responses:
+        204:
+          description: Successful Request
+        400:
+          description: Bad Request
+        404:
+          description: Workspace not found
         500:
           description: Internal Error
 
@@ -5264,15 +5264,6 @@ definitions:
       - bagitURL
       - format
 
-  PFBRequest:
-    type: object
-    properties:
-      url:
-        type: string
-        description: link to publically accessible PFB/Avro
-    required:
-      - url
-
   ConfigurationIngest:
     type: object
     properties:
@@ -6403,6 +6394,15 @@ definitions:
         type: array
         items:
           $ref: '#/definitions/MethodConfigurationName'
+
+  PFBRequest:
+    type: object
+    properties:
+      url:
+        type: string
+        description: link to publically accessible PFB/Avro
+    required:
+      - url
 
   Profile:
     type: object
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -3308,6 +3308,34 @@ paths:
         500:
           description: Internal Error
 
+  /api/workspaces/{workspaceNamespace}/{workspaceName}/importPFB:
+    post:
+      x-passthrough: false
+      tags:
+        - Entities
+      operationId: importPFB
+      summary: Import PFB data from an [Avro](https://avro.apache.org/) file
+      produces:
+        - application/json
+      parameters:
+        - $ref: '#/parameters/workspaceNamespaceParam'
+        - $ref: '#/parameters/workspaceNameParam'
+        - in: body
+          description: JSON object containing PFB URL
+          name: pfbImportRequest
+          required: true
+          schema:
+            $ref: '#/definitions/PFBRequest'
+      responses:
+        200:
+          description: Successful Request
+        400:
+          description: Bad Request
+        404:
+          description: Source Workspace not found
+        500:
+          description: Internal Error
+
   /api/workspaces/{workspaceNamespace}/{workspaceName}/importEntities:
     post:
       x-passthrough: false
@@ -5235,6 +5263,15 @@ definitions:
     required:
       - bagitURL
       - format
+
+  PFBRequest:
+    type: object
+    properties:
+      url:
+        type: string
+        description: link to publically accessible PFB/Avro
+    required:
+      - url
 
   ConfigurationIngest:
     type: object
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
@@ -10,6 +10,7 @@ import akka.pattern.pipe
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.firecloud.EntityClient._
 import org.broadinstitute.dsde.firecloud.FireCloudConfig.Rawls
+import org.broadinstitute.dsde.firecloud.model.ErrorReportExtensions.FCErrorReport
 import org.broadinstitute.dsde.firecloud.model.ModelSchema
 import org.broadinstitute.dsde.rawls.model._
 import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
@@ -353,8 +354,8 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema)(im
   def importPFB(pipeline: WithTransformerConcatenation[HttpRequest, Future[HttpResponse]],
                 workspaceNamespace: String, workspaceName: String, pfbRequest: PfbImportRequest): Future[PerRequestMessage] = {
 
-    def callArrow(url: String) = {
-      pipeline { Post(avroToRawlsURL, PfbImportRequest(url)) }
+    def callArrow = {
+      pipeline { Post(avroToRawlsURL, pfbRequest) }
     }
 
     def callRawls(entity: HttpEntity) = {
@@ -371,7 +372,7 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema)(im
     } else {
 
       val futurePerRequestMessage = for {
-        arrowResponse <- callArrow(pfbUrl.toString)
+        arrowResponse <- callArrow
         rawlsJsonEntity <- arrowResponse.status match {
           case OK => Future.successful(arrowResponse.entity)
           case _ => Future.failed(new FireCloudException(arrowResponse.status.value + ": " + arrowResponse.entity.asString))
@@ -379,11 +380,13 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema)(im
         rawlsResponse <- callRawls(rawlsJsonEntity)
         result <- rawlsResponse.status match {
           case NoContent => Future.successful(RequestComplete(NoContent))
+          case NotFound => Future.failed(new FireCloudExceptionWithErrorReport(FCErrorReport(rawlsResponse)))
           case _ => Future.failed(new FireCloudException(rawlsResponse.status.value + ": " + rawlsResponse.entity.asString))
         }
       } yield (result)
 
       futurePerRequestMessage recover {
+        case e: FireCloudExceptionWithErrorReport => RequestCompleteWithErrorReport(optAkka2sprayStatus(e.errorReport.statusCode).getOrElse(InternalServerError), e.errorReport.message, e)
         case e: Throwable => RequestCompleteWithErrorReport(InternalServerError, "Import failed. Details: " + e.getMessage, e)
       }
     }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
@@ -21,6 +21,7 @@ import org.broadinstitute.dsde.firecloud.utils.TSVLoadFile
 import spray.client.pipelining._
 import spray.http.StatusCodes._
 import spray.http._
+import spray.httpx.SprayJsonSupport._
 import spray.json.DefaultJsonProtocol._
 import spray.json._
 import spray.routing.RequestContext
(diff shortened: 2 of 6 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -3308,6 +3308,34 @@ paths:
...
+      responses:
+        200:
+          description: Successful Request
+        400:
+          description: Bad Request
+        404:
+          description: Source Workspace not found
+        500:
+          description: Internal Error
+
```
davidangb: nit: I'm trying to keep the endpoints in alphabetical order - move below `importEntities`? (I know there are other violations in this file)

src/main/resources/swagger/api-docs.yaml
```diff
@@ -5236,6 +5264,15 @@ definitions:
...
 
+  PFBRequest:
+    type: object
+    properties:
+      url:
+        type: string
+        description: link to publically accessible PFB/Avro
+    required:
+      - url
+
```
davidangb: same here: alphabetize this definition

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -340,4 +348,49 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema)(im
...
+    }
+
+    def callRawls(arrowResponse: HttpResponse) = {
+      logger.debug("calling rawls")
+      pipeline {
+        Post(FireCloudDirectiveUtils.encodeUri(Rawls.entityPathFromWorkspace(workspaceNamespace, workspaceName) + "/batchUpsert"), arrowResponse.entity)
+      }
+    }
+
+    val pfbUrl = new URL(pfbRequest.url.replace(" ", "%20"))
```
davidangb: hm, is there a less manual solution for this, something with `URLEncoder.encode` or `FireCloudDirectiveUtils.encodeUri`, or one of the `URI` methods? Alternately, assuming the end user properly encoded a uri, how did it get decoded by the time it reaches here? I'm worried the manual handling of spaces here will miss some other tricky character.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -3308,6 +3308,34 @@ paths:
...
+        - $ref: '#/parameters/workspaceNameParam'
+        - in: body
+          description: JSON object containing PFB URL
+          name: pfbImportRequest
+          required: true
+          schema:
+            $ref: '#/definitions/PFBRequest'
+      responses:
+        200:
+          description: Successful Request
```
davidangb: should this return 204 No Content on success instead?

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -340,4 +348,49 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema)(im
...
+        rawlsResponse <- arrowResponse.status match {
+          case OK =>
+            callRawls(arrowResponse)
+          case _ =>
+            Future.successful(arrowResponse)
+        }
+      } yield (rawlsResponse.status match {
+        case NoContent =>
+          logger.debug("OK response")
+          RequestComplete(OK)
```
davidangb: should this return 204 No Content on success instead?

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -340,4 +348,49 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema)(im
...
+                workspaceNamespace: String, workspaceName: String, pfbRequest: PfbImportRequest): Future[PerRequestMessage] = {
+    def callArrow = {
+      logger.debug("calling arrow")
+      pipeline {
+        Post(avroToRawlsURL, requestContext.request.entity)
+      }
+    }
+
+    def callRawls(arrowResponse: HttpResponse) = {
+      logger.debug("calling rawls")
```
davidangb: for all `logger.debug` statements in this file: either remove or make the log statements more specific/informative

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -340,4 +348,49 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema)(im
...
+            Future.successful(arrowResponse)
+        }
+      } yield (rawlsResponse.status match {
+        case NoContent =>
+          logger.debug("OK response")
+          RequestComplete(OK)
+        case _ =>
+          // Bubble up all other unmarshallable responses
+          logger.warn("Unanticipated response: " + rawlsResponse.status.defaultMessage)
+          RequestComplete(rawlsResponse)
```
davidangb: consider wrapping this into a `RequestCompleteWithErrorReport` and adding an informative top-level error message. The error message from Rawls may not make sense to an end user trying to import a PFB.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -340,4 +348,49 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema)(im
...
+  def importPFB(pipeline: WithTransformerConcatenation[HttpRequest, Future[HttpResponse]],
+                workspaceNamespace: String, workspaceName: String, pfbRequest: PfbImportRequest): Future[PerRequestMessage] = {
+    def callArrow = {
+      logger.debug("calling arrow")
+      pipeline {
+        Post(avroToRawlsURL, requestContext.request.entity)
+      }
+    }
+
+    def callRawls(arrowResponse: HttpResponse) = {
```
davidangb: nit: this should accept an `HttpEntity`, or a model object - that would make it much more testable/encapsulated

davidangb (commented):
this really should have some tests! Additional comments inline.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -3308,6 +3308,34 @@ paths:
...
+        400:
+          description: Bad Request
+        404:
+          description: Source Workspace not found
```
ahaessly: nit: it's not really a source workspace (i noticed several of the other import entities also say source workspace. i think it originated from copy entities but for that request i'm wondering what error is returned if the destination workspace is not found). anyway, just fix what you think is relevant without yak shaving

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -340,4 +348,49 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema)(im
...
+      val futureResponse = for {
+        arrowResponse <- callArrow
+        rawlsResponse <- arrowResponse.status match {
+          case OK =>
+            callRawls(arrowResponse)
+          case _ =>
+            Future.successful(arrowResponse)
+        }
+      } yield (rawlsResponse.status match {
+        case NoContent =>
```
ahaessly: i found some of this naming confusing. like the rawlsResponse might not actually be from rawls. and for some reason i associated the Future.succesful(arrowResponse) with the futureResponse val (instead of the rawlsResponse) when i first looked at the code. i don't really have any good suggestions though... maybe i've just been away from scala too long!

ahaessly (approved):
just a few comments

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
breilly2: Agree. I copied this line from `importBagit` above this in the file (there are some strong similarities between these two features). I'll test out using `FireCloudDirectiveUtils.encodeUri` instead.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
breilly2: I wrestled with this too. I think unrolling the for-comprehension avoided that particular naming issue, but overall I preferred the expressiveness of the for-comprehension.

`futureResponse` could be `futurePerRequestMessage` or `futureResult` or simply `future`. It represents the output of the function and only exists to express that the `Future.failed` case is handled using `recover`. Do any of those names seem better to you @ahaessly ?

ahaessly: i was thinking that instead of rawlsResponse, it could be something like rawlsResonseOrArrowError (although that is pretty long). anything that indicates it may be the response from the rawls or arrow call. 

as for the futureResponse val, maybe don't assign the for result to a val but wrap the for in () and then just call recover after the closing paren. unless you find that more unreadable. if you stick with using the val, i think i like futurePerRequestMessage best.

      (for {
        arrowResponse <- callArrow
        ...
      })) recover {
        ...

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
breilly2: Also, I'm not even using this value in the request to Arrow. Oops! Need to fix that.

breilly2: Unless this should be Arrow's responsibility. 🤔

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
breilly2: I partially agree. Definitely worth saying that the import failed. I still think there's value in including the underlying Rawls message. The easiest way for users to give us error info is to send us a copy of whatever error message we give them. If that included some sort of incident code that we could use to find the relevant logs, that would be really cool. Since we don't have that established practice, the next best thing is to include some detail in the message to hopefully avoid us having to dig around in the logs.

davidangb: I absolutely agree about including the underlying Rawls message; I'm just suggesting wrapping it. What about something like:
```
      } yield rawlsResponse.status match {
        case NoContent =>
          logger.debug("OK response")
          RequestComplete(OK)
        case rawlsStatus =>
          // Bubble up all other unmarshallable responses
          logger.warn("Unanticipated response: " + rawlsResponse.status.defaultMessage)
          // if we can parse Rawls' response as an ErrorReport, wrap it nicely; else just pass it on
          FCErrorReport.tryUnmarshal(rawlsResponse) match {
            case Success(rawlsErrorReport) =>
              RequestCompleteWithErrorReport(rawlsStatus, "Rawls did a baddie; original error is:", Seq(rawlsErrorReport))
            case Failure(_) =>
              RequestComplete(rawlsResponse)
          }
      }
```

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
breilly2: Turns out `FireCloudDirectiveUtils.encodeUri` does _not_ handle this. I was able to easily do this within Arrow which I think should be doing this anyway so no need for orch to do it.

The URL is a string in the JSON payload, not on the query string, so it can have any value. Historically, we encountered this in production with the bagit import so I figure it's a good idea to do it for PFB import as well.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
breilly2: My scala was definitely rusty on the first pass. It was totally weird the way I was passing the arrow error through as a "success". The whole point of mapping over a sequence of futures using a for-comprehension is to take advantage of the success bias and skip remaining steps once there's a failure. I just got hung up on some type details.

I've refactored to split those out and embraced futures and I think it reads a bit better now.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
breilly2: Thanks for the suggestion. Having to `FCErrorReport.tryUnmarshal` feels messy. Also, from reading the code, it looks like `tryUnmarshal` will not unmarshal if `response.status.isSuccess` is false which it will be in this case.

Does the updated code I pushed accomplish more-or-less what you were looking for?

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
davidangb: sounds good - as long as whatever we report back to the user is helpful, I'm totally find if it happens within Arrow.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -3343,6 +3343,34 @@ paths:
...
+          required: true
+          schema:
+            $ref: '#/definitions/PFBRequest'
+      responses:
+        204:
+          description: Successful Request
+        400:
+          description: Bad Request
+        404:
+          description: Workspace not found
```
davidangb: if Arrow returns a 404 - I assume it might, if the PFB url is bad - then this API will also return a 404 and "Workspace not found" is not correct.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -340,4 +349,43 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema)(im
...
+
+    if (!acceptableProtocols.contains(pfbUrl.getProtocol)) {
+      Future.successful(RequestCompleteWithErrorReport(StatusCodes.BadRequest, "Invalid URL protocol: must be https only"))
+    } else {
+
+      val futurePerRequestMessage = for {
+        arrowResponse <- callArrow(pfbUrl.toString)
+        rawlsJsonEntity <- arrowResponse.status match {
+          case OK => Future.successful(arrowResponse.entity)
+          case _ => Future.failed(new FireCloudException(arrowResponse.status.value + ": " + arrowResponse.entity.asString))
```
davidangb, line 379: suggest using:
```
new FireCloudExceptionWithErrorReport(ErrorReport(arrowResponse.status, arrowResponse.entity.asString))
```
since it doesn't rely on a delimited string. Same applies to the `rawlsResponse.status match` block below.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -340,4 +349,43 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema)(im
       }
     }
   }
+
+  def importPFB(pipeline: WithTransformerConcatenation[HttpRequest, Future[HttpResponse]],
+                workspaceNamespace: String, workspaceName: String, pfbRequest: PfbImportRequest): Future[PerRequestMessage] = {
+
+    def callArrow(url: String) = {
+      pipeline { Post(avroToRawlsURL, PfbImportRequest(url)) }
+    }
```
davidangb, line 359: consider having  `callArrow` accept a `PfbImportRequest`, a `URL`, or a `URI` instead of a `String` as its argument

src/main/resources/swagger/api-docs.yaml
breilly2: In that case, it will return a 500 with the message from Arrow (which could be 404 or, in the case of accessing something that doesn't exist in a Google bucket, 403).

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
davidangb: I like the revamped code better in terms of syntax/Scala-ness. It does not accomplish what I am looking for ... but it may not be necessary. Let me try to explain.

I worry that the error messages from Rawls, which in the current code are bubbled up to the user, may not be helpful. If Rawls returns "503 Service Unavailable" and we pass that to the end user, will the end user know if the problem is in Arrow, Rawls, Orch, or somewhere else? I'd love to wrap the Rawls error in a message like "Successfully retrieved PFB, but encountered an error updating entities. Original error message was: " to help with debugging.

My goal is to help the end user troubleshoot problems, where "end user" might include Workbench devs or QA who encounter test failures. If you are comfortable that the error messages you're bubbling up are enough, then great.

davidangb (commented):
couple more small comments inline

src/main/resources/swagger/api-docs.yaml
breilly2: Unfortunately, the way I wrote the code, I'm pretty sure it's going to do the same thing with a 404 from Rawls, so... I guess I need to handle at least some errors differently... maybe all.

I guess this is a good reason for those tests that you asked about.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
breilly2: Now that I'm leaving it to Arrow to sanitize the URL rather than doing that here, I can remove the argument to `callArrow` and let it use the `pfbRequest` already in scope.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
breilly2: I don't want to do this at the moment because, with the way 404 from Rawls is being communicated back up, making this a `FireCloudExceptionWithErrorReport` will cause the status code from Arrow to propagate up which I don't think is universally the right thing.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -371,19 +372,21 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema)(im
...
+        arrowResponse <- callArrow
         rawlsJsonEntity <- arrowResponse.status match {
           case OK => Future.successful(arrowResponse.entity)
           case _ => Future.failed(new FireCloudException(arrowResponse.status.value + ": " + arrowResponse.entity.asString))
         }
         rawlsResponse <- callRawls(rawlsJsonEntity)
         result <- rawlsResponse.status match {
           case NoContent => Future.successful(RequestComplete(NoContent))
+          case NotFound => Future.failed(new FireCloudExceptionWithErrorReport(FCErrorReport(rawlsResponse)))
           case _ => Future.failed(new FireCloudException(rawlsResponse.status.value + ": " + rawlsResponse.entity.asString))
```
davidangb, line 385: I still don't understand why the `NotFound` case couldn't be handled by the catchall case. Writing it somewhat differently:
```
case NoContent => Future.successful(RequestComplete(NoContent))
case errorCode:StatusCode =>
            val errorReport = FCErrorReport.tryUnmarshal(rawlsResponse) match {
              case Success(rpt) => rpt
              case Failure(_) => ErrorReport(errorCode, "Unexpected failure: " + rawlsResponse.entity.asString)
            }
            Future.failed(new FireCloudExceptionWithErrorReport(errorReport))
```
`FCErrorReport.tryUnmarshal` doesn't look for a successful status code. And in your current PR, `FCErrorReport(rawlsResponse)` could fail if the response body is invalid.

https://github.com/broadinstitute/firecloud-orchestration/blob/develop/src/main/scala/org/broadinstitute/dsde/firecloud/core/GetEntitiesWithType.scala#L72 is an example usage.

Here again - tests could help validate what the code does, and I'm okay with the current PR if it needs to go in as-is for deadline reasons.

breilly2: `tryUnmarshal` calls spray's `unmarshal` function which looks like:
```
  def unmarshal[T: FromResponseUnmarshaller]: HttpResponse ⇒ T =
    response ⇒
      if (response.status.isSuccess)
        response.as[T] match {
          case Right(value) ⇒ value
          case Left(error: MalformedContent) ⇒
            throw new PipelineException(error.errorMessage, error.cause.orNull)
          case Left(error) ⇒ throw new PipelineException(error.toString)
        }
      else throw new UnsuccessfulResponseException(response)
```
Interestingly, after writing some tests and plugging in the code you suggested, most of the tests passed. The one that failed was one where I stubbed Rawls to return a 400 response which I want to convert to 500 (it's a bad request from orch, not from the caller of `importPFB`). For the ones that succeeded, I set some breakpoints and attached a debugger and, as I expected, they always match the `Failure(_)` case. So, the code using `tryUnmarshal` has the right behavior, but the code feels misleading as to why. I believe, other than for 20x status codes, `tryUnmarshal` is guaranteed to throw an exception.

I'm happy to discuss further. For the moment, I'm moving forward with the code as it is (with a minor change to also handle 401 and 403 responses similar to 404s).

breilly2:
jenkins retest because I don't understand why you failed

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
davidangb: ah, you're right; I misread where all the code  went. You could just replace `FCErrorReport.tryUnmarshal(rawlsResponse)` with `Try(FCErrorReport(rawlsResponse))` and achieve the same goal, I think. Your call.

PR #745, opened 2019-08-22
Preferences API extensibility [no ticket; risk: low]

davidangb:
This PR has no functional changes to the end user!!!

It adds a few unit tests and refactors the code around the `/api/profile/preferences` endpoint to be easily extensible.

Currently, that API is whitelisted to only allow Thurloe keys that start with "notifications/". We do this for security - there are some Thurloe keys that end users should not modify at will. However, we also want an easily extensible way to whitelist new keys. This PR (hopefully) makes it such that it is easier for developers to add new whitelisted keys, and add new unit tests for those keys.

-----

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/RegisterService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/RegisterService.scala
@@ -86,8 +86,16 @@ class RegisterService(val rawlsDao: RawlsDAO, val samDao: SamDAO, val thurloeDao
     }
   }
 
+  /*  utility method to determine if a preferences key headed to Thurloe is valid for user input.
+      Add whitelist logic here if you need to allow more keys.
+      If we find we update this logic frequently, I suggest moving the predicates to a config file!
+   */
+  private def isValidPreferenceKey(key: String): Boolean = {
+    key.startsWith("notifications/")
+  }
+
   private def updateProfilePreferences(userInfo: UserInfo, preferences: Map[String, String]): Future[PerRequestMessage] = {
-    if (preferences.forall(_._1.startsWith("notifications/"))) {
+    if (preferences.keys.forall(isValidPreferenceKey)) {
       thurloeDao.saveKeyValues(userInfo, preferences).map(_ => RequestComplete(StatusCodes.NoContent))
     } else {
       throw new FireCloudExceptionWithErrorReport(ErrorReport(StatusCodes.BadRequest, "illegal preference key"))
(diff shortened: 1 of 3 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/RegisterService.scala
```diff
@@ -86,8 +86,16 @@ class RegisterService(val rawlsDao: RawlsDAO, val samDao: SamDAO, val thurloeDao
     }
   }
 
+  /*  utility method to determine if a preferences key headed to Thurloe is valid for user input.
+      Add whitelist logic here if you need to allow more keys.
+      If we find we update this logic frequently, I suggest moving the predicates to a config file!
+   */
+  private def isValidPreferenceKey(key: String): Boolean = {
+    key.startsWith("notifications/")
+  }
```
davidangb, line 95: if you need to add a new whitelisted key, you'd do it here, e.g.:
```
    key.startsWith("notifications/") ||
    key.equalsIgnoreCase("myNewWhitelistedKey") ||
    key.matches("my-regex-[0-9]")
```

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/RegisterApiServiceSpec.scala
```diff
@@ -0,0 +1,86 @@
...
+          "substring/notifications/arebad" -> "true"
+        )
+        assertPreferencesUpdate(payload, BadRequest)
+      }
+
+      "should succeed with an empty payload" in {
+        val payload = Map.empty[String, String]
+        assertPreferencesUpdate(payload, NoContent)
+      }
+
```
davidangb, line 70: and of course if you add new whitelisted keys, also add unit tests!

src/test/resources/reference.conf
```diff
@@ -1,7 +1,7 @@
 include "src/main/resources/reference.conf"
 
 akka {
-  loglevel = "ERROR"
+  loglevel = "OFF"
```
davidangb, line 4: this change is not strictly necessary. The unit tests I wrote below generate expected exceptions, and akka logs them to test output in an ugly way. I turned off the logging (which quiets a few other tests but not all of them). I'm open to feedback on this.

davidangb:
unrelated error in swatomation; let's just try the whole thing again: jenkins retest

jdcanas (approved):
Thanks for doing this!

breilly2 (approved):
👍 Looks good!

PR #746, opened 2019-08-26
Add importPFB tests [SATURN-987]

breilly2:

In all cases:

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
@@ -82,6 +82,7 @@ class WorkspaceApiServiceSpec extends BaseServiceSpec with WorkspaceApiService w
   private final val tsvImportPath = workspacesRoot + "/%s/%s/importEntities".format(workspace.namespace, workspace.name)
   private final val tsvImportFlexiblePath = workspacesRoot + "/%s/%s/flexibleImportEntities".format(workspace.namespace, workspace.name)
   private final val bagitImportPath = workspacesRoot + "/%s/%s/importBagit".format(workspace.namespace, workspace.name)
+  private final val pfbImportPath = workspacesRoot + "/%s/%s/importPFB".format(workspace.namespace, workspace.name)
   private final val bucketUsagePath = s"$workspacesPath/bucketUsage"
   private final val storageCostEstimatePath = s"$workspacesPath/storageCostEstimate"
   private final val tagAutocompletePath = s"$workspacesRoot/tags"
@@ -148,6 +149,7 @@ class WorkspaceApiServiceSpec extends BaseServiceSpec with WorkspaceApiService w
 
   var rawlsServer: ClientAndServer = _
   var bagitServer: ClientAndServer = _
+  var arrowServer: ClientAndServer = _
 
   /** Stubs the mock Rawls service to respond to a request. Used for testing passthroughs.
     *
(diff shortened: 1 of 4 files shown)


src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
```diff
@@ -986,6 +992,155 @@ class WorkspaceApiServiceSpec extends BaseServiceSpec with WorkspaceApiService w
...
+          .withStatusCode(400)
+          .withBody("invalid request"))
+
+        (Post(pfbImportPath, HttpEntity(MediaTypes.`application/json`, s"""{"url":"https://missing.avro"}"""))
+          ~> dummyUserIdHeaders(dummyUserId)
+          ~> sealRoute(workspaceRoutes)) ~> check {
+            status should equal(InternalServerError)
+            body.asString should include ("invalid request")
+          }
+      }
```
davidangb, line 1088: my *opinion* is that if Arrow throws a 400, orch should return a 400. In other words, if Arrow tells us the user made a bad request, we should  tell the user they made a bad request instead of telling the user that something failed on the server side. I won't block the PR  on this  - it's your call.

src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
```diff
@@ -986,6 +992,155 @@ class WorkspaceApiServiceSpec extends BaseServiceSpec with WorkspaceApiService w
...
+            .withBody("Pretend this is Rawls upsert JSON"))
+        stubRawlsService(HttpMethods.POST, s"$workspacesPath/entities/batchUpsert", BadRequest, requestBody = Some("Pretend this is Rawls upsert JSON"), body = Some("Rawls is unhappy"))
+
+        (Post(pfbImportPath, HttpEntity(MediaTypes.`application/json`, s"""{"url":"https://good.avro"}"""))
+          ~> dummyUserIdHeaders(dummyUserId)
+          ~> sealRoute(workspaceRoutes)) ~> check {
+            status should equal(InternalServerError)
+            body.asString should include ("Rawls is unhappy")
+          }
+      }
```
davidangb, line 1152: it's less clear to me what to return here. If Rawls returns a 400, I'm still tempted to return a 400 to the user, though I can certainly see why it might actually be a 500.

davidangb (commented):
minor comments inline.

Please add a test for a `PfbImportRequest` that contains a space and/or other extended characters, since that was a minor sticking point during development.

Another corner case to test is what happens if the user's PFB is extremely large and Arrow returns a huge amount of JSON. I suspect this case will be handled in various error trapping, but maybe worth a test?

src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
breilly2: There are a total of 4 requests in this whole process so it gets confusing.

* client -> Orch: If Orch can't understand the request, that is definitely the client's responsibility.
* Orch -> Arrow: If Orch does a poor job converting query parameters into the JSON that Arrow is expecting, this is a bad request that is Orch's fault, not the client's.
* Orch -> Rawls: Orch is making the call, but the guts of the request comes from Arrow. A bad request is perhaps a little bit on Orch but mostly on Arrow. Again, it's not the client's responsibility.
* Arrow -> Avro file: This is the most questionable. Arrow is the one actually making the request, but it's requesting exactly what was passed to Arrow from the client. That could be the client's responsibility. Should Arrow return 400 or 500 for this? If 400, what should Orch do with it?

There are definitely two distinct legitimate reasons why Arrow could return 400. As long as the body contains enough detail to disambiguate between the cases, we can probably choose either. So how do we choose?

I guess 400 would be something the client could potentially do something to fix themselves. Therefore, the answers to the questions are that Arrow and Orch should both lift a 400 from Arrow -> Avro file. This will potentially also lift Orch -> Arrow 400s, but the error content can indicate that it's actually not the client's fault. I will make this change.

davidangb:
or better yet - add a unit test for a user posting the empty string as a url to this API:
```
{"url":""}
```

breilly2:
The space encoding thing was sketchy to begin with. It was a response to a specific case of bad input because, at the time, we were able to respond with that fix quicker than the fix could be applied on the other end. Fully encoding the input breaks things when the input is already properly encoded (which is why the space encoding was a sketchy text substitution).

This time around, I haven't seen evidence of encoding problems and will push back on clients to send proper input if it comes up.

breilly2:
Huge input may be a problem at some point, but I think that would need to be more of an integration test and/or stress tests on Arrow and Rawls. The only problem I can think of local to Orch would be consuming huge amounts of memory because of having to keep the JSON in memory while passing it from Arrow to Rawls. That would definitely require an integration test.

(I'm not sure if using the Arrow response `HttpEntity` as the input `HttpEntity` to Rawls allows spray to stream it through, but it can't hurt to avoid deserializing the JSON into objects just to serialize them back out again.)

breilly2:
Also, what the heck coveralls!? I _added_ tests! *:sigh:*

davidangb (approved):
:+1: lgtm!

breilly2:
One test failed with a `bucket did not become readable` error, so... jenkins retest!?

PR #748, opened 2019-09-12
get-workspace allows querystrings [no ticket; risk: low]

davidangb:
This PR modifies the get workspace API passthrough to pass querystrings through to Rawls. Previously the querystring was dropped.

This is in support of broadinstitute/rawls#1132.

I am *NOT* updating any swagger docs in this PR. This is intentional, as until broadinstitute/rawls#1132 is merged, the query parameters may change. My goal with this PR  is that, by updating the passthrough, any API consumers would be able to immediately take advantage of broadinstitute/rawls#1132 as soon as it is merged, and we could take our time releasing swagger doc updates.

----

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiService.scala
@@ -79,7 +79,9 @@ trait WorkspaceApiService extends HttpService with FireCloudRequestBuilding
           pathEnd {
             get {
               requireUserInfo() { _ =>
-                passthrough(workspacePath, HttpMethods.GET)
+                extract(_.request.uri.query) { query =>
+                  passthrough(Uri(workspacePath).withQuery(query), HttpMethods.GET)
+                }
               }
             } ~
             delete {


helgridly (approved):
do we need to pass querystrings through for other services too?

devonbush:
I'm doubt it makes a difference functionally, but down below your change (lines 98-99), the extract is done prior to/outside of the requireUserInfo() call.  Should the patterns match just for consistencies' sake?

davidangb:
@helgridly my original thinking on querystrings was that anywhere the underlying service doesn't use a QS, orch shouldn't pass it out of an overabundance of security - it eliminates routes of injection or DoS. But if we find we need to keep modifying passthroughs, I agree we should just modify the core passthrough utility.

davidangb:
@devonbush I was wondering if anyone would notice that. I agree that consistency is good, and this is a micro-optimization ... as I was writing the two lines of code for this PR, I thought "well, if the user isn't authed, why should we do the work to extract a querystring that we'll never use"? So I think this PR is correct and anywhere that reverses these operations is wrong.  But I realize it's inconsistent.

devonbush (approved):

PR #753, opened 2019-10-07
changed default timeout for passthrough requests to 120 seconds

kcibul:

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/HttpClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/HttpClient.scala
@@ -16,6 +16,9 @@ import spray.http._
 import spray.routing.RequestContext
 import spray.http.StatusCodes._
 
+import scala.concurrent.duration._
+import akka.util.Timeout
+
 import org.broadinstitute.dsde.firecloud.HttpClient.PerformExternalRequest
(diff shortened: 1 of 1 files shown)


davidangb (approved):
:+1: approving ... though, please note that this affects more than just the `passthrough` utility. It affects anything using `externalHttpPerRequest` ... which I think is a fine change to make.

breilly2 (commented):
What is the goal of this change?

I think this is a dangerously slippery slope. I have also run up against the 60-second timeout and wished it were longer, however I appreciated our stance that longer-running tasks should run in the background rather than keeping the connection open. Outside of DSP, I've seen long timeouts cause people to retry intensive queries before the first query is done, causing the system to be overwhelmed with work that nothing is even listening for the results of anymore. The only available protection was to ask users, through emails and messages in the UI, to NOT retry their queries if they feel they're taking a long time.

If there is a specific endpoint that temporarily (until we can implement a better solution) needs a little extra time, I'd much rather we directly target that endpoint instead of bumping up everything that uses `externalHttpPerRequest`.

Also, if this only targets passthroughs (and a few other things), will we now have some Terra API endpoints timeout at 60s and others at 120s based entirely on the implementation detail of whether or not they pass-through to another service? And don't most of those other services also have their own 60s timeouts?

PR #756, opened 2019-10-21
AS-87: async avro import [risk: low; AS-88]

davidangb:
draft to get tests running and dockers built.

do not merge until we're sure this is the path we want to take.

this PR may be updated before it's ready for review.

This returns a payload that looks like:
```
{
  "jobId": "ce02a972-32a5-4756-87fc-f650ea861a1a",
  "url": "https://testing",
  "workspace": {
    "name": "da-110",
    "namespace": "general-dev-billing-account"
  }
}
```

and sends the same payload to Arrow.

-----

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
@@ -41,7 +41,8 @@ import scala.util.{Failure, Success, Try}
 object EntityClient {
 
   lazy val arrowRoot: String = FireCloudConfig.Arrow.baseUrl
-  lazy val avroToRawlsURL: String = s"$arrowRoot/avroToRawls"
+  lazy val arrowAppName: String = FireCloudConfig.Arrow.appName
+  lazy val avroToRawlsURL: String = s"$arrowRoot/$arrowAppName"
 
   case class ImportEntitiesFromTSV(workspaceNamespace: String,
                                    workspaceName: String,
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
@@ -395,7 +395,7 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema)(im
         // generate job UUID
         val jobId = java.util.UUID.randomUUID()
 
-        val populatedRequest = pfbRequest.copy(jobId = Option(jobId.toString), workspaceName = Option(WorkspaceName(workspaceNamespace, workspaceName)))
+        val populatedRequest = pfbRequest.copy(jobId = Option(jobId.toString), workspace = Option(WorkspaceName(workspaceNamespace, workspaceName)))
 
         // fire-and-forget call Arrow with UUID, workspace info, and presigned URL
         val gzipPipeline = addHeader (`Accept-Encoding`(gzip)) ~> sendReceive ~> decode(Gzip)
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
@@ -49,7 +49,7 @@ object EntityClient {
 
   case class ImportBagit(workspaceNamespace: String, workspaceName: String, bagitRq: BagitImportRequest)
 
-  case class ImportPFB(workspaceNamespace: String, workspaceName: String, pfbRequest: PfbImportRequest)
+  case class ImportPFB(workspaceNamespace: String, workspaceName: String, pfbRequest: PfbImportRequest, userInfo: UserInfo)
 
   def props(requestContext: RequestContext, modelSchema: ModelSchema)(implicit executionContext: ExecutionContext): Props = Props(new EntityClient(requestContext, modelSchema))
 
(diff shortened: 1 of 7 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -355,51 +354,57 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema)(im
...
-    def callRawls(entity: HttpEntity) = {
-      val authPipeline = authHeaders(requestContext) ~> pipeline
-      authPipeline {
-        Post(FireCloudDirectiveUtils.encodeUri(Rawls.entityPathFromWorkspace(workspaceNamespace, workspaceName) + "/batchUpsert"), entity)
+      if (!acceptableProtocols.contains(pfbUrl.getProtocol)) {
+        throw new FireCloudExceptionWithErrorReport(ErrorReport(BadRequest, "Invalid URL protocol: must be https only"))
       }
+
+      // TODO: add real (semantic) validation for origins, to ensure caller isn't supplying malice
+      op(pfbUrl)
```
davidangb, line 374: this `TODO` is a stub to reflect Doug's comments in the design doc that we should check hostname or other creds on the signed URL to make sure it looks like what we expect from Gen3.

ansingh7115:
Since we decided the cloud function is going to write to the bucket with the following format:
```
{
    "namespace": "general-dev-billing-account",
    "name": "matts-workspace",
    "userSubjectId": "111111111111111111",
    "userEmail": "mbemis.firecloud@gmail.com",
    "payload": [some JSON array for rawls upsert],
    "jobId": [some UUID],
    "startTime": [some timestamp]
}

```
I think Orch might also want to pass the user the `userSubjectId` and `userEmail` to the cloud function as well

davidangb:
@asingh7115 good catch! Updated. The payload to Arrow now looks like:
```
{
  "jobId": "5736761c-f711-4f3b-ba06-9a3f07c34a7a",
  "url": "https://good.avro",
  "user": {
    "userEmail": "random@site.com",
    "userSubjectId": "1234"
  },
  "workspace": {
    "name": "name",
    "namespace": "namespace"
  }
}
```

while the payload to the end user remains the same:
```
{
  "jobId": "5736761c-f711-4f3b-ba06-9a3f07c34a7a",
  "url": "https://good.avro",
  "workspace": {
    "name": "name",
    "namespace": "namespace"
  }
}
```

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -355,51 +354,64 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema)(im
...
-      new URL(pfbRequest.url)
-    } catch {
-      case e: Exception => throw new FireCloudExceptionWithErrorReport(ErrorReport(BadRequest, s"Invalid URL: ${pfbRequest.url}", e))
+    // probe Rawls with an empty upsert to this workspace. This checks that the workspace exists and that the user has
+    // permissions to perform an upsert.
+    def validateUpsertPermissions()(op: Boolean => PerRequestMessage) = {
+      val pipeline: WithTransformerConcatenation[HttpRequest, Future[HttpResponse]] = authHeaders(requestContext) ~> sendReceive
+
+      val probeUpsert = pipeline {
+        Post(FireCloudDirectiveUtils.encodeUri(Rawls.entityPathFromWorkspace(workspaceNamespace, workspaceName) + "/batchUpsert"), List.empty[String])
```
ansingh7115, line 383: This is an interesting way of checking the user can upsert to a workspace. What's the reason behind doing it this way vs. just asking Sam if the user has permissions on the workspace (since ostensibly if the workspace resource exists in Sam, the workspace should also exist)?

davidangb: Sorry! I wrote this up in the PR description but I must have overwritten it in a second tab.

I decided to do it this way for a few reasons:
* encapsulation: it keeps Rawls as the only artifact that has knowledge of which permissions need to be checked to do an upsert. Otherwise, we'd have both Orch and Rawls implementing the same permission-check.
* future-proofing: if we ever change which permission needs to be checked, this call would still do the right thing. If we implemented the Sam check ourself, we'd have to update both Rawls and Orch.
* functionality: this call to Rawls validates that the workspace exists AND that the user has permission to upsert. If we queried Sam directly for permissions, we'd either have to make a second call to Rawls to verify the workspace exists, or skip that validation step.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -355,51 +354,64 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema)(im
...
+        val arrowPayload = pfbRequest.copy(jobId = Option(jobId.toString),
+          workspace = Option(WorkspaceName(workspaceNamespace, workspaceName)),
+          user = Option(user))
 
-      futurePerRequestMessage recover {
-        case e: FireCloudExceptionWithErrorReport => RequestCompleteWithErrorReport(optAkka2sprayStatus(e.errorReport.statusCode).getOrElse(InternalServerError), e.errorReport.message, e)
-        case e: Throwable => RequestCompleteWithErrorReport(InternalServerError, "Import failed. Details: " + e.getMessage, e)
+        // fire-and-forget call Arrow with UUID, workspace info, and presigned URL
+        val gzipPipeline = addHeader (`Accept-Encoding`(gzip)) ~> sendReceive ~> decode(Gzip)
+        gzipPipeline { Post(avroToRawlsURL, arrowPayload) }
```
ansingh7115, line 407: Since we are no longer calling the Arrow service but instead calling a Cloud Function - this will have to change - the url being referenced here would be `https://us-central1-terra-arrow-dev.cloudfunctions.net/avro-import` (for dev).

davidangb: updated, here and in #broadinstitute/firecloud-develop#1889

ansingh7115 (approved):

PR #757, opened 2019-10-21
api-docs.yaml: adding number formats; making number integer where thi…

curoli:
api-docs.yaml: adding number formats; making number integer where this is clearly correct.

In all cases:

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -316,7 +316,8 @@ paths:
           in: query
           description: Snapshot ID of configuration.
           required: false
-          type: number
+          type: integer
+          format: int32
         - name: snapshotComment
           in: query
           description: Snapshot comment of AgoraEntity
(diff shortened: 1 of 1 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -7457,6 +7503,7 @@ definitions:
             default: broad-dsde-dev,
           methodVersion:
             type: integer
+            format: int64
```
davidangb: snapshotIDs are elsewhere defined as `int32`s

davidangb (commented):
one minor comment, otherwise looks good!

src/main/resources/swagger/api-docs.yaml
curoli: Thanks, good catch! I committed a fix.

davidangb (approved):
:+1: thanks again! Merge at will.

curoli:
I think I can't merge, because I don't have write access.

davidangb:
ah, of course. I'll push it through then.

PR #759, opened 2019-10-23
skip tests: update swagger responses for create and clone workspace [AS-100; risk: low]

davidangb:
updates response codes and payloads for create- and clone-workspace. Especially important for  anyone generating stubs for these API calls.

-----

In all cases:

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -2627,10 +2627,12 @@ paths:
           schema:
             $ref: '#/definitions/WorkspaceIngest'
       responses:
-        200:
-          description: No response was specified
         201:
           description: Successful Request
+          schema:
+            $ref: '#/definitions/CreateWorkspaceResponse'
+        400:
+          description: Bad request
         403:
           description: Unable to create bucket for workspace
         409:
@@ -2929,6 +2931,8 @@ paths:
       responses:
         201:
           description: Successful Request
+          schema:
+            $ref: '#/definitions/CreateWorkspaceResponse'
         400:
           description: Unable to create resources for workspace
         404:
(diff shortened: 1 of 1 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -5621,6 +5625,59 @@ definitions:
...
+      - attributes
+      - authorizationDomain
+      - bucketName
+      - createdBy
+      - createdDate
+      - isLocked
+      - lastModified
+      - name
+      - namespace
+      - workspaceId
```
ansingh7115, line 5636: When I create a workspace in orchestration I get the following:
```
{
  "attributes": {},
  "authorizationDomain": [],
  "bucketName": "fc-<REDACTED>",
  "createdBy": "ansinghfirecloud@gmail.com",
  "createdDate": "2019-10-23T19:18:58.344Z",
  "isLocked": false,
  "lastModified": "2019-10-23T19:18:58.344Z",
  "name": "test-ws-anu",
  "namespace": "abbbbbanu",
  "workflowCollectionName": "<REDACTED>",
  "workspaceId": "<REDACTED>"
}
```
which includes a `workflowCollectionName` that isn't defined in this reponse

davidangb: it's defined as a property, but not `required` since it's an `Option` in the underlying case class - and it shows up in swagger ui

ansingh7115: oh! I see :) you're right!

ansingh7115 (approved):

src/main/resources/swagger/api-docs.yaml
```diff
@@ -5004,20 +5018,6 @@ paths:
...
-    enum: [WDL, CWL]
-    required: false
-    in: formData
-  workflowTypeVersionParam:
-    name: workflowTypeVersion
-    description: The specification version for the workflow language being used. For WDL, Cromwell currently supports draft-2 and 1.0. For CWL, Cromwell currently supports v1.0.
-    required: false
-    type: string
-    enum: ["draft-2", "1.0", "v1.0"]
-    in: formData
```
davidangb, line 5020: swagger can't handle having the `in: formData` defined here in `parameters`, because it can't validate the `consumes` settings for these. So I moved them up to where they're actually used.

ansingh7115:
New changes also look good

PR #761, opened 2019-10-25
import-PFB status endpoint [AS-97; risk: low]

davidangb:
requires broadinstitute/firecloud-develop#1896

no tests; manually tested.

the code in this PR is specific to a feature under deadline and will need tech-debt refactoring after deadline.

-----

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
@@ -442,53 +442,56 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema, go
     }
 
     // list all files in the jobId's subdirectory and strip jobId prefix so we have just the filenames
-    val files = Try(googleServicesDAO.listObjectsAsRawlsSA(bucketName, jobId + "/").map(_.replace(jobId + "/",""))) match {
-      case Success(obs) => obs
+    Try(googleServicesDAO.listObjectsAsRawlsSA(bucketName, jobId + "/").map(_.replace(jobId + "/",""))) match {
+
       case Failure(ex) =>
         logger.warn(s"Error listing GCS objects for prefix gs://$bucketName/$jobId", ex)
-        List("failed to list")
+        // do not expose the bucket name in the error report we send to users
+        Future.successful(RequestCompleteWithErrorReport(InternalServerError, "Error listing bucket"))
+
+      case Success(files) =>
+        // what files do we have?
+        val (statusCode, statusString, message) = files.toSet match {
+
+          // no files yet, or subdirectory doesn't exist
+          case nothing if nothing.isEmpty =>
+            (NotFound, "NOT FOUND", "jobId not found or job not yet started")
+
+          // somebody wrote an error.json: it's an error.
+          case error if error.contains("error.json") =>
+            // try to read the error contents
+            val errorMessage = readStatusFile("error.json", "Error!")
+            (OK, "ERROR", errorMessage)
+
+          // we finished with a success.json: it's a success. Yay!
+          case success if success.contains("success.json") =>
+            // try to read the success contents
+            val successMessage = readStatusFile("success.json", "Success! Import completed.")
+            (OK, "SUCCESS", successMessage)
+
+          // we have a start marker, but do not yet have a success or error marker. It's in progress.
+          case started if started.contains("running.json") =>
+            val runningMessage = readStatusFile("running.json", "import in progress")
+            (OK, "RUNNING", runningMessage)
+
+          // we've got either/both of the upsert and the metadata, but no other markers - it's in progress.
+          case alsostarted if alsostarted == Set("upsert.json") || alsostarted == Set("metadata.json") || alsostarted == Set("upsert.json", "metadata.json") =>
+            (OK, "RUNNING", "import in progress")
+
+          // the files don't make sense; give up.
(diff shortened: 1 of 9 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -51,6 +52,7 @@ object EntityClient {
   case class ImportBagit(workspaceNamespace: String, workspaceName: String, bagitRq: BagitImportRequest)
 
   case class ImportPFB(workspaceNamespace: String, workspaceName: String, pfbRequest: PfbImportRequest, userInfo: UserInfo)
+  case class ImportPFBStatus(workspaceNamespace: String, workspaceName: String, jobId: String, userInfo: UserInfo)
```
helgridly: nit: can this be `PFBImportStatus`? i keep reading it as importing the status.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -423,4 +427,68 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema, go
...
+
+    val bucketName = FireCloudConfig.Arrow.bucketName
+
+
+    // list all files in the jobId's subdirectory and strip jobId prefix so we have just the filenames
+    val files = Try(googleServicesDAO.listObjectsAsRawlsSA(bucketName, jobId + "/").map(_.replace(jobId + "/",""))) match {
+      case Success(obs) => obs
+      case Failure(ex) =>
+        logger.warn(s"Error listing GCS objects for prefix gs://$bucketName/$jobId", ex)
+        List("failed to list")
```
helgridly: I suggest keeping the failure here and handing it below. Right now you're creating this sentinel value and then not explicitly accounting for it, leaving it to be interpreted as "a collection of three files that don't look like what we expect", which I'm not the biggest fan of.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -184,19 +185,47 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
...
     googleCredential.refreshToken()
     googleCredential.getAccessToken
   }
 
+  override def listObjectsAsRawlsSA(bucketName: String, prefix: String): List[String] = {
+    val storage = new Storage.Builder(httpTransport, jsonFactory, getRawlsServiceAccountCredential).setApplicationName(appName).build()
+    val listRequest = storage.objects().list(bucketName).setPrefix(prefix)
+    Try(executeGoogleRequest[Objects](listRequest)) match {
+      case Failure(ex) =>
+        // TODO: do anything else here?
```
helgridly: well, is there?

helgridly: also you're using a `Try` and then discarding it in favour of throwing an exception -- might as well return the `Try` from the function

helgridly (changes requested):
`List("failed to list")` really weirds me out

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
davidangb: done!

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
davidangb: done!

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
davidangb: I don't want to return the `Try` because nothing else in this class does - they all throw exceptions. But I've made the exception case do something now.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -423,4 +427,71 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema, go
...
+
+    val bucketName = FireCloudConfig.Arrow.bucketName
+
+    def readStatusFile(filename: String, default: String): String = {
+      Try(googleServicesDAO.getObjectContentsAsRawlsSA(bucketName, s"$jobId/$filename")) match {
+        case Success(msg) =>
+          if (msg.trim.nonEmpty) msg else default
+        case Failure(ex) =>
+          logger.warn(s"error reading GCS object gs://$bucketName/$jobId/$filename", ex)
+          default
```
helgridly, line 440: if you call this trying to read a `success.json` that exists but the read fails for some reason, is it correct to still return the default here, which is a message saying the import completed successfully?

helgridly (approved):
"good enough for the deadline", i'm told!

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
davidangb: discussed in person - I think it's ok to say success since the success.json read operation and import operation are separate

PR #766, opened 2019-11-07
add passthroughs for releaseHold and callcaching diff CromIAM endpoints [WA-63]

rsasch:
Closes https://broadworkbench.atlassian.net/browse/WA-63

In all cases:

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -580,7 +580,7 @@ paths:
         '200':
           description: Successful Request
           schema:
-            $ref: '#/definitions/WorkflowAbortResponse'
+            $ref: '#/definitions/WorkflowStatusResponse'
         '400':
           description: Malformed Workflow ID
         '403':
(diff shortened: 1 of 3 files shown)


kshakir (approved):
LGTM

src/main/resources/swagger/api-docs.yaml
```diff
@@ -864,6 +864,103 @@ paths:
...
+            $ref: '#/definitions/WorkflowStatusResponse'
+        '400':
+          description: Malformed Workflow ID
+        '403':
+          description: Malformed Request
+        '404':
+          description: Workflow ID Not Found
+        '500':
+          description: Internal Error
+      security:
```
davidangb: you shouldn't have to specify `security` here; it will be inherited from the global definition at line 18

src/main/resources/swagger/api-docs.yaml
```diff
@@ -864,6 +864,103 @@ paths:
...
+        - CromIAM Workflows (for Admin)
+      responses:
+        '200':
+          description: Successful Request
+          schema:
+            $ref: '#/definitions/WorkflowStatusResponse'
+        '400':
+          description: Malformed Workflow ID
+        '403':
+          description: Malformed Request
```
davidangb, line 896: 403 is usually Forbidden - is it really malformed request for this API? If that's what cromiam returns, then ok

src/main/resources/swagger/api-docs.yaml
```diff
@@ -864,6 +864,103 @@ paths:
...
+  '/api/workflows/{version}/callcaching/diff':
+    get:
+      summary: Return the hash differential between two calls
+      parameters:
+        - name: version
+          description: API Version
+          required: true
+          type: string
+          in: path
+          default: v1
```
davidangb, line 911: not a blocker for this PR: the `version` param is used for a lot of APIs and could/should be made into a reusable param in the `parameters` block  around line 5162.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -864,6 +864,103 @@ paths:
...
+        - name: indexB
+          description: Index of the second call
+          required: false
+          type: number
+          in: query
+      tags:
+        - CromIAM Workflows (for Admin)
+      responses:
+        '200':
+          description: Successful Request
```
davidangb, line 946: what payload does the 200 return?

src/main/resources/swagger/api-docs.yaml
```diff
@@ -864,6 +864,103 @@ paths:
...
+          description: Malformed query parameters
+        '404':
+          description: No matching cache entry
+        '500':
+          description: Internal Error
+      security:
+        - googleoauth:
+            - openid
+            - email
+            - profile
```
davidangb: no need to specify `security` here

davidangb (commented):
a few comments inline

src/main/resources/swagger/api-docs.yaml
rsasch: thanks – removed

src/main/resources/swagger/api-docs.yaml
rsasch: this is from the CromIAM swagger

src/main/resources/swagger/api-docs.yaml
rsasch: added in response details

src/main/resources/swagger/api-docs.yaml
rsasch: thanks – removed

rsasch:
jenkins retest

davidangb (approved):
:+1: thanks!

PR #768, opened 2019-11-21
AS-47: tsv entity import is now lenient to trailing delimiter-only lines [risk: low]

davidangb:
User attempts to import a TSV file, where the TSV contains multiple trailing delimiter-only lines, e.g.:
```
entity:some_id\tcol1\tcol2
id1\tVal1\tVal2
id2\tVal1\tVal2
\t\t
\t\t
\t\t
```
this can occur when exporting a TSV from Excel, etc.

**Expected**: the TSV imports successfully.
**Actual**: import fails with an error message "Duplicated attribute keys are not allowed"

Before this PR, the code would parse each line, extract the empty string as the entity id for the trailing lines, and fail because it believed the TSV to contain multiple rows with the same id (of the empty string)

After this PR, the TSV parser ignores any lines that consist purely of delimiters and therefore have no value.

-----

In all cases:


MatthewBemis (approved):
:shipit: 👍

PR #770, opened 2019-11-26
DFE-1: create anonymous google groups for MixPanel integration

mmorgantaylor:
Note that there is a corresponding PR in firecloud-develop that adds two config variables in firecloud-orchestration.conf.ctmpl that are used here.

This code creates an anonymized Google group to represent each user (1:1 anonymous Group : user) that we will send to Mixpanel and could later email through Mixpanel platform without revealing the user's email address or identity.

Design Requirements:
- The identity of the user that is sent to Mixpanel must not be the user’s true identity (anonymized)
- The identity we send will be a Google group email address that can be used to contact that user (i.e. the group contains the user, but the content of the group is not visible by the Mixpanel SaaS)
- The email address of this anonymous identity Google group must not be reversely engineerable (see the [SIA doc](https://docs.google.com/document/d/1X283r5ADUuqy2975TViRgyp1_6TYMjm2_etVwmOw9k0/edit) for details on the hashing algorithm)
- The Terra UI must be able to retrieve this anonymous identity Google group for any given user
- If a given user does not yet have an anonymous identity Google group it must be generated and associated with the user such that it can be retrieved later

[Design doc for this project](https://docs.google.com/document/d/1KyDvcZuQHuMj1pb2RYoCKZBD0CSpk9WceCgdv90IY_g/edit#)

In all cases:


src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -505,6 +526,50 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
...
+    val checkGroupRequest = directoryService.groups.get(groupEmail)
+    val existingGroup = Try(executeGoogleRequest(checkGroupRequest)) match {
+      case Failure(f) =>
+        logger.warn(s"Google Group $groupEmail does not exist")
+        "" // return nothing in case of failure
+      case Success(_) => {
+        groupEmail // return groupEmail
+      }
+    }
+    Option(existingGroup)
```
rushtong: Please forgive the drive by comment. I think this would be more idiomatic if it returned a boolean instead of an option-string. The caller already knows the group email, it doesn't necessarily need to know that again. Additionally, do you need to log the failure here in this underlying DAO method? Seems to me that information is more of a business logic and would be better handled in `UserService`.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -505,6 +526,50 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
...
+
+    val insertRequest = directoryService.groups.insert(newGroup)
+    val setEmail = Try(executeGoogleRequest[Group](insertRequest)) match {
+      case Failure(f) => {
+        logger.warn(s"Could not create new group called $groupEmail")
+        "" // return empty string
+      }
+      case Success(newGroupInfo) => {
+        // add targetUserEmail as member of google group - modeled after `override def addMemberToGroup` in workbench-libs HttpGoogleDirectoryDAO.scala
+        val member = new Member().setEmail(targetUserEmail).setRole("MEMBER").setDeliverySettings("ALL_MAIL")
```
rushtong: `"MEMBER"` and `"ALL_MAIL"` should be vals somewhere, similar to how `HttpGoogleDirectoryDAO` treats it.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -505,6 +526,50 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
...
+        val memberInsertRequest = directoryService.members.insert(groupEmail, member)
+
+        Try(executeGoogleRequest(memberInsertRequest)) match {
+          case Failure(f) =>
+            logger.warn(s"Could not add new member $targetUserEmail to group $groupEmail")
+            "" // return nothing
```
rushtong: Question about this failure case. If we successfully created the group, but were not able to add the member, does that leave the group as a existing group? Would future retries with the same values then fail on the `insertRequest`?

src/main/scala/org/broadinstitute/dsde/firecloud/service/UserService.scala
```diff
@@ -1,50 +1,64 @@
...
+
+
 object UserService {
   sealed trait UserServiceMessage
 
   val TerraPreferenceKey = "preferTerra"
   val TerraPreferenceLastUpdatedKey = "preferTerraLastUpdated"
+  val AnonymousGroupKey = "anonymousGroup"
+  val ContactEmailKey = "contactEmail"
+  // following lists are governed by this license (https://aaron.mit-license.org/) and come from https://github.com/aaronbassett/Pass-phrase
```
rushtong, line 33: Since we're explicitly calling out the licensing here, it would be nice to know for sure if it's compatible with Orch's BSD license.

src/main/scala/org/broadinstitute/dsde/firecloud/service/UserService.scala
```diff
@@ -110,4 +123,86 @@ class UserService(rawlsDAO: RawlsDAO, thurloeDAO: ThurloeDAO, userToken: UserInf
...
+  }
+
+  private def callCreateGoogleGroup(keys: ProfileWrapper, anonymousGroupName: String): Future[PerRequestMessage] = {
+    // define userEmail to add to google Group - check first for contactEmail, otherwise use user's login email
+    val userEmail = getProfileValue(keys, UserService.ContactEmailKey) match {
+      case None | Some("") => userToken.userEmail // if the contactEmail key doesn't exist OR (|) if there's no contactEmail set in the profile
+      case Some(contactEmail) => contactEmail // if there is a non-empty value set for contactEmail, we assume contactEmail is a valid email
+    }
+
+    googleServicesDAO.createGoogleGroup(anonymousGroupName, userEmail) match { // this will return "" if group creation is not successful
```
rushtong: Generic feedback, I think some of the inline could be grouped into method documentation instead, but that's just personal preference. The fact that it's there is great!

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
mmorgantaylor: thank you for the very valuable drive by comment! great suggestions - made all these changes.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
mmorgantaylor: I think that that failure case would leave that Google group existing but empty, and not associated with the user (since it would not be set as `anonymousGroup` in Thurloe - this logic happens in  the function `callCreateGoogleGroup` in `UserService`. So the next time this sequence happens, the process would repeat with a new group. It's not the most elegant because it does leave that empty Google group but I don't think it harms anything. Very open to suggestions, though, @rushtong!

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/GoogleServicesDAO.scala
```diff
@@ -45,6 +45,9 @@ trait GoogleServicesDAO extends ReportsSubsystemStatus {
 
   def trialBillingManagerRemoveBillingAccount(projectName: String, targetUserEmail: String): Boolean
 
+  def checkGoogleGroupExists(groupEmail: String): Boolean // Option[String]
```
MatthewBemis: is the comment here leftover code from development?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -505,6 +530,44 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
...
+
+    val insertRequest = directoryService.groups.insert(newGroup)
+    val setEmail = Try(executeGoogleRequest[Group](insertRequest)) match {
+      case Failure(_) => {
+        logger.warn(s"Could not create new group called $groupEmail")
+        "" // return empty string
+      }
+      case Success(newGroupInfo) => {
+        // add targetUserEmail as member of google group - modeled after `override def addMemberToGroup` in workbench-libs HttpGoogleDirectoryDAO.scala
+        val member = new Member().setEmail(targetUserEmail).setRole(anonymizedGroupRole).setDeliverySettings(anonymizedGroupDeliverySettings)
```
MatthewBemis: this is called `createGoogleGroup` but it does quite a bit more than that. for a DAO, i would expect separate generic and lightweight methods: one to create a google group and one to add a member to a group. there's also some logic specific to your use-case that i would be weary of including in something so generically named. could you pass in the anonymized group settings from wherever is calling this?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
rushtong: As a generic safety check, it *should* delete the group in case of user association failure, but I don't have any real sense of how much of a problem it would be if that didn't happen. Given the nature of random names, it seems extraordinarily unlikely to hit a duplicate.

src/main/scala/org/broadinstitute/dsde/firecloud/service/UserService.scala
```diff
@@ -110,4 +123,100 @@ class UserService(rawlsDAO: RawlsDAO, thurloeDAO: ThurloeDAO, userToken: UserInf
...
+    *       - double check that that google group actually exists (and try to create it if it doesn't)
+    *       - return all keys
+    *     - if that key does not exist, we:
+    *       - generate an anonymized Google group for the user
+    *       - set that as the value for `anonymousGroup` KVP
+    *       - return all keys, including new `anonymousGroup`
+    * @param userToken
+    * @return
+    */
+  def getAllUserKeys(userToken: UserInfo): Future[PerRequestMessage] = {
```
MatthewBemis, line 211: how often will this be called by Terra UI once it's fully integrated? my guess is once per full page reload since the keys should be stored in browser state. does that sound right? i'm asking because i'm trying to think about how much added volume to google group APIs this will be adding, but i think it's likely a non-issue.

this also feels a bit side-effecty to me. is there another trigger that could be used for creating the anonymized group? user registration seems like a good choice, but obviously that doesn't handle existing users. i'm not sure there's a better alternative, just thinking out loud

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/GoogleServicesDAO.scala
mmorgantaylor: yes, thanks, will fix!

src/main/scala/org/broadinstitute/dsde/firecloud/service/UserService.scala
mmorgantaylor: yes, this should be called once per full page reload. we weren't sure how to evaluate whether this would add too much volume to google group APIs?

we'd discussed your suggested alternative - putting this in user registration and migrating existing users with a one-time script - but we landed on this route and folks seemed okay with it. what do you think, @MatthewBemis ?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
mmorgantaylor: that's a good point. do you think it makes more sense to rename this function to something more specific, like `createAnonymizedGoogleGroup`, or to break it down as you suggest into the separate functions that would be reusable for other Google group purposes in future?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
mmorgantaylor: okay, I wrote a function to do so but still need to create tests to ensure it works properly

src/main/scala/org/broadinstitute/dsde/firecloud/service/UserService.scala
MatthewBemis: if it's once per full page reload then it's of no concern IMO. I just wanted to make sure that was the case since I didn't have all of the context. 

i think this approach is good- a migration script is much more difficult and not really worthwhile. this is essentially a runtime migration anyway :+1:

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
MatthewBemis: Ideally, it would be broken down into reusable code that doesn't contain any use-case specific logic. This DAO doesn't do a particularly good job at that, but the ones in workbench-libs do.

Because I don't think this code will be re-used in the near future, I would say it's okay to be in the middle-ground here, and keep the use-case specific logic but to break down the functions into more appropriate parts (which will help with readability IMO). i.e. you'll have:

* createAnonymizedGoogleGroup (like you suggested)
* addMemberToAnonymizedGoogleGroup

If you want, you can take it all the way and extract the anonymized portions too, but I wouldn't withhold my thumb for that.

src/main/scala/org/broadinstitute/dsde/firecloud/service/UserService.scala
mmorgantaylor: thanks, it's under an MIT license so we're good. I added a NOTICE.txt file with license information for those files, as per https://broadinstitute.atlassian.net/wiki/spaces/DSDE/pages/3014826/DSDE+Open+Source+Licensing

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
mmorgantaylor: I broke out an `addMemberToAnonymizedGroup` function, and that left `createGoogleGroup` as a non-case-specific function, so I didn't change its name. it wouldn't be terribly hard to add the role settings as inputs to the `addMemberToAnonymizedGroup` function, I just wasn't sure of the best way to do so (scala-wise... in python I might want those settings to be kwargs inputs). would love if you'd take another look when you get a chance and see if this mostly did what you envisioned, @MatthewBemis

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -505,6 +528,76 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
...
+      }
+      case Success(_) => {}
+    }
+  }
+
+  /**
+    * create a new Google group
+    * @param groupEmail     the name of the new Google group to be created. it must be formatted as an email address
+    * @return               Option of the groupEmail address if all above is successful, otherwise if the group creation
+    *                        step failed, return Option of an empty string
```
rushtong: Is the group email passed in the same thing that would be returned on success? If so, then this method could also just return a boolean. If they're different, then yes, this makes sense.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -505,6 +528,76 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
     }
   }
 
+  override def checkGoogleGroupExists(groupEmail: String): Boolean = {
+    val directoryService = getDirectoryManager(getDelegatedCredentialForAdminUser)
+    val checkGroupRequest = directoryService.groups.get(groupEmail)
+    Try(executeGoogleRequest(checkGroupRequest)) match {
+      case Failure(_) => false
```
rushtong: There could be a variety of failure modes here. It might make sense to check the response code and maybe log errors that are unexpected. Not having looked at the APIs, I would expect that anything other than a 404 would be a good candidate to log.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -505,6 +528,76 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
...
+    */
+  override def createGoogleGroup(groupEmail: String): Option[String] = {
+    val newGroup = new Group().setEmail(groupEmail)
+    val directoryService = getDirectoryManager(getDelegatedCredentialForAdminUser)
+
+    val insertRequest = directoryService.groups.insert(newGroup)
+    val setEmail = Try(executeGoogleRequest[Group](insertRequest)) match {
+      case Failure(_) => {
+        logger.warn(s"Could not create new group called $groupEmail")
+        "" // return empty string
```
rushtong: A more idiomatic approach is to use `Option.empty` which allows for easier case matching for callers.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -505,6 +528,76 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
...
+
+    // add targetUserEmail as member of google group - modeled after `override def addMemberToGroup` in workbench-libs HttpGoogleDirectoryDAO.scala
+    val member = new Member().setEmail(targetUserEmail).setRole(anonymizedGroupRole).setDeliverySettings(anonymizedGroupDeliverySettings)
+    val memberInsertRequest = directoryService.members.insert(groupEmail, member)
+
+    val memberEmail = Try(executeGoogleRequest(memberInsertRequest)) match {
+      case Failure(_) =>
+        logger.warn(s"Could not add new member $targetUserEmail to group $groupEmail.")
+        deleteGoogleGroup(groupEmail)
+        "" // return empty string
```
rushtong: `Option.empty` here too.

src/main/scala/org/broadinstitute/dsde/firecloud/service/UserService.scala
```diff
@@ -110,4 +123,110 @@ class UserService(rawlsDAO: RawlsDAO, thurloeDAO: ThurloeDAO, userToken: UserInf
...
+        googleServicesDAO.addMemberToAnonymizedGoogleGroup(groupEmailName, userEmail) match { // returns "" if user addition is not successful
+          case None | Some("") => {
+            Future(RequestComplete(keys))
+          }
+          case Some(addedUserEmail) => {
+            // only if the anonymized Google group was successfully created and user email added to group
+            writeAnonymousGroup(userToken, groupEmailName) // write new KVP to Thurloe
+          }
+        }
+      } // this returns a Future[PerRequestMessage]
```
rushtong: I don't think you need this comment - the return of the last execution is standard, and `Future[PerRequestMessage]` is defined as the return value in the method signature.

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockThurloeDAO.scala
```diff
@@ -110,7 +113,14 @@ class MockThurloeDAO extends ThurloeDAO {
       ),
       TRIAL_SELF_ENABLED -> baseProfile,
       TRIAL_SELF_ENABLED_ERROR -> baseProfile,
-      TRIAL_SELF_ENABLED_PREEXISTING -> baseProfile
+      TRIAL_SELF_ENABLED_PREEXISTING -> baseProfile,
+      HAVE_GOOGLE_GROUP -> baseProfile.++(Set(
+        FireCloudKeyValue(Some("anonymousGroup"), Some("existing-google-group@support.something.firecloud.org")))
+      ),
+      HAVE_EMPTY_GOOGLE_GROUP -> baseProfile.++(Set(
+      FireCloudKeyValue(Some("anonymousGroup"), Some("")))
```
rushtong: nit: indentation

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
mmorgantaylor: much cleaner, thank you!

src/test/scala/org/broadinstitute/dsde/firecloud/mock/MockGoogleServicesDAO.scala
```diff
@@ -106,6 +106,12 @@ class MockGoogleServicesDAO extends GoogleServicesDAO {
 
   override def trialBillingManagerRemoveBillingAccount(projectName: String, targetUserEmail: String): Boolean = false
 
+  override def deleteGoogleGroup(groupEmail: String): Unit = Unit
+  override def checkGoogleGroupExists(groupEmail: String): Boolean = true
+  override def createGoogleGroup(groupName: String): Option[String] = Option("new-google-group@support.something.firecloud.org")
+
+  override def addMemberToAnonymizedGoogleGroup(groupName: String, targetUserEmail: String): Option[String] = Option("user-email@something.com")
```
rushtong, line 111: In addition to the tests for the paths that are hit in `UserApiServiceSpec`, I would recommend unit tests for these dao methods as well. `HttpGoogleServicesDAOSpec` seems to make the most sense for them.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
mmorgantaylor: I changed this function to `checkUserIsInExistingGoogleGroup` and return the error code (and log non-404 failures) or 200 if successful. a 404 response results (logic in `UserService.scala`) prompts to create a brand new google group to avoid overwriting another one, and if there's a different kind of error, we do nothing.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
mmorgantaylor: I wanted to make sure we get the name of the group that's created, in case for some unimaginable reason it would change. I can't imagine why it would, so it probably would be safe to return a Boolean. But that would change the logic of our unit tests and since it works now I'd rather keep it this way unless it seems problematic @rushtong

rushtong: 👍 Sounds good to me, thanks for thinking that out!

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -505,6 +529,85 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
...
+  }
+
+  override def deleteGoogleGroup(groupEmail: String): Unit = {
+    val directoryService = getDirectoryManager(getDelegatedCredentialForAdminUser)
+    val deleteGroupRequest = directoryService.groups.delete(groupEmail)
+    Try(executeGoogleRequest(deleteGroupRequest)) match {
+      case Failure(_) => {
+        logger.warn(s"Failed to delete group $groupEmail")
+      }
+      case Success(_) => {}
```
rushtong: Could be simplified:
```
      case Failure(_) => logger.warn(s"Failed to delete group $groupEmail")
      case Success(_) =>
```

src/main/scala/org/broadinstitute/dsde/firecloud/service/UserService.scala
```diff
@@ -110,4 +122,120 @@ class UserService(rawlsDAO: RawlsDAO, thurloeDAO: ThurloeDAO, userToken: UserInf
...
+      + "@" + FireCloudConfig.FireCloud.supportDomain )
+    anonymousGroupName
+  }
+
+  private def getAllKeysFromThurloe(userToken: UserInfo): Future[ProfileWrapper] = {
+    thurloeDAO.getAllKVPs(userToken.id, userToken) map { // .getAllKVPs returns Option[ProfileWrapper]
+      case None => ProfileWrapper(userToken.id, List())
+      case Some(wrapper) => {
+        wrapper
+      }
```
rushtong: nit: Here too, could be condensed:
```
case Some(wrapper) => wrapper
```

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
mmorgantaylor: I think I can even remove the whole `case Success(_) =>` line

src/test/scala/org/broadinstitute/dsde/firecloud/mock/MockGoogleServicesDAO.scala
mmorgantaylor: there isn't currently a way to test the google calls in firecloud-orchestration. however, I added unit tests that ensure that the system still works even if those calls to google fail (in new UserServiceSpec.scala)

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -84,6 +88,8 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
   val authScopes = Seq("profile", "email")
   // the minimal scope to read from GCS
   val storageReadOnly = Seq(StorageScopes.DEVSTORAGE_READ_ONLY)
+  // scope for creating anonymized Google groups
+  val directoryScope = Seq("https://www.googleapis.com/auth/admin.directory.group")
```
rushtong: This string is also defined in `DirectoryScopes.ADMIN_DIRECTORY_GROUP`. I think this should also work and more more ideomatic:
```suggestion
  val directoryScope = Seq(DirectoryScopes.ADMIN_DIRECTORY_GROUP)
```

src/test/scala/org/broadinstitute/dsde/firecloud/service/UserServiceSpec.scala
```diff
@@ -0,0 +1,82 @@
...
+        val rqComplete = Await.
+          result(userService.setupAnonymizedGoogleGroup(keys, anonymousGroupName), 3.seconds).
+          asInstanceOf[RequestComplete[ProfileWrapper]]
+        val returnedKeys = rqComplete.response
+        returnedKeys should equal(keys)
+      }
+    }
+  }
+
+  // getAllUserKeys, writeAnonymousGroup are tested in UserApiServiceSpec
```
rushtong: nit: rm

src/test/scala/org/broadinstitute/dsde/firecloud/service/UserServiceSpec.scala
```diff
@@ -0,0 +1,82 @@
...
+      case _ => Option(groupName)
+    }
+  }
+  override def addMemberToAnonymizedGoogleGroup(groupName: String, targetUserEmail: String): Option[String] = {
+    groupName match {
+      case "makeAddMemberFail" => Option.empty
+      case _ => Option(targetUserEmail)
+    }
+  }
+}
```
rushtong: nit: newline at end of file

rushtong (approved):
Looks good to me 👍 Some minor suggestions inline. I also recommend getting an official thumb from someone else like @MatthewBemis

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -505,6 +529,67 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
...
+    *                        step failed, return Option.empty
+    */
+  override def createGoogleGroup(groupEmail: String): Option[String] = {
+    val newGroup = new Group().setEmail(groupEmail)
+    val directoryService = getDirectoryManager(getDelegatedCredentialForAdminUser)
+
+    val insertRequest = directoryService.groups.insert(newGroup)
+    Try(executeGoogleRequest[Group](insertRequest)) match {
+      case Failure(response) => {
+        val errorCode = response.asInstanceOf[GoogleJsonResponseException].getDetails.getCode
```
MatthewBemis: Can this throw exceptions other than `GoogleJsonResponseException`? A safer way to handle exceptions may be something like:

```
case Failure(response: GoogleJsonResponseException) => {
   //your existing error handling here
}
case Failure(_) => {
  //handle a generic exception here
}
case Success(newGroupInfo) => {
  //success case here
}
```

I'm not in front of a Scala IDE right now (reviewing from personal laptop) so I can't actually try this out but it _should_ allow you to get rid of the `asInstanceOf`s which could cause problems if another exception is ever thrown.

If this works, I'd also apply it to the other areas where you're using `.asInstanceOf` in error handling (except for tests, using it there is fine).

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
```diff
@@ -188,9 +188,10 @@ trait UserApiService extends HttpService with PerRequestCreator with FireCloudRe
         pathEnd {
           get {
             requireUserInfo() { userInfo =>
-              mapRequest(addFireCloudCredentials) {
-                passthrough(UserApiService.remoteGetAllURL.format(userInfo.getUniqueId), HttpMethods.GET)
```
MatthewBemis, line 192: As a sanity check, this was originally a Thurloe passthrough but now the response is constructed by Orchestration. Are these identically structured JSON responses? I'm only asking to ensure that backwards-compatibility is maintained.

MatthewBemis (approved):
Tentatively approved pending 1 question and 1 suggestion! :+1:

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
mmorgantaylor: awesome, great catch. I made this change (awesome that it removes the need to use `asInstanceOf`!) in both `createGoogleGroup` and `addMemberToAnonymizedGoogleGroup`.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala
mmorgantaylor: yes, these are identical - the construction we're using is what was being passed through, if that makes sense. we didn't change anything about how the JSON is constructed.

mmorgantaylor:
Coveralls is unhappy because of the additional lines in HttpGoogleServicesDAO, which includes calls to Google that aren't easily testable in the current firecloud-orchestration codebase.

PR #777, opened 2020-03-05
AS-155: PFB endpoints talk to import service, not arrow [risk: low]

davidangb:
**DO NOT MERGE until ready to cut over to new import service.**

Use the new import service for create-import and get-import-status endpoints.

When this PR merges, we will no longer talk to the Arrow cloud function; we will only talk to the new import service. This means that users will not be able to query for status of Arrow-based jobs. I think this is a fair tradeoff given low usage, but feel free to push back. The alternative is to keep all the Arrow config in the codebase,  and have the status endpoint first query import service, then fall back to querying Arrow if import service returns a 404.

I've left some follow-on work for future PRs to keep this one small. Still left to do:
* new endpoint for listing import jobs in a workspace
* remove Arrow from configs in firecloud-develop
* ensure that spaces in workspace names work properly

-----

In all cases:

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -3527,6 +3527,38 @@ paths:
           description: Internal Error
 
   /api/workspaces/{workspaceNamespace}/{workspaceName}/importPFB:
+    get:
+      x-passthrough: false
+      tags:
+        - Entities
+      operationId: listImportPFBJobs
+      summary: List PFB import jobs in this workspace
+      description: >
+        Lists all imports for this workspace, optionally filtered to only those imports currently in progress
+      produces:
+        - application/json
+      parameters:
+        - $ref: '#/parameters/workspaceNamespaceParam'
+        - $ref: '#/parameters/workspaceNameParam'
+        - in: query
+          description: When true, filters to only those imports currently in progress
+          name: running_only
+          required: false
+          type: boolean
+          default: false
+      responses:
+        200:
+          description: Successful Request
+        400:
+          description: Bad Request
+        401:
+          description: Unauthorized access to Workspace
+        403:
+          description: Forbidden access to Workspace
+        404:
+          description: Workspace not found
+        500:
+          description: Internal Error
     post:
       x-passthrough: false
       tags:
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -3549,12 +3549,12 @@ paths:
       responses:
         200:
           description: Successful Request
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/PFBStatusResponse'
         400:
           description: Bad Request
-        401:
-          description: Unauthorized access to Workspace
-        403:
-          description: Forbidden access to Workspace
         404:
           description: Workspace not found
         500:
@@ -6853,15 +6853,9 @@ definitions:
       message:
         type: string
         description: extended explanation for the import job.
-      timestamp:
-        type: integer
-        format: int64
-        description: time of the response
     required:
       - jobId
       - status
-      - message
-      - timestamp
 
   Profile:
     type: object
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -3621,7 +3621,7 @@ paths:
           schema:
             $ref: '#/definitions/PFBStatusResponse'
         404:
-          description: job ID not found
+          description: workspace or job ID not found
         500:
           description: Internal Error
 
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
@@ -367,20 +367,6 @@ class EntityClient(requestContext: RequestContext, modelSchema: ModelSchema, goo
     }
   }
 
-  // ask Rawls if we have permission on this workspace. For writes, Rawls will also check if the workspace is locked.
-  private def validateUpsertPermissions(userInfo: UserInfo, workspaceNamespace: String, workspaceName: String, action: String = "read"): Future[Boolean] = {
-    val pipeline: WithTransformerConcatenation[HttpRequest, Future[HttpResponse]] = authHeaders(requestContext) ~> sendRec
-
-    def enc(s: String) = java.net.URLEncoder.encode(s, "utf-8")
-
-    val checkUrl = s"${Rawls.workspacesUrl}/${enc(workspaceNamespace)}/${enc(workspaceName)}/checkIamActionWithLock/$action"
-
-    userAuthedRequest(Get(checkUrl))(userInfo: UserInfo) map {
-      case resp if resp.status == NoContent => true
-      case resp => throw new FireCloudExceptionWithErrorReport(ErrorReport(resp.status, resp.entity.asString)) // bubble up errors
-    }
-  }
-
   def importPFB(workspaceNamespace: String, workspaceName: String, pfbRequest: PfbImportRequest, userInfo: UserInfo): Future[PerRequestMessage] = {
 
     def enc(str: String) = java.net.URLEncoder.encode(str, "utf-8")
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
@@ -371,16 +371,16 @@ class EntityClient(requestContext: RequestContext, modelSchema: ModelSchema, goo
   }
 
   // ask Rawls if we have permission on this workspace. For writes, Rawls will also check if the workspace is locked.
-  private def validateUpsertPermissions(userInfo: UserInfo, workspaceNamespace: String, workspaceName: String, action: String = "read")(op: Boolean => Future[PerRequestMessage]) = {
+  private def validateUpsertPermissions(userInfo: UserInfo, workspaceNamespace: String, workspaceName: String, action: String = "read"): Future[Boolean] = {
     val pipeline: WithTransformerConcatenation[HttpRequest, Future[HttpResponse]] = authHeaders(requestContext) ~> sendRec
 
     def enc(s: String) = java.net.URLEncoder.encode(s, "utf-8")
 
     val checkUrl = s"${Rawls.workspacesUrl}/${enc(workspaceNamespace)}/${enc(workspaceName)}/checkIamActionWithLock/$action"
 
-    userAuthedRequest(Get(checkUrl))(userInfo: UserInfo) flatMap {
-      case resp if resp.status == NoContent => op(true)
-      case resp => Future(RequestComplete(resp)) // bubble up errors
+    userAuthedRequest(Get(checkUrl))(userInfo: UserInfo) map {
+      case resp if resp.status == NoContent => true
+      case resp => throw new FireCloudExceptionWithErrorReport(ErrorReport(resp.status, resp.entity.asString)) // bubble up errors
     }
   }
 
@@ -406,7 +406,7 @@ class EntityClient(requestContext: RequestContext, modelSchema: ModelSchema, goo
     // validate presigned URL
     validatePfbUrl(pfbRequest) { pfbUrl =>
       // verify workspace existence and permissions
-      validateUpsertPermissions(userInfo, workspaceNamespace, workspaceName, "write") { _ =>
+      validateUpsertPermissions(userInfo, workspaceNamespace, workspaceName, "write") flatMap { _ =>
         // the payload to Import Service sends the path
         val importServicePayload: ImportServiceRequest = ImportServiceRequest(path = pfbUrl.toString, filetype = "pfb")
         // TODO: AS-155: always encode namespace/name
@@ -426,7 +426,7 @@ class EntityClient(requestContext: RequestContext, modelSchema: ModelSchema, goo
 
   def pfbImportStatus(workspaceNamespace: String, workspaceName: String, jobId: String, userInfo: UserInfo): Future[PerRequestMessage] = {
 
-    validateUpsertPermissions(userInfo, workspaceNamespace, workspaceName) { _ =>
+    validateUpsertPermissions(userInfo, workspaceNamespace, workspaceName) flatMap { _ =>
       val importServiceUrl = s"${FireCloudConfig.ImportService.server}/$workspaceNamespace/$workspaceName/imports/$jobId"
 
       userAuthedRequest(Get(importServiceUrl))(userInfo) map {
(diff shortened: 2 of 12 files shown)


davidangb:
travis and jenkins both had network hiccups. jenkins retest.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -385,112 +405,47 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema, go
...
+      validateUpsertPermissions(userInfo, workspaceNamespace, workspaceName, "write") { _ =>
+        // the payload to Import Service sends the path
+        val importServicePayload: ImportServiceRequest = ImportServiceRequest(path = pfbUrl.toString, filetype = "pfb")
+        // TODO: AS-155: always encode namespace/name
+        val importServiceUrl = s"${FireCloudConfig.ImportService.server}/$workspaceNamespace/$workspaceName/imports"
+
+        userAuthedRequest(Post(importServiceUrl, importServicePayload))(userInfo) map {
+          case resp if resp.status == Created =>
+            val importServiceResponse = unmarshal[ImportServiceResponse].apply(resp)
+            // for backwards compatibility, we return Accepted(202), not Created(201)
```
helgridly: also because it's still correct :)

src/main/scala/org/broadinstitute/dsde/firecloud/model/Workspace.scala
```diff
@@ -34,6 +34,14 @@ case class PfbImportRequest(
   workspace: Option[WorkspaceName] = None,
   user: Option[WorkbenchUserInfo] = None)
 
+case class ImportServiceRequest(
+  path: String,
+  filetype: String)
+
+case class ImportServiceResponse(
+  id: String,
+  status: String)
```
helgridly: I think it's probably an oversight in import service if we're not also returning an error message if one exists.

src/test/scala/org/broadinstitute/dsde/firecloud/mock/MockUtils.scala
```diff
@@ -22,7 +22,7 @@ object MockUtils {
   val cromiamServerPort = 8995
   val searchServerPort = 9292
   val bagitServerPort = 9393
-  val arrowServerPort = 9394
+  val importServerPort = 9394
```
helgridly: nit: `importServicePort`? `importServiceServerPort`?

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
davidangb: yeah - I guess I'm distinguishing here that import service returns 201, but orch returns  202

src/test/scala/org/broadinstitute/dsde/firecloud/mock/MockUtils.scala
davidangb: I had this exact conversation inside my brain

davidangb:
one more time, hopefully sbt issues are gone now: jenkins retest

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -366,6 +370,22 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema, go
     }
   }
 
+  // ask Rawls if we have permission on this workspace. For writes, Rawls will also check if the workspace is locked.
+  private def validateUpsertPermissions(userInfo: UserInfo, workspaceNamespace: String, workspaceName: String, action: String = "read")(op: Boolean => Future[PerRequestMessage]) = {
```
helgridly: do we have a good reason for this function being a loan pattern instead of just returning a future?

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -366,6 +370,22 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema, go
...
+    val pipeline: WithTransformerConcatenation[HttpRequest, Future[HttpResponse]] = authHeaders(requestContext) ~> sendRec
+
+    def enc(s: String) = java.net.URLEncoder.encode(s, "utf-8")
+
+    val checkUrl = s"${Rawls.workspacesUrl}/${enc(workspaceNamespace)}/${enc(workspaceName)}/checkIamActionWithLock/$action"
+
+    userAuthedRequest(Get(checkUrl))(userInfo: UserInfo) flatMap {
+      case resp if resp.status == NoContent => op(true)
+      case resp =>
+        logger.warn(s"****** workspace permissions failed with ${resp.status.intValue} ${resp.status.defaultMessage}: ${resp}")
```
helgridly: nit: why all the asterisks? doesn't this just lead to asterisk escalation in logs?

src/main/scala/org/broadinstitute/dsde/firecloud/model/ModelJsonProtocol.scala
```diff
@@ -228,6 +228,8 @@ object ModelJsonProtocol extends WorkspaceJsonSupport {
 
   implicit val impBagitImportRequest = jsonFormat2(BagitImportRequest)
   implicit val impPfbImportRequest = jsonFormat4(PfbImportRequest)
+  implicit val impImportServiceRequest = jsonFormat2(ImportServiceRequest)
+  implicit val impImportServiceResponse = jsonFormat2(ImportServiceResponse)
```
helgridly: `impOrtServiceResponse` :laughing:

helgridly (approved):
i don't see any bugs. in your shoes i'd update import service to add the error message to responses and fix the loan pattern as part of this PR but as long as they both get done soon i don't mind

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
davidangb: thank you for this catch; I hadn't meant to leave that debugging log in.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -385,112 +403,47 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema, go
...
+            // for backwards compatibility, we return Accepted(202), even though import service returns Created(201)
+            RequestComplete(Accepted, importServiceResponse)
+          case otherResp =>
+            RequestCompleteWithErrorReport(otherResp.status, otherResp.toString)
+        }
       }
     }
   }
 
   def pfbImportStatus(workspaceNamespace: String, workspaceName: String, jobId: String, userInfo: UserInfo): Future[PerRequestMessage] = {
```
ansingh7115: Currently, in order to get the status of a job, we need the jobId. It might be cool to be able to just ask what the statuses of all jobs in my workspace are, and only have to pass in the namespace and name of the workspace. This is not for this PR, of course, just something we could consider adding in the future.

davidangb: yes absolutely! That exists in import service, and is in my list of TODOs for "new endpoint for listing import jobs in a workspace".

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -135,9 +135,10 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
       .refreshAccessToken().getTokenValue
   }
 
+  @deprecated("currently unused and targeted for deletion", "2020-03-06")
   override def getAdminIdentityToken: String = {
     getScopedServiceAccountCredentials(firecloudAdminSACreds, Seq("openid", "email"))
-      .idTokenWithAudience(EntityClient.avroToRawlsURL, List()).getTokenValue
```
ansingh7115, line 140: Is it necessary to deprecate this - can we just delete it if its not used anywhere?

ansingh7115 (approved):
Looks pretty good!

src/main/resources/swagger/api-docs.yaml
```diff
@@ -6811,15 +6843,9 @@ definitions:
...
         description: extended explanation for the import job.
-      timestamp:
-        type: integer
-        format: int64
-        description: time of the response
     required:
       - jobId
       - status
-      - message
-      - timestamp
```
davidangb, line 6832: message is optional; timestamp was never part of the payload. These changes bring the swagger def in line with reality; I am not changing any behavior here.

src/main/scala/org/broadinstitute/dsde/firecloud/model/Workspace.scala
```diff
@@ -29,11 +29,20 @@ case class EntityId(entityType: String, entityName: String)
 case class BagitImportRequest(bagitURL: String, format: String)
 
 case class PfbImportRequest(
-  url: String,
+  url: Option[String],
```
davidangb: making this an option allows us to be lenient during json deserialization of end-user request bodies, and delegate validation to import service.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -368,129 +369,30 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema, go
...
+          workspace = Some(WorkspaceName(workspaceNamespace, workspaceName))
+        )
 
-        // return Accepted with UUID
         RequestComplete(Accepted, responsePayload)
-      }
-    }
-  }
-
-  def pfbImportStatus(workspaceNamespace: String, workspaceName: String, jobId: String, userInfo: UserInfo): Future[PerRequestMessage] = {
```
davidangb, line 430: get-status is now a passthrough, implemented at the WorkspaceApiService layer.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -368,129 +369,30 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema, go
 
   def importPFB(workspaceNamespace: String, workspaceName: String, pfbRequest: PfbImportRequest, userInfo: UserInfo): Future[PerRequestMessage] = {
```
davidangb, line 370: importPFB is now simplified.  It delegates validation and permission-checking to the import service.

src/test/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceApiServiceSpec.scala
```diff
@@ -1013,194 +1013,196 @@ class WorkspaceApiServiceSpec extends BaseServiceSpec with WorkspaceApiService w
...
-          })
-      }
-
-      "should 400 if PFB URL is not https" in {
-        (Post(pfbImportPath, HttpEntity(MediaTypes.`application/json`, s"""{"url":"http://missing.avro"}"""))
-          ~> dummyUserIdHeaders(dummyUserId)
-          ~> sealRoute(workspaceRoutes)) ~> check {
-            status should equal(BadRequest)
-          }
-      }
```
davidangb, line 1031: removed some tests because we now delegate this logic to the import service; instead, we have a test to ensure we bubble up whatever errors the import service might return.

davidangb:
@ansingh7115 @helgridly re-review please, in light of broadinstitute/import-service#30

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -368,129 +369,30 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema, go
...
-        // fire-and-forget call Arrow with UUID, workspace info, and presigned URL
-        val idToken = googleServicesDAO.getAdminIdentityToken
-        val gzipPipeline = addHeader (`Accept-Encoding`(gzip)) ~> addCredentials(OAuth2BearerToken(idToken)) ~> sendReceive ~> decode(Gzip)
-        gzipPipeline { Post(avroToRawlsURL, arrowPayload) }
-
-        // the response payload to the end user omits the userid/email
-        val responsePayload = arrowPayload.copy(user = None)
+        val responsePayload:PfbImportRequest = pfbRequest.copy(
+          jobId = Some(importServiceResponse.jobId),
+          user = None,
```
helgridly: what's the `user` parameter expected to be here?

src/main/scala/org/broadinstitute/dsde/firecloud/model/Workspace.scala
helgridly: why are you using the same class for an incoming request and the response back to the user? an incoming request doesn't have a job id, and a status response doesn't have a url.

helgridly: i appreciate that this code was here before. i only just noticed it :p

helgridly (commented):
i am confused about `PfbImportRequest` (DM is good)

davidangb:
@helgridly I've split up/cleaned up the `PFBImportRequest` model object, believe it addresses all your  feedback.

helgridly (approved):

src/main/scala/org/broadinstitute/dsde/firecloud/EntityClient.scala
```diff
@@ -368,129 +369,30 @@ class EntityClient (requestContext: RequestContext, modelSchema: ModelSchema, go
...
-        throw new FireCloudExceptionWithErrorReport(ErrorReport(BadRequest, "Invalid URL protocol: must be https only"))
-      }
+    def enc(str: String) = java.net.URLEncoder.encode(str, "utf-8")
 
-      // TODO: add real (semantic) validation for origins, to ensure caller isn't supplying malice
-      op(pfbUrl)
-    }
+    // the payload to Import Service sends "path" and filetype.  Here, we force-hardcode filetype because this API
+    // should only be used for PFBs.
+    val importServicePayload: ImportServiceRequest = ImportServiceRequest(path = pfbRequest.url.getOrElse(""), filetype = "pfb")
```
ansingh7115, line 376: It feels weird to submit an empty string for the url, knowing it will fail - but I also understand that you're pushing all validation down to import service

davidangb: I could make it an `Option` and omit sending the entire key ... would that be better?

ansingh7115: up to you - this is fine too!

davidangb:
unrelated OOM on integration tests, so: jenkins retest

PR #783, opened 2020-04-30
AS-206: Data Repo Integration [Feature Branch]

MatthewBemis:
This is the feature branch for App Services' Data Repo Integration epic (AS-206).

This is a work in progress. Most (if not all) Data Repo integration work should branch off of this branch.

Corresponding feature branches:

rawls: `FEATURE-as-data-repo-integration`
firecloud-develop: TBD

PR #785, opened 2020-05-14
AS-273 Add a passthrough for listSnapshots endpoint in Rawls

zloery:

In all cases:

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -4106,6 +4106,47 @@ paths:
         - Snapshots
       summary: Add a reference to a snapshot in the Terra Data Repo
       operationId: createDataRepoSnapshot
+    get:
+      x-passthrough: true
+      x-passthrough-target: rawls
+      responses:
+        '200':
+          description: OK
+          schema:
+            $ref: '#/definitions/DataRepoSnapshotList'
+        '404':
+          description: Workspace not found or user lacks permissions
+          schema:
+            $ref: '#/definitions/ErrorReport'
+        '500':
+          description: Internal Error
+          schema:
+            $ref: '#/definitions/ErrorReport'
+      parameters:
+      - in: path
+        description: The workspace namespace
+        name: namespace
+        required: true
+        type: string
+      - in: path
+        description: The workspace name
+        name: name
+        required: true
+        type: string
+      - in: query
+        name: offset
+        description: The number of items to skip before starting to collect the result set.
+        type: int
+        required: true
+      - in: query
+        name: limit
+        description: The number of items to return.
+        type: int
+        required: true
+      tags:
+      - Snapshots
+      summary: List references to snapshots in the Terra Data Repo
+      operationId: listDataRepoSnapshot
 
   /api/workspaces/{workspaceNamespace}/{workspaceName}/snapshots/{snapshotId}:
     get:
@@ -7662,6 +7703,15 @@ definitions:
         type: object
       cloningInstructions:
         type: string
+  DataRepoSnapshotReferenceList:
+    description: A list of DataRepoSnapshotReferences
+    properties:
+      resources:
+        description: A list of Data Reference definitions from Workspace Manager
+        type: array
+        items:
+          $ref: '#/definitions/DataRepoSnapshotReferences'
+
 
   Workflow:
     description: ''
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -4123,16 +4123,8 @@ paths:
           schema:
             $ref: '#/definitions/ErrorReport'
       parameters:
-      - in: path
-        description: The workspace namespace
-        name: namespace
-        required: true
-        type: string
-      - in: path
-        description: The workspace name
-        name: name
-        required: true
-        type: string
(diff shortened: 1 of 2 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/webservice/SnapshotApiService.scala
```diff
@@ -19,7 +19,10 @@ trait SnapshotApiService extends FireCloudDirectives with UserInfoDirectives {
       pathPrefix("workspaces" / Segment / Segment / "snapshots") { (namespace, name) =>
         pathEnd {
           post {
-            passthrough(SnapshotApiService.createDataRepoSnapshotURL(namespace, name), HttpMethods.POST)
+            passthrough(SnapshotApiService.dataRepoSnapshotBaseURL(namespace, name), HttpMethods.POST)
+          } ~
+          get {
+            passthrough(SnapshotApiService.dataRepoSnapshotBaseURL(namespace, name), HttpMethods.GET)
```
MatthewBemis: I don't think the `limit` and `offset` query params are getting passed through to rawls here

zloery: whoops, good catch!

MatthewBemis (approved):
lgtm

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/SnapshotApiService.scala
```diff
@@ -20,6 +21,11 @@ trait SnapshotApiService extends FireCloudDirectives with UserInfoDirectives {
         pathEnd {
           post {
             passthrough(SnapshotApiService.createDataRepoSnapshotURL(namespace, name), HttpMethods.POST)
+          } ~
+          get {
+            parameters("offset", "limit") { (offset, limit) =>
+              passthrough(SnapshotApiService.listDataRepoSnapshotURL(namespace, name, offset, limit), HttpMethods.GET)
+            }
```
davidangb, line 28: for passthrough apis with a querystring, we typically handle it by passing the querystring as-is. See https://github.com/broadinstitute/firecloud-orchestration/blob/develop/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiService.scala#L61 for an example.

Your solution would work too; this is more about consistency in the codebase. If there is a reason to extract/verify only specific params, I'd use your solution; passing it through untouched means we can add/change params in Rawls without needing changes here in orch.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -4106,6 +4106,47 @@ paths:
...
+      - in: path
+        description: The workspace namespace
+        name: namespace
+        required: true
+        type: string
+      - in: path
+        description: The workspace name
+        name: name
+        required: true
+        type: string
```
davidangb: for namespace and name, use:
```
      parameters:
```
to normalize the params and cut down on lines of code. See the create-snapshot definition just previous to this as an example.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/SnapshotApiService.scala
zloery: SGTM - it's better to pass it through untouched in this case

davidangb: this should not pass through to `FireCloudConfig.Rawls.baseUrl` - it should pass through to `listDataRepoSnapshotURL(namespace, name)`, and `listDataRepoSnapshotURL` should be ignorant of any querystring.

davidangb (commented):
one change needed inline.

Also - do not merge, right? If we merge this it'll make it to prod, and we don't yet have an enable/disable flag for these apis in orchestration.

zloery:
Now that you mention it, it might be important to actually call the right Rawls endpoint...
Also, this is merging to a feature branch so it shouldn't get to production until we merge the [whole branch](https://github.com/broadinstitute/firecloud-orchestration/pull/783), right?

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/SnapshotApiService.scala
```diff
@@ -20,6 +21,13 @@ trait SnapshotApiService extends FireCloudDirectives with UserInfoDirectives {
         pathEnd {
           post {
             passthrough(SnapshotApiService.createDataRepoSnapshotURL(namespace, name), HttpMethods.POST)
+          } ~
+          get {
+            pathPrefix("workspaces" / Segment / Segment / "snapshots") { (namespace, name) =>
```
davidangb: sorry, one more tweak - need to remove the `pathPrefix("workspaces" / Segment / Segment / "snapshots")` here. This is already inside a `pathPrefix("workspaces" / Segment / Segment / "snapshots")` at line 20. All the nested path directives (including the `pathEnd` at line 21) combine to define the path - this extra line makes it invalid!

zloery: good catch!

davidangb (approved):
:+1: ... and you're totally right, this is merging to a feature branch, I retract my previous fear about merging this!

PR #786, opened 2020-05-22
[CA-854] url encode emails when making requests to Sam with email in path

marctalbott:
Ticket: [CA-854](https://broadworkbench.atlassian.net/browse/CA-854)

Orch should be url encoding email addresses when it makes calls to Sam that include the email in the url. This is causing problems for a user with a % in their email address. I'm not actually sure that this code is even used, but Orch should be doing it no matter what.

In all cases:


andy7i (approved):
same question as this: https://github.com/broadinstitute/rawls/pull/1217#discussion_r429328574

otherwise 👍

dvoet (approved):

PR #787, opened 2020-05-27
AS-363: upgrade to ES client 5.6.16 [risk: low]

davidangb:
the diff on this looks somewhat large, but a lot of it is renaming/reorganizing. Try reviewing commit-by-commit for an easier view.

This PR does 2 things:
1. Resolves sbt warnings about dependency version conflicts, by applying exclude directives
2. Upgrades the elasticsearch transport client to 5.6.16. Anything after 5.6.1 includes an upgrade to Lucene QueryParsers 6.6.1 (https://www.elastic.co/guide/en/elasticsearch/reference/5.6/release-notes-5.6.1.html). In turn, this allowed me to override the transitive dependency and upgrade to Lucene QueryParsers 6.6.2, which resolves a proactive OSWAP security warning.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -9,12 +9,14 @@ object Dependencies {
   def excludeGuavaJDK5(m: ModuleID): ModuleID = m.exclude("com.google.guava", "guava-jdk5")
 
   val rootDependencies = Seq(
-    // proactively pull in latest versions of Jackson libs and Netty codec, instead of relying on the versions
+    // proactively pull in latest versions of these libraries, instead of relying on the versions
     // specified as transitive dependencies, due to OWASP DependencyCheck warnings for earlier versions.
     "com.fasterxml.jackson.core"     % "jackson-annotations" % jacksonV,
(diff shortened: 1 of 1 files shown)


project/Dependencies.scala
```diff
@@ -39,16 +44,19 @@ object Dependencies {
...
 
-    "org.elasticsearch.client"       % "transport"           % "5.4.3",
-
-    "com.google.guava"               % "guava"               % "28.1-android",
+    "org.elasticsearch.client"       % "transport"           % "5.6.16"
+      exclude("io.netty", "netty-codec")
+      exclude("io.netty", "netty-transport")
+      exclude("io.netty", "netty-resolver")
+      exclude("io.netty", "netty-buffer")
+      exclude("io.netty", "netty-common"),
```
davidangb, line 52: I tried an `excludeAll(ExclusionRule(organization = "io.netty"))`, but that caused problems at runtime - there's some netty library in there we DON'T want to exclude.

MatthewBemis (approved):

zarsky-broad (approved):
Wasn't sure how best to test locally, the fact that the CI/Jenkins tests passed seems like a good enough indicator to me.

PR #793, opened 2020-07-22
AS-439: fix file upload/download definitions [risk: no]

davidangb:
This fixes file uploads, and cleans up some file downloads, in swagger-ui.

Example of file upload *before* this PR:
before-upload

and *after* this PR:
after-upload


zarsky-broad (approved):
Passes validation and makes sense to me.

zloery (approved):
LGTM

PR #797, opened 2020-08-26
AS-465: integration test for security-related http headers

davidangb:
followup to broadinstitute/firecloud-develop#2251, which addresses [AS-465](https://broadworkbench.atlassian.net/browse/AS-465).

Reviewer(s):
* are these tests worth it? We'd have to add these tests to every service I updated (i.e. agora, orchestration, rawls, sam, thurloe, leo, job manager). Even then, we could really only test at the load balancer level, and not test each individual instance. Note that our security scans will scan every instance as well as the load balancers, so our scans and our tests would not match.
* do our security scans equate to test coverage? They run weekly, and will notify us if these headers are missing. They scan production, so if we rely solely on scans and there is any regression to headers, we'd be reactive instead of proactive; the regression would have already made it to prod.

My opinion is we should just rely on the security scans, not add integration tests, as the cost:benefit of integration tests is not worth it.

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/HttpHeaderSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/HttpHeaderSpec.scala
@@ -1,0 +1,58 @@
+package org.broadinstitute.dsde.test.api.orch
+
+import akka.http.scaladsl.model.HttpHeader
+import org.broadinstitute.dsde.test.OrchConfig.Users
+import org.broadinstitute.dsde.workbench.auth.AuthToken
+import org.broadinstitute.dsde.workbench.service.Orchestration
+import org.broadinstitute.dsde.workbench.config.{Credentials, UserPool}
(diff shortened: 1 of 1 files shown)


davidangb:
@rtitle @ansingh7115 @s-rubenstein as reviewers of the original PR, do you have opinions on the questions posed in the description here? @emandakh what about you?

ansingh7115:
I'm fine relying on the scans, if these tests aren't that useful. 
The likelihood of headers being missing seems low

emandakh:
Agreed, I think leaving it to security scans is fine.  That process knows more about the vulnerabilities than the hard-coded test.  There might be need to change the headers legitimately (in still secure ways) in the future and we don't want that to fail the tests.

emandakh:
Also, in the future, we'll look into adding the security scans into test environments as part of the general release testing suite, so it's not a reactionary measure anymore.

davidangb:
closing; we'll rely on scans for this instead of integration tests.

PR #799, opened 2020-09-09
AoU RW-3627: Expose Rawls DELETE /api/user/billing/{projectName}

jmthibault79:
AoU RW wants to do a better job of cleaning up our resources.  We have a one-workspace-per-project paradigm, so it's safe to delete our projects after we delete the associated workspace.  Until now, we have used a batch cleanup job to do this.

With https://precisionmedicineinitiative.atlassian.net/browse/RW-3627 (please let me know if you need access) we will now be calling the Terra delete-project endpoint after calling Terra's delete-workspace whenever we delete an AoU RW workspace.  We prefer to call FC-Orch instead of Rawls directly for abstraction reasons - we don't currently make any Rawls calls.

This PR adds the Rawls call we need (implemented for https://broadworkbench.atlassian.net/browse/CA-540) to Orchestration.

Tested on a FIAB:
* created a few billing projects in FC-UI
* hit the Rawls endpoint with the same user to delete one
* hit the new Orch endpoint with the same user to delete another
* observed that the billing projects were no longer present in FC-UI 

In all cases:

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -329,8 +329,7 @@ paths:
   /api/user/billing/{projectName}:
     delete:
       tags:
-        - billing
-        - users
+        - Billing
       summary: delete billing project
       description: delete billing project
       operationId: deleteBillingProject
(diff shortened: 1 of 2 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -326,6 +326,47 @@ paths:
             - profile
       x-passthrough: true
       x-passthrough-target: rawls
+  /api/user/billing/{projectName}:
```
jmthibault79, line 329: Copied from Rawls

MatthewBemis (approved):

PR #414, opened 2017-03-27
GAWB-1772: use spray host connector for ontology, which gives us a throttle

davidangb:
* allow spray-can client [host-level API](http://spray.io/documentation/1.2.4/spray-can/http-client/host-level/#hostlevelapi) in RestJsonClient;
* reduce logging levels through the indexing process;
* remove deprecated blocking-API call in HttpOntologyDAO;
* add unit test

I've run the unit test a bunch of times and it's been stable, but I am slightly worried that Travis will be slow or wonky and we'll see flaky results.

NOTE: github's diff is uglier than reality - there's a lot of small changes in here but github makes it look bad.

Also see a few code comments inline.

- [x] **Submitter**: Include the JIRA issue number in the PR description
- [x] **Submitter**: Make sure Swagger is updated if API changes
- [x] **Submitter**: If updating admin endpoints, also update [firecloud-admin-cli](https://github.com/broadinstitute/firecloud-admin-cli)
- [x] **Submitter**: Check documentation and code comments. Add explanatory PR comments if helpful.
- [x] **Submitter**: JIRA ticket checks:
  * Acceptance criteria exists and is met
  * Note any changes to implementation from the description
  * Add notes on what you've tested
- [x] **Submitter**: Update RC_XXX release ticket with any config or environment changes necessary
- [x] **Submitter**: Update FISMA documentation if changes to:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
- [x] **Submitter**: If you're adding new libraries, sign us up to security updates for them
- [x] Tell the tech lead (TL) that the PR exists if they wants to look at it
- [x] Anoint a lead reviewer (LR). **Assign PR to LR**
* Review cycle:
  * LR reviews
  * Rest of team may comment on PR at will
  * **LR assigns to submitter** for feedback fixes
  * Submitter rebases to develop again if necessary
  * Submitter makes further commits. DO NOT SQUASH
  * Submitter updates documentation as needed
  * Submitter **reassigns to LR** for further feedback
- [x] **TL** sign off
- [x] **LR** sign off
- [x] **Product Owner** sign off
- [x] **Assign to submitter** to finalize
- [x] **Submitter**: Verify all tests go green, including CI tests
- [x] **Submitter**: Squash commits and merge to develop
- [x] **Submitter**: Delete branch after merge
- [x] **Submitter**: **Test this change works on dev environment after deployment**. YOU own getting it fixed if dev isn't working for ANY reason!
[...]

PR #802, opened 2020-09-23
AS-120: Spray -> akka-http migration

MatthewBemis:
This is pretty much complete from my side, with a few small pieces still to come:

* ~~15 unit test failures. Most of these I know the source of and will fix them. There are a couple that baffle me.~~
* ~~A few open questions in the comments about whether or not I should push this PR even further and do even more Api/Service/DAO layer separation. I'm inclined to clean them up in a follow-up and call them out-of-scope for this PR. This may mean that I leave DsdeHttpDAO in until that happens.~~
* ~~Need to re-implement [appendTimestampOnFailure](https://github.com/broadinstitute/firecloud-orchestration/pull/802/files#diff-4b37a7c9e1a1cd38d6533a7fa935972f76233a5e86fe8b4f9895700f27678898R119) and get [logRequestResult](https://github.com/broadinstitute/firecloud-orchestration/pull/802/files#diff-4b37a7c9e1a1cd38d6533a7fa935972f76233a5e86fe8b4f9895700f27678898R94) working without consuming the entity more than once. These would result in the passthrough directive breaking.~~
* ~~A few commented out lines or TODOs here and there, I don't think there are very many left...~~
* ~~Need to rebase. Saving this until later, there's not really much that currently concerns me with the conflicting files.~~
* ~~After unit tests succeed, need to run this through integration tests and I'd like to click around a lot on my own (I've already started the clicking around Swagger part).~~

Resources consulted for this migration:

https://github.com/broadinstitute/rawls/pull/860
https://doc.akka.io/docs/akka-http/current/migration-guide/migration-from-spray.html
https://engineering.zalando.com/posts/2016/05/migrating-spray-akka-http.html

-------------

In all cases:


project/Dependencies.scala
```diff
@@ -25,19 +27,21 @@ object Dependencies {
     "com.getsentry.raven"            % "raven-logback"       % "7.8.6",
     "com.typesafe.scala-logging"    %% "scala-logging"       % "3.7.2",
 
-    "org.broadinstitute.dsde.vault" %% "vault-common"        % "0.1-19-ca8b927"
-      exclude("io.spray", "spray-routing_2.11"),
+    "org.parboiled" % "parboiled-core" % "1.2.0",
```
MatthewBemis, line 30: this used to be a part of `vault-common`

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudApiService.scala
```diff
@@ -0,0 +1,158 @@
...
+import org.broadinstitute.dsde.firecloud.model.{ModelSchema, UserInfo, WithAccessToken}
+import org.broadinstitute.dsde.firecloud.service._
+import org.broadinstitute.dsde.firecloud.utils.StandardUserInfoDirectives
+import org.broadinstitute.dsde.firecloud.webservice._
+import org.slf4j.LoggerFactory
+
+import scala.concurrent.ExecutionContext
+import scala.language.postfixOps
+
+trait FireCloudApiService extends CookieAuthedApiService
```
MatthewBemis, line 41: I'm migrating this to work more like how rawls works. It's not done, and I still need to break out some of the monitors which are commented out below

src/main/scala/org/broadinstitute/dsde/firecloud/core/GetEntitiesWithType.scala
```diff
@@ -33,6 +38,7 @@ object GetEntitiesWithType {
 class GetEntitiesWithTypeActor(requestContext: RequestContext) extends Actor with FireCloudRequestBuilding {
```
MatthewBemis: As noted in the PR description, I haven't really attempted to rewrite this yet

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/SwaggerApiService.scala
```diff
@@ -0,0 +1,64 @@
...
+
+import akka.http.scaladsl.model.{ContentTypes, HttpEntity}
+import akka.http.scaladsl.server
+import akka.http.scaladsl.server.Directives._
+import akka.stream.scaladsl.Flow
+import akka.util.ByteString
+import com.typesafe.scalalogging.LazyLogging
+import org.broadinstitute.dsde.firecloud.FireCloudConfig
+
+trait SwaggerApiService extends LazyLogging {
```
MatthewBemis, line 12: I broke all of this out from the old boot code, following rawls as a guide

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -79,7 +87,9 @@ object GooglePriceListJsonProtocol extends DefaultJsonProtocol {
 }
 import org.broadinstitute.dsde.firecloud.dataaccess.GooglePriceListJsonProtocol._
 
-object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuilding with LazyLogging {
+object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuilding with LazyLogging with DsdeHttpDAO with SprayJsonSupport {
```
MatthewBemis: this file is still a bit rough. there's some actor stuff I need to work out before it's finished, but it's mostly done

project/Dependencies.scala
```diff
@@ -2,11 +2,13 @@ import sbt._
 
 object Dependencies {
   val akkaV = "2.4.19"
-  val sprayV = "1.3.4"
+  val akkaHttpV = "10.1.12"
```
davidangb, line 5: ah, I was curious why you didn't go to akka-http 10.2.x ... and realize that 10.2 would require Scala 2.12.

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudApiService.scala
```diff
@@ -0,0 +1,158 @@
...
+        val contentType = response.header[HttpHeader.`Content-Type`].map{_.contentType}.getOrElse(ContentTypes.`application/json`)
+        response.withEntity(HttpEntity(contentType, withTimestamp.toJson.prettyPrint + "\n"))
+      } catch {
+        // usually a failure to parse, if the response isn't JSON (e.g. HTML responses from Google)
+        case e: Exception => response
+      }
+    } else response
+  }
+
+  // wraps route rejections in an ErrorReport
```
davidangb: comment about nothing?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpLogitDAO.scala
```diff
@@ -30,12 +33,14 @@ class HttpLogitDAO(logitUrl: String, logitApiKey: String)
...
-
-    pipeline map { logitResponse =>
+    val httpRequest = HttpRequest(
+      method = HttpMethods.PUT,
+      uri = logitUrl,
+      headers = List(RawHeader("LogType", Metrics.LOG_TYPE), RawHeader("ApiKey", logitApiKey)),
+      entity = metric
+    )
+
+    http.singleRequest(httpRequest) map { logitResponse =>
```
davidangb: can you use `executeRequestRaw` here?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
```diff
@@ -48,7 +51,7 @@ class HttpRawlsDAO( implicit val system: ActorSystem, implicit val executionCont
   }
 
   override def isLibraryCurator(userInfo: UserInfo): Future[Boolean] = {
-    userAuthedRequest( Get(rawlsCuratorUrl) )(userInfo) map { response =>
+    userAuthedRequest(Get(rawlsCuratorUrl))(userInfo) map { response =>
```
davidangb: yeah, I definitely see the inconsistency/overlap between the `RestJsonClient` and `DsdeHttpDAO` traits

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
```diff
@@ -138,20 +143,8 @@ class HttpRawlsDAO( implicit val system: ActorSystem, implicit val executionCont
     authedRequestToObject[Seq[AgoraConfigurationShort]](Get(rawlsWorkspaceMethodConfigsUrl(ns, name)), true)
   }
 
-  override def createProject(projectName: String, billingAccount: String)(implicit userToken: WithAccessToken): Future[Boolean] = {
-    val create = CreateRawlsBillingProjectFullRequest(projectName, billingAccount)
-    userAuthedRequest(Post(FireCloudConfig.Rawls.authUrl + "/billing", create)).map { resp =>
-      resp.status.isSuccess
-    }
-  }
-
```
davidangb, line 147: did `createProject` disappear?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpSamDAO.scala
```diff
@@ -119,12 +128,14 @@ class HttpSamDAO( implicit val system: ActorSystem, implicit val executionContex
     }
   }
 
+  //TODO: do I really need to break out the singleRequest for this?
```
davidangb: would `executeRequest[HttpResponse]` work here, instead of `http.singleRequest`?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/RawlsDAO.scala
```diff
@@ -83,7 +83,6 @@ trait RawlsDAO extends LazyLogging with ReportsSubsystemStatus {
 
   def deleteWorkspace(workspaceNamespace: String, workspaceName: String)(implicit userToken: WithAccessToken): Future[WorkspaceDeleteResponse]
 
-  def createProject(projectName: String, billingAccount: String)(implicit userToken: WithAccessToken): Future[Boolean]
```
davidangb, line 86: repeating my previous comment - was this unused?

src/main/scala/org/broadinstitute/dsde/firecloud/model/ErrorReport.scala
```diff
@@ -38,16 +42,3 @@ object RequestCompleteWithErrorReport {
...
-  def apply(statusCode: StatusCode, message: String) =
-    HttpResponse(statusCode, ErrorReport(statusCode, message).toJson.compactPrint)
-
-  def apply(statusCode: StatusCode, throwable: Throwable) =
-    HttpResponse(statusCode, ErrorReport(statusCode, throwable).toJson.compactPrint)
-
-  def apply(statusCode: StatusCode, message: String, throwable: Throwable) =
-    HttpResponse(statusCode, ErrorReport(statusCode, message, throwable).toJson.compactPrint)
-
-}
```
davidangb, line 53: oh huh, we never used these?

src/main/scala/org/broadinstitute/dsde/firecloud/service/EntityService.scala
```diff
@@ -51,7 +52,10 @@ trait EntityService extends HttpService with PerRequestCreator with FireCloudDir
                             entityType = copyRequest.entityType,
                             entityNames = copyRequest.entityNames)
                           val extReq = Post(FireCloudConfig.Rawls.workspacesEntitiesCopyUrl(linkExistingEntitiesBool), copyMethodConfig)
-                          externalHttpPerRequest(requestContext, extReq)
+
+                          executeRequestRaw(userInfo.accessToken)(extReq).flatMap { resp =>
+                            requestContext.complete(resp)
+                          }
```
davidangb: can this construct simplify to just `executeRequestRaw(userInfo.accessToken)(extReq)`, i.e. omit the `requestContext.complete`? (you may also have to omit the `requestContext =>` prior to it)

src/main/scala/org/broadinstitute/dsde/firecloud/service/EntityService.scala
```diff
@@ -86,13 +90,16 @@ trait EntityService extends HttpService with PerRequestCreator with FireCloudDir
...
                   val entityQueryUri = FireCloudConfig.Rawls.
                     entityQueryUriFromWorkspaceAndQuery(workspaceNamespace, workspaceName, entityType).
-                    withQuery(requestUri.query)
+                    withQuery(requestUri.query())
                   val extReq = Get(entityQueryUri)
-                  externalHttpPerRequest(requestContext, extReq)
+
+                  executeRequestRaw(userInfo.accessToken)(extReq).flatMap { resp =>
+                    requestContext.complete(resp)
+                  }
```
davidangb, line 100: same here, can you omit the `requestContext.complete`?

src/main/scala/org/broadinstitute/dsde/firecloud/service/MethodConfigurationService.scala
```diff
@@ -45,69 +42,78 @@ object MethodConfigurationService {
...
+                    destination = Option(MethodConfigurationId(
+                      name = ingest.destinationName,
+                      namespace = ingest.destinationNamespace,
+                      workspaceName = Option(WorkspaceName(
+                        namespace = workspaceNamespace,
+                        name = workspaceName)))))
+                  val extReq = Post(MethodConfigurationService.remoteCopyFromMethodRepoConfigUrl, copyMethodConfig)
+
+                  executeRequestRaw(userInfo.accessToken)(extReq).flatMap { resp =>
+                    requestContext.complete(resp)
```
davidangb: can you omit the `requestContext.complete`? I'll stop commenting about this; if you're able to make the change, please make it everywhere it applies.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/HttpClientUtils.scala
```diff
@@ -0,0 +1,85 @@
...
+import org.broadinstitute.dsde.rawls.RawlsExceptionWithErrorReport
+import org.broadinstitute.dsde.rawls.model.{ErrorReport, ErrorReportSource}
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.Success
+
+/**
+  * Created by rtitle on 7/7/17.
+  */
+trait HttpClientUtils extends LazyLogging {
```
davidangb: eek, RestJsonClient vs DsdeHttpDAO vs HttpClientUtils

src/main/scala/org/broadinstitute/dsde/firecloud/utils/RestJsonClient.scala
```diff
@@ -44,54 +53,54 @@ trait RestJsonClient extends FireCloudRequestBuilding with PerformanceLogging {
...
-      case (true, true) => addFireCloudCredentials ~> addHeader (`Accept-Encoding`(gzip)) ~> sr ~> decode(Gzip)
-      case (true, false) => addHeader (`Accept-Encoding`(gzip)) ~> sr ~> decode(Gzip)
-      case (false, true) => addFireCloudCredentials ~> sr
-      case _ => sr
+    //    val sr = if (connector.isDefined) {
+    //      sendReceive(connector.get)(executionContext, futureTimeout = 60.seconds)
+    //    } else {
+    //      sendReceive
+    //    }
+    //TODO: ask David An about the above connector code
```
davidangb: at one point, we used a [Spray host connector](http://spray.io/documentation/1.2.4/spray-can/http-client/host-level/) to throttle simultaneous requests to ontology, because we were overwhelming it: #414. But we eventually removed that code and used a different approach: #544

src/main/scala/org/broadinstitute/dsde/firecloud/utils/StandardUserInfoDirectives.scala
```diff
@@ -1,46 +1,45 @@
...
 
-
+  private def getWorkbenchUserEmailId(userInfo:UserInfo):Future[Option[RegistrationInfoV2]] = {
+    if (isServiceAccount(userInfo.userEmail)) {
+      samDAO.getRegistrationStatusV2(userInfo)
+    }
+    else {
+      Future.successful(None)
+    }
+  }
```
davidangb: this is a good change to make ... but it is also a functional change. I say we keep it in but look out for any subtle behavior changes.

davidangb (commented):
some comments inline. I think overall this is the right approach and you've done good work to get to this point. I see your complaint about the inconsistencies - I defer to you as to whether it's easier to refactor all the inconsistencies in place, or try to align everything so there's a common set of working code.

I suggest the next best step is getting tests working, so you can validate how well this all functions!

src/main/scala/org/broadinstitute/dsde/firecloud/utils/RestJsonClient.scala
rushtong: If it helps to not worry about overloading ontology with a simpler approach, please do. We've moved to a k8s cluster and can spin up as many as needed to handle the volume.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
MatthewBemis: yeah it was unused as of #800 and had managed to escape my scalpel

src/main/scala/org/broadinstitute/dsde/firecloud/service/EntityService.scala
MatthewBemis: it looks like `complete { executeRequestRaw(userInfo.accessToken)(extReq) }` will work, will make this change across the board

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpImportServiceDAO.scala
```diff
@@ -0,0 +1,54 @@
...
+import org.broadinstitute.dsde.firecloud.model.{ImportServiceRequest, ImportServiceResponse, PfbImportRequest, PfbImportResponse, RequestCompleteWithErrorReport, UserInfo}
+import org.broadinstitute.dsde.firecloud.service.FireCloudDirectiveUtils
+import org.broadinstitute.dsde.firecloud.service.PerRequest.{PerRequestMessage, RequestComplete}
+import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
+import org.broadinstitute.dsde.firecloud.utils.{HttpClientUtilsStandard, RestJsonClient}
+import org.broadinstitute.dsde.rawls.model.WorkspaceName
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class HttpImportServiceDAO(implicit val system: ActorSystem, implicit val materializer: Materializer, implicit val executionContext: ExecutionContext)
```
MatthewBemis, line 19: broke this out into it's own DAO for better isolation. was previously implemented within `EntityClient` (now called `EntityService`)

src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
```diff
@@ -124,30 +102,15 @@ object EntityClient {
 
 }
 
-class EntityClient(requestContext: RequestContext, modelSchema: ModelSchema, googleServicesDAO: GoogleServicesDAO)(implicit val executionContext: ExecutionContext)
-  extends Actor with RestJsonClient with TSVFileSupport with LazyLogging {
+class EntityService(rawlsDAO: RawlsDAO, importServiceDAO: ImportServiceDAO, modelSchema: ModelSchema, googleServicesDAO: GoogleServicesDAO)(implicit val executionContext: ExecutionContext)
```
MatthewBemis: renamed to `EntityService` from `EntityClient` for consistency.

there are a few places in this PR where i try to do some separation of Api/Service/DAO layers. there's definitely more i can do there, but I only wanted to tackle the ones that were directly in my way. will take care of it in a followup.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
```diff
@@ -368,29 +326,20 @@ class EntityClient(requestContext: RequestContext, modelSchema: ModelSchema, goo
   }
 
   def importPFB(workspaceNamespace: String, workspaceName: String, pfbRequest: PfbImportRequest, userInfo: UserInfo): Future[PerRequestMessage] = {
```
MatthewBemis, line 324: moved everything in this method to a new `ImportServiceDAO`

src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
```diff
@@ -368,29 +326,20 @@ class EntityClient(requestContext: RequestContext, modelSchema: ModelSchema, goo
...
-        val importServiceResponse = unmarshal[ImportServiceResponse].apply(resp)
-        // for backwards compatibility, we return Accepted(202), even though import service returns Created(201),
-        // and we return a different response payload than what import service returns.
-
-        val responsePayload:PfbImportResponse = PfbImportResponse(
-          jobId = importServiceResponse.jobId,
-          url = pfbRequest.url.getOrElse(""),
-          workspace = WorkspaceName(workspaceNamespace, workspaceName)
-        )
+  def getEntitiesWithType(workspaceNamespace: String, workspaceName: String, userInfo: UserInfo): Future[PerRequestMessage] = {
```
MatthewBemis, line 328: this used to live in `GetEntitiesWithType.scala`. i've moved it here and greatly simplified the code

src/main/scala/org/broadinstitute/dsde/firecloud/model/package.scala
```diff
@@ -1,43 +1,31 @@
...
-  implicit def spray2akkaStatus(spray: SprayStatusCode): AkkaStatusCode = {
-    AkkaStatusCode.int2StatusCode(spray.intValue)
-  }
+  import org.broadinstitute.dsde.rawls.model.WorkspaceJsonSupport._
+  import spray.json._
 
-  implicit def akka2sprayStatus(akka: AkkaStatusCode): SprayStatusCode = {
-    SprayStatusCode.int2StatusCode(akka.intValue)
-  }
+  implicit val defaultErrorReportRejectionHandler = RejectionHandler.default.mapRejectionResponse {
```
MatthewBemis, line 19: i believe these rejection handlers are correctly reimplemented but i'm not certain that both of them get properly utilized. TBD.

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudApp.scala
```diff
@@ -1,25 +1,118 @@
...
+import org.broadinstitute.dsde.firecloud.model.{ModelSchema, UserInfo, WithAccessToken}
+import org.broadinstitute.dsde.firecloud.service._
+import org.broadinstitute.dsde.workbench.util.health.HealthMonitor
+import org.elasticsearch.client.transport.TransportClient
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.duration._
 
-object FireCloudApp extends LazyLogging {
+object FireCloudApp extends App with LazyLogging {
```
MatthewBemis: this is analogous to `Boot.scala` in rawls. maybe i'll rename it so we have some consistency? TBD

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudServiceActor.scala
```diff
@@ -1,228 +0,0 @@
...
-import org.broadinstitute.dsde.firecloud.webservice._
-import org.broadinstitute.dsde.workbench.util.health.HealthMonitor
-import org.elasticsearch.client.transport.TransportClient
-
-import scala.concurrent.duration._
-import scala.concurrent.ExecutionContext.Implicits.global
-import scala.language.postfixOps
-
-
-class FireCloudServiceActor extends HttpServiceActor with FireCloudDirectives
```
MatthewBemis, line 22: as noted above, this class has been removed and replaced with `FireCloudApp.scala` and `FireCloudApiService.scala`

src/main/scala/org/broadinstitute/dsde/firecloud/core/AgoraPermissionHandler.scala
```diff
@@ -1,137 +0,0 @@
...
-          case ListOwner => Owner
-          case ListAll => Owner
-          case _ => NoAccess
-        }
-      }
-    }
-  }
-}
-
-class AgoraPermissionActor (requestContext: RequestContext) extends Actor
```
MatthewBemis, line 69: see `AgoraPermissionService.scala` for the new implementation

src/main/scala/org/broadinstitute/dsde/firecloud/core/GetEntitiesWithType.scala
```diff
@@ -1,90 +0,0 @@
...
-
-import scala.concurrent.Future
-import scala.util.{Failure, Success}
-
-object GetEntitiesWithType {
-  case class ProcessUrl(url: String)
-  def props(requestContext: RequestContext): Props = Props(new GetEntitiesWithTypeActor(requestContext))
-}
-
-class GetEntitiesWithTypeActor(requestContext: RequestContext) extends Actor with FireCloudRequestBuilding {
```
MatthewBemis, line 33: as noted, the code in here has been simplified and rolled into `EntityService.scala`

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/DsdeHttpDAO.scala
```diff
@@ -0,0 +1,61 @@
...
+import org.broadinstitute.dsde.firecloud.utils.{HttpClientUtils, HttpClientUtilsStandard}
+import org.broadinstitute.dsde.rawls.RawlsExceptionWithErrorReport
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+
+/**
+  * Created by dvoet on 9/2/15.
+  */
+trait DsdeHttpDAO extends LazyLogging {
```
MatthewBemis: lifted from rawls. I am second-guessing whether or not we actually need this _and_ `RestJsonClient.scala` in orch. i have been slowly removing usage of this and favoring `RestJsonClient` where possible

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpAgoraDAO.scala
```diff
@@ -38,6 +40,18 @@ class HttpAgoraDAO(config: FireCloudConfig.Agora.type)(implicit val system: Acto
     authedRequestToObject[List[EntityAccessControlAgora]]( Post(getMultiEntityPermissionUrl(entityType), entities) )
   }
 
+  override def batchCreatePermissions(inputs: List[EntityAccessControlAgora])(implicit userInfo: UserInfo): Future[List[EntityAccessControlAgora]] = {
```
MatthewBemis, line 43: new methods since we're no longer firing off direct (but non-passthrough) requests from `AgoraPermissionActor` anymore

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -94,13 +86,38 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
...
   lazy private val firecloudAdminSACreds = ServiceAccountCredentials
     .fromStream(new FileInputStream(FireCloudConfig.Auth.firecloudAdminSAJsonFile))
 
+  def getAdminUserAccessToken = {
+    getScopedServiceAccountCredentials(firecloudAdminSACreds, authScopes)
+      .refreshAccessToken().getTokenValue
+  }
+}
+
+class HttpGoogleServicesDAO(implicit val system: ActorSystem, implicit val materializer: Materializer, implicit val executionContext: ExecutionContext) extends GoogleServicesDAO with FireCloudRequestBuilding with LazyLogging with DsdeHttpDAO with SprayJsonSupport {
```
MatthewBemis: this was an object but i moved most of the code into a class and only what was necessary stays inside of a companion object

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -360,15 +352,15 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
...
+
+                  executeRequestRaw(userAuthToken.accessToken)(extReq).map { proxyResponse =>
+                    proxyResponse.header[`Content-Type`] match {
                       case Some(ct) =>
-                        RequestCompleteWithHeaders((proxyResponse.status, proxyResponse.entity),
-                          HttpHeaders.`Content-Type`(ct.contentType)
-                        )
-                      case None => RequestComplete((proxyResponse.status, proxyResponse.entity))
+                        RequestComplete(proxyResponse.status)
+//                        RequestCompleteWithHeaders((proxyResponse.status, proxyResponse.entity),
```
MatthewBemis: i see now that there are still some comments in here that i need to clean up

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -94,19 +93,30 @@ class ExportEntitiesByTypeActor(rawlsDAO: RawlsDAO,
...
-      }.recover { case t: Throwable =>
-        // Stream exceptions have to be handled by directly closing out the RequestContext responder
-        handleStreamException(t)
+
+        val contentType = ContentType(MediaTypes.`text/tab-separated-values`, HttpCharsets.`UTF-8`)
+        val fileName = entityType + ".tsv"
+        val disposition = `Content-Disposition`.apply(ContentDispositionTypes.attachment, Map("filename" -> fileName))
+
+        streamSingularType(entityQueries, metadata, headers).map { source =>
+          HttpResponse(entity = HttpEntity.fromFile(contentType, source.toJava), headers = List(keepAlive, disposition))
```
MatthewBemis, line 117: `fromFile` should utilize chunking (with the default chunk size) but admittedly i haven't looked up what that default chunk size is

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -119,56 +129,64 @@ class ExportEntitiesByTypeActor(rawlsDAO: RawlsDAO,
...
-  private def handleStreamException(t: Throwable): Unit = {
-    logger.info("handling exception", t)
-    val message = t match {
-      case f: FireCloudExceptionWithErrorReport => s"FireCloudException: Error generating entity download: ${f.errorReport.message}"
-      case _ => s"FireCloudException: Error generating entity download: ${t.getMessage}"
-    }
-    ctx.responder ! MessageChunk(message)
-    ctx.responder ! ChunkedMessageEnd
-  }
+  private def streamSingularType(entityQueries: Seq[EntityQuery], metadata: EntityTypeMetadata, entityHeaders: IndexedSeq[String]): Future[File] = {
```
MatthewBemis, line 142: i re-implemented this to basically mimic what `streamCollectionType` was already doing. i wanted to kill off the actor sending rows as chunks because i thought that implementation was harder to understand and honestly i just couldn't get it to work. i think i better understand it all now but unless there are objections to this implementation i would like to go this route

src/main/scala/org/broadinstitute/dsde/firecloud/service/FireCloudDirectives.scala
```diff
@@ -36,19 +36,18 @@ trait FireCloudDirectives extends spray.routing.Directives with PerRequestCreato
   def encodeUri(path: String): String = FireCloudDirectiveUtils.encodeUri(path)
 
   private def generateExternalHttpPerRequestForMethod(uri: Uri, inMethod: HttpMethod) = {
```
MatthewBemis: currently, there is a bug in here somewhere that can cause an akka-http EntitySource to be materialized twice, which results in an exception being thrown pretty much any time the passthrough directive is used. i haven't successfully tracked it down but maybe i just have fatigue from staring at this code for far too long and it'll be obvious to fresh eyes

src/main/scala/org/broadinstitute/dsde/firecloud/service/FireCloudDirectives.scala
```diff
@@ -36,19 +36,18 @@ trait FireCloudDirectives extends spray.routing.Directives with PerRequestCreato
...
-    val outMethod = new RequestBuilder(inMethod)
     // POST, PUT, PATCH
     if (inMethod.isEntityAccepted) {
```
MatthewBemis: oddly,` HttpMethods.GET.isEntityAccepted` returns `true` now (and DELETE as well)

```
scala> HttpMethods.GET.isEntityAccepted
res3: Boolean = true
```

looking into the akka-http implementation, GET and DELETE "tolerate" entities, and for the other methods it's "expected". if we want to be tolerant of these potential cases, i propose removing the branching here

we could alternatively keep the branching and call `getRequestEntityAcceptance` on `inMethod` and see if the value is `Expected`

src/main/scala/org/broadinstitute/dsde/firecloud/service/MethodConfigurationService.scala
```diff
@@ -45,69 +42,72 @@ object MethodConfigurationService {
...
+                  methodRepoSnapshotId = ingest.configurationSnapshotId,
+                  destination = Option(MethodConfigurationId(
+                    name = ingest.destinationName,
+                    namespace = ingest.destinationNamespace,
+                    workspaceName = Option(WorkspaceName(
+                      namespace = workspaceNamespace,
+                      name = workspaceName)))))
+                val extReq = Post(MethodConfigurationService.remoteCopyFromMethodRepoConfigUrl, copyMethodConfig)
+
+                complete { executeRequestRaw(userInfo.accessToken)(extReq) }
```
MatthewBemis: i tried to do this in most places but looks like i missed one. i'd like to rename this trait to `MethodConfigurationApiService` and lighten up the routes a lot, pushing this logic down to a `MethodConfigurationService` and underlying `AgoraDAO`

i don't think that we should be generating any external HTTP calls from this layer unless it's a passthrough directive

src/main/scala/org/broadinstitute/dsde/firecloud/service/UserService.scala
```diff
@@ -221,4 +205,11 @@ class UserService(rawlsDAO: RawlsDAO, thurloeDAO: ThurloeDAO, googleServicesDAO:
       }
     }
   }
+
+  def getUserProfileGoogle(userToken: UserInfo): Future[PerRequestMessage] = {
```
MatthewBemis, line 209: this call was previously made directly from the route/API layer, which as i noted in other places, i don't like

src/main/scala/org/broadinstitute/dsde/firecloud/utils/HttpClientUtils.scala
MatthewBemis: Originally lifted from rawls but as with DsdeHttpDAO, I don't think we actually need it in orch. i have one more place in HttpGoogleServicesDAO and a few usages of DsdeHttpDAO that i can hopefully rip out and then i think this will disappear

src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingActor.scala
```diff
@@ -1,44 +0,0 @@
...
-
-object StreamingActor {
-  case class FirstChunk(httpData: HttpData, remaining: Int)
-  case class NextChunk(httpData: HttpData, remaining: Int)
-  case class Ok(remaining: Int)
-  def props(ctx: RequestContext, contentType: ContentType, fileName: String) =
-    Props(new StreamingActor(ctx, contentType, fileName))
-}
-
-class StreamingActor(ctx: RequestContext, contentType: ContentType, fileName: String) extends Actor with LazyLogging {
```
MatthewBemis, line 17: as noted in ExportEntitiesByTypeActor, I killed this actor off in favor of another approach

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/CromIamApiService.scala
```diff
@@ -65,7 +65,7 @@ trait CromIamApiService extends HttpService with FireCloudRequestBuilding with F
             }
           }
         } ~
-        path("backend" / "metadata" / Segment.repeat(Slash)) { operationId =>
+        path("backend" / "metadata" / Segment.repeat(0,1024, Slash)) { operationId =>
```
MatthewBemis, line 68: `repeat` now requires a min and max # of times to repeat

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/EntityApiService.scala
```diff
@@ -40,18 +42,19 @@ trait EntityService extends HttpService with PerRequestCreator with FireCloudDir
...
                           val copyMethodConfig = new EntityCopyDefinition(
                             sourceWorkspace = copyRequest.sourceWorkspace,
                             destinationWorkspace = WorkspaceName(workspaceNamespace, workspaceName),
                             entityType = copyRequest.entityType,
                             entityNames = copyRequest.entityNames)
                           val extReq = Post(FireCloudConfig.Rawls.workspacesEntitiesCopyUrl(linkExistingEntitiesBool), copyMethodConfig)
-                          externalHttpPerRequest(requestContext, extReq)
+
+
+                          complete { executeRequestRaw(userInfo.accessToken)(extReq) }
```
MatthewBemis: i don't like executing these requests right from here. will refactor a bit more

src/main/scala/org/broadinstitute/dsde/firecloud/service/FireCloudDirectives.scala
MatthewBemis: found where it was being consumed twice.  the `logRequestResult` directive which is wrapped around every route in `FireCloudApiService` was grabbing the entity to log it, which was throwing this exception. i've removed the use of that logging directive for now. i don't think it's critical for the first round of review and i'll figure out what's going on for followup review

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/EntityApiService.scala
MatthewBemis: actually i will leave whether or not i do this refactor up to the reviewers. i think this PR has moved orch in a better direction in terms of standardizing how we make external requests and separate our service layers, but i am doubting that i need to do all of it right now in this PR. this could easily be followup cleanup and make for a more manageable review...

src/main/scala/org/broadinstitute/dsde/firecloud/utils/StandardUserInfoDirectives.scala
MatthewBemis: i ended up scrapping it for now. it was having a ripple effect across some other classes that used the trait. i might have screwed something up with that but i propose doing this in a followup PR instead

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -360,15 +353,12 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
...
+
+                  executeRequestRaw(userAuthToken.accessToken)(extReq).map { proxyResponse =>
+                    proxyResponse.header[`Content-Type`] match {
                       case Some(ct) =>
-                        RequestCompleteWithHeaders((proxyResponse.status, proxyResponse.entity),
-                          HttpHeaders.`Content-Type`(ct.contentType)
-                        )
-                      case None => RequestComplete((proxyResponse.status, proxyResponse.entity))
+                        RequestCompleteWithHeaders(proxyResponse, `Content-Type`(ct.contentType))
+                      case None => RequestComplete(proxyResponse)
```
MatthewBemis, line 357: i think that this and the above line are equivalent to their previous implementations but have not tested

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -390,31 +380,31 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
...
-                        RequestCompleteWithHeaders(StatusCodes.TemporaryRedirect, HttpHeaders.Location(Uri(redirectUrl)))
+                        RequestCompleteWithHeaders(StatusCodes.TemporaryRedirect, Location(Uri(redirectUrl)))
 
                     }
                   }
                 }
 
               case _ =>
                 // the user does not have access to the object.
-                val responseStr = objectResponse.entity.asString.replaceAll("\n","")
```
MatthewBemis, line 401: this will no longer be logged. consuming the entity to put it in the logs was a PITA but maybe i wasn't going about it the right way

src/main/scala/org/broadinstitute/dsde/firecloud/model/ErrorReport.scala
```diff
@@ -1,33 +1,34 @@
...
-    }
+      //TODO
+//      val (message, causes) = Try(Unmarshal(response.entity).to[ErrorReport]) match {
+//        case Failure(re) => (re.getMessage, Seq(re))
+//        case Success(err) => (response.entity.toString, Seq.empty)
+//      }
 
-    def tryUnmarshal(response: HttpResponse) =
-      Try { unmarshal[ErrorReport].apply(response) }
+      new ErrorReport(ers.source, "TODO", Option(response.status), Seq.empty, Seq.empty, None)
```
MatthewBemis: just realized I left this TODO, i'll address it tomorrow

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/StaticNotebooksApiService.scala
```diff
@@ -1,31 +1,39 @@
...
+          requestContext =>
+            // call Calhoun and pass its response back to our own caller
+            // can't use passthrough() here because that demands a JSON response
+
+            //TODO: ensure mediatype is still honored, but respondWithMediaType was deprecated
+            //because it was an anti-pattern
+
+            val extReq = Post(calhounStaticNotebooksURL, requestContext.request.entity)
+            executeRequestRaw(userInfo.accessToken)(extReq).flatMap { resp =>
+              requestContext.complete(resp)
```
MatthewBemis, line 27: i need to actually test that this returns `text/html` still. if not i'll have to make sure the content type negotiation is working, because `respondWithMediaType` was removed in akka-http since it was an anti-pattern

ansingh7115:
I agree that further separation of the api, service and dao layers should be considered out of scope for this PR and for this ticket.

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudApp.scala
ansingh7115: I like the consistency of similar naming, but it's not super duper important

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudApp.scala
```diff
@@ -1,25 +1,118 @@
...
 
   // we need an ActorSystem to host our application in
   val system = ActorSystem("FireCloud-Orchestration-API")
   val timeoutDuration = FiniteDuration(FireCloudConfig.HttpConfig.timeoutSeconds, SECONDS)
```
ansingh7115: `system` and `timeoutDuration` are not used  anywhere

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/DsdeHttpDAO.scala
ansingh7115: not to mention dsde no longer exists...  :)

project/Dependencies.scala
davidangb: ah, I love that we no longer depend on `vault-common`

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudApiService.scala
```diff
@@ -0,0 +1,179 @@
...
+      workspaceRoutes ~
+      notificationsRoutes ~
+      statusRoutes ~
+      ga4ghRoutes ~
+      pathPrefix("api") {
+        apiRoutes
+      } ~
+      // insecure cookie-authed routes
+      cookieAuthedRoutes
+  }
```
davidangb, line 193: I haven't cross-referenced, but some of these look like they should be in `apiRoutes`. If they code the `/api/` prefix themselves then it's fine (if inconsistent)

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/DsdeHttpDAO.scala
davidangb: :+1: for only having one of the two;  I don't think that should be a blocker for this PR

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
davidangb: I say as long as this file works for the end user, we're good. We should revisit at some point - I think there may be a much simpler streaming solution now that we're on a more recent version of akka-http.

src/main/scala/org/broadinstitute/dsde/firecloud/service/FireCloudDirectives.scala
davidangb: GET and DELETE entities were added as tolerated in updates to the HTTP spec, so it makes sense that spray (super old) didn't think they were allowed but akka-http does. I agree with removing this branching, or rewriting the logic for it.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/EntityApiService.scala
davidangb: I agree on getting this PR in and merged, and saving non-critical refactors for later. If the refactor helps YOU merge this PR it's worth it though!

davidangb (commented):
submitting comments for a "half review" ... I made it through about half the files and hit a wall. So ... here are a few comments, I will continue this review tomorrow!

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudApp.scala
MatthewBemis: just renamed it!

src/main/scala/org/broadinstitute/dsde/firecloud/utils/HttpClientUtils.scala
MatthewBemis: I just removed DsdeHttpDAO and HttpClientUtils. RestJsonClient is used everywhere now

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/DsdeHttpDAO.scala
MatthewBemis: it's gone now! replaced all usages of DsdeHttpDAO code with RestJsonClient

davidangb:
jenkins retest

davidangb:
jenkins retest, to doublecheck the failures before I start tackling them

src/main/scala/org/broadinstitute/dsde/firecloud/model/package.scala
davidangb: I tested and refactored these - I am confident they are (it is) working now.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/StaticNotebooksApiService.scala
davidangb: I tested this empirically, and see a response header of `content-type: text/html; charset=UTF-8 `

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudApiService.scala
davidangb: I double-checked, these are fine

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -343,8 +331,9 @@ object HttpGoogleServicesDAO extends GoogleServicesDAO with FireCloudRequestBuil
           // user is known to Google. Extract the user's email and SID from the response, for logging
           import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol.impOAuthUser
           import spray.json._
-          val oauthUser:Try[OAuthUser] = Try(userResponse.entity.asString.parseJson.convertTo[OAuthUser])
+          val oauthUser = Try(Unmarshaller.stringUnmarshaller.map(_.parseJson.convertTo[OAuthUser]))
           val userStr = (oauthUser getOrElse userResponse.entity).toString
+          userResponse.discardEntityBytes()
```
davidangb, line 336: here - and a few places below - it's important to `discardEntityBytes` if we don't consume the response entity in any other way. Else, akka stalls holding on to that entity.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -119,56 +129,64 @@ class ExportEntitiesByTypeActor(rawlsDAO: RawlsDAO,
...
+    val tempEntityFile: File = File.newTemporaryFile(prefix = entityType)
+    val entitySink: Sink[ByteString, Future[IOResult]] = FileIO.toPath(tempEntityFile.path)
```
davidangb, line 144: do you have any idea when (if?) these temp files are cleaned up?

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockAgoraDAO.scala
```diff
@@ -33,4 +35,46 @@ class MockAgoraDAO extends AgoraDAO {
...
+    } else {
+      List.empty
+    }
+
+    // methods endpoints return the mock data in reverse order - this way we can differentiate methods vs. configs
+    if (pathString.startsWith("/api/v1/methods/")) {
+      Future.successful(rawData.reverse)
+    } else {
+      Future.successful(rawData)
+    }
```
davidangb, line 57: this is really confusing, but it migrates pre-existing test code (which I originally wrote so I claim fault)

src/main/scala/org/broadinstitute/dsde/firecloud/service/FireCloudDirectives.scala
```diff
@@ -18,10 +20,25 @@ object FireCloudDirectiveUtils {
...
+   *     - if headers that control the http protocol such as Accept-Encoding or Connection
+   *        are passed through, they may not reflect reality. The original request may have
+   *        come from a browser that supports different encodings or connection protocols
+   *        than the service-to-service request we're about to make.
+   *     - if headers that inform the target such as User-Agent, Referer or X-Forwarded-For
+   *        are passed through, they will be misleading, as they reflect the original request
+   *        and not the service-to-service request we're about to make.
+   */
+  final val allowedPassthroughHeaders = List(Authorization, `Content-Type`).map(_.lowercaseName)
+
```
davidangb, line 38: This is a functional change from before this PR. Previously we would only send the `Authorization` header. I think it is worthwhile to add the `Content-Type` header, but am totally open to argument we shouldn't do it. We definitely need to filter out the Host header though, since that causes problems with Import Service hosted on appengine.

Before this PR we created a new outbound HttpRequest based on the inbound method, URI, and Authorization header. In this PR we are reusing the inbound HttpRequest, but filtering its headers.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
davidangb: default is 8192 I believe, specified in `akka.stream.scaladsl.FileIO.fromPath`

src/main/scala/org/broadinstitute/dsde/firecloud/service/FireCloudDirectives.scala
davidangb: I've fixed up `logRequestResult`!

davidangb (approved):
giving this a :+1: for all of Matt's work, even though I also have commits on this PR (we'll need more thumbs than mine)

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
MatthewBemis: they're within the docker container at `/tmp` (just verified on dev), so they'd get destroyed any time we redeploy but i suspect they'd also get cleaned up at least once a month by the OS

MatthewBemis:
👍 for David's work! thanks for carrying this to the finish line!

ansingh7115 (approved):
Looked at from a high level organization of classes / separation of concerns, focusing on JsonRestClient and entity related changes. Those look good to me! Good job on the tag-teaming - this is amazing!

ungwudik:
A few notes about testing:

- I'm assuming people outside our team will know this change is coming (including QA) so people are aware when looking at automated tests, alerts, etc.
- Looks like code coverage decreased. Something to investigate there?
- Can we easily roll back if we run into unexpected issues with this upgrade as it's promoted to other environments?

PR #809, opened 2020-10-22
[CA-1058] update rawls-model

andy7i:

This PR contains a rawls-model update which led to:
- code changes
- unit test changes
- fixes for sbt assembly merge issues

verified locally using a call to cloneWorkspace with `noWorkspaceOwner` which did not work with the old rawls-model.
https://bvdp-saturn-dev.appspot.com/#workspaces/aj-test-noworkspaceowner-bp/test-localorch-cloneWorkspace--noWorkspaceOwner-true-10-23

In all cases:

### DIFF
--- a/project/Merging.scala
+++ b/project/Merging.scala
@@ -5,6 +5,7 @@ object Merging {
     case x if x.endsWith("io.netty.versions.properties") => MergeStrategy.discard
     // we target Java 8, which does not use module-info.class. Some dependencies (Jackson) cause assembly problems on module-info
     case x if x.endsWith("module-info.class") => MergeStrategy.discard
+    case x if x.contains("javax/activation") => MergeStrategy.first
     case x => oldStrategy(x)
   }
 }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
@@ -73,7 +73,7 @@ trait LibraryServiceSupport extends DataUseRestrictionSupport with LazyLogging {
           case Some(s:AttributeString) =>
             val orspAttrs = restrictionMap.getOrElse(s.value, Map.empty[AttributeName, Attribute])
             val newAttrs = replaceDataUseAttributes(ws.attributes.getOrElse(Map.empty), orspAttrs)
-            ws.copy(attributes = Some(newAttrs))
+            ws.copy(attributes = Option(newAttrs))
           case _ =>
             // this workspace does not have an ORSP id; leave it untouched
             ws
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
@@ -69,11 +69,11 @@ trait LibraryServiceSupport extends DataUseRestrictionSupport with LazyLogging {
 
       val annotatedWorkspaces = workspaces map { ws =>
         // does this workspace have an orsp id?
-        ws.attributes.get(orspIdAttribute) match {
+        ws.attributes.getOrElse(Map.empty).get(orspIdAttribute) match {
           case Some(s:AttributeString) =>
             val orspAttrs = restrictionMap.getOrElse(s.value, Map.empty[AttributeName, Attribute])
-            val newAttrs = replaceDataUseAttributes(ws.attributes, orspAttrs)
-            ws.copy(attributes = newAttrs)
+            val newAttrs = replaceDataUseAttributes(ws.attributes.getOrElse(Map.empty), orspAttrs)
+            ws.copy(attributes = Option(newAttrs))
           case _ =>
             // this workspace does not have an ORSP id; leave it untouched
             ws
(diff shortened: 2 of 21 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
```diff
@@ -69,11 +69,11 @@ trait LibraryServiceSupport extends DataUseRestrictionSupport with LazyLogging {
...
       val annotatedWorkspaces = workspaces map { ws =>
         // does this workspace have an orsp id?
-        ws.attributes.get(orspIdAttribute) match {
+        ws.attributes.getOrElse(Map.empty).get(orspIdAttribute) match {
           case Some(s:AttributeString) =>
             val orspAttrs = restrictionMap.getOrElse(s.value, Map.empty[AttributeName, Attribute])
-            val newAttrs = replaceDataUseAttributes(ws.attributes, orspAttrs)
-            ws.copy(attributes = newAttrs)
+            val newAttrs = replaceDataUseAttributes(ws.attributes.getOrElse(Map.empty), orspAttrs)
+            ws.copy(attributes = Option(newAttrs))
```
andy7i, line 76: `Option(newAttrs))` or `Some(newAttrs))` ?

src/main/scala/org/broadinstitute/dsde/firecloud/utils/PermissionsSupport.scala
```diff
@@ -49,7 +49,7 @@ trait PermissionsSupport {
     tryIsAdmin(userInfo) flatMap { isadmin =>
       if (!isadmin) {
         rawlsDAO.getWorkspace(workspaceNamespace, workspaceName)(userInfo.asInstanceOf[WithAccessToken]) map { ws =>
-          ws.accessLevel >= neededLevel
+          ws.accessLevel.getOrElse(WorkspaceAccessLevels.NoAccess) >= neededLevel
```
andy7i: Not 100% sure about this

andy7i (commented):
self-review

src/main/scala/org/broadinstitute/dsde/firecloud/utils/PermissionsSupport.scala
MatthewBemis: i think it's a fairly safe assumption to make, but a more idiomatic approach may be to use pattern matching:

```
          ws.accessLevel match {
            case Some(accessLevel) => accessLevel >= neededLevel
            case None => false
          }
```

pattern matching may also be useful in some other areas here too if you want to short-circuit any sort of evaluation that would be done by supplying a default in `getOrElse`. sorry i didn't mention this earlier when we were slacking!

project/Merging.scala
```diff
@@ -5,6 +5,7 @@ object Merging {
     case x if x.endsWith("io.netty.versions.properties") => MergeStrategy.discard
     // we target Java 8, which does not use module-info.class. Some dependencies (Jackson) cause assembly problems on module-info
     case x if x.endsWith("module-info.class") => MergeStrategy.discard
+    case x if x.contains("javax/activation") => MergeStrategy.first
```
andy7i, line 8: how much would it matter if we replaced this section:
```
{
    case x if x.endsWith("io.netty.versions.properties") => MergeStrategy.discard
    // we target Java 8, which does not use module-info.class. Some dependencies (Jackson) cause assembly problems on module-info
    case x if x.endsWith("module-info.class") => MergeStrategy.discard
    case x if x.contains("javax/activation") => MergeStrategy.first
    case x => oldStrategy(x)
  }
```

with:

```
{
 case x => MergeStrategy.first
}
```
https://stackoverflow.com/questions/25144484/sbt-assembly-deduplication-found-error

src/main/scala/org/broadinstitute/dsde/firecloud/utils/PermissionsSupport.scala
andy7i: totally! thanks for the help!

s-rubenstein (approved):
We paired on a lot of this, but it looks good to me. I don't have an answer to your question though.

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
MatthewBemis: The way you originally had it is generally preferred because it safely handles null values. `Option(null)` will equate to `None`, but `Some(null)` will throw an exception. I pretty much always use `Option(foo)` now because of this, even if I know that `foo` isn't going to be null

project/Merging.scala
MatthewBemis: It's pretty rare that these duplicates pop up so personally I prefer dealing with them on a case by case basis rather than handling every case silently. I think it could lead to some tricky debugging if the merge strategy chooses something that the developer isn't expecting

src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
andy7i: good to know -- thanks for the info!

project/Merging.scala
davidangb: +1 to Matt's sentiment!

MatthewBemis (approved):
:+1:

PR #814, opened 2020-11-13
AS-120 followup: not-quite-exhaustive handling of discardEntityBytes

MatthewBemis:
I have not yet determined if `requestContext.complete(response)` properly consumes `response`....there are a few places that still do that. This takes care of all of the other places, I believe

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
@@ -7,6 +7,7 @@ import java.util.zip.{ZipEntry, ZipFile}
 
 import akka.http.scaladsl.model.StatusCodes._
 import akka.http.scaladsl.model.{HttpResponse, StatusCodes}
+import akka.stream.Materializer
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.firecloud.EntityService._
 import org.broadinstitute.dsde.firecloud.FireCloudConfig.Rawls
@@ -29,7 +30,7 @@ import scala.util.{Failure, Success, Try}
 
 object EntityService {
 
-  def constructor(app: Application)(modelSchema: ModelSchema)(implicit executionContext: ExecutionContext) =
+  def constructor(app: Application)(modelSchema: ModelSchema)(implicit executionContext: ExecutionContext, materializer: Materializer) =
     new EntityService(app.rawlsDAO, app.importServiceDAO, modelSchema)
 
   def colNamesToAttributeNames(headers: Seq[String], requiredAttributes: Map[String, String]): Seq[(String, Option[String])] = {
@@ -98,7 +99,7 @@ object EntityService {
 
 }
 
-class EntityService(rawlsDAO: RawlsDAO, importServiceDAO: ImportServiceDAO, modelSchema: ModelSchema)(implicit val executionContext: ExecutionContext)
+class EntityService(rawlsDAO: RawlsDAO, importServiceDAO: ImportServiceDAO, modelSchema: ModelSchema)(implicit val executionContext: ExecutionContext, val materializer: Materializer)
   extends TSVFileSupport with LazyLogging {
 
   val format = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZZ")
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
@@ -211,6 +211,7 @@ class EntityService(rawlsDAO: RawlsDAO, importServiceDAO: ImportServiceDAO, mode
     response map { response =>
       response.status match {
         case NoContent =>
+          response.discardEntityBytes()
           logger.debug("OK response")
           RequestComplete(OK, entityType)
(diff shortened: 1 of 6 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -356,6 +356,7 @@ class HttpGoogleServicesDAO(implicit val system: ActorSystem, implicit val mater
                   val extReq = Get(gcsApiUrl)
 
                   userAuthedRequest(extReq)(userAuthToken).map { proxyResponse =>
+                    proxyResponse.discardEntityBytes()
                     proxyResponse.header[`Content-Type`] match {
                       case Some(ct) =>
                         RequestCompleteWithHeaders(proxyResponse, `Content-Type`(ct.contentType))
```
davidangb, line 362: this is surprising to me ... sending the response object to the client does not consume the entity?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
```diff
@@ -162,6 +165,7 @@ class HttpRawlsDAO(implicit val system: ActorSystem, implicit val materializer:
     val url = editBillingMembershipURL(projectId, role, email)
 
     userAuthedRequest(Put(url), true) flatMap { resp =>
+      resp.discardEntityBytes()
```
davidangb, line 168: should this be inside the `if (resp.status.isSuccess) {` clause? Does the `FCErrorReport(resp).flatMap { errorReport =>`  properly consume the entity?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
```diff
@@ -176,6 +180,7 @@ class HttpRawlsDAO(implicit val system: ActorSystem, implicit val materializer:
     val url = editBillingMembershipURL(projectId, role, email)
 
     userAuthedRequest(Delete(url), true) flatMap { resp =>
+      resp.discardEntityBytes()
```
davidangb, line 183: same question here, does creating the error report consume the entity?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpSamDAO.scala
```diff
@@ -106,6 +106,7 @@ class HttpSamDAO( implicit val system: ActorSystem, val materializer: Materializ
 
   private def userAuthedRequestToUnit(request: HttpRequest)(implicit userInfo: WithAccessToken): Future[Unit] = {
     userAuthedRequest(request).map { resp =>
+      resp.discardEntityBytes()
```
davidangb, line 109: same question about errorreport

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
MatthewBemis: I probably shouldn't have speculated, I will verify this now (along with the other unclear bits)

src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
```diff
@@ -211,6 +212,7 @@ class EntityService(rawlsDAO: RawlsDAO, importServiceDAO: ImportServiceDAO, mode
     response map { response =>
       response.status match {
         case NoContent =>
+          response.discardEntityBytes()
```
ansingh7115, line 215: what does discardEntityBytes do?

MatthewBemis: the comments in the akka-http code explain it _much_ better than i could, so here they are: 

```     
* Discards the entities data bytes by running the `dataBytes` Source contained by the `entity` of this HTTP message.
     *
     * Note: It is crucial that entities are either discarded, or consumed by running the underlying [[akka.stream.scaladsl.Source]]
     * as otherwise the lack of consuming of the data will trigger back-pressure to the underlying TCP connection
     * (as designed), however possibly leading to an idle-timeout that will close the connection, instead of
     * just having ignored the data.
```

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
MatthewBemis: im a bit confused by this. while testing, i'm forcing it to take the FCErrorReport code path, and it works with or without discarding the bytes manually. is this the same thing we were talking about yesterday with the `logRequests` directive?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
MatthewBemis: I believe you are right, otherwise we would be seeing this in other places that I have not seen it

PR #815, opened 2020-11-13
EXPERIMENTAL/DNM: auto-batching of batch entity upserts/updates between Orch and Rawls

davidangb:
Drafting a PR so others can provide feedback on the approach.

TSV import and Bagit import both generate potentially large JSON request bodies which Orch sends to Rawls. It is possible for this request body to be so large that Rawls rejects it (though we just tweaked the size limit in broadinstitute/rawls#1311). It is also possible for the request body to be large enough that Rawls accepts it but is slow to operate on that much data.

This PR - which is very much in first-draft/sketch form - takes that single large JSON request body, and splits it up into multiple request bodies, testing that each subset is smaller than a configurable byte-size limit. It then sends each of those request bodies serially to Rawls and aggregates the responses.

This would affect:
* TSV import
* Bagit/BDBag import (these are .zip files containing multiple TSVs)

This would not affect:
* PFB import
* any calls from a client directly to Rawls' batchUpsert API, bypassing orch

Dear reader: is this a good idea? Should I continue on this approach? Are you aware of any libraries/utilities that do this json-splitting already, instead of my hand-written algorithm?

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
@@ -28,7 +28,9 @@ import org.joda.time.DateTime
 import spray.json.DefaultJsonProtocol._
 import spray.json._
 
+import scala.concurrent.duration.Duration
 import scala.concurrent.{ExecutionContext, Future}
+import scala.util.Try
 import scala.util.control.NonFatal
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
```diff
@@ -186,18 +188,127 @@ class HttpRawlsDAO(implicit val system: ActorSystem, implicit val materializer:
...
-    val request = Post(FireCloudDirectiveUtils.encodeUri(Rawls.entityPathFromWorkspace(workspaceNamespace, workspaceName)+"/batchUpsert"),
-      HttpEntity(MediaTypes.`application/json`,upserts.toJson.toString))
+  private def aggregateResponses(requests: Seq[HttpRequest])(implicit userToken: UserInfo): Future[HttpResponse] = {
+
+    case class ResponseStub(statusCode: StatusCode, bodyString: String)
+
+    // send all of the batches. This needs to be done serially to ensure that prior entities are committed before
+    // any entities that depend on them.
+    // If we are certain these are all the same entityType, they could be sent in parallel. But, sending in parallel
+    // could cause contention, so ...
```
MatthewBemis, line 198: 👍  for this reasoning to not send in parallel even in this case

src/main/scala/org/broadinstitute/dsde/firecloud/utils/JsonUtils.scala
```diff
@@ -0,0 +1,99 @@
...
+
+import com.typesafe.scalalogging.LazyLogging
+import spray.json.{JsArray, JsValue}
+
+import scala.annotation.tailrec
+
+class JsonUtils extends LazyLogging {
+
+  // how close to the byte size limit should we get to be considered a good batch?
+  private final val TOLERANCE = .8
```
MatthewBemis, line 11: 👍  for a good tolerance

MatthewBemis (commented):
reviewing purely from a high-level, i approve of the approach

I also would love if there was a library that would handle all of the json grouping/splitting, but I don't know of one. I can do a little searching myself

PR #820, opened 2020-12-01
[AoU RW-5987] Add passthrough for Rawls addProjectToServicePerimeter

jmthibault79:
AoU is prototyping a change of our process from specifying a Service Perimeter at project creation time to adding it when we are ready to use it to create a workspace.  We will need this endpoint to do so. 

In all cases:

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -6729,7 +6729,7 @@ paths:
   /servicePerimeters/{servicePerimeterName}/projects/{projectName}:
     put:
       tags:
-        - servicePerimeters
+        - Service Perimeters
       summary: Add a project to a service perimeter
       description: |
         Add a project to a service perimeter, must have 'add_to_service_perimeter'
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -6726,6 +6726,61 @@ paths:
                   $ref: '#/components/schemas/ToolClass'
       x-passthrough: true
       x-passthrough-target: agora
+  /servicePerimeters/{servicePerimeterName}/projects/{projectName}:
+    put:
+      tags:
+        - servicePerimeters
+      summary: Add a project to a service perimeter
+      description: |
+        Add a project to a service perimeter, must have 'add_to_service_perimeter'
+        action on project and 'add_project' action on perimeter, included in owner
+        role for both
+      operationId: addProjectToServicePerimeter
+      parameters:
+        - name: servicePerimeterName
+          in: path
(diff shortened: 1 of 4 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -6726,6 +6726,61 @@ paths:
                   $ref: '#/components/schemas/ToolClass'
       x-passthrough: true
       x-passthrough-target: agora
+  /servicePerimeters/{servicePerimeterName}/projects/{projectName}:
+    put:
+      tags:
+        - Service Perimeters
```
calbach, line 6732: I'm surprised this works, but it appears to be the convention in the file 🤷‍♂️

calbach (approved):
Looks semantically correct to me, though I'm unqualified to review Scala

src/main/resources/swagger/api-docs.yaml
```diff
@@ -6726,6 +6726,61 @@ paths:
...
+        Add a project to a service perimeter, must have 'add_to_service_perimeter'
+        action on project and 'add_project' action on perimeter, included in owner
+        role for both
+      operationId: addProjectToServicePerimeter
+      parameters:
+        - name: servicePerimeterName
+          in: path
+          description: |
+            Fully qualified google service perimeter name in the form of
+            accessPolicies/[POLICY NUMBER]/servicePerimeters/[NAME], url encoded
```
MatthewBemis, line 6744: Who is encoding this parameter? I believe we could avoid requiring an encoded perimeter name by doing something like this in the route:

```
path("servicePerimeters" / Segment.repeat(4, Slash) / "projects" / Segment) { (servicePerimeterName, projectName) =>

    val qualifiedName = servicePerimeterName.mkString("/")
    ...
````

I realize that this may have upstream and downstream implications so if we _have_ to work with a url-encoded parameter then consider this totally optional

MatthewBemis (approved):
This looks good to me. I left one comment which is optional.

src/main/resources/swagger/api-docs.yaml
jmthibault79: AFAICT this is only used to generate the section headings in the Swagger UI.  It isn't used for code generation like in AoU RW.

src/main/resources/swagger/api-docs.yaml
jmthibault79: Interesting!  I had not seen that feature.

I'd rather keep it as-is since we'd have to re-encode to send it on to Rawls anyway.  Unless we are OK with making a breaking change to the Rawls API too.  Doesn't seem worth it to me in this case.

PR #823, opened 2020-12-15
[DO NOT MERGE] Convert entity to strict entity in export methodconfig endpoint

MatthewBemis:
Disclaimer: I'm only opening this PR to help illustrate the bug and to discuss it. I think this needs a bit more discussion before proposing a real fix.

This is a potential fix for a bug that occurs if a user tries to export a workflow (née method config) to their workspace. In rare cases, they will get a "Substream cannot be materialized twice" error. This does not seem to happen with smaller workflows, only with large ones that have many inputs. In production and dev, 300 was a sufficient number of inputs to trigger the error. While testing with a local orch, I had to push the number to 1000 inputs to get a fairly repeatable error. 

Digging in, it seems that it is actually possible to unmarshal a request entity twice without converting it into a strict entity, but only if it is small enough to fit inside of akka-http's internal buffer. That would explain why this only occurs with a large payload and why it sometimes works and sometimes doesn't (the buffer might not always be near the limit).

The fix proposed here wraps the route in a directive that converts the request entity to strict- guaranteeing that it will be stored in memory rather than streamed. It's not ideal, but it's still bound to the akka-http `max-to-strict-bytes` config setting so we would have some protection against OOMs.

I am concerned that this is not localized to just this endpoint, but I haven't tested that other endpoints are vulnerable. I suspect that somewhere under the hood, either `entity(as[Foo])` and/or somewhere in our `generateExternalHttpRequestForMethod` method, we are somehow reading the request entity more than once. If that's the case, then it's possible that other endpoints could experience this bug, specifically if they meet the following criteria:

* Passthrough
* POST
* We look at the payload before passing it into the `passthrough` directive 

This needs a bit more investigation, though.

Possible solutions:

A) There may be a bug in `generateExternalHttpRequestForMethod`, fix that instead (best case)
B) Adopt this change (and apply it to other endpoints that meet the above criteria. only acceptable if it's deemed that there is no bug in option A)
C) Convert this passthrough to use a DAO which doesn't seem to be susceptible (not ideal, ignores the problem)

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiService.scala
@@ -19,6 +19,7 @@ import org.slf4j.{Logger, LoggerFactory}
 import spray.json.DefaultJsonProtocol._
 
 import scala.concurrent.ExecutionContext
+import scala.concurrent.duration._
 
 trait WorkspaceApiService extends FireCloudRequestBuilding with FireCloudDirectives with StandardUserInfoDirectives {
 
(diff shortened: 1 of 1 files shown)


davidangb:
if this problem is triggered by overall size - the number of inputs - I wonder if it has to do with whether or not the request is chunked or not. I haven't looked at the code to see how this could affect us, just a thought!

PR #826, opened 2020-12-16
AS-543: Remove Logit metrics code

MatthewBemis:

In all cases:


davidangb:
love this PR ... aaaand, I don't know who owns these metrics now, we should make sure they're absolutely not being used by anyone anywhere

MatthewBemis:
@davidangb I'll see if I can figure that out. Another benefit of removing this is that I believe it's the only thing that's still calling the old rawls admin statistics endpoint. I would need to double check with Chet to make sure it's definitely no longer needed for NCI reporting but I suspect that it's not based on looking through Logit. It would be really nice to get rid of it- it's a bunch of pretty ugly code and I'd like if nobody had to think about maintaining it.

I have a rawls branch to remove that [here](https://github.com/broadinstitute/rawls/compare/mb-remove-stats-endpoint), but it can sit on the back burner for a bit.

davidangb:
@MatthewBemis did you ever follow up on this? It would be great to merge this.

MatthewBemis:
@davidangb Thanks for the reminder, I just sent out a Slack message and will follow up via email as well depending on the response the Slack msg gets

MatthewBemis:
This is now clear for merge (after review of course)

davidangb (approved):
:+1: thanks!

As a follow-on, could you clean up the .conf file, too? https://github.com/broadinstitute/firecloud-develop/blob/dev/base-configs/firecloud-orchestration/firecloud-orchestration.conf.ctmpl#L343

MatthewBemis:
@davidangb good call, just opened https://github.com/broadinstitute/firecloud-develop/pull/2395

ansingh7115 (approved):

PR #830, opened 2021-01-15
AS-560: don't use passthrough when extracting entity [risk: low]

davidangb:
Explanation of this PR is here: https://github.com/broadinstitute/firecloud-orchestration/wiki/Passthroughs#passthroughs-and-entities

Matt was spot on in his work on broadinstitute/firecloud-orchestration#823, and this PR builds on his.

This problem happens when:
* a route uses both  `entity(as[...])` and `passthrough`
* the inbound request entity is large enough that akka treats it as a stream, not an in-memory object (`HttpEntity.Default`  vs `HttpEntity.Strict`).  This condition is intermittent based on network conditions and the current state of akka's buffers -  in practical terms, it is unpredictably intermittent.

I searched the codebases for all uses of `entity(as[...])`. The route I changed in this PR is the only one that also uses `passthrough`.

Finally, I _*tried*_ to write a unit test for this, but could not get the test to fail. There may be something  different about running real server vs. running via testkit:
```
    "POST on /workspaces/.../.../methodconfigs with a non-strict entity succeeds" in {
      // create a large http entity, and explicitly make it HttpEntity.Default
      val manyinputs: Map[String, AttributeString] = (1 to 99999).map { idx =>
        s"input$idx" -> AttributeString(s"this.column$idx")
      }.toMap

      val config: MethodConfiguration = MethodConfiguration("namespace", "name", None, None, manyinputs, Map(),
        MethodRepoMethod("methodNamespace", "methodName", 1))
      val configJsonString: String = config.toJson.prettyPrint
      val testBytes: Array[Byte] = configJsonString.getBytes
      val byteString: ByteString = new ByteStringBuilder().putBytes(testBytes).result()
      val byteSource: Source[ByteString, Any] = Source.apply(List(byteString))
      val ent = HttpEntity.apply(ContentTypes.`application/json`, testBytes.length.toLong, byteSource)
      assert(ent.isDefault(), "test setup incorrect; fixture entity not of type HttpEntity.Default")

      stubRawlsService(HttpMethods.POST, methodconfigsPath, ImATeapot)

      Post(methodconfigsPath, ent) ~> dummyUserIdHeaders(dummyUserId) ~> sealRoute(workspaceRoutes) ~> check {
        status should be (ImATeapot)
      }
    }
```

---

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiService.scala
@@ -97,12 +97,14 @@ trait WorkspaceApiService extends FireCloudRequestBuilding with FireCloudDirecti
                     }
                   } ~
                     post {
-                      requireUserInfo() { _ =>
+                      requireUserInfo() { userInfo =>
                         entity(as[MethodConfiguration]) { methodConfig =>
-                          if (!methodConfig.outputs.exists { param => param._2.value.startsWith("this.library:") || param._2.value.startsWith("workspace.library:")})
-                            passthrough(workspacePath + "/methodconfigs", HttpMethods.GET, HttpMethods.POST)
-                          else
+                          if (!methodConfig.outputs.exists { param => param._2.value.startsWith("this.library:") || param._2.value.startsWith("workspace.library:")}) {
+                            val passthroughReq = Post(workspacePath + "/methodconfigs", methodConfig)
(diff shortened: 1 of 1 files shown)


davidangb:
jenkins retest

MatthewBemis (approved):

davidangb:
jenkins/swat tests are in bad shape right now, I will re-run shortly once they recover

davidangb:
is jenkins happier now? jenkins retest.

PR #839, opened 2021-03-04
SA health check resiliency/logging [AS-667]

davidangb:
AS-667.

* modifies the low-level `RestJsonClient.resultsToObject` to bubble up detail about deserialization/unmarshalling errors. This is called from many daos/services, so it's worth attention from reviewers.
* modifies the `is_admin_sa_registered` health check to log more info, including the aforementioned newly-exposed deserialization error messages
* modifies the `is_admin_sa_registered` health check to retry one time the SA's registration lookup, if the first lookup failed but the self-healing registration attempt resulted in a 409 conflict

This PR modifies the message displayed in orch's /status API in the case where orch cannot successfully read the admin SA's registration status.  Previously, this would typically result in orch showing an "Admin SA already exists!" message, because orch tried to self-heal and register the SA after it couldn't verify it was registered. Now, orch will display the error it  encountered when trying to read the registration. For instance, if orch can't deserialize Sam's registration response, orch will display a DeserializationException message.

before:
```
'is_admin_sa_registered': {
	'messages': [
		'Admin SA already exists!'
	],
	'ok': false
}
```

after:
```
'is_admin_sa_registered': {
	'messages': [
		'Error on registration status for Admin SA: could not deserialize response: Object is missing required member 'userDetails' (Sam)'
	],
	'ok': false
}
```

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/HealthChecks.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/HealthChecks.scala
@@ -60,8 +60,8 @@ class HealthChecks(app: Application, registerSAs: Boolean = true)
         } else {
           Future.successful(Option(s"$name already exists!"))
         }
-      case errorReport: FireCloudExceptionWithErrorReport =>
-        Future.successful(Option(s"Error on registration status for $name: ${errorReport.getMessage}"))
+      case e: FireCloudExceptionWithErrorReport =>
+        Future.successful(Option(s"Error on registration status for $name: ${e.errorReport.message} (${e.errorReport.source})"))
       case e: Exception =>
         Future.successful(Option(s"Error on registration status for $name: ${e.getMessage}"))
     }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/HealthChecks.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/HealthChecks.scala
@@ -31,8 +31,7 @@ class HealthChecks(app: Application, registerSAs: Boolean = true)
     val lookup = manageRegistration(name, app.samDAO.getRegistrationStatus(token))
     lookup flatMap {
       case Some(err) if registerSAs =>
-        logger.warn(s"SA registration lookup found: $err")
-        logger.info(s"attempting to register $name ...")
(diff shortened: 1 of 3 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/utils/RestJsonClient.scala
```diff
@@ -112,6 +112,9 @@ trait RestJsonClient extends FireCloudRequestBuilding with PerformanceLogging {
       response.status match {
         case s if s.isSuccess =>
           Unmarshal(response.entity).to[T].recoverWith {
+            case de: DeserializationException =>
+              throw new FireCloudExceptionWithErrorReport(
+                ErrorReport(s"could not deserialize response: ${de.msg}"))
```
davidangb, line 117: the most common case of an error in unmarshalling is `DeserializationException`; this bubbles up the underlying error.

src/main/scala/org/broadinstitute/dsde/firecloud/HealthChecks.scala
```diff
@@ -28,28 +28,44 @@ class HealthChecks(app: Application, registerSAs: Boolean = true)
...
   private def maybeRegisterServiceAccount(name: String, token: AccessToken): Future[Option[String]] = {
-    val lookup = manageRegistration(name, app.samDAO.getRegistrationStatus(implicitly(token)))
+    val lookup = manageRegistration(name, app.samDAO.getRegistrationStatus(token))
     lookup flatMap {
-      case Some(_) if registerSAs =>
+      case Some(err) if registerSAs =>
+        logger.warn(s"SA registration lookup found: $err")
         logger.info(s"attempting to register $name ...")
-        manageRegistration(name, app.samDAO.registerUser(implicitly(token)))
+        manageRegistration(name, app.samDAO.registerUser(token))
```
davidangb, line 35: `implicitly`  is not needed here - the sam dao code may have changed since this was written, or we didn't need `implicitly` in the first place, I don't know

davidangb:
@s-rubenstein do you have opinions on this PR, since you ran into the issue? I can get code review within the team, but welcome any feedback (code-level or feature-level) if you have it

ansingh7115 (approved):

s-rubenstein:
This approach looks good to me!

PR #844, opened 2021-04-01
handle non-strict responses from Import Service [QA-1230]

davidangb:
This solves _*one*_ of the test failures called out in QA-1230.

When `HttpImportServiceDAO.importPFB()` received an error response (non-`Created` status code) from Import Service, it only handled strict http entities properly. For those strict entities, it bubbled up the response body from Import Service.

However, for non-strict entities, e.g. `HttpEntity.Default`, it would call `toString()` on the entire `HttpResponse` and use *that* as its response body. This was a very different response.

In practice, this behavior meant that:
* strict entities from import service would result in a message like "Cannot perform the action write on $projectName/$workspaceName" to the end user
* non-strict entities from import service would result in a message like "HttpResponse(403 Forbidden,List(Date: Wed, 31 Mar 2021 17:37:40 GMT, Server: gunicorn/20.0.4),HttpEntity.Default(text/html; charset=UTF-8,153 bytes total),HttpProtocol(HTTP/1.1))" to the end user

And then in turn, this would cause a swatomation test to fail, because the test inspected the response body: https://github.com/broadinstitute/firecloud-orchestration/blob/dcaaa90f2c3ba70391033e0b8ecfacb9cd33df57/automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/PFBImportSpec.scala#L153

Akka can and does decide on its own, based on available memory and network conditions, whether the entity it receives is strict or not (see https://doc.akka.io/docs/akka-http/current/common/http-model.html#httpentity). Thus we must be able to handle both cases.

**_Reviewer:_** codecov seems to hate this PR. You might want to "hide annotations".

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpImportServiceDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpImportServiceDAO.scala
@@ -15,6 +15,7 @@ import org.broadinstitute.dsde.firecloud.utils.RestJsonClient
 import org.broadinstitute.dsde.rawls.model.WorkspaceName
 
 import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.duration._
 
 class HttpImportServiceDAO(implicit val system: ActorSystem, implicit val materializer: Materializer, implicit val executionContext: ExecutionContext)
   extends ImportServiceDAO with RestJsonClient with SprayJsonSupport {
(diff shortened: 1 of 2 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpImportServiceDAO.scala
```diff
@@ -45,12 +53,20 @@ class HttpImportServiceDAO(implicit val system: ActorSystem, implicit val materi
...
       case otherResp =>
         // see if we can extract errors
-        val responseString = otherResp.entity match {
-          case HttpEntity.Strict(_, data) => data.utf8String
-          case _ => otherResp.toString()
+        val responseStringFuture = otherResp.entity match {
+          case HttpEntity.Strict(_, data) =>
+            Future.successful(data.utf8String)
+          case nonStrictEntity =>
+            nonStrictEntity.toStrict(10.seconds).map(_.data.utf8String)
```
davidangb, line 60: calling `toStrict` can be a memory hog for very large responses. In this case, we know that the responses from import service are sane and small, so this feels ok.

ungwudik (approved):
Looks good to me. Thanks for adding the tests. 

It may be worth adding a comment to the ticket QA-1230…to note that this fix is covering one out of potentially-different root causes of the 403 errors that were reported. If we (QA) see 403s in the future, we'll aim to separate them as much as possible.

MatthewBemis (approved):

PR #845, opened 2021-04-01
specs should be no-arg [AS-713]

davidangb:
`HttpGoogleServicesDAOSpec` has not been running its tests, because it needs to be no-arg to be picked up by the test runner.


MatthewBemis (approved):
Oh god... nice catch

PR #846, opened 2021-04-02
[CA-1222] Set iats for Shibboleth generated test JWTs to distant future

marctalbott:
Ticket: [CA-1222](https://broadworkbench.atlassian.net/browse/CA-1222)
* Change MockThurloeDAO so that it overwrites existing keys to prevent duplicate keys with different values.
* Update Shibboleth-generated JWTs to have iat in distant future. This will remove the need to rotate these tokens on a monthly basis
* Update relinking unit test to not check that expiration date is 30 days in future. The implementation here has changed and we now set expiration date to 30 days past the iat instead of 30 days from the current time

In all cases:


marctalbott:
Including this note from the firecloud-automated-testing to give some additional context on these changes

> We considered reverting the change in Orch so that it would generate the expiration date based on the current date rather than the token's iat, but we opted for this approach because it doesn't allow for reusing the same token without relinking with the NIH through Shibboleth.

> There was also some discussion of whether Orch should accept tokens with an iat thousands of years in the future. Orch trusts that these JWTs have been generated by Shibboleth because they have been signed with its private key. Shibboleth generates the iat automatically when it generates the JWT. Given that, I am confident that tokens with modified iats (like these test ones) will not crop up in the real world without Shibboleth's private key being exposed or some bug in the library Shibboleth uses to generate the JWTs which would constitute a far larger security breach.

MatthewBemis (approved):
LGTM

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockThurloeDAO.scala
```diff
@@ -134,7 +134,11 @@ class MockThurloeDAO extends ThurloeDAO {
   }
 
   override def saveKeyValues(userInfo: UserInfo, keyValues: Map[String, String]): Future[Try[Unit]] = {
-    val newKVsForUser = (userInfo.id -> (mockKeyValues(userInfo.id) ++ keyValues.map { case (key, value) => FireCloudKeyValue(Option(key), Option(value))}))
+    val newKVsForUser = userInfo.id -> (mockKeyValues(userInfo.id).filter {
+      case FireCloudKeyValue(Some(key), _) => !keyValues.contains(key)
+      case FireCloudKeyValue(_, _) => false
```
andy7i, line 139: good catch.. I spent a while trying to understand why we need this -- we need to join 2 sets of type ` Set[FireCloudKeyValue]` and we can't automatically replace key-value pairs because they're in the `FireCloudKeyValue` case class and not in a standard data structure like a Map.

andy7i (approved):

PR #850, opened 2021-04-26
PROD-528: Use user pet service account to fetch bucket information

yonghaoy:
We are getting this error when fetching bucket information to calculate bucket cost:
`firecloud-prod@broad-dsde-prod.iam.gserviceaccount.com does not have storage.buckets.get access to the Google Cloud Storage bucket.`
This is not captured in lower env because in lower env, `firecloud-env@ SA` has permission to those bucket. 

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
@@ -1,7 +1,6 @@
 package org.broadinstitute.dsde.firecloud.dataaccess
 
-import java.io.FileInputStream
-
+import java.io.{ByteArrayInputStream, FileInputStream}
 import akka.actor.ActorSystem
 import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport
 import akka.http.scaladsl.model.StatusCodes._
@@ -187,8 +186,11 @@ class HttpGoogleServicesDAO(implicit val system: ActorSystem, implicit val mater
     storage.objects().get(bucketName, objectKey).executeMediaAsInputStream
   }
 
-  def getBucket(bucketName: String, userAuthToken: WithAccessToken): Option[Bucket] = {
-    val storage = new Storage.Builder(httpTransport, jsonFactory, new HttpCredentialsAdapter(userAuthToken).setApplicationName(appName).build()
+  def getBucket(bucketName: String, petKey: String): Option[Bucket] = {
+    val keyStream = new ByteArrayInputStream(petKey.getBytes)
+    val credential = ServiceAccountCredentials.fromStream(keyStream).createScoped(storageReadOnly.asJava)
+
+    val storage = new Storage.Builder(httpTransport, jsonFactory, new HttpCredentialsAdapter(credential)).setApplicationName(appName).build()
 
     Try(executeGoogleRequest[Bucket](storage.buckets().get(bucketName))) match {
       case Failure(ex) =>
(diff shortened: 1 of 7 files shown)


wnojopra (approved):
LGTM

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -187,8 +186,11 @@ class HttpGoogleServicesDAO(implicit val system: ActorSystem, implicit val mater
     storage.objects().get(bucketName, objectKey).executeMediaAsInputStream
   }
 
-  def getBucket(bucketName: String): Option[Bucket] = {
-    val storage = new Storage.Builder(httpTransport, jsonFactory, new HttpCredentialsAdapter(getBucketServiceAccountCredential)).setApplicationName(appName).build()
+  def getBucket(bucketName: String, petKey: String): Option[Bucket] = {
+    val keyStream = new ByteArrayInputStream(petKey.getBytes)
+    val credential = ServiceAccountCredentials.fromStream(keyStream).createScoped(storageReadOnly.asJava)
```
davidangb: can this use `getScopedServiceAccountCredentials` to avoid copy/paste?
```
val credential = getScopedServiceAccountCredentials(ServiceAccountCredentials.fromStream(keyStream), storageReadOnly)
```

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -50,7 +51,8 @@ class WorkspaceService(protected val argUserToken: WithAccessToken, val rawlsDAO
 
   def getStorageCostEstimate(workspaceNamespace: String, workspaceName: String): Future[RequestComplete[WorkspaceStorageCostEstimate]] = {
     rawlsDAO.getWorkspace(workspaceNamespace, workspaceName) flatMap { workspaceResponse =>
-      googleServicesDAO.getBucket(workspaceResponse.workspace.bucketName) match {
+      samDao.getPetServiceAccountKeyForUser(userToken, GoogleProject(workspaceNamespace)) flatMap { petKey =>
+      googleServicesDAO.getBucket(workspaceResponse.workspace.bucketName, petKey) match {
```
davidangb: nit: indentation

davidangb (approved):
:+1: approving. I have two small comments inline, feel free to tackle those in a follow-on PR if you need to get this out in a hotfix.

davidangb:
@yonghaoy one more question: does this actually need to execute as the user's pet? Does the user have the right permissions? If so, it could be simpler to run as the user instead of the pet.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
yonghaoy: Done

yonghaoy:
> @yonghaoy one more question: does this actually need to execute as the user's pet? Does the user have the right permissions? If so, it could be simpler to run as the user instead of the pet.

Hi @davidangb Actually I am not sure if user has right permission(not familiar with general user pet vs user account actually). The only thing I know: The credential need to have permission to the workspace bucker(Terra Bucket Reader?) Not sure if user has that permission, but I know user pet has that. @MatthewBemis may know?

MatthewBemis:
I think the user should work but I'm not sure that we'd be able to use the Google java lib with just the user's access token, but maybe there's some support for alternative creds buried in there that I haven't seen.

That said, we could definitely just make a direct HTTP call to Google using the users token.

This is probably worthy of some follow-up optimization once the hotfix goes out.

davidangb:
I think you can do `val credential = OAuth2Credentials.create(accessToken)` to pass on the user's access token ... but after letting it roll around a bit more, the problem will be that the user's access token has the wrong scopes (doesn't have storage-read scope), so the pet is probably the right approach.

PR #861, opened 2021-07-12
async TSV uploads [AS-576]

davidangb:
AS-576

This PR adds an "async" query parameter to the TSV upload API. For upsert and membership TSVs, Orch will now parse the TSV, write the Rawls payload to Import Service's bucket, then trigger an async import via Import Service.

Open question: update TSVs (which only update, do not insert) are not supported by Import Service. If an end user specifies "async=true" but sends an update TSV, what should we do? As-is in this PR, we silently ignore the async instruction and import that TSV synchronously; we could instead return a BadRequest. Or something else?

Integration tests will follow in AS-878.

This functionality depends on a Terraform/IAM change to grant the relevant SA access to write to the bucket: AS-876.


davidangb:
jenkins retest

davidangb:
jenkins retest

project/Dependencies.scala
```diff
@@ -34,8 +34,10 @@ object Dependencies {
       exclude("com.typesafe.scala-logging", "scala-logging_2.12")
       exclude("com.typesafe.akka", "akka-stream_2.12")
       exclude("com.google.code.findbugs", "jsr305")
+      exclude("bio.terra", "workspace-manager-client")
       excludeAll(excludeAkkaHttp, excludeSprayJson),
-    excludeGuava("org.broadinstitute.dsde.workbench" %% "workbench-util"  % "0.3-12b7791-SNAP"),
+    excludeGuava("org.broadinstitute.dsde.workbench" %% "workbench-util"  % "0.5-4bc7050"),
```
davidangb, line 39: version 0.3 had dependency conflicts. Latest is 0.6, but 0.6 also has dependency conflicts because it pulls in later versions of akka. So, I upgraded to version 0.5

project/Dependencies.scala
```diff
@@ -34,8 +34,10 @@ object Dependencies {
       exclude("com.typesafe.scala-logging", "scala-logging_2.12")
       exclude("com.typesafe.akka", "akka-stream_2.12")
       exclude("com.google.code.findbugs", "jsr305")
+      exclude("bio.terra", "workspace-manager-client")
```
davidangb, line 37: added this exclusion to avoid dependency conflicts.

project/Dependencies.scala
```diff
@@ -74,7 +76,12 @@ object Dependencies {
...
 
-    "org.scalatest"                 %% "scalatest"           % "3.2.5"   % "test",
-    "org.mock-server"                % "mockserver-netty"    % "3.10.8"  % "test"
+    "org.scalatest"                 %% "scalatest"           % "3.2.9"   % "test",
+    "org.mock-server"                % "mockserver-netty"    % "3.10.8"  % "test",
+    // jaxb-api needed by WorkspaceApiServiceSpec.bagitService() method
+    "javax.xml.bind"                 % "jaxb-api"            % "2.3.1"   % "test",
+    // provides testing mocks
+    "com.google.cloud"               % "google-cloud-nio"    % "0.123.7" % "test",
+    "org.scalatestplus"             %% "mockito-3-4"         % "3.2.9.0" % "test"
```
davidangb, line 85: I'm not sure why we suddenly need to explicitly specify jaxb; I think it was previously supplied by workspace-manager-client, which is now excluded above.

project/Merging.scala
```diff
@@ -3,8 +3,12 @@ import sbtassembly.{MergeStrategy, PathList}
 object Merging {
   def customMergeStrategy(oldStrategy: (String) => MergeStrategy):(String => MergeStrategy) = {
     case x if x.endsWith("io.netty.versions.properties") => MergeStrategy.discard
+    case x if x.contains("native-image/io.netty") => MergeStrategy.first
+
+    // TODO: we no longer target Java 8, reassess this:
     // we target Java 8, which does not use module-info.class. Some dependencies (Jackson) cause assembly problems on module-info
     case x if x.endsWith("module-info.class") => MergeStrategy.discard
+
```
davidangb, line 11: added this TODO comment because I don't want to scope-creep this PR too much

project/Merging.scala
```diff
@@ -3,8 +3,12 @@ import sbtassembly.{MergeStrategy, PathList}
 object Merging {
   def customMergeStrategy(oldStrategy: (String) => MergeStrategy):(String => MergeStrategy) = {
     case x if x.endsWith("io.netty.versions.properties") => MergeStrategy.discard
+    case x if x.contains("native-image/io.netty") => MergeStrategy.first
```
davidangb, line 6: workbench-libs google2 brings in a shadowed netty, this merge strategy is now necessary.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
```diff
@@ -192,16 +187,50 @@ class EntityService(rawlsDAO: RawlsDAO, importServiceDAO: ImportServiceDAO, mode
...
+
+    // write rawlsCalls to import service's bucket
+    // TODO: Rawls SA needs write permission, not just read; Storage Object Creator is fine
+    val dataBytes = rawlsCalls.toJson.prettyPrint.getBytes(StandardCharsets.UTF_8)
+    val insertedObject = googleServicesDAO.writeObjectAsRawlsSA(bucketToWrite, fileToWrite, dataBytes)
+    val gcsPath = s"gs://${insertedObject.bucketName.value}/${insertedObject.objectName.value}"
+
+    // TODO: this is functional, but the class name "PfbImportRequest" is misleading; rename it?
+    val importRequest = PfbImportRequest(Option(gcsPath))
+    importServiceDAO.importBatchUpsertJson(workspaceNamespace, workspaceName, importRequest)(userInfo)
```
davidangb, line 234: renaming the `PfbImportRequest` class causes small changes across many lines of code; I'd like to do that as a separate follow-on PR, to keep this one's diff as small as possible

src/test/scala/org/broadinstitute/dsde/firecloud/EntityServiceSpec.scala
```diff
@@ -77,7 +86,7 @@ class EntityServiceSpec extends BaseServiceSpec with BeforeAndAfterEach {
       val zip = new ZipFile("src/test/resources/testfiles/bagit/duplicate_participants_nested_testbag.zip")
 
       val ex = intercept[FireCloudExceptionWithErrorReport] {
-        EntityService.unzipTSVs("duplicate_participants_nested_testbag", zip) { (participants, samples) =>
+        EntityService.unzipTSVs("duplicate_participants_nested_testbag", zip) { (_, _) =>
```
davidangb, line 89: this change is not necessary as part of this PR, but it resolved an IntelliJ warning and I was trying to make this entire file clean

src/test/scala/org/broadinstitute/dsde/firecloud/EntityServiceSpec.scala
```diff
@@ -88,11 +97,125 @@ class EntityServiceSpec extends BaseServiceSpec with BeforeAndAfterEach {
...
+          val entityService = getEntityService(mockImportServiceDAO = testImportDAO)
+          val response =
+            entityService.importEntitiesFromTSV("workspaceNamespace", "workspaceName",
+              tsvData, userToken, isAsync = true).futureValue
+          response shouldBe testImportDAO.successDefinition
+        }
+    }
+
+    // should this case throw a BadRequest? or just ignore the async instruction and update synchronously?
+    "should return ??? for (async=true + update TSV)" is pending
```
davidangb: this is the big open question I have about this PR!

davidangb:
no space left on device - jenkins retest

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -174,6 +180,51 @@ class HttpGoogleServicesDAO(implicit val system: ActorSystem, implicit val mater
...
   }
 
+  /**
+    * Uploads the supplied data to GCS, using the Rawls service account credentials
+    * @param bucketName target bucket name for upload
+    * @param objectKey target object name for upload
+    * @param objectContents byte array of the data to upload
+    * @return path to the uploaded GCS object
+    */
+  override def writeObjectAsRawlsSA(bucketName: GcsBucketName, objectKey: GcsObjectName, objectContents: Array[Byte]): GcsPath = {
```
MatthewBemis, line 190: i know this is not a pattern introduced here but yuck re: rawls SA being used in orch

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
```diff
@@ -200,6 +200,7 @@ object FireCloudConfig {
 
   object ImportService {
     lazy val server: String = if (config.hasPath("importService.server")) config.getString("importService.server") else ""
+    lazy val bucket: String = if (config.hasPath("importService.bucketName")) config.getString("importService.bucketName") else "[importService.bucketName not specified]"
```
sagehen03: I'm wondering if we should throw an exception here and kill the process if the bucket name is not specified in the config?  Generally, I prefer failing early though I suspect the downstream error due to a bucket with the name [importService.bucketName not specified] will be pretty clear.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -174,6 +180,51 @@ class HttpGoogleServicesDAO(implicit val system: ActorSystem, implicit val mater
...
+    val uploadAttempt = storageResource.use { storageService =>
+      // create the destination pipe to which we will write the file
+      // N.B. workbench-libs' streamUploadBlob does not allow setting the Content-Type, so we don't set it
+      val uploadPipe = storageService.streamUploadBlob(bucketName, GcsBlobName(objectKey.value))
+      // stream the data to the destination pipe
+      dataStream.through(uploadPipe).compile.lastOrError
+    }
+
+    // execute the upload
+    uploadAttempt.unsafeRunSync()
```
sagehen03, line 222: I'm surprised we don't have existing scala code for writing a byte array to a bucket :thinking:

src/test/scala/org/broadinstitute/dsde/firecloud/EntityServiceSpec.scala
sagehen03: I think we should throw a BadRequest.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpImportServiceDAO.scala
```diff
@@ -21,10 +21,17 @@ class HttpImportServiceDAO(implicit val system: ActorSystem, implicit val materi
...
+
+  override def importBatchUpsertJson(workspaceNamespace: String, workspaceName: String, pfbRequest: PfbImportRequest)(implicit userInfo: UserInfo): Future[PerRequestMessage] = {
+    doImport(workspaceNamespace, workspaceName, pfbRequest, "rawlsjson")
+  }
 
+  private def doImport(workspaceNamespace: String, workspaceName: String, pfbRequest: PfbImportRequest, filetype: String)(implicit userInfo: UserInfo): Future[PerRequestMessage] = {
     // the payload to Import Service sends "path" and filetype.  Here, we force-hardcode filetype because this API
     // should only be used for PFBs.
-    val importServicePayload: ImportServiceRequest = ImportServiceRequest(path = pfbRequest.url.getOrElse(""), filetype = "pfb")
+    val importServicePayload: ImportServiceRequest = ImportServiceRequest(path = pfbRequest.url.getOrElse(""), filetype = filetype)
```
sagehen03: Could we throw an exception if pfbRequest.url is null?  Setting it to empty string will cause an error further from the calling code in import service I believe?

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAOSpec.scala
```diff
@@ -134,4 +144,54 @@ class HttpGoogleServicesDAOSpec extends AnyFlatSpec with Matchers with PrivateMe
...
+    val objectName = GcsObjectName("folder/object.json")
+
+    val objectContents = "Hello world".getBytes(StandardCharsets.UTF_8)
+
+    assertResult(GcsPath(bucketName, objectName)) {
+      gcsDAO.streamUploadObject(localStorage, bucketName, objectName, objectContents)
+    }
+  }
+
+  it should "throw error for an unsuccessful object upload" in {
```
sagehen03, line 163: This test feels like we we're just verifying that if we mock Storage to throw an exception that it does in fact throw an exception  i.e. testing the mock framework rather than our code, but it's possible that I'm missing something.

sagehen03 (approved):
Just a few questions mostly for my own self improvement.  Looks good.

src/test/scala/org/broadinstitute/dsde/firecloud/EntityServiceSpec.scala
MatthewBemis: +1 to bad request

MatthewBemis (approved):
lgtm!

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
davidangb: Good catch; this is leftover jamboree code from before the conf file changes merged. I've updated to fail if bucketName (or server) is not found.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpImportServiceDAO.scala
davidangb: Another good catch. Previously (https://github.com/broadinstitute/firecloud-orchestration/pull/777/files#r395754726) I'd been passing on potentially-invalid payloads to Import Service, and letting Import Service do all the validation and produce all the error messages. But now that we're handling both TSVs (which orch mostly owns) and PFBs (which Import Service owns) I agree about validating here. I've updated the code.

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAOSpec.scala
davidangb: The intention of this test is to ensure that if Storage throws an exception, we bubble that exception up through the other layers inside `streamUploadObject`. I admit there's only a thin additional layer, but I'd like to keep the test, unless you object

davidangb:
@MatthewBemis @sagehen03 this PR now returns BadRequest if the user specifies async=true with an update TSV … however, I haven't put too much effort into that code, since it'll get replaced to work properly very soon. I've also made a few other tweaks based on Drew's feedback. Thanks, and ready for re-review!

sagehen03 (approved):
lgtm

PR #862, opened 2021-07-15
remove authorizations that are not present

dvoet:
https://broadworkbench.atlassian.net/browse/POL-62

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/OrchestrationApiSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/OrchestrationApiSpec.scala
@@ -81,10 +81,6 @@ class OrchestrationApiSpec extends FreeSpec with Matchers with ScalaFutures with
 
         Orchestration.NIH.addUserInNIH(OrchConfig.Users.genericJsonWebTokenKey)
 
-        verifyDatasetPermissions(Set(NihDatasetPermission("TCGA", true), NihDatasetPermission("TARGET", true)))
-
-        Orchestration.NIH.syncWhitelistFull
-
         verifyDatasetPermissions(Set(NihDatasetPermission("TCGA", false), NihDatasetPermission("TARGET", false)))
       }
     }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpSamDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpSamDAO.scala
@@ -106,7 +106,9 @@ class HttpSamDAO( implicit val system: ActorSystem, val materializer: Materializ
 
   private def userAuthedRequestToUnit(request: HttpRequest)(implicit userInfo: WithAccessToken): Future[Unit] = {
     userAuthedRequest(request).map { resp =>
-      if(resp.status.isSuccess) Future.successful(())
+      if(resp.status.isSuccess) Future.successful({
+        resp.discardEntityBytes()
+      })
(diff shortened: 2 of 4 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpSamDAO.scala
```diff
@@ -106,7 +106,9 @@ class HttpSamDAO( implicit val system: ActorSystem, val materializer: Materializ
 
   private def userAuthedRequestToUnit(request: HttpRequest)(implicit userInfo: WithAccessToken): Future[Unit] = {
     userAuthedRequest(request).map { resp =>
-      if(resp.status.isSuccess) Future.successful(())
+      if(resp.status.isSuccess) Future.successful({
+        resp.discardEntityBytes()
+      })
```
dvoet, line 111: this should fix errors in logs like
```[WARN] [07/15/2021 17:00:22.965] [FireCloud-Orchestration-API-akka.actor.default-dispatcher-16] [FireCloud-Orchestration-API/Pool(shared->https://sam.dsde-dev.broadinstitute.org:443)] [1 (WaitingForResponseEntitySubscription)]Response entity was not subscribed after 10 seconds. Make sure to read the response `entity` body or call `entity.discardBytes()` on it -- in case you deal with `HttpResponse`, use the shortcut `response.discardEntityBytes()`. PUT /api/group/TCGA-dbGaP-Authorized/member Strict(2 bytes) -> 201 Created Chunked```

**Comment by dvoet** [comment]
jenkins retest

File: src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpSamDAO.scala (hunk 1 above)
Comment (davidangb) [reply src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpSamDAO.scala:111]: thank you - we tried to catch all these when migrating from spray->akka but know we missed a few

File: automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/OrchestrationApiSpec.scala [hunk 2]
```diff
@@ -81,10 +81,6 @@ class OrchestrationApiSpec extends FreeSpec with Matchers with ScalaFutures with
 
         Orchestration.NIH.addUserInNIH(OrchConfig.Users.genericJsonWebTokenKey)
 
-        verifyDatasetPermissions(Set(NihDatasetPermission("TCGA", true), NihDatasetPermission("TARGET", true)))
-
-        Orchestration.NIH.syncWhitelistFull
-
```
dvoet, line 87: @MatthewBemis remember adding this test 3 years ago? I don't know if you wrote it then or just moved it from somewhere else but this part seems wrong and exactly what this PR is fixing. When a user links a different NIH account they should lose access to anything the original has access to but the new does not.

MatthewBemis: I think I wrote it but I don't remember the specifics. If I had to guess, this test was backfill for functionality that had already existed in prod for a while so I probably made the dangerous assumption that it was all correct, and then just shaped the test around that incorrect existing behavior. Or it's bad copy-paste from another test case. Either way, it's definitely wrong

dvoet: 👍

davidangb (approved):
code changes lgtm; I did not test it myself

PR #874, opened 2021-08-21
Handle new consent status format

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DUOS-1411

This PR adds support for an updated Consent Status that includes an overall OK state, a Degraded state, moves the dropwizard checks into a list of Systems, and adds Sam as a new dependency. This will parse both the new form and the old

See also: https://github.com/DataBiosphere/consent/pull/1235

---

In all cases:


JVThomas (approved):
Looks good. New status are added and there's filtering on the error messages to prevent internals from leaking out.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/ReportsSubsystemStatus.scala
```diff
@@ -24,15 +25,27 @@ trait ReportsSubsystemStatus extends SprayJsonSupport {
...
+          Unmarshal(resp).to[Map[String, DropwizardHealth]].map { dwStatus =>
+            val ok = dwStatus.values.forall(_.healthy)
+            val errors = dwStatus.
+              filter(dw => !dw._2.healthy).
+              map(dw => s"Error in ${dw._1}: ${dw._2.message.getOrElse("unspecified error")}").toList
+            if (ok)
+              SubsystemStatus(ok, None)
+            else
+              SubsystemStatus(ok, Some(errors))
+          }
```
rushtong, line 48: This is the original logic to parsing consent status that is now used only when the new parsing fails. When https://github.com/DataBiosphere/consent/pull/1235 is merged, the new status format will take effect.

MatthewBemis (approved):
LGTM

PR #879, opened 2021-09-01
AS-245: Improve resilience of storage cost estimate endpoint

MatthewBemis:
Improvements:

- only call the pricelist once when booting, then cache it as a `val`
- if the pricelist isn't available or can't be parsed, default to a hardcoded json and log an error so we can fix it at our leisure

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
@@ -466,7 +466,7 @@ class HttpGoogleServicesDAO(priceListUrl: String, defaultPriceList: GooglePriceL
   //Because the price list is brittle and Google sometimes changes the names of keys in the JSON, there is a
   //default cached value in configuration to use as a backup. If we fallback to it, the error will be logged
   //but users will probably not notice a difference. They're cost *estimates*, after all.
-  val fetchPriceList: Future[GooglePriceList] = {
+  lazy val fetchPriceList: Future[GooglePriceList] = {
     val httpReq = Get(priceListUrl)
 
     unAuthedRequestToObject[GooglePriceList](httpReq).recover {
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
@@ -104,7 +104,7 @@ object HttpGoogleServicesDAO {
   }
 }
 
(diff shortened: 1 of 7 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -461,10 +461,19 @@ class HttpGoogleServicesDAO(implicit val system: ActorSystem, implicit val mater
...
+  //every time we want to calculate a cost estimate (which happens extremely often in the Terra UI)
+  //Because the price list is brittle and Google sometimes changes the names of keys in the JSON, there is a
+  //default cached value in configuration to use as a backup. If we fallback to it, the error will be logged
+  //but users will probably not notice a difference. They're cost *estimates*, after all.
+  val fetchPriceList: Future[GooglePriceList] = {
+    val httpReq = Get(priceListUrl)
+
+    unAuthedRequestToObject[GooglePriceList](httpReq).recover {
+      case t: Throwable =>
+        logger.warn(s"Unable to fetch/parse latest Google price list. A cached (possibly outdated) value will be used instead. Error: ${t.getMessage}")
```
davidangb: nit: make this an error, to ensure we get notified via Sentry or otherwise about failures

davidangb (approved):
one nit, otherwise :+1:

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -461,10 +461,19 @@ class HttpGoogleServicesDAO(implicit val system: ActorSystem, implicit val mater
...
-  def fetchPriceList(implicit executionContext: ExecutionContext): Future[GooglePriceList] = {
-    val httpReq = Get(FireCloudConfig.GoogleCloud.priceListUrl)
-
-    unAuthedRequestToObject[GooglePriceList](httpReq)
+  //Why is this a val? Because the price lists do not change very often. This prevents making an HTTP call to Google
+  //every time we want to calculate a cost estimate (which happens extremely often in the Terra UI)
+  //Because the price list is brittle and Google sometimes changes the names of keys in the JSON, there is a
+  //default cached value in configuration to use as a backup. If we fallback to it, the error will be logged
+  //but users will probably not notice a difference. They're cost *estimates*, after all.
+  val fetchPriceList: Future[GooglePriceList] = {
```
davidangb: sorry, one more thought: should this be `lazy`?

If it's `lazy`, then we'd avoid querying Google every time a new `HttpGoogleServicesDAO` is constructed, for instance if someone wrote a set of unit tests that constantly make new daos.

If it's NOT `lazy`, then by eagerly querying Google on construction of the DAO, we avoid any potential pauses or race conditions when the price list is needed.

I lean towards making it lazy.

davidangb (approved):
re-:+1:

ansingh7115 (approved):

PR #882, opened 2021-09-03
flaky bagit test [AS-922]

davidangb:
**REVIEWER**: I suggest you ignore whitespace, the diff is much easier to read!

-----

The test 
```
WorkspaceService BagIt Tests
- should unbundle a bagit containing both participants and samples
```
has flakily failed in both GHA and Jenkins runs. I cannot get it to fail locally so it's hard to reproduce, but this PR includes some defensive coding and some additional test cases around that coding, hopefully this either stabilizes it or helps to identify the cause. 

In this PR:
* change the method we use for downloading the BagIt. Instead of ProcessBuilder `#>` and `!!` we use `java.io`/`java.nio` channels
* a handful of new error-trapping cases and log entries

The process to import a BagIt is:
1. download BagIt to temp file
2. unzip temp file and read its contents
3. insert the contents

In all the failure cases I saw, the downloaded temp file was 0 bytes. This could be a problem with downloading (which is why I reimplemented the download method), or potentially a problem with something deleting temp files while they're still in use, or even a problem with a full disk.


davidangb:
jenkins retest

.github/workflows/unittest.yml
```diff
@@ -21,9 +21,10 @@ jobs:
       uses: coursier/cache-action@v5
 
     - name: Set up JDK 11
-      uses: actions/setup-java@v1
+      uses: actions/setup-java@v2
       with:
-        java-version: 11
+        distribution: 'adopt-hotspot'
+        java-version: '11'
```
davidangb, line 27: not necessary as part of this PR, but I was trying things and this felt good to keep in

davidangb:
jenkins retest

src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
```diff
@@ -305,41 +306,62 @@ class EntityService(rawlsDAO: RawlsDAO, importServiceDAO: ImportServiceDAO, goog
...
-                  for {
-                    // This should vomit back errors from rawls.
-                    participantResult <- participantsStr.map(ps => importEntitiesFromTSV(workspaceNamespace, workspaceName, ps, userInfo)).getOrElse(Future.successful(RequestComplete(OK)))
-                    sampleResult <- samplesStr.map(ss => importEntitiesFromTSV(workspaceNamespace, workspaceName, ss, userInfo)).getOrElse(Future.successful(RequestComplete(OK)))
-                  } yield {
-                    participantResult match {
-                      case RequestComplete((OK, _)) => sampleResult
-                      case _ => participantResult
+            // download the file
+            val readFromBagit = Channels.newChannel(bagitURL.openStream())
```
ungwudik, line 322: I'm not too familiar with `java.nio.channels.Channels`, but I'm assuming it's generally a better approach than the one being replaced?

davidangb: it's definitely less mysterious, and offers more insight, such as the number of bytes downloaded

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiServiceSpec.scala
```diff
@@ -1021,14 +1027,35 @@ class WorkspaceApiServiceSpec extends BaseServiceSpec with WorkspaceApiService w
...
+      }
+
+      "should 400 if a bagit is unzippable" in {
+        stubRawlsService(HttpMethods.POST, s"$workspacesPath/entities/batchUpsert", NoContent)
+        (Post(bagitImportPath, HttpEntity(MediaTypes.`application/json`, s"""{"bagitURL":"https://localhost:$bagitServerPort/notazip.zip", "format":"TSV" }"""))
+          ~> dummyUserIdHeaders(dummyUserId)
+          ~> sealRoute(workspaceRoutes)) ~> check {
+          status should equal(BadRequest)
+          val errRpt = responseAs[ErrorReport]
+          errRpt.message shouldBe s"Problem with BDBag: zip END header not found"
```
ungwudik, line 1057: I'm probably missing something…but how do we know to test for this particular error string? If it's external, would it always stay the same?

ungwudik (approved):

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiServiceSpec.scala
davidangb: the "zip END header not found" string comes from `java.util.zip.ZipEntry` - or one of the related classes in `java.util.zip`. Since it's part of Java itself, I don't expect it to change frequently or carelessly - but yeah, it's external

PR #884, opened 2021-09-03
enable async for update tsvs [AS-916]

davidangb:
Enables async imports for update TSVs, where previously we only supported entity and membership TSVs.


src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpImportServiceDAO.scala
```diff
@@ -21,18 +21,18 @@ class HttpImportServiceDAO(implicit val system: ActorSystem, implicit val materi
   extends ImportServiceDAO with RestJsonClient with SprayJsonSupport {
 
   override def importPFB(workspaceNamespace: String, workspaceName: String, pfbRequest: AsyncImportRequest)(implicit userInfo: UserInfo): Future[PerRequestMessage] = {
-    doImport(workspaceNamespace, workspaceName, pfbRequest, "pfb")
+    doImport(workspaceNamespace, workspaceName, isUpsert = true, pfbRequest, "pfb")
```
davidangb, line 24: PFB imports are always upsert.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
```diff
@@ -201,23 +201,17 @@ class EntityService(rawlsDAO: RawlsDAO, importServiceDAO: ImportServiceDAO, goog
...
-      if (!isUpsert) {
-        // until Rawls & Import Service implement update-only TSVs, throw a BadRequest if the user requested async
-        // but sent an update TSV.
-        Future(RequestCompleteWithErrorReport(BadRequest, "Update-only TSVs cannot use async mode"))
-      } else {
-        asyncUpsert(workspaceNamespace, workspaceName, rawlsCalls, userInfo).recover {
-          case e: Exception =>
-            RequestCompleteWithErrorReport(InternalServerError, "Unexpected error during async TSV import", e)
-        }
+      asyncUpsert(workspaceNamespace, workspaceName, isUpsert, rawlsCalls, userInfo).recover {
```
ansingh7115: Should we change code comments on lines 201-202?

MatthewBemis (approved):

src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
davidangb: yes, we should! Good catch; removed

src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
```diff
@@ -201,23 +201,17 @@ class EntityService(rawlsDAO: RawlsDAO, importServiceDAO: ImportServiceDAO, goog
...
+          RequestCompleteWithErrorReport(InternalServerError, "Unexpected error during async TSV import", e)
       }
     } else {
       val rawlsResponse = rawlsDAO.batchUpsertEntities(workspaceNamespace, workspaceName, entityType, rawlsCalls)(userInfo)
       handleBatchRawlsResponse(entityType, rawlsResponse)
     }
   }
 
-  private def asyncUpsert(workspaceNamespace: String, workspaceName: String,
+  private def asyncUpsert(workspaceNamespace: String, workspaceName: String, isUpsert: Boolean,
```
sagehen03: Seeing a method called `asyncUpsert` with a parameter called `isUpsert` is jarring to me.  It's called `asyncUpsert` so  `isUpsert` should always be true at least for the easiest comprehension.  I understand that the isUpsert name is baked in to Rawls and ImportService already so doing something like changing the param to `isUpdateOnly` and inverting the logic here in Orch will make the code less consistent with Rawls and ImportService.  What if we renamed this method to `asyncUpdateOrInsert` or just just `asyncUpdate`?  I'll leave it to you to decide if there's a sensible way to address my overloaded upsert hangup.

sagehen03 (approved):
One question around our overloaded use of "upsert", but not a blocker.

ansingh7115 (approved):

src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
davidangb: no, that's good feedback, and my fault - I had that method name left over from when it truly was upsert-only, and failed to rename it!

Also, wow, your comment called out another problem in the synchronous path, which I am fixing too.

sagehen03: Cool, good catch. High five! :hand:

sagehen03 (approved):

PR #880, opened 2021-09-01
AS-878: Integration test for async TSV upload

ansingh7115:
\<your comments for this PR go here\>

Have you read [CONTRIBUTING.md](../CONTRIBUTING.md) lately? If not, do that first.

I, the developer opening this PR, do solemnly pinky swear that:

- [ ] I've followed [the instructions](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/CONTRIBUTING.md#api-changes) if I've made any changes to the API, _especially_ if they're breaking changes
- [ ] I've updated the RC_XXX release ticket with any manual steps required to release this change
- [ ] I've updated the [FISMA documentation](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/CONTRIBUTING.md#fisma-documentation-changes) if I've made any security-related changes, including auth, encryption, or auditing

In all cases:

- [ ] Get two thumbsworth of review and PO signoff if necessary
- [ ] Verify all tests go green
- [ ] Squash and merge; you can delete your branch after this **unless it's for a hotfix**. In that case, don't delete it!
- [ ] Test this change deployed correctly and works on dev environment after deployment

PR #887, opened 2021-09-08
AS-924: Support attribute arrays in TSVs

MatthewBemis:
Support attribute arrays in TSV uploads, AS-924.

We only support:
* arrays whose elements are all the same datatype
* arrays of strings, numbers, booleans, entity references
* empty arrays

We will throw BadRequest if the TSV contains anything else.

If the value in the TSV is malformed json, e.g. `["foo">`, we will treat it as a String. In other words, we have no way to detect that you *meant* for it to be an array.

if your existing data does not contain the attribute in question, we will create the attribute.

If your existing data contains the attribute in question, we will replace it with the array in the TSV regardless of the pre-existing datatype. If your existing data already contains an array, we will replace it with the new array (we do not merge arrays)

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupport.scala
@@ -154,7 +154,7 @@ trait TSVFileSupport {
             val listNameEntry = "attributeListName" -> AttributeString(attributeName)
             def listValEntry( attr: Attribute ) = "newMember" -> attr
 
-            listElements match {
+            val addElements = listElements match {
               case elements if elements.forall(_.isInstanceOf[JsString]) =>
                 val elementsTyped: List[JsString] = elements.asInstanceOf[List[JsString]]
                 elementsTyped.map(jsstr => Map(addListMemberOperation, listNameEntry, listValEntry(AttributeString(jsstr.value))))
@@ -169,6 +169,8 @@ trait TSVFileSupport {
 
               case _ => throw new FireCloudException("Mixed-type entity attribute lists are not supported.")
             }
+            val removeOldListOp = Seq(Map(removeAttrOperation,nameEntry))
+            removeOldListOp ++ addElements
           }
           case _ => Seq(Map(upsertAttrOperation,nameEntry,valEntry(AttributeString(value))))
         }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupport.scala
@@ -176,6 +176,9 @@ trait TSVFileSupport {
   def generateAttributeArrayOperations(attributeValue: String, attributeName: String): Seq[Map[String, Attribute]] = {
     val listElements = attributeValue.parseJson.convertTo[JsArray].elements.toList
 
+    def addListEntry(attrVal: AttributeValue) =
+      Map(addListMemberOperation, listNameEntry(attributeName), listValEntry(attrVal))
+
     //if the list is empty, short-circuit and just replace any existing list with an empty list
     if(listElements.isEmpty) {
       Seq(Map(removeAttrOperation, nameEntry(attributeName)), Map(createAttrValueListOperation, nameEntry(attributeName)))
@@ -183,15 +186,15 @@ trait TSVFileSupport {
       val addElements = listElements match {
         case elements if elements.forall(_.isInstanceOf[JsString]) =>
           val elementsTyped: List[JsString] = elements.asInstanceOf[List[JsString]]
-          elementsTyped.map(jsstr => Map(addListMemberOperation, listNameEntry(attributeName), listValEntry(AttributeString(jsstr.value))))
+          elementsTyped.map(jsstr => addListEntry(AttributeString(jsstr.value)))
 
         case elements if elements.forall(_.isInstanceOf[JsNumber]) =>
           val elementsTyped: List[JsNumber] = elements.asInstanceOf[List[JsNumber]]
-          elementsTyped.map(jsnum => Map(addListMemberOperation, listNameEntry(attributeName), listValEntry(AttributeNumber(jsnum.value))))
+          elementsTyped.map(jsnum => addListEntry(AttributeNumber(jsnum.value)))
 
         case elements if elements.forall(_.isInstanceOf[JsBoolean]) =>
           val elementsTyped: List[JsBoolean] = elements.asInstanceOf[List[JsBoolean]]
-          elementsTyped.map(jsbool => Map(addListMemberOperation, listNameEntry(attributeName), listValEntry(AttributeBoolean(jsbool.value))))
+          elementsTyped.map(jsbool => addListEntry(AttributeBoolean(jsbool.value)))
 
         case _ => throw new FireCloudException("Mixed-type entity attribute lists are not supported.")
       }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupport.scala
@@ -136,6 +136,11 @@ trait TSVFileSupport {
   val createRefListOperation: (String, AttributeString) = "op" -> AttributeString("CreateAttributeEntityReferenceList")
   val createAttrValueListOperation: (String, AttributeString) = "op" -> AttributeString("CreateAttributeValueList")
 
+  def nameEntry(attributeName: String) = "attributeName" -> AttributeString(attributeName)
+  def valEntry(attr: Attribute) = "addUpdateAttribute" -> attr
+  def listNameEntry(attributeName: String) = "attributeListName" -> AttributeString(attributeName)
+  def listValEntry(attr: Attribute) = "newMember" -> attr
+
   /**
     * colInfo is a list of (headerName, refType), where refType is the type of the entity if the headerName is an AttributeRef
     * e.g. on TCGA Pairs, there's a header called case_sample_id where the refType would be Sample */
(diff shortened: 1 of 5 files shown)


src/test/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupportSpec.scala
```diff
@@ -51,4 +55,30 @@ class TSVFileSupportSpec extends AnyFreeSpec with TSVFileSupport {
     }
   }
 
+  "setAttributesOnEntity" - {
```
MatthewBemis, line 60: all of these tests need more detailed assertions. i just did bare minimum to have some reasonable confidence that things worked. i've also tested with a local orch and tsvs import correctly

i'll add more detailed tests before merging, of course

src/main/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupport.scala
```diff
@@ -143,10 +146,31 @@ trait TSVFileSupport {
...
-          case _ => Map(upsertAttrOperation,nameEntry,valEntry(AttributeString(value)))
+          case "__DELETE__" => Seq(Map(removeAttrOperation,nameEntry))
+          case value if modelSchema.isAttributeArray(value) => {
+            val listElements = value.parseJson.convertTo[JsArray].elements.toList
+            val listNameEntry = "attributeListName" -> AttributeString(attributeName)
+            def listValEntry( attr: Attribute ) = "newMember" -> attr
+
+            listElements match {
+              case elements if elements.forall(_.isInstanceOf[JsString]) =>
+                val elementsTyped: List[JsString] = elements.asInstanceOf[List[JsString]]
```
MatthewBemis: is there a cleaner way to do this than `asInstanceOf`ing everything?

i tried coercing the types by making the case statements look something like: 

`case elements: List[JsString] if elements.forall(_.isInstanceOf[JsString]) =>`

but it didn't work because it causes premature casts during pattern matching...

davidangb: What about something like this?
```
          case value if modelSchema.isAttributeArray(value) =>
            val listElements = value.parseJson.convertTo[JsArray].elements.toList

            if (listElements.isEmpty) {
              // TODO: check for zero-element array and skip other processing, what should this return?
              Seq()
            } else {
              // validate that all elements are the same type
              val elClasses = listElements.map(_.getClass).toSet
              // TODO: this fails on list-of-booleans, because it resolves to Set(JsTrue, JsFalse)
              if (elClasses.size > 1) {
                // probably should remove the output of what classes we actually found
                throw new FireCloudException(s"Mixed-type entity attribute lists are not supported. Found: ${elClasses.map(_.getName).mkString(",")}")
              }

              val listNameEntry = "attributeListName" -> AttributeString(attributeName)
              def listValEntry( attr: Attribute ) = "newMember" -> attr

              // since all elements are the same type, we can just match on each one, trusting they'll be the same
              listElements.map {
                case jsstr:JsString => Map(addListMemberOperation, listNameEntry, listValEntry(AttributeString(jsstr.value)))
                case jsnum:JsNumber => Map(addListMemberOperation, listNameEntry, listValEntry(AttributeNumber(jsnum.value)))
                case jsbool:JsBoolean => Map(addListMemberOperation, listNameEntry, listValEntry(AttributeBoolean(jsbool.value)))
                case _ => throw new FireCloudException("Only string, number, and boolean attribute lists are supported")
              }
            }
```

Note the important TODO inline; this fails the list-of-booleans unit test … but I think you get where I'm going with it and maybe this will help?

EDIT: can fix the list-of-booleans issue with
```
if (elClasses.size > 1 && elClasses.diff(Set(JsTrue.getClass, JsFalse.getClass)).nonEmpty) {
```
if that's not too ugly

src/test/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupportSpec.scala
davidangb: Could you make sure to test with a local orch all the way through Rawls for the following case?
* workspace already has entities in it
* entities already have an attribute "foo", which could be a string or could be an array (pls test both)
* TSV upload attempts to write an array to attribute "foo"

I was looking at the entity-update code in Rawls as part of my other PRs, and wonder about some of the behavior: if a list already exists, does the TSV replace it or append to it? I would expect to replace it, but I think we may append? What about the case where a String exists in its place?

MatthewBemis: Oh yikes. Yeah I think we append in the case of an existing list, and error in the case of an existing single value. I'll definitely double-check this.

MatthewBemis: Just to continue the discussion of expected behavior:

I agree with you that it should replace an existing list. Maybe Orch should generate a `RemoveAttribute` in front of creating the array. I believe Rawls processes the `op`s in order, so this should work.

In the case of an existing single-value, I would expect it to throw an error because there's a type mismatch. If there's agreement on that, then I think we're probably all set here, but again I will double check the actual behavior later this morning.

MatthewBemis: I've confirmed the following about the current state:

* uploading a TSV with an array appends to it rather than replace it.
* uploading a TSV with an array doesn't replace an existing single-value string, and throws this error: `AttributeName(default,sample_type) of OV-04-1336-TP is not a list`


It seems like sticking a `RemoveAttribute` op in front of the list add ops fixes both of these cases and allows for replacement. Do you think that's the appropriate behavior for both cases? I originally thought that it may be best to throw an error in the case of replacing a single value with a list, but I think I've changed my mind and replacement should be seamless and one step


```
            val removeOldListOp = Seq(Map(removeAttrOperation,nameEntry))
            removeOldListOp ++ addElements
```

davidangb: I agree on overwriting in either case, and the implementation of adding a `RemoveAttribute` first sounds good to me. I'd like to hear product's opinion on this too, but it all seems right to me.

davidangb:
We could probably leverage the work in #880 to add integration tests for TSVs containing arrays … sometime after this PR merges

src/main/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupport.scala
MatthewBemis: I'm not sure that I find this more readable than the current approach. Is casting with `asInstanceOf` a dangerous operation to do if I've already checked for types using `isInstanceOf`? I generally try to avoid it, but if it's not a meltdown risk, then I'd lean towards keeping it as-is

re: the TODO, I will add some logic to better handle empty lists

MatthewBemis:
jenkins retest ; fiab start failed

src/main/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupport.scala
```diff
@@ -143,10 +147,38 @@ trait TSVFileSupport {
...
+                case elements if elements.forall(_.isInstanceOf[JsBoolean]) =>
+                  val elementsTyped: List[JsBoolean] = elements.asInstanceOf[List[JsBoolean]]
+                  elementsTyped.map(jsbool => Map(addListMemberOperation, listNameEntry, listValEntry(AttributeBoolean(jsbool.value))))
+
+                case _ => throw new FireCloudException("Mixed-type entity attribute lists are not supported.")
+              }
+              val removeOldListOp = Seq(Map(removeAttrOperation, nameEntry))
+              removeOldListOp ++ addElements
+            }
+          }
```
sagehen03: Could we refactor this block of code into its own function or maybe even multiple functions? Shorter methods whose names describe what's going on are nice I think--like built in comments that are less likely to go out of date.  However, that's my java mindset speaking up; maybe there are other considerations with scala code.

src/test/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupportSpec.scala
```diff
@@ -51,4 +55,77 @@ class TSVFileSupportSpec extends AnyFreeSpec with TSVFileSupport {
...
+
+      //assert that these operations always remove the existing attribute by this name
+      resultingOpNames.head shouldBe AttributeString("RemoveAttribute")
+
+      //assert that all of the remaining operations are adding a list member
+      resultingOpNames.tail.map { op =>
+        op shouldBe AttributeString("AddListMember")
+      }
+
+      println(resultingOps)
```
davidangb: nit: remove `println`

src/test/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupportSpec.scala
```diff
@@ -51,4 +55,77 @@ class TSVFileSupportSpec extends AnyFreeSpec with TSVFileSupport {
...
+      val resultingOpNames = resultingOps.operations.map(ops => ops("op"))
+
+      //assert that these operations always remove the existing attribute by this name
+      resultingOpNames.head shouldBe AttributeString("RemoveAttribute")
+
+      //assert that all of the remaining operations are adding a list member
+      resultingOpNames.tail.map { op =>
+        op shouldBe AttributeString("AddListMember")
+      }
+    }
```
davidangb, line 109: I'd love to see somewhat deeper validation in these tests. Something like:
```
    def validateListMemberOps(ops: Seq[Map[String, Attribute]], expectedClass: Class[_], expectedAttrName: String = "arrays") = {
      ops.map { op =>
        op.keySet should contain theSameElementsAs List("attributeListName", "newMember", "op")
        op("attributeListName") shouldBe AttributeString(expectedAttrName)
        val element = op("newMember")
        element.getClass shouldBe(expectedClass)
      }
    }
```
and then within each test, you can call:
```
validateListMemberOps(resultingOps.operations.tail, AttributeString("").getClass)
```
(substituting `AttributeString`, `AttributeNumber`, `AttributeBoolean` as appropriate)

src/main/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupport.scala
MatthewBemis: I broke this out along with some smaller pieces. 


I was also hoping to also use type parameters to slim this function down (so there aren't 3 extremely similar `case` branches), but I don't think it will work in the way that I wanted

This doesn't handle the mixed-type exception, but I was thinking something roughly like this (does not compile):

```
  def generateAttributeArrayOperations[T, U](attributeValue: String, attributeName: String): Seq[Map[String, Attribute]] = {
    val listElements = attributeValue.parseJson.convertTo[JsArray].elements.toList

    //if the list is empty, short-circuit and just replace any existing list with an empty list
    if(listElements.isEmpty) {
      Seq(Map(removeAttrOperation, nameEntry(attributeName)), Map(createAttrValueListOperation, nameEntry(attributeName)))
    } else {
      val elementsTyped: List[T] = listElements.asInstanceOf[List[T]]
      val elementsToAdd = elementsTyped.map(jsval => Map(addListMemberOperation, listNameEntry(attributeName), listValEntry(U(jsval.value))))
      
      val removeOldListOp = Seq(Map(removeAttrOperation, nameEntry(attributeName)))
      removeOldListOp ++ elementsToAdd
    }
  }
```

MatthewBemis:
The current state:

Unless there are more comments regarding the state of the application code, I think all that's left is to add the deeper unit test validation that @davidangb suggested.

I have done quite a bit of manual validation with a local Orch as well and feel pretty good about the fix, but it probably wouldn't hurt for someone else to try out various test cases.

So really, as of right now, this needs deeper validation in the unit tests, someone else to do additional manual testing, and then it should be mergeable. I've told Janki that this should be do-able for next weeks release. I believe PPW is going to land on dev on 9/14 so we should really try to beat that because they want a week of burn-in before releasing.

MatthewBemis:
jenkins retest ; flaky fiab start

src/main/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupport.scala
```diff
@@ -139,14 +148,13 @@ trait TSVFileSupport {
...
+    val ops = for { (attributeValue,(attributeName,refTypeOpt)) <- row.tail zip colInfo if refTypeOpt.isDefined || !attributeValue.isEmpty } yield {
       refTypeOpt match {
-        case Some(refType) => Map(upsertAttrOperation,nameEntry,valEntry(AttributeEntityReference(refType,value)))
-        case None => value match {
-          case "__DELETE__" => Map(removeAttrOperation,nameEntry)
-          case _ => Map(upsertAttrOperation,nameEntry,valEntry(AttributeString(value)))
+        case Some(refType) => Seq(Map(upsertAttrOperation,nameEntry(attributeName),valEntry(AttributeEntityReference(refType,attributeValue))))
+        case None => attributeValue match {
+          case "__DELETE__" => Seq(Map(removeAttrOperation,nameEntry(attributeName)))
+          case value if modelSchema.isAttributeArray(value) => generateAttributeArrayOperations(value, attributeName)
```
ungwudik, line 157: If `modelSchema.isAttributeArray(value)` finds that the the TSV is a malformed json, should we let the user know?

src/test/scala/org/broadinstitute/dsde/firecloud/mock/MockTSV.scala
```diff
@@ -316,6 +316,13 @@ object MockTSVLoadFiles {
   val missingFields1 = TSVLoadFile("foo", Seq("foo", "bar", "baz"), Seq(Seq("biz", "", "buz")))
   val missingFields2 = TSVLoadFile("foo", Seq("foo", "bar", "baz"), Seq(Seq("", "", "buz"), Seq("abc", "123", "")))
 
+  val entityWithAttributeBooleanArray = TSVLoadFile("array", Seq("array"), Seq(Seq("bla", """[false,true,true]""")))
+  val entityWithAttributeNumberArray = TSVLoadFile("array", Seq("array"), Seq(Seq("bla", """[1,2,3]""")))
+  val entityWithAttributeStringArray = TSVLoadFile("array", Seq("array"), Seq(Seq("bla", """["foo","bar","baz"]""")))
+  val entityWithAttributeMixedArray = TSVLoadFile("array", Seq("array"), Seq(Seq("bla", """[false,"foo",1]""")))
+  val entityWithAttributeArrayOfObjects = TSVLoadFile("array", Seq("array"), Seq(Seq("bla", """[{"one":"two"},{"three":"four"},{"five":"six"}]""")))
+  val entityWithEmptyAttributeArray = TSVLoadFile("array", Seq("array"), Seq(Seq("bla", """[]""")))
```
ungwudik: Should there also be a mock here for malformed json?

ungwudik: And a unit test for unit as well

ungwudik (approved):
Approving, with a few comments.

src/test/scala/org/broadinstitute/dsde/firecloud/mock/MockTSV.scala
davidangb: @ungwudik what do you mean by "unit test for unit"?

I don't think it's possible to cleanly distinguish a malformed json array. Your typical TSV will contain mostly strings and numbers … all of which technically do not parse as json arrays. So, if your TSV contains `"hello world"` and `123`, both of those cells will fail to parse as arrays. We can't warn or alert the user on those, because they're totally valid, they're just not json arrays. Inside the Scala logic, we can't really distinguish between `"hello world"` and `["hello", "world">` … there's no indicator for "close to" a json array.

ungwudik: Sorry, typo. I meant "And a unit test for **it(malformed json array)** as well"

sagehen03 (approved):

ungwudik (approved):

sagehen03:
Changes to support arrays of json objects look good to me.

sagehen03 (approved):

PR #890, opened 2021-09-15
AS-923: Remove original pricelist test

MatthewBemis:
I would like to "fix" this test by removing it entirely. I don't think it's particularly useful after #879 was merged. While that PR made the application code resilient to changes by Google, it didn't make this particular test resilient in the same manner. Instead, the improved resilience is tested with a new unit test (below this one). This test is vulnerable to Google changing key names which will cause the test to fail and the testing pipeline to break until the application code is updated. IMO, that defeats a part of the purpose of writing resilient code in the first place.

In summary, I feel comfortable removing this coverage because:

1) The code is resilient, and that resilience is tested
2) This unit test can't really tell us anything that we don't already know. Odds are very high that if this code breaks, it will be broken on production as well and we'll get plenty of Sentry alerts about it

In all cases:

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAOSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAOSpec.scala
@@ -41,22 +41,6 @@ class HttpGoogleServicesDAOSpec extends AnyFlatSpec with Matchers with PrivateMe
 
   behavior of "HttpGoogleServicesDAO"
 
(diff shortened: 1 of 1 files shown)


ungwudik (approved):
LGTM. Is there an example somewhere that shows the Sentry alerts firing?

MatthewBemis:
@ungwudik not yet, but the error is thrown [here](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala#L474)

davidangb (approved):

PR #896, opened 2021-10-27
AS-906 Orch added optional field in request body for TOS

jordansafer:
On the API side, I added an optional termsOfService field to the BasicProfile object passed in the request body. This allows the client to optionally provide a String for terms of service in the request to Orch. If the user provides a string, this is propagated, and the terms of service string is included in the request to Sam, as part of the message body. The change to add the parameter is made in `model/Profile.scala` and passed in `service/RegisterService.scala`. The call to Sam is updated in `dataaccess/HttpSamDAO.scala`.

Testing:
- Updated unit tests and made sure they are completing successfully
- Ran a local Terra and local Orch on this version and registered a new user

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/HealthChecks.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/HealthChecks.scala
@@ -32,7 +32,7 @@ class HealthChecks(app: Application, registerSAs: Boolean = true)
     lookup flatMap {
       case Some(err) if registerSAs =>
         logger.warn(s"SA registration lookup failed; attempting to register $name. Lookup failure was: $err")
-        manageRegistration(name, app.samDAO.registerUser(token))
+        manageRegistration(name, app.samDAO.registerUser(token, None))
 
       case registerMessage =>
         Future.successful(registerMessage)
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/HealthChecks.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/HealthChecks.scala
@@ -11,6 +11,7 @@ import scala.concurrent.{ExecutionContext, Future}
 
 object HealthChecks {
   val adminSaRegistered = Subsystems.Custom("is_admin_sa_registered")
+  val termsOfServiceUrl = "app.terra.bio/#terms-of-service"
 }
 
 class HealthChecks(app: Application, registerSAs: Boolean = true)
@@ -32,7 +33,7 @@ class HealthChecks(app: Application, registerSAs: Boolean = true)
     lookup flatMap {
       case Some(err) if registerSAs =>
         logger.warn(s"SA registration lookup failed; attempting to register $name. Lookup failure was: $err")
-        manageRegistration(name, app.samDAO.registerUser(token, None))
+        manageRegistration(name, app.samDAO.registerUser(Some(termsOfServiceUrl))(token))
 
       case registerMessage =>
         Future.successful(registerMessage)
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpSamDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpSamDAO.scala
@@ -34,7 +34,7 @@ class HttpSamDAO( implicit val system: ActorSystem, val materializer: Materializ
     authedRequestToObject[Seq[UserPolicy]](Get(samListResources("workspace")), label=Some("HttpSamDAO.listWorkspaceResources"))
   }
 
-  override def registerUser(implicit userInfo: WithAccessToken, termsOfService: Option[String]): Future[RegistrationInfo] = {
+  override def registerUser(termsOfService: Option[String])(implicit userInfo: WithAccessToken): Future[RegistrationInfo] = {
     authedRequestToObject[RegistrationInfo](Post(samUserRegistrationUrl, termsOfService), label=Some("HttpSamDAO.registerUser"))
   }
 
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpSamDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpSamDAO.scala
@@ -3,9 +3,7 @@ package org.broadinstitute.dsde.firecloud.dataaccess
 import java.net.URLEncoder
 import java.nio.charset.StandardCharsets.UTF_8
 import akka.actor.ActorSystem
-import akka.http.scaladsl.Http
 import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport
-import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
 import akka.http.scaladsl.model.HttpRequest
 import akka.http.scaladsl.unmarshalling.Unmarshal
 import akka.stream.Materializer
@@ -36,8 +34,8 @@ class HttpSamDAO( implicit val system: ActorSystem, val materializer: Materializ
     authedRequestToObject[Seq[UserPolicy]](Get(samListResources("workspace")), label=Some("HttpSamDAO.listWorkspaceResources"))
   }
 
-  override def registerUser(implicit userInfo: WithAccessToken): Future[RegistrationInfo] = {
-    authedRequestToObject[RegistrationInfo](Post(samUserRegistrationUrl), label=Some("HttpSamDAO.registerUser"))
+  override def registerUser(implicit userInfo: WithAccessToken, termsOfService: Option[String]): Future[RegistrationInfo] = {
+    authedRequestToObject[RegistrationInfo](Post(samUserRegistrationUrl, termsOfService), label=Some("HttpSamDAO.registerUser"))
   }
 
   override def getRegistrationStatus(implicit userInfo: WithAccessToken): Future[RegistrationInfo] = {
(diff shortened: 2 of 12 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/HealthChecks.scala
```diff
@@ -32,7 +32,7 @@ class HealthChecks(app: Application, registerSAs: Boolean = true)
     lookup flatMap {
       case Some(err) if registerSAs =>
         logger.warn(s"SA registration lookup failed; attempting to register $name. Lookup failure was: $err")
-        manageRegistration(name, app.samDAO.registerUser(token))
+        manageRegistration(name, app.samDAO.registerUser(token, None))
```
davidangb: This line attempts to register the admin SA. By passing `None` here, wouldn't this guarantee that the registration would fail, since it doesn't include the appropriate URL to the ToS?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpSamDAO.scala
```diff
@@ -36,8 +34,8 @@ class HttpSamDAO( implicit val system: ActorSystem, val materializer: Materializ
     authedRequestToObject[Seq[UserPolicy]](Get(samListResources("workspace")), label=Some("HttpSamDAO.listWorkspaceResources"))
   }
 
-  override def registerUser(implicit userInfo: WithAccessToken): Future[RegistrationInfo] = {
-    authedRequestToObject[RegistrationInfo](Post(samUserRegistrationUrl), label=Some("HttpSamDAO.registerUser"))
+  override def registerUser(implicit userInfo: WithAccessToken, termsOfService: Option[String]): Future[RegistrationInfo] = {
+    authedRequestToObject[RegistrationInfo](Post(samUserRegistrationUrl, termsOfService), label=Some("HttpSamDAO.registerUser"))
```
davidangb, line 38: did you intend for `termsOfService` to be an implicit argument? To make it non-implicit, follow the pattern from e.g. `getUserIds` lower down in this file; you'd have something like:
```Scala
override def registerUser(termsOfService: Option[String])(implicit userInfo: WithAccessToken): Future[RegistrationInfo] = …
```

src/test/scala/org/broadinstitute/dsde/firecloud/model/ProfileSpec.scala
```diff
@@ -91,7 +92,8 @@ class ProfileSpec extends AnyFreeSpec with Matchers {
             programLocationState = "",
             programLocationCountry = "",
             pi = "",
-            nonProfitStatus = ""
+            nonProfitStatus = "",
+            Option.empty
```
davidangb: up to you: you can use `None` instead of `Option.empty`; I typically find `None` easier to read

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/RegisterApiServiceSpec.scala
```diff
@@ -71,6 +73,42 @@ final class RegisterApiServiceSpec extends BaseServiceSpec with RegisterApiServi
       }
 
     }
+
+    "register-profile API" - {
+      implicit val impBasicProfile: RootJsonFormat[BasicProfile] = jsonFormat12(BasicProfile)
```
davidangb: you should be able to `import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol.impBasicProfile`  instead of creating a new formatter here.

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/RegisterApiServiceSpec.scala
```diff
@@ -71,6 +73,42 @@ final class RegisterApiServiceSpec extends BaseServiceSpec with RegisterApiServi
...
+
+      "should succeed with no terms of service" in {
+        val payload = makeBasicProfile(false)
+        Post("/register/profile", payload) ~> dummyUserIdHeaders("RegisterApiServiceSpec") ~> sealRoute(registerRoutes) ~> check {
+          status should be(OK)
+        }
+      }
```
davidangb, line 83: ah, interesting. In the live environments, this will fail. But in this test spec, registering a user with no terms of service succeeds, because the `MockSamDAO` is hardcoded to succeed. I suggest one of the following:
* change the test name and/or add code comments explaining why this test is expected to pass even though live envs would fail;
* either modify `MockSamDAO.registerUser` or create a new mock Sam DAO to use in this test case; have the mock check for the presence of the TOS value, pass/fail the registration attempt appropriately, and update the tests. This is obviously more work, but it would test if we are properly propagating the user's TOS argument all the way to Sam.

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiServiceSpec.scala
```diff
@@ -48,7 +48,8 @@ class UserApiServiceSpec extends BaseServiceSpec with RegisterApiService with Us
     programLocationState = randomAlpha(),
     programLocationCountry = randomAlpha(),
     pi = randomAlpha(),
-    nonProfitStatus = randomAlpha()
+    nonProfitStatus = randomAlpha(),
+    termsOfService = Option.empty
```
davidangb: same here, consider using `None` instead; this is purely code style

davidangb (commented):
some comments inline, nice work on this so far!

This PR will also require an update to the swagger definition for the `POST /register/profile` api, to indicate the new TOS field: https://github.com/broadinstitute/firecloud-orchestration/blob/3cb3cb15090e37e72d546a4cdb5d592c4d065a2a/src/main/resources/swagger/api-docs.yaml#L7044

src/main/scala/org/broadinstitute/dsde/firecloud/HealthChecks.scala
jordansafer: Ah, yes! This is bad, I'm depending on the fact that not including the ToS url is still supported in multiple places, and shouldn't be (here and in the tests). Fixing!

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpSamDAO.scala
jordansafer: ah no! This should not be implicit- fixing as well!

src/test/scala/org/broadinstitute/dsde/firecloud/model/ProfileSpec.scala
jordansafer: Will use None, agreed it is easier to read :)

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/RegisterApiServiceSpec.scala
jordansafer: Trying this-

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/RegisterApiServiceSpec.scala
jordansafer: I'm updating the Sam DAO to check for the presence of the ToS value, I agree, we should be requiring this from the Sam side. Psychologically I was thinking the Sam DAO ought to accept it both ways, since current Sam accepts it both ways, but realistically we should be verifying this works when Sam requires the param, since the plan is to require it-

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiServiceSpec.scala
jordansafer: agreed and fixing!

jordansafer:
jenkins retest

MatthewBemis (approved):
Nice! LGTM

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockSamDAO.scala
```diff
@@ -21,9 +22,14 @@ class MockSamDAO extends SamDAO {
...
-  override def registerUser(implicit userInfo: WithAccessToken): Future[RegistrationInfo] = enabledUserInfo
+  override def registerUser(termsOfService: Option[String])(implicit userInfo: WithAccessToken): Future[RegistrationInfo] =
+    if (termsOfService.contains(termsOfServiceUrl)) enabledUserInfo
+    else Future.failed(new FireCloudExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.Forbidden,
+      s"You must accept the Terms of Service in order to register.")))
 
-  override def getRegistrationStatus(implicit userInfo: WithAccessToken): Future[RegistrationInfo] = enabledUserInfo
+  override def getRegistrationStatus(implicit userInfo: WithAccessToken): Future[RegistrationInfo] =
+    if (userInfo.accessToken.token.equals("new")) unknownUserInfo
+    else enabledUserInfo
```
davidangb, line 32: is this `if` condition exercised? I see that `RegisterApiServiceSpec` has two tests that send a token containing "new" - but those tests only assert on the http response status, they don't look at the RegistrationInfo details.

davidangb (approved):
one minor question inline, not a blocker, otherwise looks good and thanks for sticking with this!

I see this also needs a rebase, sorry :(

src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockSamDAO.scala
jordansafer: Yes, it's exercised- when the "unknownUserInfo" is returned, this drives Orch to call Sam (in this case MockSam) to create a new user. If the ToS URL isn't included in the request, than mockSam returns an exception. However, if the token isn't "new", MockSam will return that the user is already known, so Orch won't try to create the user, and no exception will be thrown if there's no ToS parameter.

PR #903, opened 2021-11-19
AS-908 Add pass through API for ToS text from Orch to Sam

jordansafer:
Added an API at /tos/text for getting the ToS text from SAM as part of the Terms of Service architecture changes being worked on. https://broadworkbench.atlassian.net/browse/AS-908

Testing:
- Unit tested, locally tested, and added an integration test

In all cases:

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/RegistrationApiSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/RegistrationApiSpec.scala
@@ -1,12 +1,14 @@
 package org.broadinstitute.dsde.test.api.orch
 
+import akka.actor.ActorSystem
 import akka.http.scaladsl.model.HttpMethods.GET
 import akka.http.scaladsl.model.{HttpRequest, StatusCodes}
 import akka.http.scaladsl.unmarshalling.Unmarshal
+import akka.testkit.TestKitBase
 import org.broadinstitute.dsde.test.OrchConfig.Users
 import org.broadinstitute.dsde.workbench.auth.AuthToken
 import org.broadinstitute.dsde.workbench.service.Orchestration
-import org.broadinstitute.dsde.workbench.config.{Credentials, UserPool}
+import org.broadinstitute.dsde.workbench.config.{Credentials, ServiceTestConfig, UserPool}
 import org.broadinstitute.dsde.workbench.fixture.BillingFixtures
 import org.broadinstitute.dsde.workbench.service.Sam.sendRequest
 import org.broadinstitute.dsde.workbench.service.{Sam, Thurloe}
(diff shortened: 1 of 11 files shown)


automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/RegistrationApiSpec.scala
```diff
@@ -54,5 +60,17 @@ class RegistrationApiSpec extends FreeSpec with Matchers with ScalaFutures with
...
+
+    "should return terms of services with no auth token" in {
+      val req = HttpRequest(GET, ServiceTestConfig.FireCloud.orchApiUrl + s"tos/text")
+      val response = sendRequest(req)
+
+      val textFuture = Unmarshal(response.entity).to[String]
+
+      response.status shouldEqual StatusCodes.OK
+      whenReady(textFuture) { text =>
+        text should include("Terms as of February 12, 2020.")
```
MatthewBemis: I am worried that a change to the ToS will break this test. Is there something else that we could assert that would be resilient to ToS changes?

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/RegisterApiService.scala
```diff
@@ -43,4 +43,14 @@ trait RegisterApiService extends Directives with RequestBuilding with StandardUs
         }
       }
     }
+
+  val tosRoutes: Route =
+    pathPrefix("tos") {
+      path("text") {
+        get {
+          complete {
+            registerServiceConstructor().getTermsOfServiceText }
```
MatthewBemis: could this use the `passthrough` directive instead? Then you wouldn't need to add anything to `SamDAO`

sagehen03 (approved):
Works for me!

automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/RegistrationApiSpec.scala
jordansafer: This makes sense. Truthfully, the only thing we know for sure is that the response status is OK. The assertion about the text contents is could always flake. The same issue is present in that SAM integration tests I added, which I can update-

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/RegisterApiService.scala
jordansafer: Oh cool! Testing this out-

automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/RegistrationApiSpec.scala
MatthewBemis: Could it test that the text is not empty? That seems like a safe thing for us to assume in the tests and it gives _some_ information

jordansafer: Yes, it can, I'll test that

MatthewBemis (approved):
LGTM!

PR #912, opened 2022-01-14
increase max TSV columns 1024->2000 [no ticket yet]

davidangb:
provisional PR; not sure we want this yet

increase the max allowed number of columns for TSV import to 2000. This also affects BDBag/Bagit import, which uses TSV under the covers.

Tested manually using a TSV with 1681 columns.


sagehen03 (approved):

davidangb:
we've decided to not change this setting.

PR #913, opened 2022-02-02
[WOR-68] ToS passthrough

marctalbott:
Ticket: [WOR-68(https://broadworkbench.atlassian.net/browse/WOR-68)
Add passthrough to Sam to accept the ToS

Have you read [CONTRIBUTING.md](../CONTRIBUTING.md) lately? If not, do that first.

I, the developer opening this PR, do solemnly pinky swear that:

- [ ] I've followed [the instructions](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/CONTRIBUTING.md#api-changes) if I've made any changes to the API, _especially_ if they're breaking changes
- [ ] I've updated the RC_XXX release ticket with any manual steps required to release this change
- [ ] I've updated the [FISMA documentation](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/CONTRIBUTING.md#fisma-documentation-changes) if I've made any security-related changes, including auth, encryption, or auditing

In all cases:

- [ ] Get two thumbsworth of review and PO signoff if necessary
- [ ] Verify all tests go green
- [ ] Squash and merge; you can delete your branch after this **unless it's for a hotfix**. In that case, don't delete it!
- [ ] Test this change deployed correctly and works on dev environment after deployment

PR #918, opened 2022-02-28
Revert "[WOR-68] ToS passthrough"

MatthewBemis:
Reverts broadinstitute/firecloud-orchestration#913

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/RegistrationApiSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/RegistrationApiSpec.scala
@@ -1,10 +1,8 @@
 package org.broadinstitute.dsde.test.api.orch
 
 import akka.actor.ActorSystem
-import akka.http.javadsl.model.RequestEntity
-import akka.http.scaladsl.model.headers.{Authorization, OAuth2BearerToken}
-import akka.http.scaladsl.model.HttpMethods.{GET, POST}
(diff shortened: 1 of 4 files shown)


ungwudik (approved):
lgtm

aherbst-broad (approved):

rtitle:
I am sort of wondering if it was this test which was causing the 401s, not the ToS enablement in Sam itself. Example flow:

1. All users are registered and tos-accepted by `fiab-start`
2. This Orch test executes which disables a random user (say Sirius) in Sam
3. Before the test re-enables the user, another swatomation test (Rawls, Orch, Leo) makes an Orch call as Sirius (all the tests call Orch to set up billing projects/workspaces so it's pretty common)
4. The LDAP check fails because the user is disabled, and the LDAP result is cached in orch-proxy
5. Even when the user is re-enabled in Sam, the LDAP value is cached for ~5 minutes causing mass 401s for any test which happens to use that user.

I haven't 100% proved this, so it could be off. But it kind of matches my observations in logs. E.g. it looks like the user getting the 401s matches the user this test tries to disable. It also might explain why it's intermittent: the user is randomly chosen, and it's sensitive to timing.

Luckily this test is already reverted so we can observe the effect. :) I don't see any mass 401s since this PR was merged (a few Leo failures, but looks unrelated). Interested how it looks in the morning..

MatthewBemis:
@rtitle wow yeah you might be right. Things still look good since reverting this part of ToS. The PR to disable ToS entirely had a 401 test failure, but it was still using this version of Orch with this test. Everything makes sense now!

PR #924, opened 2022-03-15
update the TOS accept body description and add fields to the Enabled entity

jmthibault79:
A small update to match the changes made when I adapted it to AoU here: https://github.com/all-of-us/workbench/pull/6454

In all cases:

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -7190,7 +7190,7 @@ paths:
       summary: Passes through to Sam API and accepts Terra Terms of Service
       operationId: acceptTermsOfService
       requestBody:
-        description: url
+        description: Must equal the (double-quoted) url of the Terra Terms of Service
         content:
           'application/json':
             schema:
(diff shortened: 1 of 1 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -8138,6 +8138,12 @@ components:
         allUsersGroup:
           type: boolean
           description: User is a member of the "All Users" group?
+        adminEnabled:
+          type: boolean
+          description: User is a member of the admin group?
```
MatthewBemis, line 8143: I was confused by the naming of this `adminEnabled` field when I first saw it, but it's actually to indicate whether or not a Terra admin has explicitly disabled a user in the system. So maybe a more fitting description would be "User is enabled in the system?" or something similar

MatthewBemis (approved):
lgtm pending re-wording the `adminEnabled` description

src/main/resources/swagger/api-docs.yaml
jmthibault79: thanks, I was just guessing.  Updated.

jmthibault79: ... but somehow it got lost in the squash/merge.  Trying again: #926

PR #928, opened 2022-03-18
SU-34: Optionally allow for TSV parser to process blank values as removals

MatthewBemis:

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiService.scala
@@ -114,10 +114,14 @@ trait WorkspaceApiService extends FireCloudRequestBuilding with FireCloudDirecti
                   post {
                     requireUserInfo() { userInfo =>
                       parameter("async" ? "false") { asyncStr =>
-                        formFields('entities) { entitiesTSV =>
-                          complete {
-                            val isAsync = java.lang.Boolean.valueOf(asyncStr) // for lenient parsing
-                            entityServiceConstructor(FlexibleModelSchema).importEntitiesFromTSV(workspaceNamespace, workspaceName, entitiesTSV, userInfo, isAsync)
+                        parameter("processBlanksAsNull" ? "false") { processBlanksStr =>
+                          formFields('entities) { entitiesTSV =>
+                            complete {
+                              val isAsync = java.lang.Boolean.valueOf(asyncStr) // for lenient parsing
+                              val processBlanksAsNull = java.lang.Boolean.valueOf(processBlanksStr) // for lenient parsing
+                              println(processBlanksAsNull)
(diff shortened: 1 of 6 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiService.scala
```diff
@@ -114,10 +114,13 @@ trait WorkspaceApiService extends FireCloudRequestBuilding with FireCloudDirecti
                   post {
```
MatthewBemis, line 114: I only did this for flexibleImportEntities. Should I bother with `importEntities`?

davidangb: Terra UI calls `importEntities` instead of `flexibleImportEntities` if the end user checks the _Create participant, sample, and pair associations_ box. I assume we also want to support blanks-as-removals for that use case.

davidangb: Terra UI calls `importEntities` instead of `flexibleImportEntities` if the end user checks the _Create participant, sample, and pair associations_ box. I assume we also want to support blanks-as-removals for that use case.

davidangb (commented):
with the exception of also adding this option to the `importEntities` API, the code in here looks :+1: to me.

My main questions are about the UX - how will this be presented to users? How will it be release-noted and communicated? Will users be confused that to delete a column, they need to add the column to the TSV headers but leave all cells blank, as opposed to omitting the column entirely from the TSV (which I realize is a tricky/difficult back end implementation)? None of those questions is in scope for this PR and you don't have to answer them, just thinking out loud.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiService.scala
MatthewBemis: Ah! I will add support for `importEntities`, then. Thanks!

MatthewBemis:
@davidangb I definitely agree that there are still plenty of TBD UX details, which will be figured out with the UX team as part of SUP-248 now that backend support is "unlocked" with this PR. We'll make sure to run it by you and your team as well, likely during an upcoming team sync

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiService.scala
MatthewBemis: Added in aa2a4825f115942b9603ed36e4143ddaa7845fb5

davidangb (approved):
:+1: assuming tests pass

note to self: huh, we never enabled async TSVs for importEntities either??? Oops

MatthewBemis:
jenkins retest (build failed)

nawatts (approved):

MatthewBemis:
tests passed, but fiab-stop failed :/

jenkins retest

PR #929, opened 2022-03-22
[AJ-339] Use AttributeBoolean and AttributeNumber classes during TSV upload

vitzlo:
When uploading a TSV into a Terra data table, all non-array values are created as strings. This PR makes a best-effort attempt to detect booleans and numbers in the uploaded data and create values in data tables respecting those data types.

Points of note (because I trust others will form more judicious opinions on these than me):
- Should a `+` prepended to an otherwise currently valid number be respected as an integer?
- Should we scan for integers that are too large here? Letting Rawls throw an error vs. acknowledging as a string
- Should we scan for booleans in a case-insensitive fashion?


davidangb (commented):
what you have here looks like it will get the job done - good job on figuring out all that Scala! Some thoughts on how to make it better:

I suggest breaking out all your new logic into a standalone method that accepts a String and returns an AttributeValue (which is the parent of AttributeBoolean, AttributeNumber, and AttributeString). This would be something like:
```scala
def stringToTypedAttribute(input: String): AttributeValue = {
   …
}
```

Then, inside `def setAttributesOnEntity`, where you currently have `case _ => …`, you would have:

```scala
case scalar => Seq(Map(upsertAttrOperation,nameEntry(attributeName),valEntry(stringToTypedAttribute(scalar))))
```

creating a separate method like this keeps the code clean, but more importantly it makes it easily unit-testable. Which brings me to my next point … 

Could you please add unit tests for this? You could add good coverage for the new `def stringToTypedAttribute()` method, and then probably one higher-level test that sends in TSV data to `setAttributesOnEntity` and inspects the final results. #887 might be useful here as an example.

Finally - when parsing the string to see if it would work as a number or boolean, let's not reinvent the wheel. Instead of writing regexes for these, I suggest using `java.lang.Integer.parseInt` and `java.lang.Double.parseDouble` (I think we should use double instead of float). These throw errors if the input isn't a valid int or double, so in Scala you'd do something like:
```scala
Try(java.lang.Integer.parseInt(input)) match {
      case Success(intValue) => AttributeNumber(intValue)
      case Failure(_) => // continue looking for doubles, booleans
    }
```
unfortunately, `java.lang.Boolean.parseBoolean` is too lenient for our use case, so I don't think we can use it. I wonder if any of the utility libraries currently in orch have better boolean parsing.

Happy to talk through these f2f if that's helpful!

davidangb:
Update: not sure this is significantly better, but you could use Apache's commons-lang, which is already in Orch. This would look like:
```scala
import org.apache.commons.lang3.BooleanUtils

…

Option(BooleanUtils.toBoolean(input.trim.toLowerCase, "true", "false")) match {
    case Some(b) => AttributeBoolean(b)
    case _ => AttributeString(input)
}
```

src/main/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupport.scala
```diff
@@ -155,15 +172,15 @@ trait TSVFileSupport {
...
+        case Some(refType) => Seq(Map(upsertAttrOperation, nameEntry(attributeName), valEntry(AttributeEntityReference(refType, attributeValue))))
         case None =>
           attributeValue match {
-            case "__DELETE__" => Seq(Map(removeAttrOperation,nameEntry(attributeName)))
-            case value if deleteEmptyValues && value.trim.isEmpty => Seq(Map(removeAttrOperation,nameEntry(attributeName)))
+            case "__DELETE__" => Seq(Map(removeAttrOperation, nameEntry(attributeName)))
+            case value if deleteEmptyValues && value.trim.isEmpty => Seq(Map(removeAttrOperation, nameEntry(attributeName)))
             case value if modelSchema.isAttributeArray(value) => generateAttributeArrayOperations(value, attributeName)
-            case _ => Seq(Map(upsertAttrOperation,nameEntry(attributeName),valEntry(AttributeString(attributeValue))))
+            case _ => Seq(Map(upsertAttrOperation, nameEntry(attributeName), valEntry(stringToTypedAttribute(attributeValue))));
```
davidangb: nit: drop the semicolon at the end of the line. Scala allows is but it is frowned upon in Scala-land.

src/test/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupportSpec.scala
```diff
@@ -168,6 +210,32 @@ class TSVFileSupportSpec extends AnyFreeSpec with TSVFileSupport {
...
+      val secondBoolOp = resultingOpsSecond.operations.head
+      secondBoolOp("addUpdateAttribute") shouldBe AttributeBoolean(false)
+
+      val secondNumOp = resultingOpsSecond.operations(1)
+      secondNumOp("addUpdateAttribute") shouldBe AttributeNumber(3.14)
+
+      val secondStringOp = resultingOpsSecond.operations(2)
+      secondStringOp("addUpdateAttribute") shouldBe AttributeString(",")
+    }
+  }
```
davidangb, line 262: We could make this somewhat smaller and less copy-paste. How about this?
```scala
    "create AttributeBoolean and AttributeNumber when applicable" in {
      val colInfo = colNamesToAttributeNames(MockTSVLoadFiles.entityWithBooleanAndNumberAttributes.headers, Map.empty)
      val resultingOpsFirst = setAttributesOnEntity("foo", None, MockTSVLoadFiles.entityWithBooleanAndNumberAttributes.tsvData.head, colInfo, FlexibleModelSchema)
      val resultingOpsSecond = setAttributesOnEntity("foo", None, MockTSVLoadFiles.entityWithBooleanAndNumberAttributes.tsvData(1), colInfo, FlexibleModelSchema)

      val expectedOpsFirst = List(AttributeBoolean(true), AttributeNumber(0), AttributeString("string"))
      val expectedOpsSecond = List(AttributeBoolean(false), AttributeNumber(3.14), AttributeString(","))

      resultingOpsFirst.operations.map(_("addUpdateAttribute")) should contain theSameElementsInOrderAs expectedOpsFirst
      resultingOpsSecond.operations.map(_("addUpdateAttribute")) should contain theSameElementsInOrderAs expectedOpsSecond
    }
```

in the `resultingOpsFirst.operations.map(_("addUpdateAttribute"))`, we're using the underscore inside the `map` as shorthand/sugar. This is equivalent to `resultingOpsFirst.operations.map(x => x("addUpdateAttribute"))`.

also, the `should contain theSameElementsInOrderAs` is a best practice in Scalatest when working with collections, because in some collection cases checking equality all the way down is difficult. However, in this specific case using `shouldBe` instead of `should contain theSameElementsInOrderAs` works just as well.

src/test/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupportSpec.scala
```diff
@@ -58,6 +58,48 @@ class TSVFileSupportSpec extends AnyFreeSpec with TSVFileSupport {
     }
   }
 
+  "stringToTypedAttribute" - {
+    val booleanTestCases = Map("true" -> AttributeBoolean(true), "True" -> AttributeBoolean(true), "tRue" -> AttributeBoolean(true),
+      "false" -> AttributeBoolean(false), "False" -> AttributeBoolean(false), "FALSE" -> AttributeBoolean(false)
+    )
```
davidangb, line 73: a few more suggested test cases: "  true " to test whitespace trimming; "yes", "no", "0", and "1" to test common boolean-ish strings.

also, purely for readability I suggest putting each case on its own line (and do the same for integers and doubles)
```scala
val booleanTestCases = Map(
      "true" -> AttributeBoolean(true),
      "  true " -> AttributeBoolean(true),
      "True" -> AttributeBoolean(true),
      "tRue" -> AttributeBoolean(true),
      "false" -> AttributeBoolean(false),
      "False" -> AttributeBoolean(false),
      "FALSE" -> AttributeBoolean(false),
      "yes" -> AttributeString("yes"),
      "no" -> AttributeString("no"),
      "0" -> AttributeNumber(0),
      "1" -> AttributeNumber(1)
    )
```

src/test/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupportSpec.scala
```diff
@@ -58,6 +58,48 @@ class TSVFileSupportSpec extends AnyFreeSpec with TSVFileSupport {
...
+      "525600" -> AttributeNumber(525600), "525,600" -> AttributeString("525,600"), "525_600" -> AttributeString("525_600"),
+      "-525600" -> AttributeNumber(-525600), "-525,600" -> AttributeString("-525,600"), "-525_600" -> AttributeString("-525_600"),
+      "0" -> AttributeNumber(0), "00" -> AttributeNumber(0), "-0" -> AttributeNumber(0), "-00" -> AttributeNumber(0)
+    )
+    val doubleTestCases = Map(
+      "4.2" -> AttributeNumber(4.2), ".42" -> AttributeNumber(0.42), "0.42" -> AttributeNumber(0.42),
+      "42." -> AttributeNumber(42), "42.0" -> AttributeNumber(42), "." -> AttributeString("."),
+      "-4.2" -> AttributeNumber(-4.2), "-.42" -> AttributeNumber(-0.42), "-0.42" -> AttributeNumber(-0.42),
+      "-42." -> AttributeNumber(-42), "-42.0" -> AttributeNumber(-42), "-." -> AttributeString("-."),
+    )
```
davidangb, line 104: additional test cases for integers and doubles - it's always good to test the boundaries:
```scala
      Int.MaxValue.toString -> AttributeNumber(Int.MaxValue),
      Int.MinValue.toString -> AttributeNumber(Int.MinValue),

      Double.MaxValue.toString -> AttributeNumber(Double.MaxValue),
      Double.MinValue.toString -> AttributeNumber(Double.MinValue),
      Double.MinPositiveValue.toString -> AttributeNumber(Double.MinPositiveValue)
```
as a bonus, the "toStrings" on some of these generate scientific notation, so you'll get test cases such as "'4.9E-324'"!

src/test/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupportSpec.scala
davidangb: and another thought: we `trim()` inside of `stringToTypedAttribute()` when looking for booleans, but we don't trim when looking for numbers. This feels inconsistent … what do you think?

src/test/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupportSpec.scala
```diff
@@ -58,6 +58,48 @@ class TSVFileSupportSpec extends AnyFreeSpec with TSVFileSupport {
...
+      case (input, expected) => s"int test: $input" in {
+        stringToTypedAttribute(input) shouldBe expected
+      }
+    }
+
+    doubleTestCases foreach {
+      case (input, expected) => s"double test: $input" in {
+        stringToTypedAttribute(input) shouldBe expected
+      }
+    }
```
davidangb, line 129: I know this was my suggestion during pairing, but in practice having one test case for each input MAY be overwhelming. You end up with test output like:
```
stringToTypedAttribute
[info] - boolean test: false
[info] - boolean test: yes
[info] - boolean test: True
[info] - boolean test: False
[info] - boolean test: FALSE
[info] - boolean test: 1
[info] - boolean test:   true 
[info] - boolean test: tRue
[info] - boolean test: 0
[info] - boolean test: true
[info] - boolean test: no
[info] - int test: -525,600
[info] - int test: -0
[info] - int test: 525,600
[info] - int test: -525600
[info] - int test: 525_600
[info] - int test: 525600
[info] - int test: -00
[info] - int test: 0
```
you could instead make a single test case that loops through the different input values, and use `withClue` to help with any failures. This would look something like:
```scala
    "should detect double values" in {
      doubleTestCases foreach {
        case (input, expected) => withClue(s"for input value '$input'") {
          stringToTypedAttribute(input) shouldBe expected
        }
      }
    }
``` 
and would give you shorter output, like:
```
[info] stringToTypedAttribute
[info] - should detect boolean values
[info] - should detect integer values
[info] - should detect double values *** FAILED ***
[info]   for input value '4.9E-324' AttributeNumber(4.9E-324) was not equal to AttributeNumber(-1.7976931348623157E+308) (TSVFileSupportSpec.scala:111)
```
( I created an intentionally failing test to illustrate the output of `withClue` ).

The change in behavior is that the latter approach will fail the test on the first failed assertion, whereas the former approach will always test all input values, potentially resulting in multiple failures. I could go either way but am leaning towards the latter.

Finally - by convention, Scalatest test names start/involve the word "should", so even if you kept with the former approach, you'd want something like `case (input, expected) => s"should handle ints for input: $input" in {` instead of `case (input, expected) => s"int test: $input" in {`.

davidangb (commented):
Some suggestions on additional useful test cases, and a question about the use of `trim()` - these are the priority pieces of feedback.

There's a bunch of other suggestions about idiomatic Scala and code clarity which are less important but thought would be good for discussion!

src/test/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupportSpec.scala
vitzlo: i assumed there are stricter rules on what should be reasonably considered an int/number based on where the string is coming from (i.e. if there's whitespace in a boolean string taking it in as a boolean is ok because the meaning stays invariant, but the presence of whitespace in a float string might imply a non-number use case?). but honestly now that you mention it i don't think it's that much of a differentiator and i can't think of a concrete example where whitespace does imply a string use case so i'm on board to trim for numbers as well

davidangb: I dug a bit more into the TSV parser. We configure the parser's behavior here: 
https://github.com/broadinstitute/firecloud-orchestration/blob/93211ee80c595e9ae9b1489546da099738e97dc5/src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVParser.scala#L18-L28

in that block of code, we do not explicitly call either `settings.setIgnoreTrailingWhitespaces()` or `settings.setIgnoreLeadingWhitespaces()`, which means we use the defaults … which are true, according to my read of the library code. Which means … I believe values pulled out of a TSV are already trimmed, and we could leave it all out of our code. This is worth verifying in practice, of course!

vitzlo: leaving all the manual trimming actually causes failures for exactly one test group: the booleans.

```scala
val booleanTestCases = Map(
  ...
  "  true  " -> AttributeBoolean(true),
  " false " -> AttributeBoolean(false),
  ...
)
```

the error message reads `should handle potential boolean:   true  AttributeString(  true  ) was not equal to AttributeBoolean(true)`, and adding the `.trim` passes the tests, so I suppose we'll leave it in there. not sure why it passes for the ints and doubles, I combed through the parseInt() method and couldn't find any explicit trimming or acknowledgment of it.

davidangb: if we rely on the TSV parser to trim values, then `"  true " -> AttributeBoolean(true),` is no longer a valid or worthwhile unit test … we'll never get string input values that have leading/trailing whitespace.

My suggested approach:
1. stand up a locally-running Orch on the dev branch (not this PR branch)
2. craft a TSV that contains whitespace-padded values
3. add logging statements inside `setAttributesOnEntity` of your locally-running Orch to see what values are actually being processed

If it's true that the TSV parser automatically trims values, then we can remove all trimming from `stringToTypedAttribute` in this branch, AND remove any unit test inputs for padded values.

If this is easier to talk about in person instead of endless github threads, I'm game!

vitzlo: saw this comment after the push, i'd be down to f2f later today if you're free

davidangb (approved):
nice!

jordansafer (approved):
Looks good, my lack of knowledge concern/question- what happens if a user expects their data to be stored as strings, but we've now converted it to ints/doubles/bools, etc. Will they be able to access it normally from the same workflows? Or is there potential for downstream negative impact- @davidangb

PR #934, opened 2022-04-08
[SU-85] Change "FireCloud" to "Terra" in Orch swagger

MatthewBemis:
There are a few instances of "FireCloud" that I didn't change to Terra, I left those either to maintain consistency with other texts or because changing them will actually break something (i.e. firecloud vs flexible entity schema)

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -46,7 +46,7 @@ paths:
     post:
       tags:
         - Billing
-      summary: create billing project in FireCloud and google
+      summary: create billing project in Terra and google
       operationId: createBillingProjectFull
       requestBody:
         description: create project request
@@ -57,10 +57,10 @@ paths:
         required: true
       responses:
         204:
-          description: Successfully Created Billing Project in FireCloud and Google
+          description: Successfully Created Billing Project in Terra and Google
           content: {}
         400:
-          description: FireCloud billing user must be an admin of the billing account
+          description: Terra billing user must be an admin of the billing account
           content:
             application/json:
               schema:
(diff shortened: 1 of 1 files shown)


MatthewBemis:
Holding off on pushing this forward until we confirm with AoU that this doesn't impact their swagger codegen

MatthewBemis:
AoU says: "no we don't think this would affect us"

nawatts (approved):

src/main/resources/swagger/api-docs.yaml
```diff
@@ -6632,8 +6632,8 @@ paths:
       tags:
         - Workspaces
       summary: |
-        Get all tags used in FireCloud (for autocomplete)
-      description: list of all tags used in FireCloud and their associated frequencies
+        Get all workspace tags (for autocomplete)
+      description: Get all tags for workspaces that a user has access to (for autocomplete)
```
MatthewBemis, line 6636: Since I'm here, I harmonized this text with https://github.com/broadinstitute/rawls/pull/1689

MatthewBemis:
jenkins retest

src/main/resources/swagger/api-docs.yaml
```diff
@@ -7380,7 +7381,7 @@ paths:
           description: |
             An identifier of the tool version for this particular tool registry, for example `v1`.
 
-              *In FireCloud, this must be an integer representing the FireCloud snapshot id.*
```
jordansafer, line 7383: nit: Terra snapshot id sounds bit confusing- could this be more clear/specific?

jordansafer (approved):

src/main/resources/swagger/api-docs.yaml
MatthewBemis: `method snapshot id` might be more accurate, I will think about this a bit more but change it to that if I can't think of anything

MatthewBemis:
jenkins retest

jordansafer (approved):

MatthewBemis:
jenkins retest

PR #935, opened 2022-04-12
[TOAZ-110] Add backend oauth2 routes to support B2C integration in Terra UI

rtitle:
[Design doc](https://docs.google.com/document/d/1fnfLSbHWMKi8F__zJm-p7C2URlXIEKzB0-bgyrf-qBg/edit#heading=h.t8i9f7a84yrl) 
[Ticket](https://broadworkbench.atlassian.net/browse/TOAZ-110)
Friends with: https://github.com/DataBiosphere/terra-ui/pull/2949

This PR adds 2 backend routes to facilitate migrating the system to B2C authentication.

1. `GET /oauth2/authorize` -- redirects to the OAuth2 authorize endpoint
2. `POST /oauth2/token` -- proxies request to the OAuth2 token endpoint

Terra UI (and eventually the Orch swagger page -- out of scope for now) will call these routes when it does the oauth2 flow. 

The reason for doing this is as follows:

* Previously we only had 1 authorizationUrl (`https://accounts.google.com/o/oauth2/v2/auth`) but now we are in the situation where we have a Google auth url _plus_ a B2C url for each environment. It's unwieldy/impossible to make the authorizationUrl dynamic in all our UIs+Swagger pages. With this change the backend controls redirecting to the right authorization endpoint.
* It allows us to inject things securely as needed. For example Google oauth (but not B2C) requires us to send a `client_secret` in the authorization_code flow. We don't want to put the secret in the UI, so we can have the backend read it from vault and inject it to the token request.

Note Google is still enabled for now. I have tested this alongside the above Terra UI PR which switches to [react-oidc-context](https://github.com/authts/react-oidc-context). The UI piece has some TODOs but I believe these Orch endpoints are working well.

----

In all cases:


src/main/resources/swagger/api-docs.yaml
```diff
@@ -3361,24 +3361,6 @@ paths:
           description: Internal Server Error
           content: {}
       x-passthrough: false
-  /api/refresh-token-status:
```
rtitle, line 3364: Decided to remove these deprecated oauth APIs.

Reminds me -- need to document the new APIs in Swagger.

rtitle:
I am comparing this with the Java version of this functionality in TDR here: https://github.com/DataBiosphere/jade-data-repo/pull/1259/

Have a couple improvements in mind, going to push more commit(s).

rtitle:
Actually gonna close this for now.. thinking more, I might try and put this functionality in workbench-libs from the start since more services will need it.

PR #930, opened 2022-03-29
[AJ-267] scala 2.13

MatthewBemis:
\<your comments for this PR go here\>

Have you read [CONTRIBUTING.md](../CONTRIBUTING.md) lately? If not, do that first.

I, the developer opening this PR, do solemnly pinky swear that:

- [ ] I've followed [the instructions](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/CONTRIBUTING.md#api-changes) if I've made any changes to the API, _especially_ if they're breaking changes
- [ ] I've updated the RC_XXX release ticket with any manual steps required to release this change
- [ ] I've updated the [FISMA documentation](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/CONTRIBUTING.md#fisma-documentation-changes) if I've made any security-related changes, including auth, encryption, or auditing

In all cases:

- [ ] Get two thumbsworth of review and PO signoff if necessary
- [ ] Verify all tests go green
- [ ] Squash and merge; you can delete your branch after this **unless it's for a hotfix**. In that case, don't delete it!
- [ ] Test this change deployed correctly and works on dev environment after deployment

PR #938, opened 2022-04-22
use  Java 17, fix compile warnings [AJ-382]

davidangb:
1. Update Orchestration to use and compile to Java 17. See companion PR broadinstitute/firecloud-develop#2879 for the local dev environment.
2. Removes compile-time warnings about ["in" vs slash syntax ](https://www.scala-sbt.org/1.x/docs/Migrating-from-sbt-013x.html#slash)in sbt project build files
3. Removes, from the automation subproject, compile-time warnings left over from the Scala 2.13 upgrade at #930
4. Upgrades sbt from 1.4.9 to 1.6.2


automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/OrchestrationApiSpec.scala
```diff
@@ -69,7 +69,7 @@ class OrchestrationApiSpec
         verifyDatasetPermissions(Set(NihDatasetPermission("TCGA", true), NihDatasetPermission("TARGET", true)))
         Orchestration.NIH.addUserInNIH(OrchConfig.Users.genericJsonWebTokenKey)
         verifyDatasetPermissions(Set(NihDatasetPermission("TCGA", false), NihDatasetPermission("TARGET", false)))
-      } finally resetNihLinkToInactive
+      } finally resetNihLinkToInactive()
```
davidangb, line 72: these are Scala 2.13 fixes that we missed in #930

automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/StorageApiSpec.scala
```diff
@@ -30,7 +29,6 @@ import scala.language.implicitConversions
 
 class StorageApiSpec extends AnyFreeSpec with StorageApiSpecSupport with Matchers with LazyLogging {
   implicit val system: ActorSystem = ActorSystem()
-  implicit val materializer: ActorMaterializer = ActorMaterializer()
```
davidangb, line 33: a Scala 2.13 fix that we missed in #930 - ActorMaterializer is now automatically found from the ActorSystem, we don't need to create one ourselves

build.sbt
```diff
@@ -14,4 +14,4 @@ Revolver.enableDebugging(port = 5051, suspend = false)
 // itself rather than sbt, but they are not passed by default to the application, which is a forked
 // process. This passes them through to the "re-start" command, which is probably what a developer
 // would normally expect.
-javaOptions in reStart ++= sys.env("JAVA_OPTS").split(" ").toSeq
+reStart / javaOptions ++= sys.env("JAVA_OPTS").split(" ").toSeq
```
davidangb, line 17: see https://www.scala-sbt.org/1.x/docs/Migrating-from-sbt-013x.html#slash

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.4.9
+sbt.version=1.6.2
```
davidangb, line 1: #930 upgraded various docker base images to `hseeberger/scala-sbt:eclipse-temurin-11.0.14.1_1.6.2_2.13.8` - which uses sbt 1.6.2. However, this build.properties still specified 1.4.9. Let's fix that.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM us.gcr.io/broad-dsp-gcr-public/base/jre:11-debian
+FROM us.gcr.io/broad-dsp-gcr-public/base/jre:17-debian
```
sagehen03, line 1: I know this predates this change, but why do we use the debian image here and temurin elsewhere?

sagehen03 (approved):
looks good as far as I can tell.  Trusting the compiler to save us in pre-prod environments if there are any issues.

Dockerfile
davidangb: I think this is just naming conventions from `dsp-appsec-blessed-images`. The "-debian" image is built on top of Temurin: https://github.com/broadinstitute/dsp-appsec-blessed-images/blob/main/jre/Dockerfile.17-debian

PR #945, opened 2022-05-05
build cleanup and streamlining [AJ-414]

davidangb:
cleaning up after #942.

This PR includes:
* addition of `-d clean reload update` to the jenkins build - I was seeing some build hangs, the `clean reload update` should fend off a class of hangs, and `-d` will help to debug them
* cleaner exclusion rules for the automation subdirectory (see details below)

This PR does _not_ upgrade all dependencies. I plan to follow with another PR for that; some of those will take some investigation and updating.

### Cleaner Excludes

All of the following applies to the automation subdirectory; this is not in the main root-directory build.

**The problem:** `com.readytalk:metrics-statsd-common:4.2.0` was failing to download - 404 in maven central and elsewhere. This caused builds to fail.

**The hacky fix in #942:** exclude all `com.readytalk` artifacts from all of workbench-model, workbench-google, workbench-service-test. This worked but was a big ugly hammer.

**The better fix:** `com.readytalk:metrics-statsd-common:4.2.0` is now stashed in our local Artifactory instance, so we do not need to exclude it at all. 

Additionally, there is a dependency from workbench-service-test -> workbench-google. Prior to #942, we _tried_ to exclude workbench-google from workbench-service-test. This was faulty - the exclude was missing an underscore. This is now fixed.


davidangb:
jenkins retest

davidangb:
jenkins retest

davidangb:
fiab-start failure accepting TOS for a user; jenkins retest

automation/project/Settings.scala
```diff
@@ -32,7 +32,7 @@ object Settings {
   //     (removed on April 22, 2018) `G` - show reminder of failed and canceled tests with full stack traces at the end of log file
   // `-fWD` - causes test results to be written to the summary.log with test duration but without colored text
   val testSettings = List(
-    testOptions in Test += Tests.Argument("-oFD", "-u", "test-reports", "-fWD", "test-reports/TEST-summary.log")
+    Test / testOptions += Tests.Argument("-oFD", "-u", "test-reports", "-fWD", "test-reports/TEST-summary.log")
```
davidangb, line 35: cleans up build warnings about deprecated syntax

davidangb:
I aborted the automation and integration runs in Jenkins; they had no errors but were hung. Jenkins problem? Jenkins retest.

davidangb:
fiab-start failure; jenkins retest

davidangb:
jenkins retest - are the three orch failures real or transient?

src/docker/install.sh
```diff
@@ -5,7 +5,7 @@ set -e
 ORCH_DIR=$1
 cd $ORCH_DIR
 
-sbt -batch compile
+sbt -batch -d clean reload update compile
```
davidangb, line 8: this _may_ be the source of some build wonkiness - we weren't cleaning before building. There may have been stale caches with incorrect info in them. sbt _should_ handle that but it also feels correct to have a clean step here.

The `-d` might be overkill but was helpful during this PR, so I'm leaving it in.

davidangb:
jenkins retest

src/docker/install.sh
jordansafer: Thank you!!

jordansafer (approved):

PR #946, opened 2022-05-05
remove admin SA registration from health checks - [QA-1855]

davidangb:
Don't automatically verify/register the admin SA's registration during health checks.

This SA is already registered into FiaBs during the fiab-populate process here:
https://github.com/broadinstitute/firecloud-develop/blob/5dec185aceafdca33d27e335d1997f612a8bbdbe/run-context/fiab/scripts/fiab-populate/create_groups.sh#L26
We see this in fiab-start logs with output such as:
```
12:48:25 The service account firecloud-qa@broad-dsde-qa.iam.gserviceaccount.com is now registered with FireCloud. You can share workspaces with this address, or use it to call APIs.
```

and in live envs, this SA should have been registered long ago. Does it add value for Orchestration to continually ping Sam to see if this SA stays registered?

Note that https://github.com/broadinstitute/firecloud-develop/pull/2910 will change where in the fiab-populate process the SA is registered, but it will still be registered.

---

This PR addresses QA-1855 - potentially only partially - by removing a source of resource leakage inside FiaBs. The resource leak occurs because the fiab-unpopulate process unregisters the admin SA, which in turn should clean up the SA's proxy group for that fiab. However, because Orchestration is still running inside the fiab, it could easily run its health checks, which would then re-register the SA and create a new proxy group. This new proxy group would never be cleaned up; the fiab-unpopulate code has already run. There may be other resource leaks; this is not guaranteed to be a comprehensive fix.


rtitle:
Even if this SA still needs to be registered it makes sense to pull it out of status checks. The register-on-status-check caused me some headache when I was testing B2C..

rtitle (approved):
This looks good to me -- nice that this simplifies things a bit.

jordansafer (approved):

PR #947, opened 2022-05-09
TOAZ-129: Support Google logins through B2C

rtitle:
Orch uses `UserInfo.id` for talking to Thurloe and potentially other things. Previously this was the google user id, but with B2C enabled it's the B2C user id. Change it to still use the google id if present to support existing Google users in the system.

We'll probably need to make a similar change for other services which use the googleId.

Tested in Terra UI + fiab with B2C enabled.

In all cases:


rtitle:
@Qi77Qi Leo may want to do this too (not sure if we use the user's googleId for anything..)

Qi77Qi (approved):
Lgtm! I'll start working on integrating b2c in Leo this sprint!

rtitle:
@Qi77Qi fyi, I think this code in Leo also needs to change to handle a Google token _or_ B2C token: https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/google/GoogleOAuth2Interpreter.scala#L24

Do we do something similar in relay-listener?

PR #954, opened 2022-05-24
[SU-108] Add endpoint to unlink NIH account

MatthewBemis:

In all cases:

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -3143,10 +3143,10 @@ paths:
         204:
           description: Successfully unlinked NIH account
         404:
-          description: linkedNihUsername key missing from status
+          description: NIH account link not found
           content: {}
         500:
-          description: Internal server error retrieving NIH status
+          description: Internal server error unlinking NIH account
           content: {}
       x-passthrough: false
   /api/notifications/general:
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -3133,6 +3133,22 @@ paths:
           description: Internal server error retrieving NIH status
           content: {}
       x-passthrough: false
+  /api/nih/unlink:
+    delete:
+      tags:
+        - NIH
+      summary: Unlinks a user's NIH account from their Terra account
+      operationId: NIHUnlink
+      responses:
+        204:
+          description: Successfully unlinked NIH account
+        404:
+          description: linkedNihUsername key missing from status
+          content: {}
+        500:
+          description: Internal server error retrieving NIH status
+          content: {}
+      x-passthrough: false
   /api/notifications/general:
     get:
       tags:
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -3133,7 +3133,7 @@ paths:
           description: Internal server error retrieving NIH status
           content: {}
       x-passthrough: false
-  /api/nih/unlink:
+  /api/nih/account:
     delete:
       tags:
         - NIH
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
@@ -80,6 +80,15 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
     saveKeyValues(userInfo, profilePropertyMap).map(_ => ())
   }
 
+  override def deleteKeyValue(forUserId: String, keyName: String, callerToken: WithAccessToken): Future[Try[Unit]] = {
+    wrapExceptions {
+      userAuthedRequest(Delete(UserApiService.remoteDeleteKeyURL.format(forUserId, keyName)), useFireCloudHeader = true, label = Some("HttpThurloeDAO.deleteKeyValue"))(callerToken) map { response =>
+        if(response.status.isSuccess) Try(())
+        else Try(throw new FireCloudException(s"Unable to delete key ${keyName} from user profile"))
+      }
+    }
+  }
+
   private def wrapExceptions[T](codeBlock: => Future[T]): Future[T] = {
     codeBlock.recover {
       case t: Throwable => {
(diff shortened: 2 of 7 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -3133,6 +3133,22 @@ paths:
           description: Internal server error retrieving NIH status
           content: {}
       x-passthrough: false
+  /api/nih/unlink:
+    delete:
```
nawatts, line 3137: Nit: `delete unlink` seems like a double negative. Would `delete link` or `post unlink` be a bit more conventional?

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
```diff
@@ -127,6 +128,29 @@ class NihService(val samDao: SamDAO, val thurloeDao: ThurloeDAO, val googleDao:
     thurloeDao.saveKeyValues(userInfo, profilePropertyMap)
   }
 
+  private def unlinkNihAccount(userInfo: UserInfo): Future[Unit] = {
+    val nihKeys = Set("linkedNihUsername", "linkExpireTime")
+
+    Future.traverse(nihKeys) { nihKey =>
+      thurloeDao.deleteKeyValue(userInfo.id, nihKey, userInfo)
+    } map { results =>
+      if(!results.forall(_.isSuccess))
```
nawatts: Is it a concern if only one key is successfully deleted? Does Thurloe provide an option to delete multiple keys in the same request?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
```diff
@@ -80,6 +80,15 @@ class HttpThurloeDAO ( implicit val system: ActorSystem, implicit val executionC
     saveKeyValues(userInfo, profilePropertyMap).map(_ => ())
   }
 
+  override def deleteKeyValue(forUserId: String, keyName: String, callerToken: WithAccessToken): Future[Try[Unit]] = {
```
nawatts, line 83: Out of curiosity, why is the return value here wrapped in a `Try`?

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
```diff
@@ -127,6 +128,29 @@ class NihService(val samDao: SamDAO, val thurloeDao: ThurloeDAO, val googleDao:
...
+        throw new FireCloudExceptionWithErrorReport(ErrorReport(StatusCodes.InternalServerError, "Unable to unlink NIH account"))
+    }
+  }
+
+  def unlinkNihAccountAndSyncSelf(userInfo: UserInfo): Future[Unit] = {
+    for {
+      _ <- unlinkNihAccount(userInfo)
+      _ <- ensureWhitelistGroupsExists()
+      _ <- Future.traverse(nihWhitelists) {
+        whitelist => removeUserFromNihWhitelistGroup(WorkbenchEmail(userInfo.userEmail), whitelist).recoverWith {
```
nawatts, line 152: Similar to above, is it a concern if deleting the keys in Thurloe succeeds but removing from the groups fails?

nawatts (approved):

src/main/resources/swagger/api-docs.yaml
MatthewBemis: I agree, I don't like putting the verb in the path either. I mainly did this to keep it consistent with how other endpoints are structured throughout orch/rawls. I'd be happier if it was `DELETE /api/nih/account`  which is more RESTful.

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
MatthewBemis: I might be able to do this with https://thurloe.dsde-dev.broadinstitute.org/#/Key-value%20pairs/setKey

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
MatthewBemis: Deleting the keys but not removing from the group is okay imo, the process that syncs NIH groups will realize it within 6 hours and correct it.

src/main/resources/swagger/api-docs.yaml
MatthewBemis: Made it nice in 6c7c6a0858d7a849bccd80daa7960d679fadd365

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
MatthewBemis: That endpoint doesn't appear to do what I want (it doesn't support removing keys), so we're stuck with the 2 calls to deleteKey. It isn't a concern though- we will still be able to properly update access to the NIH groups if only one key is removed. The only side effect will be that the user will see a value for the un-deleted key in their profile, but they can unlink again to clear it if they'd like.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpThurloeDAO.scala
MatthewBemis: Solely for consistency with the other DAO method that manipulates keys

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
```diff
@@ -127,6 +128,29 @@
...
   }
 
+  private def unlinkNihAccount(userInfo: UserInfo): Future[Unit] = {
+    val nihKeys = Set("linkedNihUsername", "linkExpireTime")
+
+    Future.traverse(nihKeys) { nihKey =>
+      thurloeDao.deleteKeyValue(userInfo.id, nihKey, userInfo)
+    } map { results =>
+      if(!results.forall(_.isSuccess))
+        throw new FireCloudExceptionWithErrorReport(ErrorReport(StatusCodes.InternalServerError, "Unable to unlink NIH account"))
```
tlangs: It looks like this is swallowing the exception thrown by `ThurloeDao.deleteKeyValue`, and will lose the key that failed to be deleted.

tlangs (changes requested):
Looks good, but I'm worried about losing information in the exceptions thrown by `ThurloeDao`

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
MatthewBemis: Addressed in 7d4194a6dc0d60f90e49aebff2cc6afb5d9e0702

tlangs: Thanks so much!

tlangs (approved):

PR #955, opened 2022-05-26
Aj400

calypsomatic:
AJ-400 
Creates a helper trait with functions for failing out of tests that use eventually more quickly and puts them in use in a single test file to begin with.

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/AsyncImportSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/AsyncImportSpec.scala
@@ -13,9 +13,11 @@ import org.broadinstitute.dsde.workbench.model.ErrorReport
 import org.broadinstitute.dsde.workbench.model.ErrorReportJsonSupport.ErrorReportFormat
 import org.broadinstitute.dsde.workbench.service.{AclEntry, Orchestration, RestException, WorkspaceAccessLevel}
 import org.parboiled.common.FileUtils
+import org.scalatest.Assertion
 import org.scalatest.OptionValues._
 import org.scalatest.concurrent.PatienceConfiguration.{Interval, Timeout}
 import org.scalatest.concurrent.{Eventually, ScalaFutures}
+import org.scalatest.exceptions.TestFailedException
 import org.scalatest.freespec.AnyFreeSpec
 import org.scalatest.matchers.should.Matchers
 import org.scalatest.time.{Minutes, Seconds, Span}
@@ -23,6 +25,7 @@ import spray.json._
 
 import java.util.UUID
 import java.util.concurrent.TimeUnit
+import javax.xml.parsers.FactoryConfigurationError
 import scala.io.Source
 import scala.util.{Failure, Success, Try}
 
@@ -46,6 +49,46 @@ class AsyncImportSpec
 
   lazy private val expectedEntities: Map[String, EntityTypeMetadata] = Source.fromResource("AsyncImportSpec-pfb-expected-entities.json").getLines().mkString
     .parseJson.convertTo[Map[String, EntityTypeMetadata]]
+
+
+
+  //TODO: These two methods and exception can be put in a utility class for other tests to use
+  class TestEscapedException(
+                              message: String,
+                              failedCodeStackDepth: Int,
+                            ) extends TestFailedException(message, failedCodeStackDepth){
+    def this(message: String) = this(message, 0)
+  }
+
+  /**
+    * Functions as an assertion unless some fail-immediately condition is met,
+    * in which case it throws a test-failing error
+    * @param assertion Scalatest assertion to be checked for success or failure
+    * @param failCondition Condition that, if met, will end an asynchronous test with failure immediately
+    * @param failMessage Message to report in test failure
+    * @return test assertion
+    */
+  def assertWithFailure(assertion: => Assertion, failCondition: Boolean, failMessage: String): Assertion =
+    if (failCondition) throw new FactoryConfigurationError(failMessage)
+    else assertion
+
+  /**
+    * Wraps a function in an eventually wrapped in a try-catch;
+    * Functions as an eventually but will fail test without retrying if test-ending errors thrown in the test function
+    * @param timeout
+    * @param interval
+    * @param testFunction Test to run; should include an assertWithFailure statement
+    * @tparam T the result of the testFunction
+    * @return The result of the testFunction (an assertion) if an error is not thrown
+    */
+  def eventuallyWithFail[T](timeout: Timeout, interval: Interval, testFunction: => T): T =
+    try {
+      eventually(timeout, interval){
+        testFunction
+      }
+    } catch {
+      case ex:FactoryConfigurationError => throw new TestEscapedException(ex.getMessage)
+    }
 
   "Orchestration" - {
 
@@ -235,7 +278,7 @@ class AsyncImportSpec
 
     // poll for completion as owner
     withClue(s"import job $importJobId failed its eventually assertions on job status: ") {
-      eventually(timeout = Timeout(scaled(Span(10, Minutes))), interval = Interval(scaled(Span(5, Seconds)))) {
+      eventuallyWithFail(timeout = Timeout(scaled(Span(10, Minutes))), interval = Interval(scaled(Span(5, Seconds))), {
         val requestId = scala.util.Random.alphanumeric.take(8).mkString // just to assist with logging
         logger.info(s"[$requestId] About to check status for import job $importJobId. Elapsed: ${humanReadableMillis(System.currentTimeMillis()-startTime)}")
         val resp: HttpResponse = Orchestration.getRequest(s"${workspaceUrl(projectName, workspaceName)}/importJob/$importJobId")
@@ -244,8 +287,8 @@ class AsyncImportSpec
         respStatus shouldBe StatusCodes.OK
         val importStatus = blockForStringBody(resp).parseJson.asJsObject.fields.get("status").value
         logger.info(s"[$requestId] Import Service job status for import job $importJobId is [$importStatus]")
-        importStatus shouldBe JsString ("Done")
-      }
+        assertWithFailure(importStatus shouldBe JsString ("Done"), importStatus.equals(JsString ("Error")), "Import error occurred, failing test")
+      })
     }
 
     logger.info(s"$projectName/$workspaceName import job $importJobId completed for file $importFilePath in " +
--- a/automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/AsyncImportSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/AsyncImportSpec.scala
@@ -68,11 +68,11 @@ class AsyncImportSpec
             }
 
             // poll for completion as owner
-            eventuallyWithFail{
+            eventuallyWithEscape{
             val resp: HttpResponse = Orchestration.getRequest( s"${workspaceUrl(projectName, workspaceName)}/importPFB/$importJobId")
             resp.status shouldBe StatusCodes.OK
             val importStatus = blockForStringBody(resp).parseJson.asJsObject.fields.get("status").value
-            assertWithFailure(
+            assertWithEscape(
                importStatus shouldBe JsString("Done"),
               importStatus.equals(JsString("Error")),
               "Error encountered during import, failing test early"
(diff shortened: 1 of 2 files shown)


automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/AsyncTestUtils.scala
```diff
@@ -0,0 +1,79 @@
...
+  /**
+    * Wraps a function in an eventually wrapped in a try-catch;
+    * Functions as an eventually but will fail test without retrying if test-ending errors thrown in the test function
+    * @param timeout
+    * @param interval
+    * @param testFunction Test to run; should include a failAndEscape (could be within assertWithEscape)
+    * @tparam T the result of the testFunction
+    * @return The result of the testFunction (an assertion) if an error is not thrown
+    */
+  def eventuallyWithEscape[T](timeout: Timeout, interval: Interval, testFunction: => T): T =
```
jordansafer: nit: I'd vote to put the timeout/interval arguments second, and then make them optional. Something like:
`def eventuallyWithEscape[T](testFunction: => T, timeout: Timout=None, interval=None): T =`
And then combine the two functions. Maybe there's an even fancier technique out there to mirror the eventually functions? (Is it possible to have a curried function where the first parameter is either the timeout or the function? I think this would be more work than it's worth)

automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/AsyncImportSpec.scala
```diff
@@ -67,11 +68,16 @@ class AsyncImportSpec
...
+            val importStatus = blockForStringBody(resp).parseJson.asJsObject.fields.get("status").value
+            assertWithEscape(
+               importStatus shouldBe JsString("Done"),
+              importStatus.equals(JsString("Error")),
+              "Error encountered during import, failing test early"
+            )
+          }
 
             // inspect data entities and confirm correct import as owner
             eventually {
```
jordansafer, line 82: Should we migrate this eventually call too?

jordansafer (approved):
Looks good!!

automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/AsyncTestUtils.scala
calypsomatic: In the original eventually they're implicit https://github.com/scalatest/scalatest/blob/7aeda262e249a66d2319b6b2f5daa1b7d0cf9a4f/jvm/core/src/main/scala/org/scalatest/concurrent/Eventually.scala#L414
so that if you don't supply them it grabs them from PatienceConfig.  So I was relying on the `eventually` inside `eventuallyWithEscape` to take care of this in the second function that only has one parameter.  I don't quite understand how implicit works so I didn't want to try doing that myself, but I'm happy to rewrite/rearrange these functions however works best.

automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/AsyncImportSpec.scala
calypsomatic: For this one and a few others it wasn't obvious to me what a fail condition would be.  I just updated the ones that are checking an import status, since it's easy to know that "Error" means it's not ever going to get to done.  This one is checking all the contents of a map.

automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/AsyncImportSpec.scala
```diff
@@ -67,11 +68,16 @@ class AsyncImportSpec
...
+            eventuallyWithEscape{
+            val resp: HttpResponse = Orchestration.getRequest( s"${workspaceUrl(projectName, workspaceName)}/importPFB/$importJobId")
+            resp.status shouldBe StatusCodes.OK
+            val importStatus = blockForStringBody(resp).parseJson.asJsObject.fields.get("status").value
+            assertWithEscape(
+               importStatus shouldBe JsString("Done"),
+              importStatus.equals(JsString("Error")),
+              "Error encountered during import, failing test early"
+            )
+          }
```
davidangb: nit: indentation/formatting

EDIT: I'm seeing some funky formatting in a few places, maybe give this a `Code -> Reformat Code` in IntelliJ?

automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/AsyncImportSpec.scala
```diff
@@ -244,8 +255,8 @@ class AsyncImportSpec
         respStatus shouldBe StatusCodes.OK
         val importStatus = blockForStringBody(resp).parseJson.asJsObject.fields.get("status").value
         logger.info(s"[$requestId] Import Service job status for import job $importJobId is [$importStatus]")
-        importStatus shouldBe JsString ("Done")
-      }
+        assertWithEscape(importStatus shouldBe JsString ("Done"), importStatus.equals(JsString ("Error")), "Import error occurred, failing test")
+      })
```
davidangb: this changes how the parens are balanced - it moves one close-paren from line 249 down to line 259. See my comment below on how to avoid that …

automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/AsyncTestUtils.scala
jordansafer: I see. I would still vote for optional arguments in that case, to reduce duplication. We could just call eventually with the timeout arguments if they're not none, but leave them out otherwise.

automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/AsyncImportSpec.scala
jordansafer: Makes sense!

automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/AsyncImportSpec.scala
```diff
@@ -31,6 +31,7 @@ class AsyncImportSpec
   extends AnyFreeSpec
     with Matchers
     with Eventually
+    with AsyncTestUtils
```
davidangb, line 33: nit: since `AsyncImportSpec` extends `AsyncTestUtils`, and `AsyncTestUtils` extends `Eventually`, you could remove the `Eventually` from `AsyncImportSpec`

automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/AsyncTestUtils.scala
davidangb: if you change this signature to be:
```scala
  def eventuallyWithEscape[T](timeout: Timeout, interval: Interval)(testFunction: => T)
```
you can avoid rebalancing the parens in `AsyncImportSpec`. BUT, see the other thread on just copying the method signature from ScalaTest's `eventually`.

davidangb: I think the best thing to do here is to copy the signature of the `eventually` that you're wrapping, which is [this one](https://github.com/scalatest/scalatest/blob/7aeda262e249a66d2319b6b2f5daa1b7d0cf9a4f/jvm/core/src/main/scala/org/scalatest/concurrent/Eventually.scala#L312). Since its signature is
```scala
def eventually[T](timeout: Timeout, interval: Interval)(fun: => T)(implicit retrying: Retrying[T], pos: source.Position): T
```
you'd have:
```scala
def eventuallyWithEscape[T](timeout: Timeout, interval: Interval)(testFunction: => T)(implicit retrying: Retrying[T], pos: source.Position): T
```
( do the same for the `eventuallyWithEscape` at line 40 too )

FYI, IntelliJ's `View -> Show Implicit Hints` and sometimes `View -> Expand Implicit Hints` can be helpful to follow all the sneaky implicits through! Mostly I hate implicits.

automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/AsyncTestUtils.scala
```diff
@@ -0,0 +1,79 @@
...
+    } catch {
+      case ex:FactoryConfigurationError => throw new TestEscapedException(ex.getMessage)
+    }
+
+  /**
+    * Will fail a test that uses eventually instead of retrying
+    * @param message Error message
+    * @return
+    */
+  def failAndEscape(message: String): Nothing = { throw new FactoryConfigurationError(message) }
```
davidangb: not necessary for this PR, because this is all rather prospective at this point: I could see some point in the future also adding a 
```scala
def failAndEscape(exception: Exception): Nothing = throw new FactoryConfigurationError(exception)
```
method that allows the original exception to propagate. Of course this would need some corresponding support in `eventuallyWithEscape` to check if the `FactoryConfigurationError` contains an exception and rethrow it.

automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/AsyncTestUtils.scala
```diff
@@ -0,0 +1,79 @@
...
+    * in which case it throws a test-failing error
+    * @param assertion Scalatest assertion to be checked for success or failure
+    * @param failCondition Condition that, if met, will end an asynchronous test with failure immediately
+    * @param failMessage Message to report in test failure
+    * @return test assertion
+    */
+  def assertWithEscape(assertion: => Assertion, failCondition: Boolean, failMessage: String): Assertion =
+    if (failCondition) failAndEscape(failMessage)
+    else assertion
+}
```
davidangb, line 81: does this work if `failCondition` is also a pass-by-name `Assertion`? Like so:
```scala
  def assertWithEscape(assertion: => Assertion, failCondition: => Assertion, failMessage: String): Assertion = {
    try {
      failCondition
    } catch {
      case ex: Exception => failAndEscape(ex)
    }
    assertion
  }
```

then callers could do something like:
```scala
assertWithEscape(
  importStatus shouldBe JsString("Done"),
  importStatus should not be JsString("Error"),
  "Error encountered during import, failing test early")
```

davidangb (commented):
nice job on this! Comments inline …

automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/AsyncTestUtils.scala
calypsomatic: I like that idea!

automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/AsyncImportSpec.scala
```diff
@@ -228,42 +238,42 @@ class AsyncImportSpec
...
         val respStatus = resp.status
-        logger.info(s"[$requestId] HTTP response status for import job $importJobId status request is [${respStatus.intValue()}]. Elapsed: ${humanReadableMillis(System.currentTimeMillis()-startTime)}")
+        logger.info(s"[$requestId] HTTP response status for import job $importJobId status request is [${respStatus.intValue()}]. Elapsed: ${humanReadableMillis(System.currentTimeMillis() - startTime)}")
         respStatus shouldBe StatusCodes.OK
         val importStatus = blockForStringBody(resp).parseJson.asJsObject.fields.get("status").value
         logger.info(s"[$requestId] Import Service job status for import job $importJobId is [$importStatus]")
-        importStatus shouldBe JsString ("Done")
-      }
+        assertWithEscape(importStatus shouldBe JsString("Done"), importStatus should not equal JsString("Error"), "Import error occurred, failing test")
+      })
```
davidangb: now that the signatures for `eventuallyWithEscape` mirror those of `eventually`, please revert the changes to parens here - move the close-paren from line 258 to line 248

automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/AsyncTestUtils.scala
```diff
@@ -0,0 +1,94 @@
...
+    * @param assertion     Scalatest assertion to be checked for success or failure
+    * @param failCondition Condition that, if met, will end an asynchronous test with failure immediately
+    * @param failMessage   Message to report in test failure
+    * @return test assertion
+    */
+  def assertWithEscape(assertion: => Assertion, failCondition: => Assertion, failMessage: String): Assertion = {
+    try {
+      failCondition
+    } catch {
+      case ex: Exception => failAndEscape(failMessage)
```
davidangb: this will lose all details from the thrown exception; how about:
```scala
case ex: Exception => failAndEscape(s"$failMessage - underlying error message is: ${ex.getMessage}")
```
or something like that?

automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/AsyncImportSpec.scala
calypsomatic: Whoops! Forgot that part.  Thanks!

davidangb (approved):
:+1: assuming tests pass!

PR #956, opened 2022-06-03
delete storageservicespec tests that mostly just call Google

sagehen03:
I'm recommending deleting these tests because in terms of Broad code there are three things that these tests exercise:

1. https://github.com/broadinstitute/firecloud-orchestration/blob/219bdb1a7c4c85df5350f27be9292bb66233a355/src/main/scala/org/broadinstitute/dsde/firecloud/service/StorageService.scala#L37-L52
2. https://github.com/broadinstitute/firecloud-orchestration/blob/219bdb1a7c4c85df5350f27be9292bb66233a355/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/StorageApiService.scala#L19-L28
3. https://github.com/broadinstitute/workbench-libs/blob/258d4835c0cee75218cfb3718ea8624a0350a395/serviceTest/src/test/scala/org/broadinstitute/dsde/workbench/service/RestClient.scala#L143-L148

I've added unit test coverage for 1 as part of this PR.  We can add coverage for 2 (though I could probably use some Scala test help in doing so), and I hope there's unit test coverage somewhere for our RestClient that ensures that exceptions and messages are propagated as expected to the caller.  Essentially, my argument for removing these tests is twofold, first, we can rely on unit tests (either existing or new) to replace the testing of Broad code that occurs with these FIAB tests.  Second, calling Google doesn't really work well as a test.  Google can in theory change their code so that the assertions in these tests will fail (I would bet that they would never do that without versioning, deprecation warnings, sunset schedule etc), but even if they did these tests don't serve the function of protecting us from releasing buggy code.  If Google changes their API behavior, they break our tests and they simultaneous break production.  The tests don't save protect us under that scenario.  I could see some use for this type of code for the sake of monitoring so long as the application can degrade gracefully under those circumstances (not likely for terra and gcp I think).

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/StorageApiSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/StorageApiSpec.scala
@@ -54,43 +54,6 @@ class StorageApiSpec extends AnyFreeSpec with StorageApiSpecSupport with Matcher
   // check for a 403 when requesting non-existent objects, which may seem wrong at first glance to anyone
   // reading this code.
 
-  "metadata endpoint" - {
-
-    "should return metadata response for a public object" in {
-      implicit val authToken: AuthToken = student.makeAuthToken()
-      val result:ObjectMetadata = Orchestration.storage.getObjectMetadata(publicUrl.bucketName, publicUrl.objectName)
-      assertResult(publicUrl.bucketName.value) { result.bucket }
-      assertResult(publicUrl.objectName.value) { result.name }
-    }
-
-    "should return metadata response for an object I have permissions to" in {
-      implicit val token = student.makeAuthToken()
-      withSmallFile { smallFile =>
-        setStudentAndSA(smallFile, student)
-        val result:ObjectMetadata = Orchestration.storage.getObjectMetadata(smallFile.bucketName, smallFile.objectName)
-        assertResult(smallFile.bucketName.value) { result.bucket }
-        assertResult(smallFile.objectName.value) { result.name }
-      }
-    }
-
-    "should return 403 for a non-existent object" in {
-      implicit val authToken: AuthToken = student.makeAuthToken()
-      val requestEx = intercept[RestException] {
-        Orchestration.storage.getObjectMetadata(nonExistent.bucketName, nonExistent.objectName)
-      }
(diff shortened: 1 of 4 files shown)


automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/StorageApiSpec.scala
```diff
@@ -54,43 +54,6 @@ class StorageApiSpec extends AnyFreeSpec with StorageApiSpecSupport with Matcher
   // check for a 403 when requesting non-existent objects, which may seem wrong at first glance to anyone
   // reading this code.
```
calypsomatic, line 56: This note here says that these tests are checking that errors get propagated to the user.  Would we want to have one test here that mocks google with an error code just to check error propagation?

davidangb:
> We can add coverage for 2 (though I could probably use some Scala test help in doing so

Take a look at the tests under `org.broadinstitute.dsde.firecloud.webservice` … `PermissionReportApiSpec` might be a good example. It sets up a `val testApp = app.copy(…` pointing at a couple mock DAOs, then runs some route tests. `StatusApiServiceSpec` is a smaller/clearer example of running route tests.

davidangb (approved):
Changes in here look good to me!

If it's simple enough to add some unit tests (which should execute quickly) for the `/api/storage/.../.…` route, great. If it seems expensive to write those, I'd check how often these APIs are called in practice - if infrequent, it doesn't feel worth the cost:benefit.

automation/src/test/scala/org/broadinstitute/dsde/test/api/orch/StorageApiSpec.scala
sagehen03: We could add something like 
```
  "should propagate exception" in {
      val error = intercept[FireCloudExceptionWithErrorReport] {
        Await.result(storageService.getObjectStats("exception", "bar"), Duration.Inf)
      }
      assert(error.getMessage.contains("No access"))
    }
```
But I'm not convinced it tests much worthwhile.  Were you thinking along those ^^ lines or something else?

calypsomatic: Sure, I was just trying to honor the comment.  If it's not worth it, I'm not attached to it.

jordansafer (approved):

PR #957, opened 2022-06-08
Delete more orch FiaB tests

sagehen03:
[Here's](https://github.com/broadinstitute/firecloud-orchestration/blob/a1cdc2fbe4ec3ef7dadc389c67d3d5dc6012a755/src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala#L37-L53) the Orch code under test and that code is covered by [this spec](https://github.com/broadinstitute/firecloud-orchestration/blob/a1cdc2fbe4ec3ef7dadc389c67d3d5dc6012a755/src/test/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiServiceSpec.scala) 

Looking into the Rawls calls that Orch makes: 

1. [getWorkspace](https://github.com/broadinstitute/rawls/blob/98485d0cca4d4956d4eef5c7a6f6bff055de24a5/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala#L203) is tested at the service level but not the API level--so we could pull the tests in [WorkspaceServiceSpec](https://github.com/broadinstitute/rawls/blob/ee05d4ead20587eacbcf097d6b1367fdd1ae552b/core/src/test/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceServiceSpec.scalal) up into [WorkspaceApiServiceSpec](https://github.com/broadinstitute/rawls/blob/5a4d291fed54f6256a0240de6d84e49ac8263098/core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiServiceSpec.scala)
2. [getBucketUsage](https://github.com/broadinstitute/rawls/blob/98485d0cca4d4956d4eef5c7a6f6bff055de24a5/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala#L2005) is covered starting at the API level in [WorkspaceApiServiceSpec](https://github.com/broadinstitute/rawls/blob/5a4d291fed54f6256a0240de6d84e49ac8263098/core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiServiceSpec.scala) so I don't think we need any action there.


calypsomatic (approved):

davidangb (approved):
I am not concerned about the coverage of `getWorkspace` here. These tests are specifically written to assert that readers do not see storage cost estimates, but owners and writers do. Any other code coverage is a side effect and we shouldn't really count on the side effect coverage.

https://github.com/broadinstitute/rawls/blob/5a4d291fed54f6256a0240de6d84e49ac8263098/core/src/test/scala/org/broadinstitute/dsde/rawls/webservice/WorkspaceApiServiceSpec.scala#L1966-L1999 sure looks like it handles what these tests are supposed to do. :+1:!

PR #959, opened 2022-06-24
AJ-479 accomodate larger TSV uploads and handle lists of lists

sagehen03:
remove column width check and store a list of lists

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVParser.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVParser.scala
@@ -19,7 +19,7 @@ object TSVParser {
     // Automatically detect what the line separator is (e.g. \n for Unix, \r\n for Windows).
     settings.setLineSeparatorDetectionEnabled(true)
     settings.setMaxColumns(1024)
-    settings.setMaxCharsPerColumn(16384)
+    settings.setMaxCharsPerColumn(-1)
     settings.getFormat.setDelimiter('\t')
     // By default, the CsvParser returns null for missing fields, however the application expects the
     // empty string. These replace all nulls with the empty string.
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVParser.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVParser.scala
@@ -19,7 +19,8 @@ object TSVParser {
     // Automatically detect what the line separator is (e.g. \n for Unix, \r\n for Windows).
     settings.setLineSeparatorDetectionEnabled(true)
     settings.setMaxColumns(1024)
-    settings.setMaxCharsPerColumn(-1)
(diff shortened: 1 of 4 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVParser.scala
```diff
@@ -19,7 +19,7 @@ object TSVParser {
     // Automatically detect what the line separator is (e.g. \n for Unix, \r\n for Windows).
     settings.setLineSeparatorDetectionEnabled(true)
     settings.setMaxColumns(1024)
-    settings.setMaxCharsPerColumn(16384)
+    settings.setMaxCharsPerColumn(-1)
```
davidangb: I still think it's better to set this to a finite but large value - 128MB, even - as a DoS protection.

Consider also setting [setErrorContentLength](http://docs.univocity.com/parsers/2.7.3/com/univocity/parsers/common/CommonSettings.html#setErrorContentLength-int-) to some reasonable value … if we're increasing column size, we don't want to arbitrarily increase error message size too.

sagehen03: Why do you think it's better?  I feel like an attacker would quickly defeat this half measure--the error message would even give them a clue to keep their columns less than a certain size.  If your answer is some defense is better than none, I would argue that this defense creates support tickets and work for devs and would only stop a lazy/half interested attacker who would stop at one attempt.  It feels like building a wall around half the castle, and I think there's a strong chance another customer will come along and need to insert data larger than whatever threshold we pick this time at some point in the future.

davidangb: yeah, my stance is that some defense is better than none. Absolutely this would not stop a malicious attacker. It does provide some protection, though, against accidents. We've seen instances, for example, where a user attempted to insert the _contents_ of a bam file instead of the _path_ to that file. Or, where a user made a mistake with TSV delimiters and created very large columns. 

The limit of 16,777,216 in the latest commit is well above the limit of 65,535 imposed at the database level in the `value_string` column. The only use case where the 16,777,216 limit will be relevant is if the user is inserting raw json. If we realize we have users who need to insert > 16MB of json into the entity service, I *want* the system to throw an error - there are so many other places this could cause problems, such as the UI; we haven't designed for that. So, if it comes to that, let's encounter errors and make conscious choices about which features to build.

sagehen03: Fair points!

davidangb (approved):
If the user uploads a TSV that is over the (new) column limit, is the TSV content included in the error response? If so, I think we need to set [setErrorContentLength](http://docs.univocity.com/parsers/2.7.3/com/univocity/parsers/common/CommonSettings.html#setErrorContentLength-int-) to a low-ish value to avoid 16MB-large error responses. If the error response does not include the user's column value, all set.

proactive :+1:

PR #962, opened 2022-06-28
Entity type/attribute rename/delete passthroughs [SUP-474]

davidangb:
Copying the release notes from [SUP-474](https://broadworkbench.atlassian.net/browse/SUP-474):
Three new APIs are now available at api.firecloud.org; these were previously available only at rawls.dsde-prod.broadinstitute.org. The three APIs are:
* rename entity type, aka rename data table, at `PATCH /api/workspaces/{workspaceNamespace}/{workspaceName}/entityTypes/{entityType}`
* delete all entities of type, aka delete data table, at `DELETE /api/workspaces/{workspaceNamespace}/{workspaceName}/entityTypes/{entityType}`
* rename entity attribute, aka rename data table column, at `PATCH /api/workspaces/{workspaceNamespace}/{workspaceName}/entityTypes/{entityType}/attributes/{attributeName}`

---

no tests, because these are purely passthroughs. I have tested empirically by running Orch locally and seeing that the requests flow through to the right place in Rawls.

PR #964, opened 2022-07-05
passthrough for rename entity [SUP-691]

davidangb:
Followup to #962.

I misunderstood the original Jira ticket, SUP-474, and added the wrong passthroughs in #962 (though those passthroughs are still valuable).

In this PR, now associated with SUP-691, I am adding a passthrough for rename-entity, which is the one originally requested by the user.

Tested manually by running a local orchestration and seeing it work. No additional automated tests since this is just a passthrough.

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -4527,6 +4527,44 @@ paths:
       x-passthrough: true
       x-passthrough-target: rawls
       x-codegen-request-body-name: expression
+  /api/workspaces/{workspaceNamespace}/{workspaceName}/entities/{entityType}/{entityName}/rename:
+    post:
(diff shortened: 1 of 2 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -8419,6 +8457,15 @@ components:
...
           description: Entity Name
+    EntityName:
+      required:
+        - name
+      type: object
+      properties:
+        name:
+          type: string
+          description: The name of the entity
+      description: ""
```
davidangb, line 8468: swagger changes copied from Rawls, since that's where the API is defined

calypsomatic (approved):

jordansafer (approved):

PR #966, opened 2022-07-13
SU-146: Make valid preference keys configurable, add support for starredWorkspaces key

MatthewBemis:
The purpose of this PR is to add a new allowed Thurloe preference key, `starredWorkspaces`. This key will store the list of workspace IDs that the user has starred. We've chosen to put this in Thurloe because this feature needs to be per-user (so a workspace attribute isn't ideal, and readers need to be able to star workspaces as well) and it must be preserved across browsers/computers (so local storage is not suitable).

Terra UI will look something like this:
Screen Shot 2022-07-13 at 3 36 40 PM

In all cases:

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -41,6 +41,8 @@ userprofile {
   getAll="/thurloe/%s"
   getQuery="/thurloe"
   delete="/thurloe/%s/%s"
+  validPreferenceKeyPrefixes=["notifications/"]
+  validPreferenceKeys=["starredWorkspaces"]
 }
 
 googlecloud {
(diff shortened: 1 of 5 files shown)


nawatts (approved):

davidangb (commented):
:+1: for the syntax in this PR, assuming tests pass.

BUT … could you verify the max character count for Thurloe values? If https://github.com/broadinstitute/thurloe/blob/b47f580ce8af306a9a9bae6f5f85a5450ae05701/src/main/migrations/changesets/db_schema.xml#L17-L19 is the limit, that would start to break at ~64 workspace ids, depending on how you're storing them in an array. I suspect some users will run over that limit.

MatthewBemis:
@davidangb I did check this and felt that the 2048 chars would be enough. The rationale is that if you have >50 starred workspaces, you're probably either using the feature "incorrectly" _or_ we need to add some additional organization features like folders. We'll need to enforce that limit in the UI, however

davidangb (approved):

davidangb:
@MatthewBemis cool, you're on it … thumbed, then!

PR #967, opened 2022-07-13
Rawls response for delete-workspace is just a string [AJ-492]

davidangb:
Waaaaay back in https://github.com/broadinstitute/rawls/pull/1555, we changed Rawls' response payload for the delete-workspace API from a `PerRequestMessage` to a `String`. (It has since changed to an `Option[String]` but that's mostly irrelevant). Code link [here](https://github.com/broadinstitute/rawls/pull/1555/files#diff-786b2fbecdc6fe48eddc0850e532a3dfa215568f835aebdd3b9b7fed4f785bb4R355-R357).

Orchestration, however, is not a pure passthrough for delete-workspace, since it needs to handle Library unpublish duties if a user deletes a workspace. Orchestration is "strongly typed" in that it deserializes Rawls' response into a model object.

Before this PR, it was attempting to deserialize the `String` response from Rawls into a `WorkspaceDeleteResponse`. This would fail, and Orch would trigger its `recover` handler, prepending "You cannot delete this workspace: " to the Rawls response.

## Impact
All delete-workspace requests that ran through Orch would appear to fail, even if Rawls actually succeeded to delete the workspace. In the case of a successful deletion, Orch would return a 202 status code but still say "You cannot delete this workspace". This caused a lot of noise in swatomation tests, because it looked as if the tests were littered with cleanup failures.

## This PR
Remove the `WorkspaceDeleteResponse` model and, everywhere Orch works with the Rawls response, use an `Option[String]` instead.

### Alternatives Considered
_These are all larger refactors than I wanted to tackle in this PR ..._
* Refactor to remove deserialization of Rawls' response entirely, and just pass on the response.
* Refactor delete-workspace to be a passthrough with a post-response fire-and-forget unpublish

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
@@ -216,8 +216,8 @@ class HttpRawlsDAO(implicit val system: ActorSystem, implicit val materializer: 
     }
   }
 
-  override def deleteWorkspace(workspaceNamespace: String, workspaceName: String)(implicit userToken: WithAccessToken): Future[WorkspaceDeleteResponse] = {
-    authedRequestToObject[WorkspaceDeleteResponse](Delete(getWorkspaceUrl(workspaceNamespace, workspaceName)))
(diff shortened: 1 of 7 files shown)


calypsomatic (approved):

jordansafer (approved):

MatthewBemis:
Thank you for fixing the bug that I introduced! 😄

davidangb:
@MatthewBemis I reviewed that PR and didn't catch it either!

PR #818, opened 2020-11-23
AS-542: Use gh actions for unit tests [risk: low]

davidangb:
Use GH Actions for unit tests instead of Travis.

Because the GH Actions -> Coveralls integration for sbt is weak (see for instance https://github.com/lemurheavy/coveralls-public, issue 1388), this PR also:
* switches from `scoverage` to `sbt/sbt-jacoco` to generate coverage reports
* switches from coveralls to codecov as our hosted coverage service

Note that the Travis checks for this PR will fail, because I've removed the .travis.yml; this is expected until we disable Travis for this repo. Once this PR is approved, I'll change which checks are required for PRs.

-----

Have you read [CONTRIBUTING.md](../CONTRIBUTING.md) lately? If not, do that first.

I, the developer opening this PR, do solemnly pinky swear that:

- [ ] I've followed [the instructions](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/CONTRIBUTING.md#api-changes) if I've made any changes to the API, _especially_ if they're breaking changes
- [ ] I've updated the RC_XXX release ticket with any manual steps required to release this change
- [ ] I've updated the [FISMA documentation](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/CONTRIBUTING.md#fisma-documentation-changes) if I've made any security-related changes, including auth, encryption, or auditing

In all cases:

- [ ] Get two thumbsworth of review and PO signoff if necessary
- [ ] Verify all tests go green
- [ ] Squash and merge; you can delete your branch after this **unless it's for a hotfix**. In that case, don't delete it!
- [ ] Test this change deployed correctly and works on dev environment after deployment

PR #971, opened 2022-07-19
fix code coverage [AJ-509]

davidangb:
Code coverage was pretty broken. This repo has been at 0% coverage for … as long as codecov has been tracking PRs: https://app.codecov.io/gh/broadinstitute/firecloud-orchestration/pulls. This may go all the way back to #818 .

If nobody noticed or cared that coverage was broken, it begs the question: should we even have code coverage? I'm keeping it in place for now because it doesn't add _too_ much overhead.

### Coverage generation
switch from `sbt-jacoco` to `sbt-scoverage`. JaCoCo was consistently generating 0% coverage and I couldn't figure out why. We use scoverage elsewhere (e.g. Rawls) so I just switched. 

### Codecov uploading
Bonus fix rom https://github.com/codecov/codecov-action:
> As of February 1, 2022, v1 has been fully sunset and no longer functions

Which doesn't seem to be true; the v1 action was succeeding at uploading our 0% coverage reports. Nonetheless, let's update to their recommended version.

### DIFF
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -34,6 +34,6 @@ jobs:
         FIRE_CLOUD_ID: 123
       run: sbt jacoco
 
-    - uses: codecov/codecov-action@v1
+    - uses: codecov/codecov-action@v3
       with:
         file: ./target/scala-2.13/jacoco/report/jacoco.xml
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -32,8 +32,7 @@ jobs:
         RAWLS_URL_ROOT: http://localhost:8990
         THURLOE_URL_ROOT: http://localhost:8991
         FIRE_CLOUD_ID: 123
-      run: sbt jacoco
+      run: sbt clean coverage test coverageReport
 
     - uses: codecov/codecov-action@v3
-      with:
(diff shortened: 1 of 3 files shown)


.github/workflows/unittest.yml
```diff
@@ -32,8 +32,7 @@ jobs:
...
         THURLOE_URL_ROOT: http://localhost:8991
         FIRE_CLOUD_ID: 123
-      run: sbt jacoco
+      run: sbt clean coverage test coverageReport
 
-    - uses: codecov/codecov-action@v1
-      with:
-        file: ./target/scala-2.13/jacoco/report/jacoco.xml
+    - uses: codecov/codecov-action@v3
+      if: ${{ always() }}
```
calypsomatic, line 38: Does it not always run without an 'if'?

davidangb: by default, if a previous step of the job fails, subsequent steps will not run - unless the subsequent jobs' `if` clause says it should: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsif

in the case of this job, it is probably fine to skip upload if a previous step fails but I was rewriting the action anyway so I added it as a best practice.

calypsomatic (approved):

PR #973, opened 2022-08-04
quote TSV values if they contain a tab [AJ-523]

davidangb:
If an entity attribute value contains a tab, wrap that value in double-quotes when outputting it to a TSV. This prevents the embedded tab from breaking the format of the TSV.

I've added/updated unit tests as well as tested manually by running locally.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
@@ -4,9 +4,6 @@ import org.broadinstitute.dsde.rawls.model._
 import org.broadinstitute.dsde.firecloud.model._
 import org.broadinstitute.dsde.firecloud.service.TsvTypes
 
-import scala.util.{Failure, Success, Try}
-import spray.json.{JsString, JsValue}
-
 object TSVFormatter {
 
   /**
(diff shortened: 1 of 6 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
```diff
@@ -68,14 +77,17 @@ object TSVFormatter {
...
+    * @param value The JsValue to stringify
+    * @return Resultant string value
     */
   def cleanValue(value: JsValue): String = {
-    val regex = "^\"|\"$".r
-    regex.replaceAllIn(value.toString(), "")
+    value match {
+      case s:JsString => s.value
+      case x => x.toString()
+    }
```
davidangb, line 90: this change feels more accurate to the intention of the method, and it is necessary to allow the double-quoted values through to the TSV output.

davidangb:
jenkins retest

davidangb:
Jenkins retest

davidangb:
jenkins retest

src/main/scala/org/broadinstitute/dsde/firecloud/utils/TSVFormatter.scala
```diff
@@ -68,14 +65,27 @@ object TSVFormatter {
...
     *
-    * @param value The JsValue to remove leading and trailing quotes from.
-    * @return Trimmed string value
+    * @param value The input attribute to make safe
+    * @return the safe value
     */
-  def cleanValue(value: JsValue): String = {
-    val regex = "^\"|\"$".r
-    regex.replaceAllIn(value.toString(), "")
+  def tsvSafeAttribute(attribute: Attribute): String = tsvSafeString(AttributeStringifier(attribute))
```
calypsomatic, line 74: Do we no longer need to trim the quotes?

davidangb: We explicitly need to NOT trim the quotes. If we trimmed quotes, then it would prevent any values-containing-a-tab from being quoted in the final TSV.

Elsewhere in the PR, we switch from using `value.toString()` to `AttributeStringifier` to turn a `JsValue` into a string. That change prevents `JsString` objects from being double-quoted … even though that wasn't actually a problem.

sagehen03 (approved):

calypsomatic (approved):

PR #980, opened 2022-08-08
user-enabled checks [AJ-510]

davidangb:
- [ ] TODO: update the error handler to provide better error messages for the APIs that have implicit protection and don't use `requireEnabledUser`

-----

This PR adds the ability to wrap an API with a Sam call. The Sam call will verify if the current user is enabled or not; if the user is not enabled, the API will short-circuit and return a 401.

In this PR, I protect only a few APIs, in `UserApiService` and `ShareLogApiService`. Orch will need to implement this for multiple additional APIs; I'll do that in a follow-on PR once this is approved and merged and shown to be working well.

See the `TODO` comments in `EnabledUserDirectives` for areas where I'd love additional feedback.

response for an unregistered user:
```json
{
  "causes": [],
  "message": "User is not registered.",
  "source": "Orchestration-enabled-check",
  "stackTrace": [],
  "statusCode": 401
}
```

response for a registered but disabled user:
```json
{
  "causes": [],
  "message": "User is disabled.",
  "source": "Orchestration-enabled-check",
  "stackTrace": [],
  "statusCode": 401
}
```

compare also to https://github.com/broadinstitute/agora/pull/323 and https://github.com/broadinstitute/cromwell/pull/6826

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/utils/EnabledUserDirectives.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/utils/EnabledUserDirectives.scala
@@ -32,8 +32,8 @@ trait EnabledUserDirectives extends LazyLogging  with SprayJsonSupport {
           innerRoute
         case Success(false) =>
           logger.warn(s"User ${userInfo.userEmail} is disabled: $uri")
-          // TODO: do we want a more helpful message like "user is disabled"?
-          throwErrorReport(StatusCodes.Forbidden, StatusCodes.Forbidden.defaultMessage)
+          // the 401/"User is disabled." response mirrors what Sam returns in this case.
+          throwErrorReport(StatusCodes.Unauthorized, "User is disabled.")
         case Failure(ex) => ex match {
           case apiex:ApiException =>
             logger.error(s"Exception checking enabled status for user ${userInfo.userEmail}: (${apiex.getMessage}) while calling $uri", apiex)
@@ -54,7 +54,11 @@ trait EnabledUserDirectives extends LazyLogging  with SprayJsonSupport {
     val sam = new UsersApi(apiClient)
     val userStatus = sam.getUserStatusInfo
     // TODO: is it enough to check this one `getEnabled` boolean?
-    // TODO: retries?
+    // TODO: should we use retries here?
+    /* TODO: instead of directly calling Sam's userStatusInfo API, should we call a "real" api and, if that api returns
+        401, echo its response? That would allow Orch to always give the same error response that Sam gives, but
+        it is likely to be more expensive in the common case of the user being enabled.
+     */
     userStatus.getEnabled
   }
 
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/utils/EnabledUserDirectives.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/utils/EnabledUserDirectives.scala
@@ -2,12 +2,11 @@ package org.broadinstitute.dsde.firecloud.utils
 
 import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport
 import akka.http.scaladsl.model.{StatusCode, StatusCodes}
-import akka.http.scaladsl.server.Directives.{complete, extractUri}
+import akka.http.scaladsl.server.Directives.extractUri
 import akka.http.scaladsl.server.{RequestContext, Route, RouteResult}
 import com.typesafe.scalalogging.LazyLogging
-import org.broadinstitute.dsde.firecloud.{FireCloudConfig, FireCloudExceptionWithErrorReport}
 import org.broadinstitute.dsde.firecloud.model.UserInfo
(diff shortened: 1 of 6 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/utils/EnabledUserDirectives.scala
```diff
@@ -0,0 +1,79 @@
...
+    val sam = new UsersApi(apiClient)
+    val userStatus = sam.getUserStatusInfo
+    // TODO: is it enough to check this one `getEnabled` boolean?
+    // TODO: should we use retries here?
+    /* TODO: instead of directly calling Sam's userStatusInfo API, should we call a "real" api and, if that api returns
+        401, echo its response? That would allow Orch to always give the same error response that Sam gives, but
+        it is likely to be more expensive in the common case of the user being enabled.
+     */
+    userStatus.getEnabled
+  }
```
davidangb, line 115: I'm thinking of potentially moving this call inside `HttpSamDAO`, and then adding a `val samDAO: SamDAO` to this trait, which would require any class that extends it to provide a `SamDAO` implementation. That would help with injecting mocks during unit tests. Otherwise, each unit test that runs against a `requireEnabledUser`-protected API will need to spin up its own mockserver, as seen in `ShareLogApiServiceSpec`.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/EnabledUserDirectives.scala
```diff
@@ -0,0 +1,79 @@
...
+    * Queries Sam to see if the current user is enabled. If the user is disabled,
+    * responds with Forbidden and prevents the rest of the route from executing.
+    *
+    * @param userInfo credentials for the current user
+    * @param samBaseUrl where to find Sam - used for unit testing
+    * @return n/a
+    */
+  def requireEnabledUser(userInfo: UserInfo, samBaseUrl: String = FireCloudConfig.Sam.baseUrl)(innerRoute: RequestContext => Future[RouteResult]): Route = {
+    extractUri { uri =>
+      Try(getUserEnabled(userInfo.accessToken.token, samBaseUrl)) match {
```
rtitle: I think this should probably wrap the blocking Sam call in a Future instead of a Try. Or alternatively use the async client (see [example](https://github.com/broadinstitute/rawls/pull/1946/files#r930090541)).

src/main/scala/org/broadinstitute/dsde/firecloud/utils/EnabledUserDirectives.scala
```diff
@@ -0,0 +1,79 @@
...
+    }
+  }
+
+  private def getUserEnabled(token: String, samBaseUrl: String): Boolean = {
+    val apiClient = new ApiClient()
+    apiClient.setAccessToken(token)
+    apiClient.setBasePath(samBaseUrl)
+    val sam = new UsersApi(apiClient)
+    val userStatus = sam.getUserStatusInfo
+    // TODO: is it enough to check this one `getEnabled` boolean?
```
rtitle: My 2 cents:

1. I think checking `getEnabled` is sufficient
2. I would vote against proactively adding retries
3. I think the `getUserStatusInfo` API is correct, and probably cheapest. We would only use this directive if there is _not_ otherwise a Sam API being called, right?

davidangb: 3. correct - we only need `requireEnabledUser` for a small minority of Orch's APIs.

davidangb:
@rtitle I've updated to use the Sam client async method - is this what you were looking for?

src/main/scala/org/broadinstitute/dsde/firecloud/utils/EnabledUserDirectives.scala
```diff
@@ -0,0 +1,126 @@
...
+    val sam = new UsersApi(apiClient)
+    val callback = new SamApiCallback[UserStatusInfo]
+    // TODO: should we use retries here?
+    /* TODO: instead of directly calling Sam's userStatusInfo API, should we call a "real" api and, if that api returns
+        401, echo its response? That would allow Orch to always give the same error response that Sam gives, but
+        it is likely to be more expensive in the common case of the user being enabled.
+     */
+    sam.getUserStatusInfoAsync(callback)
+    // TODO: is it enough to check this one `getEnabled` boolean?
+    callback.future map (userStatus => Boolean.unbox(userStatus.getEnabled))
```
aednichols, line 114: FWIW that's what I independently decided I should do in CromIAM, not 1000% sure it's right

calypsomatic (approved):

rtitle (approved):
> @rtitle I've updated to use the Sam client async method - is this what you were looking for?

LGTM, thank you.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/EnabledUserDirectives.scala
dvoet: done :)

PR #992, opened 2022-08-19
[SU-186] Add new profile fields, remove unused ones

MatthewBemis:

In all cases:


src/main/scala/org/broadinstitute/dsde/firecloud/model/Profile.scala
```diff
@@ -29,25 +29,20 @@ case class BasicProfile (
...
     contactEmail: Option[String],
     institute: String,
-    institutionalProgram: String,
     programLocationCity: String,
     programLocationState: String,
     programLocationCountry: String,
-    pi: String,
-    nonProfitStatus: String,
-    termsOfService: Option[String]
+    termsOfService: Option[String],
```
MatthewBemis, line 35: This `termsOfService` stuff seems to be unused from an earlier implementation of ToS in Sam.

I plan to verify that and rip it out but not as part of this PR

src/main/scala/org/broadinstitute/dsde/firecloud/model/Profile.scala
```diff
@@ -29,25 +29,20 @@ case class BasicProfile (
...
+    researchArea: Option[String]
```
MatthewBemis, line 36: Making this new field optional for the sake of backwards compatibility. Don't want to break anyone's scripts.

src/main/scala/org/broadinstitute/dsde/firecloud/model/Profile.scala
```diff
@@ -29,25 +29,20 @@ case class BasicProfile (
     title: String,
     contactEmail: Option[String],
     institute: String,
-    institutionalProgram: String,
```
MatthewBemis, line 32: It may be nice to also remove these fields from FireCloud UI. Removing them here doesn't break anything in FireCloud UI, but they won't be honored anymore when filled out.

nawatts (approved):

davidangb (approved):

PR #996, opened 2022-08-23
streaming/flexible passthrough [AJ-350]

davidangb:
This is a rewrite of Orchestration's passthrough capabilities. This new passthrough directive:
* uses akka streaming instead of using a store-and-forward materialized approach
* passes everything under a specified path to its target, eliminating the need to define every single passthrough route

I have updated `BillingApiService`, `CromIamApiService`, and `EntityApiService` with example usages of the new passthrough capability - these should illustrate how it can reduce code.

A drawback of passing through using a wildcard is that the passthrough could expose APIs in the underlying service that were previously hidden.

TODOs:
- [x] does the change to `CromIamApiService` expose any CromIAM endpoints that we would like to leave unexposed?


davidangb:
fiab-start transient failure:
```
10:43:30 Error response from daemon: Get https://registry-1.docker.io/v2/broadinstitute/dsp-toolbox/manifests/latest: Get https://auth.docker.io/token?account=dsdejenkins&scope=repository%3Abroadinstitute%2Fdsp-toolbox%3Apull&service=registry.docker.io: EOF
10:43:30 Build step 'Execute shell' marked build as failure
```
Jenkins retest.

sagehen03 (approved):
Have you been able to test this end to end and confirm memory usage is down for orch with this change?

src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingPassthrough.scala
```diff
@@ -0,0 +1,183 @@
...
+    // This condition should only be caused by developer error in which the streamingPassthrough
+    // directive is incorrectly configured inside a route.
+    if (!requestUri.path.startsWith(localBasePath)) {
+      throw new Exception(s"request path doesn't start properly: ${requestUri.path} does not start with $localBasePath")
+    }
+
+    // recursively loops through the segments of the actual path, and removes
+    // all segments present in the base path. Returns the result:
+    // every part of the path, minus the base.
+    // This should be equivalent to `actual.toString.replace(base.toString, "")`
```
jordansafer: I'm strongly in favor of the string conversion approach for readability. I'm not sure the recursive approach would be faster, since it's adding a bunch of function calls. My one concern with string conversion is that the solution will fail if the base path is for any reason duplicated in the middle of the path

src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingPassthrough.scala
```diff
@@ -0,0 +1,183 @@
...
+    // Convert the URI to the one suitable for the remote system
+    val targetUri = convertToRemoteUri(req.uri, localBasePath, remoteBaseUri)
+
+    // Remove unwanted headers. Akka automatically adds a Timeout-Access to the request
+    // for internal use in managing timeouts; see akka.http.server.request-timeout.
+    // This header is not legal to pass to an external remote system.
+    val targetHeaders = req.headers.filter(_.isNot(`Timeout-Access`.lowercaseName))
+
+    // TODO: what should this log?
+    streamingPassthroughLogger.info(s"${req.method} ${req.uri} => $targetUri")
```
jordansafer: I like what we're logging, I'd log a brief explanation for readability in the logs, along these lines:
```suggestion
    streamingPassthroughLogger.info(s"Passthrough API called. Forwarding call: ${req.method} ${req.uri} => $targetUri")
```

jordansafer (approved):

src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingPassthrough.scala
davidangb: switched to string conversion, made sure to only replace the first instance, and added a unit test for this.

davidangb:
@sagehen03 
> Have you been able to test this end to end and confirm memory usage is down for orch with this change?

I have tested functionality end to end, but I have not tested memory usage. My goal with this PR is to improve time-to-first-byte, support any underlying (e.g. Rawls) APIs that stream, and reduce code complexity when defining routes - memory usage isn't a goal. Do you anticipate any problems here?

jordansafer (approved):
LGTM :)

sagehen03 (approved):

aednichols (commented):
To answer [this question](https://broadinstitute.slack.com/archives/C1EH66VCM/p1662654689315229), CromIAM is designed to be world-facing with all sensitive calls checked with Sam.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/BillingApiService.scala
```diff
@@ -1,86 +1,23 @@
...
-                  (get | put | delete) {
-                    extract(_.request.method) { method =>
-                      passthrough(s"$v2BillingUrl/$projectId/spendReportConfiguration", method)
-                    }
-                  }
-                }
-            }
-        }
+      // all paths under /api/billing pass through to the same path in Rawls
+      streamingPassthrough(Uri.Path("/api/billing") -> Uri(FireCloudConfig.Rawls.authUrl + "/billing"))
```
aherbst-broad, line 16: I'm not grokking this -- will the /billing/v2 endpoints deleted above still be passed through with this approach?

aherbst-broad (commented):
Had a question about the v2 passthroughs

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/BillingApiService.scala
davidangb: yes - consider this as a wildcard; it will pass `${orchestration}/api/billing*` to `${rawls}/api/billing*`.

davidangb: @aherbst-broad does this work for you?

aherbst-broad: Yes, looks good.

aherbst-broad (approved):

PR #1009, opened 2022-09-06
ID-150 When Sam fails, Orch Fails

tlangs:
Make failed requests to Sam not get swallowed

In all cases:


Ghost-in-a-Jar:
Does it make sense to add a small test for this or add a check to an existing test?

tlangs:
@Ghost-in-a-Jar Not sure if that's more trouble than it's worth. The tests don't actually talk to Sam proper, instead mocking out the Sam calls. Adding a test case to test our own mock would be kinda redundant. 

I suppose an integration test could be added, but through my manual testing, I'm pretty confident in this change.

Ghost-in-a-Jar (approved):

Shakespeared (approved):

PR #1010, opened 2022-09-06
[AJ-537] support optional syncPermissions param

jordansafer:
This PR adds support for optionally syncing permissions in TDR snapshot import-by-copy. The importJob request will now contain an optional boolean field specifying whether the TDR permissions sync is enabled.

Tested end to end with TDR, Terra-UI, Orch, import service, rawls. I tested passing true, false, and no param paths.

In all cases:


davidangb:
I'm imagining that we may add more flags to the import process parallel to `tdrSyncPermissions`. AJ-136 is an outstanding (and low priority) feature request that would add a flag; we may need other ones for handling sensitive data imports. What do you think about nesting `tdrSyncPermissions` under an `options` key-value map where we could add other flags later? Since `tdrSyncPermissions` is irrelevant for TSV and PFB imports, this would also make it less glaring for those imports.

davidangb (commented):
syntax looks good to me. I'm still wondering about nesting the `tdrSyncPermissions` flag under an `options` map - what do you think about that?

jordansafer:
> I'm imagining that we may add more flags to the import process parallel to `tdrSyncPermissions`. [AJ-136](https://broadworkbench.atlassian.net/browse/AJ-136) is an outstanding (and low priority) feature request that would add a flag; we may need other ones for handling sensitive data imports. What do you think about nesting `tdrSyncPermissions` under an `options` key-value map where we could add other flags later? Since `tdrSyncPermissions` is irrelevant for TSV and PFB imports, this would also make it less glaring for those imports.

I like the idea of using an options map! I'll add it. I did think the tdrSyncPermissions looked messy.

davidangb (approved):
:+1: assuming tests pass

sagehen03 (approved):

PR #1023, opened 2022-09-20
DDO-2343 Make proxy passthrough logic set Host header to destination

choover-broad:
Configure passthrough proxy logic to update Host header to point at the destination, instead of preserving the source request's Host header.

(This was breaking Orch proxying in BEEs, which use the Nginx Ingress controller, which uses the Host header to route requests)

----

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingPassthrough.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingPassthrough.scala
@@ -102,7 +102,7 @@ trait StreamingPassthrough
       filter(_.isNot(`Timeout-Access`.lowercaseName)).
       filter(_.isNot(Host.lowercaseName))
 
-    val targetHeaders = filteredHeaders :+ Host.apply(targetUri.authority)
+    val targetHeaders = filteredHeaders :+ Host(targetUri.authority.host)
 
     // TODO: what should this log?
     streamingPassthroughLogger.info(s"Passthrough API called. Forwarding call: ${req.method} ${req.uri} => $targetUri")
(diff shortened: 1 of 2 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingPassthrough.scala
```diff
@@ -90,10 +91,18 @@ trait StreamingPassthrough
...
+    // Timeout-Access: Akka automatically adds a Timeout-Access to the request
     // for internal use in managing timeouts; see akka.http.server.request-timeout.
     // This header is not legal to pass to an external remote system.
-    val targetHeaders = req.headers.filter(_.isNot(`Timeout-Access`.lowercaseName))
+    //
+    // Host: The Nginx ingress controller used in Terra's BEE environments uses the Host header for routing,
+    // so we remove and set it with targetUri host
+    val filteredHeaders = req.headers.
+      filter(_.isNot(`Timeout-Access`.lowercaseName)).
+      filter(_.isNot(Host.lowercaseName))
```
davidangb: I suggest consolidating this into one `filter` operation which checks both conditions:
```scala
    val filteredHeaders = req.headers.filter { hdr =>
      hdr.isNot(`Timeout-Access`.lowercaseName) &&
        hdr.isNot(Host.lowercaseName)
    }
```

davidangb (approved):
This looks good to me and is pretty much what I would have done! One optimization suggestion inline, but it's pretty minor.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingPassthrough.scala
choover-broad: integrated this, thanks!

jack-r-warren (approved):

PR #1025, opened 2022-09-23
fix space-encoding in streamingPassthrough; support submissions [AJ-600]

davidangb:
Two changes in this PR:

1. the `streamingPassthrough` directive threw an exception when the remote URL base path contained a space, i.e. was not a valid URI. This is programmer error, not end user error - the programmer should never allow an invalid URI as a remote URL. This PR fixes it by encoding workspace namespaces and names (the user inputs that could contain spaces) in the remote URLs.

2. `SubmissionApiService` now uses `streamingPassthrough` and is much simplified. This also allows `SubmissionApiServiceNegativeSpec` to be deleted.


src/main/scala/org/broadinstitute/dsde/firecloud/webservice/EntityApiService.scala
```diff
@@ -67,11 +68,11 @@ trait EntityApiService extends FireCloudDirectives
                 }
               } ~
               pathPrefix(Segment) { entityType =>
-                streamingPassthrough(baseRawlsEntitiesUrl + "/" + entityType)
+                streamingPassthrough(FireCloudConfig.Rawls.entityPathFromWorkspace(escapePathSegment(workspaceNamespace), escapePathSegment(workspaceName)) + "/" + entityType)
```
calypsomatic, line 71: I see that `baseRawlsEntitiesUrl` is used throughout this file - why does it only need to be replaced here?

davidangb: The other usages are working correctly (and have been working correctly for a while), because the old `passthrough` directive already has some encoding capabilities. The new `streamingPassthrough` requires the encoding in this PR.

calypsomatic (approved):

sagehen03 (approved):

PR #1038, opened 2022-10-07
update scoverage plugin [AJ-629]

davidangb:
this is a prerequisite to #1028


davidangb:
whoops, this is a dupe of #1037. Closing.

PR #1063, opened 2022-11-18
[WOR-552] update rawls-model dependency

ghost:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-552
Orch parses the status response from rawls so had to update the rawls model dependency and fix tests  
- updated `rawls-model` dependency to use the new version (https://github.com/broadinstitute/rawls/pull/2121) which adds BPM and WSM to the rawls status check 
-  used workbench-libs version of`Notifications` instead of `rawls-model `
- updated akka version from `2.5.32` to `2.6.19` (same version as rawls) to fix `java.lang.NoSuchMethodError` when running unit tests

firecloud-orchestration Swagger : 
Screen Shot 2022-11-23 at 3 57 20 PM

In all cases:

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -51,6 +51,7 @@ object Dependencies {
     "org.broadinstitute.dsde.workbench" %% "workbench-google2" % s"0.24-$workbenchLibsHash",
     "org.broadinstitute.dsde.workbench" %% "workbench-oauth2" % s"0.2-$workbenchLibsHash",
     "org.broadinstitute.dsde.workbench" %% "sam-client"       % "0.1-ef83073",
+    "org.broadinstitute.dsde.workbench" %% "workbench-notifications" %s"0.3-${workbenchLibsHash}",
 
     "com.typesafe.akka"   %%  "akka-actor"           % akkaV,
     "com.typesafe.akka"   %%  "akka-slf4j"           % akkaV,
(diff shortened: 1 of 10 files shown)


project/Dependencies.scala
```diff
@@ -41,7 +41,7 @@ object Dependencies {
     "com.typesafe.scala-logging"    %% "scala-logging"       % "3.9.5",
 
     "org.parboiled" % "parboiled-core" % "1.4.1",
-    excludeGuava("org.broadinstitute.dsde"       %% "rawls-model"         % "0.1-e0584dbdc")
+    excludeGuava("org.broadinstitute.dsde"       %% "rawls-model"         % "0.1-34e8799b-SNAP")
```
cahrens: So does the Rawls PR have to merge first, and then the version number gets updated here?

ghost: As far as I know, I have to merge this PR first then merge the Rawls PR and update with an official version here but I could be wrong

ehigham (approved):

project/Dependencies.scala
```diff
@@ -41,7 +41,7 @@ object Dependencies {
     "com.typesafe.scala-logging"    %% "scala-logging"       % "3.9.5",
 
     "org.parboiled" % "parboiled-core" % "1.4.1",
-    excludeGuava("org.broadinstitute.dsde"       %% "rawls-model"         % "0.1-e0584dbdc")
+    excludeGuava("org.broadinstitute.dsde"       %% "rawls-model"         % "0.1-902d77f8-SNAP")
```
cahrens, line 44: I think I already asked this, but just to be clear will ask again.

So the plan is that you merge the Rawls PR first and then update the version string here so it isn't a SNAP version?

cahrens: Ah, I saw your previous response.

cahrens (approved):

PR #1071, opened 2022-12-06
[RW-9102] Add /api/convert/rmd api endpoint in fireclouid-orch

yonghaoy:
Add convert rmd API into firecloud-orch: 
https://github.com/DataBiosphere/calhoun/blob/dev/static/api-docs.yaml#L76

AoU is using firecloud-orch for most Terra calls including convert Jupyter Notebook: 
https://github.com/all-of-us/workbench/blob/main/api/src/main/resources/firecloud.yaml#L933

As we are supporting RStudio via Leo GKE, we need to expose convertRmd in firecloud-orch

Not sure if that is the best api interface. Or shall I do `staticNotebook/conver/rmd`? 

In all cases:

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -3055,8 +3055,11 @@ paths:
           description: Success
           content: { }
         500:
-          description: Internal Error
-          content: { }
+          description: FireCloud Internal Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
       x-passthrough: true
       x-passthrough-target: calhoun
       x-codegen-request-body-name: rmd
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -3012,6 +3012,55 @@ paths:
       x-passthrough: true
       x-passthrough-target: calhoun
       x-codegen-request-body-name: notebook
+
+  /api/staticRMarkdown/convert:
+    post:
+      tags:
+        - Static R Markdown files
+      summary: |
+        Converts an R Markdown file to HTML. Accepts the content of an .Rmd file in the request body.
+      description: |
+        Converts an R Markdown file to HTML. Accepts the content of an .Rmd file in the request body.
+      operationId: convertRmd
+      requestBody:
+        description: R Markdown file
+        content:
+          text/plain:
+            schema:
+              type: string
+              format: binary
+              example: |
+                ---
+                title: "test-rmd"
+                output: html_document
+                ---
+                ```{r setup, include=FALSE}
+                knitr::opts_chunk$set(echo = TRUE)
+                ```
+                ## R Markdown
+                This is an R Markdown document. Markdown is a simple formatting syntax for authoring HTML, PDF, and MS Word documents. For more details on using R Markdown see <http://rmarkdown.rstudio.com>.
+                When you click the **Knit** button a document will be generated that includes both content as well as the output of any embedded R code chunks within the document. You can embed an R code chunk like this:
+                ```{r cars}
+                summary(cars)
+                ```
+                ## Including Plots
+                You can also embed plots, for example:
+                ```{r pressure, echo=FALSE}
+                plot(pressure)
+                ```
+                Note that the `echo = FALSE` parameter was added to the code chunk to prevent printing of the R code that generated the plot.
+        required: true
+      responses:
+        200:
+          description: Success
+          content: { }
+        500:
+          description: Internal Error
+          content: { }
+      x-passthrough: true
+      x-passthrough-target: calhoun
+      x-codegen-request-body-name: rmd
+
   /api/storage/{bucket}/{object}:
     get:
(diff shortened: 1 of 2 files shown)


Qi77Qi:
@yonghaoy can u add a bit more context why this needs to be added to orch

yonghaoy:
> @yonghaoy can u add a bit more context why this needs to be added to orch

updated. PTAL! Thanks

src/main/resources/swagger/api-docs.yaml
```diff
@@ -3012,6 +3012,55 @@ paths:
       x-passthrough: true
       x-passthrough-target: calhoun
       x-codegen-request-body-name: notebook
+
+  /api/staticRmd/convert:
+    post:
+      tags:
+        - Static R Markdown files
```
davidangb: nit: this creates an entirely new section in the swagger UI. Is this your intent, as opposed to putting this endpoint into one of the existing sections? If you do want to create an entirely new section, please add this tag to the `tags:` declaration at line 13; that will ensure proper ordering of sections instead of putting "Static R Markdown files" last.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/StaticNotebooksApiService.scala
```diff
@@ -30,4 +31,21 @@ trait StaticNotebooksApiService extends FireCloudDirectives with StandardUserInf
...
   }
+
+  val staticRmdRoutes: Route = {
+    path("staticRmd" / "convert") {
+      requireUserInfo() { userInfo =>
+        post {
+          requestContext =>
+            // call Calhoun and pass its response back to our own caller
+            // can't use passthrough() here because that demands a JSON response
+            // and we expect this to return text/html
```
davidangb: would `streamingPassthrough` work for you, e.g.
https://github.com/broadinstitute/firecloud-orchestration/blob/4bfd24ee001e783209147cc23be87cf15931ba0b/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/SubmissionApiService.scala#L11

src/main/resources/swagger/api-docs.yaml
```diff
@@ -3012,6 +3012,55 @@ paths:
...
+                ```
+                Note that the `echo = FALSE` parameter was added to the code chunk to prevent printing of the R code that generated the plot.
+        required: true
+      responses:
+        200:
+          description: Success
+          content: { }
+        500:
+          description: Internal Error
+          content: { }
```
davidangb: do the 200 and 500 really return `{}` as their response? Would the 500 actually return a `'#/components/schemas/ErrorReport'`? and can you give any more detail on the 200?

src/main/resources/swagger/api-docs.yaml
yonghaoy: ah, thanks for letting me know how to create new section. Question about new section vs reuse staticNotebook: 
I don't have preference/opinion about new section or not. I was trying to follow calhoun API interface, and come up with this section. https://github.com/DataBiosphere/calhoun/blob/dev/static/api-docs.yaml#L76

What I feel: 

1. New section is cleaner and easy to read (I couldn't figure out a nicer way to put RMD convert into existing staticNotebook section without a large refactor/reorg on the existing endpoints)
2. Keep creating new section may not be the right choice if we end up with extending calhoun to support other file artifacts. 

I am new to calhoun and this FOS endpoints, any feedback would be helpful!

src/main/resources/swagger/api-docs.yaml
yonghaoy: Actually I don't know... I was following staticNotebook pattern and don't know the historical reason of designing that endpoint like that.. (same reason for `passthrough`  comment above). 
maybe @rtitle remembers?

src/main/resources/swagger/api-docs.yaml
davidangb: I can't really comment on whether or not this deserves a new section or not - that question is really about the content and behavior and intent of this endpoint, and that's a question for the implementing team. I'm just raising the question and whatever the calhoun team decides is ok with me.

src/main/resources/swagger/api-docs.yaml
davidangb: here also, I don't personally have any history. I'd suggest though that if the `staticNotebook` definition is lacking detail, fixing the detail for this API is better than copying over the lack of detail

davidangb:
@yonghaoy is this PR still valid? If not, could you close it out?

yonghaoy:
> @yonghaoy is this PR still valid? If not, could you close it out?

oh sorry @davidangb This PR is still valid. Just come back from vacation and catching up things. Totally forget this PR. I will polish it a bit and send it out again. 
Thanks for reminding!

yonghaoy:
Per offline chart with @jdcanas I will hold off this PR and ask AoU Security see if we can just call Calhoun directly.

davidangb:
removing myself as a reviewer so I can keep my review requests clean; feel free to re-add me if appropriate.

PR #1072, opened 2022-12-09
[WOR-646] Update swagger page

ghost:
Ticket: https://broadworkbench.atlassian.net/browse/WOR-646
Related to [Add cloudPlatform to workspace details endpoint ](https://github.com/broadinstitute/rawls/pull/2151)

In all cases:

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -9192,6 +9192,7 @@ components:
         - workspaceId
         - googleProject
         - workspaceVersion
+        - cloudPlatform
       type: object
       properties:
         attributes:
@@ -9258,6 +9259,9 @@ components:
         cloudPlatform:
           type: string
           description: The cloud platform of the workspace
+          enum:
+            - GCP
+            - AZURE
       description: ""
     WorkspaceListResponse:
       type: object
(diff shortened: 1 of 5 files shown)


cahrens (approved):

src/main/resources/swagger/api-docs.yaml
```diff
@@ -9255,6 +9255,9 @@ components:
           description: The date all files were successfully transferred to the workspace bucket if applicable
             in yyyy-MM-ddTHH:mm:ss.SSSZZ format
           format: date-time
+        cloudPlatform:
+          type: string
+          description: The cloud platform of the workspace
```
cahrens, line 9260: Actually, wondering if we should add the possible values, like we do for the billing project response.

cloudPlatform:
          type: string
          enum:
            - GCP
            - AZURE

cahrens: And should this be listed under "required"?

cahrens (changes requested):
Had a few comments/suggestions. These will likely require some cleanup in Rawls too.

src/main/resources/swagger/api-docs.yaml
marctalbott: It should not be required as the cloud platform is optional on the underlying `WorkspaceDetails` case class in Rawls and may not be present in the response

marctalbott (approved):

cahrens (approved):

PR #1080, opened 2023-01-05
[RW-9349] Add Sam ToS Details passthrough

jmthibault79:
Add a passthrough endpoint for Sam `register/user/v2/self/termsOfServiceDetails` (added in https://github.com/broadinstitute/sam/pull/935)

The All of Us Researcher Workbench development team prefers to call Firecloud-Orchestration endpoints instead of calling Terra services directly.

Have you read [CONTRIBUTING.md](../CONTRIBUTING.md) lately? If not, do that first.

I, the developer opening this PR, do solemnly pinky swear that:

- [ ] I've followed [the instructions](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/CONTRIBUTING.md#api-changes) if I've made any changes to the API, _especially_ if they're breaking changes
- [ ] I've updated the RC_XXX release ticket with any manual steps required to release this change
- [ ] I've updated the [FISMA documentation](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/CONTRIBUTING.md#fisma-documentation-changes) if I've made any security-related changes, including auth, encryption, or auditing

In all cases:

- [ ] Get two thumbsworth of review and PO signoff if necessary
- [ ] Verify all tests go green
- [ ] Squash and merge; you can delete your branch after this **unless it's for a hotfix**. In that case, don't delete it!
- [ ] Test this change deployed correctly and works on dev environment after deployment

PR #1086, opened 2023-01-10
append routes [RW-9349]

davidangb:
The routes under `"register" / "user"` were not appended to each other; the last one `"v2" / "self" / "termsOfServiceDetails"` was the only one that worked. This PR fixes that.

I manually tested each of the `"register" / "user"` routes running locally.

Followup to #1080


jmthibault79 (approved):
ohhhh.  yes that makes sense.  I was not *adding* the v2 route, but instead *overwriting what was already there*.  thank you.

PR #1087, opened 2023-01-12
[SUP-757]  Fix for Rawls workflow passthough

JVThomas:
Partially addresses [SUP-757](https://broadworkbench.atlassian.net/browse/SUP-757)

The initial ticket describes a UI bug where the "Compute Details" button in Job Manager's UI did nothing. Looking into the bug, I noticed that the button did make the expected outbound request only to receive a 404 response.

Further exploration lead me here to where I think the JM requests eventually land, where all of the workflow endpoint routes are redirected with a CromIAM based URI assigned and the request path appended to the constructed remote request. 

The sole Rawls based endpoint is different not just because it requires a Rawls based url, but the path itself needs to be reconstructed to match the expected pattern on Rawls.

I'm new to Scala and Akka, so I'd appreciate any feedback given (especially for the path construction part).

In all cases:

[SUP-757]: https://broadworkbench.atlassian.net/browse/SUP-757?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingPassthrough.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingPassthrough.scala
@@ -125,7 +125,7 @@ trait StreamingPassthrough
     * @param modifiedRemotePath a modified path value to be used over requestUri.path.toString if provided
     * @return the URI suitable for sending to the remote system
     */
-  def convertToRemoteUri(requestUri: Uri, localBasePath: Uri.Path, remoteBaseUri: Uri, modifiedRemotePath: Option[String]): Uri = {
+  def convertToRemoteUri(requestUri: Uri, localBasePath: Uri.Path, remoteBaseUri: Uri, modifiedRemotePath: Option[String] = None): Uri = {
     // Ensure the incoming request starts with the localBasePath. Abort if it doesn't.
     // This condition should only be caused by developer error in which the streamingPassthrough
     // directive is incorrectly configured inside a route.
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingPassthrough.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingPassthrough.scala
@@ -89,7 +89,7 @@ trait StreamingPassthrough
     * @param req the request inbound to Orchestration
     * @return the outbound request to be sent to another service
     */
-  def transformToPassthroughRequest(localBasePath: Uri.Path, remoteBaseUri: Uri, remotePath: Option[String])(req: HttpRequest): HttpRequest = {
+  def transformToPassthroughRequest(localBasePath: Uri.Path, remoteBaseUri: Uri, remotePath: Option[String] = None)(req: HttpRequest): HttpRequest = {
     // Convert the URI to the one suitable for the remote system
     val targetUri = convertToRemoteUri(req.uri, localBasePath, remoteBaseUri, remotePath)
     // Remove unwanted headers:
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingPassthrough.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingPassthrough.scala
@@ -10,7 +10,7 @@ import akka.http.scaladsl.server.directives.{BasicDirectives, RouteDirectives}
 import akka.stream.scaladsl.{Sink, Source}
 import com.google.common.net.UrlEscapers
 import com.typesafe.scalalogging.Logger
-import org.broadinstitute.dsde.firecloud.FireCloudExceptionWithErrorReport
+import org.broadinstitute.dsde.firecloud.{FireCloudConfig, FireCloudExceptionWithErrorReport}
 import org.broadinstitute.dsde.rawls.model.{ErrorReport, ErrorReportSource}
 import org.slf4j.LoggerFactory
 
@@ -92,7 +92,6 @@ trait StreamingPassthrough
   def transformToPassthroughRequest(localBasePath: Uri.Path, remoteBaseUri: Uri)(req: HttpRequest): HttpRequest = {
     // Convert the URI to the one suitable for the remote system
     val targetUri = convertToRemoteUri(req.uri, localBasePath, remoteBaseUri)
-
     // Remove unwanted headers:
     // Timeout-Access: Akka automatically adds a Timeout-Access to the request
     // for internal use in managing timeouts; see akka.http.server.request-timeout.
@@ -112,6 +111,25 @@ trait StreamingPassthrough
 
     // return a copy of the inbound request, using the new URI and new headers.
     req.withUri(targetUri).withHeaders(targetHeaders)
+  }
+
+  // TODO: get feedback on this method. Not happy having rawls specific request handling here (rather have it on CromiamApiService)
+  /**
+   * Private method that modifies the requestPath for the one workflow passthrough to Rawls
+   * due to the fact that the Rawls endpoint has a different URI construction, making suffix
+   * passthroughs inept
+   * @param requestUri
+   * @param remoteBaseUri
+   * @return the requestPath (String) suitable to be sent to the remote system
+   */
+  private def requestStringConstruction(requestUri: Uri, remoteBaseUri: Uri) : String = {
+    val requestPath = requestUri.path.toString
+    if (
+        FireCloudConfig.Rawls.authUrl.contains(remoteBaseUri.authority.toString()) &&
+        requestPath.contains("/backend/metadata/")
+    ) {
+      requestPath.replace("backend/metadata", "genomics")
+    } else requestPath
   }
 
   /**
@@ -135,9 +153,8 @@ trait StreamingPassthrough
 
     // find every part of the actual request path, minus the base.
     val baseString = localBasePath.toString
-    val requestString = requestUri.path.toString
+    val requestString = requestStringConstruction(requestUri, remoteBaseUri)
     val remainder = Uri.Path(requestString.replaceFirst(baseString, ""))
-
     // append the remainder to the remoteBaseUri's path
     val remotePath = remoteBaseUri.path ++ remainder
 
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingPassthrough.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingPassthrough.scala
@@ -60,13 +60,28 @@ trait StreamingPassthrough
   }
 
   /**
+   * Passes through, to a remote server, all requests that match or start with the
+   * supplied local path with a supplied remotePath for remote Uri construction
+   *
+   * @param passthroughMapping in the form `localBasePath -> remoteBaseUri`, where
+   *                           the localBasePath is the Orchestration-local path
+   *                           which must be matched for passthroughs, and remoteBaseUri
+   *                           is the fully-qualified URL to a remote system
+   *                           to use as target for passthrough requests.
+   *  @param pathOverride as the pre-defined path to use when constructing the remote path value
+   */
+  def streamingPassthroughWithPathRedirect(passthroughMapping: (Uri.Path, Uri), pathOverride: String): Route = {
+    passthroughImpl(passthroughMapping._1, passthroughMapping._2, Option(pathOverride))
+  }
+
+  /**
     * The passthrough implementation:
     *   - `mapRequest` to transform the incoming request to what we want to send to the remote system
     *   - `extractRequest` so we have the transformed request as an object
    *    - `remotePathOverride` to provide a pre-configured path if remote path structure is different from local
     *   - call the remote system and reply to the user via `routeResponse` streaming
    */
-  def passthroughImpl(localBasePath: Uri.Path, remoteBaseUri: Uri, remotePathOverride: Option[String] = None): Route = {
+  private def passthroughImpl(localBasePath: Uri.Path, remoteBaseUri: Uri, remotePathOverride: Option[String] = None): Route = {
     mapRequest(transformToPassthroughRequest(localBasePath, remoteBaseUri, remotePathOverride)) {
       extractRequest { req =>
         complete {
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingPassthrough.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingPassthrough.scala
@@ -56,7 +56,7 @@ trait StreamingPassthrough
     *                           to use as target for passthrough requests.
     */
   def streamingPassthrough(passthroughMapping: (Uri.Path, Uri)): Route = {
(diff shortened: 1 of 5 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingPassthrough.scala
```diff
@@ -114,6 +113,25 @@ trait StreamingPassthrough
...
+   */
+  private def requestStringConstruction(requestUri: Uri, remoteBaseUri: Uri) : String = {
+    val requestPath = requestUri.path.toString
+    if (
+        FireCloudConfig.Rawls.authUrl.contains(remoteBaseUri.authority.toString()) &&
+        requestPath.contains("/backend/metadata/")
+    ) {
+      requestPath.replace("backend/metadata", "genomics")
+    } else requestPath
+  }
```
jgainerdewar: It looks like this replaces the passthrough to CromIAM's metadata endpoint with a passthrough to the Rawls genomics endpoint. I think we need to retain the endpoint that goes to CromIAM's metadata endpoint, I expect that's used by Job Manager to get all the other information to display, and maybe some other things as well. 

I think we want to support both the CromIAM metadata passthrough, and a Rawls genomics passthrough.

aednichols: It feels like we're re-implementing functionality built into Akka HTTP here.

Its route DSL is designed to do things like match a pattern in the path and handle it a specific way.

aednichols (commented):
It would be confidence-inspiring to identify the PR introducing the bug in the description; it used to work, and knowing why it broke is related to knowing we have the right fix.

aednichols:
Also, this routes stuff is tricky, so please do not feel bad if you find it confusing.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingPassthrough.scala
JVThomas: I don't know what you mean by CromIAM endpoint. Do you mean `api/workflows/{version}/{id}/metadata`? If so, it (passthrough) works locally for me
Screen Shot 2023-01-13 at 2 22 38 PM

If not, then I'm not sure. CromIAM does not have an endpoint that follows the structure `{version}/{id}/backend/metadata/{operationId}`
Screen Shot 2023-01-13 at 2 38 12 PM



To Adam's point, this is why I don't like my implementation, but I'm not sure how to tackle the path modification. The `streamingPassthrough (unaltered)` method will always access the path value directly from the request object and appends it to the remoteUrl, which does not work for Rawls since its endpoint is structured differently. 

At the same time, I don't want to change `streamingPassthrough's` core functionality since it is valid for every other route. I would like to access the request object at a higher level (service ideally, within the Route directive), I just don't know how.

JVThomas:
> It would be confidence-inspiring to identify the PR introducing the bug in the description; it used to work, and knowing why it broke is related to knowing we have the right fix.

I thought it was this one, with the changes to `CromIamApiService.scala`. I can see that the previous implementation used the `rawlsWorkflowRoot` for the rawls endpoint.
https://github.com/broadinstitute/firecloud-orchestration/pull/996

aednichols:
Yeah, David quite reasonably did not notice the one `$rawlsWorkflowRoot` among the other `$workflowRoot` entries. Pretty sure I originally wrote that code and set this trap 🥇

src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingPassthrough.scala
```diff
@@ -89,10 +102,9 @@ trait StreamingPassthrough
     * @param req the request inbound to Orchestration
     * @return the outbound request to be sent to another service
     */
-  def transformToPassthroughRequest(localBasePath: Uri.Path, remoteBaseUri: Uri)(req: HttpRequest): HttpRequest = {
+  def transformToPassthroughRequest(localBasePath: Uri.Path, remoteBaseUri: Uri, remotePath: Option[String])(req: HttpRequest): HttpRequest = {
```
jgainerdewar: If you set a default value for `remotePath` here, you should be able to avoid adding `None` to all the call sites. 
```
def transformToPassthroughRequest(localBasePath: Uri.Path, remoteBaseUri: Uri, remotePath: Option[String] = None)
```

src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingPassthrough.scala
```diff
@@ -135,9 +148,8 @@ trait StreamingPassthrough
 
     // find every part of the actual request path, minus the base.
     val baseString = localBasePath.toString
-    val requestString = requestUri.path.toString
+    val requestString = if(modifiedRemotePath.isEmpty) requestUri.path.toString() else modifiedRemotePath.get
```
jgainerdewar: We generally want to avoid direct `get` calls, even in cases like this where logically it's impossible for them to throw an error. The Scala way to do this is:
```
val requestString = modifiedRemotePath.getOrElse(requestUri.path.toString)
```

src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingPassthrough.scala
```diff
@@ -10,7 +10,7 @@ import akka.http.scaladsl.server.directives.{BasicDirectives, RouteDirectives}
 import akka.stream.scaladsl.{Sink, Source}
 import com.google.common.net.UrlEscapers
 import com.typesafe.scalalogging.Logger
-import org.broadinstitute.dsde.firecloud.FireCloudExceptionWithErrorReport
+import org.broadinstitute.dsde.firecloud.{FireCloudExceptionWithErrorReport}
```
jgainerdewar: This looks leftover from a different set of imports, these braces should be removed.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingPassthrough.scala
```diff
@@ -75,6 +75,19 @@ trait StreamingPassthrough
     }
   }
 
+  /**
+   * The Rawls passthrough implementation:
+   *   - Similar to the passthroughImpl except it supplies a remotePath to be used in downstream uri construction
+   */
+  def rawlsPassthrough(localBasePath: Uri.Path, remoteBaseUri: Uri, rawlsPath: String): Route = {
```
jgainerdewar: Rather than having a Rawls-specific passthrough function, I wonder if it would make sense to incorporate this optional remotePath into the existing passthrough. It could be useful in other places.

If we don't want to do that, I suggest renaming this to something like `passthroughWithRemotePathOverride` since it seems generically useful.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/CromIamApiService.scala
```diff
@@ -14,15 +13,20 @@ trait CromIamApiService extends FireCloudRequestBuilding
...
   // Adam Nichols, 2019-02-04
 
-  val cromIamApiServiceRoutes: Route =
-    pathPrefix( "workflows" / Segment ) { _ =>
-      streamingPassthrough(Uri.Path(s"/api/workflows/v1") -> Uri(workflowRoot))
+  val cromIamApiServiceRoutes: Route = {
+    val localBase = s"/api/workflows/v1"
+    pathPrefix("workflows" / Segment / Segment / "backend" / "metadata" / Segments) {(version, workflowId, operationSegments) =>
+      val suffix = operationSegments.mkString("/")
+      rawlsPassthrough(Uri.Path(localBase), Uri(rawlsWorkflowRoot), s"/${workflowId}/genomics/${suffix}")
```
aednichols:
```suggestion
      passthrough(s"$rawlsWorkflowRoot/$workflowId/genomics/${operationSegments.mkString("/")}", HttpMethods.GET)
```

aednichols (commented):
I'm not really sure what the new `rawlsPassthrough()` method gets us compared to reviving the `passthrough()` call used ~in~ before https://github.com/broadinstitute/firecloud-orchestration/pull/996.

If we want to do a slightly larger refactor, which I'm onboard with, we should move the Rawls passthrough entirely out of the file titled "CromIAM".

JVThomas:
> I'm not really sure what the new `rawlsPassthrough()` method gets us compared to reviving the `passthrough()` call used ~in~ before #996.
> 
> If we want to do a slightly larger refactor, which I'm onboard with, we should move the Rawls passthrough entirely out of the file titled "CromIAM".

The downstream methods perform some necessary updates on the `HttpRequest` instance. For example,`transformToPassthroughRequest` removes Akka applied headers, assigns `targetUri` as the `Host` header for BEE environments, and creates a new req instance for the passthrough by copying the incoming request instance and applying the generated Uri updates.

That said, you're right about not needing a separate function. Janet had also suggested updating the existing `passthroughImpl` function which makes sense. It makes the function a bit more flexible without compromising its existing functionality.

src/test/scala/org/broadinstitute/dsde/firecloud/utils/StreamingPassthroughSpec.scala
```diff
@@ -57,38 +57,38 @@ class StreamingPassthroughSpec extends AnyFreeSpec
...
-  "convertToTargetUri" - {
+  "convertToRemoteUri" - {
     "should calculate a remainder" in {
       val requestUri = Uri("http://localhost:8123/foo/bar/baz/qux")
       val localBasePath = Path("/foo/bar")
       val remoteBaseUri = Uri("https://example.com/api/version/foo")
 
       val expected = Uri("https://example.com/api/version/foo/baz/qux")
-      convertToRemoteUri(requestUri, localBasePath, remoteBaseUri) shouldBe expected
+      convertToRemoteUri(requestUri, localBasePath, remoteBaseUri, None) shouldBe expected
```
jgainerdewar, line 67: I think the `None` inputs can be removed in this file? Unless you're deliberately making clear what the test input is.

jgainerdewar (approved):
Left one optional comment, but this looks good to me!

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/CromIamApiService.scala
```diff
@@ -14,15 +13,20 @@ trait CromIamApiService extends FireCloudRequestBuilding
...
-    pathPrefix( "workflows" / Segment ) { _ =>
-      streamingPassthrough(Uri.Path(s"/api/workflows/v1") -> Uri(workflowRoot))
+  val cromIamApiServiceRoutes: Route = {
+    val localBase = s"/api/workflows/v1"
+    pathPrefix("workflows" / Segment / Segment / "backend" / "metadata" / Segments) {(version, workflowId, operationSegments) =>
+      val suffix = operationSegments.mkString("/")
+      passthroughImpl(Uri.Path(localBase), Uri(rawlsWorkflowRoot), Option(s"/${workflowId}/genomics/${suffix}"))
+    } ~
+    pathPrefix( "workflows" / Segments ) { segmentsArr =>
+      streamingPassthrough(Uri.Path(localBase) -> Uri(workflowRoot))
```
aednichols: I'm sympathetic to not wanting to refactor the single Rawls route into its own file, but I think we should do _something_ to stop perpetuating the confusion caused by intermixed Rawls and CromIAM routes.

Maybe something like this pseudocode.
```
val cromIamServiceRoutes = { ... }
val rawlsServiceRoutes = { ... } // This Rawls route is an oddball included for historical reasons
val routes = cromIamServiceRoutes ~ rawlsServiceRoutes

File: src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingPassthrough.scala [hunk 9]
```diff
@@ -63,10 +63,11 @@ trait StreamingPassthrough
     * The passthrough implementation:
     *   - `mapRequest` to transform the incoming request to what we want to send to the remote system
     *   - `extractRequest` so we have the transformed request as an object
+   *    - `remotePathOverride` to provide a pre-configured path if remote path structure is different from local
     *   - call the remote system and reply to the user via `routeResponse` streaming
    */
-  private def passthroughImpl(localBasePath: Uri.Path, remoteBaseUri: Uri): Route = {
-    mapRequest(transformToPassthroughRequest(localBasePath, remoteBaseUri)) {
+  def passthroughImpl(localBasePath: Uri.Path, remoteBaseUri: Uri, remotePathOverride: Option[String] = None): Route = {
```
aednichols: We could make a new function like `streamingRedirect(mapping: (Uri.Path, Uri))` that is clear about the fact that it sends the request to a different path, rather than just passing through.

Then, implement its body with with a call to the private `passthroughImpl` that supports `remotePathOverride`.

Finally, replace
```
passthroughImpl(Uri.Path(localBase), Uri(rawlsWorkflowRoot), Option(s"/${workflowId}/genomics/${suffix}"))
```
with
```
streamingRedirect(Uri.Path(localBase) -> Uri(s"$rawlsWorkflowRoot/${workflowId}/genomics/${suffix}"))
```

(The thought that sent me down this path was that functions with `impl` in the name are usually intended to be the hidden implementation behind a more ergonomic public facade.)

aednichols (commented):
This is definitely on the right track but I have some optional suggestions to think about. I promise I'm not usually this pedantic about semantics, but in this case we're cleaning up a high profile bug that was caused by particularly unclear semantics. 

---

Note to self:

I was wondering how the `/backend/metadata`  endpoint came to be since it's not in Cromwell/CromIAM. It appeared in Orchestration first https://github.com/broadinstitute/firecloud-orchestration/pull/723 and was adopted by Job Manager a few months later https://github.com/DataBiosphere/job-manager/pull/661.

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/CromIamApiService.scala
JVThomas: Yeah, that makes sense. The main reason I want to avoid a separate file is because eventually the new file will need to be imported and referenced in `FireCloudApiService`. Due to the overlapping root path structure, I'm sure that the order in which the files are defined there will matter, so we'll have the problem there.

By keeping the routes defined in `CromIamService`, we can at least define the problem and the context in the same spot reasonably well.

src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingPassthrough.scala
JVThomas: Wasn't aware of the `impl` naming convention, that's good to know. It lines up with why the method was defined `private` to begin with.

aednichols (approved):
Thank you for the revisions!

src/test/scala/org/broadinstitute/dsde/firecloud/utils/StreamingPassthroughSpec.scala
JVThomas: Yeah, I usually write out the parameters on tests just so the function outline/context is clear

PR #1096, opened 2023-01-25
[SU-281] Azure preview welcome email

MatthewBemis:
I tested that the existing email flow still works and also sort of pseudo-tested that the new Azure account flow works by changing around some of the pattern matching logic to fire off an Azure email if the user logs in as a legacy GCP user. So we know the Orch->Thurloe->SendGrid flow works fine for the new Azure email and as long as the pattern matching logic in `generateWelcomeEmail` is correct, then all should be well.

Other relevant PRs for context:

https://github.com/broadinstitute/workbench-libs/pull/1229
https://github.com/broadinstitute/thurloe/pull/131
https://github.com/broadinstitute/firecloud-develop/pull/3224

In all cases:

### DIFF
--- a/project/Merging.scala
+++ b/project/Merging.scala
@@ -13,6 +13,7 @@ object Merging {
     case x if x.contains("javax/annotation") => MergeStrategy.first
 
     case x if x.endsWith("kotlin-stdlib.kotlin_module") => MergeStrategy.first
+    case x if x.contains("bouncycastle") => MergeStrategy.first
     case x if x.endsWith("kotlin-stdlib-common.kotlin_module") => MergeStrategy.first
     case x if x.endsWith("arrow-git.properties") => MergeStrategy.concat
 
--- a/project/Merging.scala
+++ b/project/Merging.scala
@@ -12,6 +12,10 @@ object Merging {
     case x if x.contains("javax/activation") => MergeStrategy.first
     case x if x.contains("javax/annotation") => MergeStrategy.first
 
+    case x if x.endsWith("kotlin-stdlib.kotlin_module") => MergeStrategy.first
+    case x if x.endsWith("kotlin-stdlib-common.kotlin_module") => MergeStrategy.first
+    case x if x.endsWith("arrow-git.properties") => MergeStrategy.concat
+
     // For the following error:
     // Error: (assembly) deduplicate: different file contents found in the following:
     // Error: /home/sbtuser/.cache/coursier/v1/https/repo1.maven.org/maven2/com/google/protobuf/protobuf-java/3.19.4/protobuf-java-3.19.4.jar:google/protobuf/struct.proto
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
@@ -15,8 +15,8 @@ import com.google.api.client.googleapis.json.GoogleJsonResponseException
 import com.google.api.client.googleapis.services.AbstractGoogleClientRequest
 import com.google.api.client.http.HttpResponseException
 import com.google.api.client.json.jackson2.JacksonFactory
-import com.google.api.services.admin.directory.model.{Group, Member}
-import com.google.api.services.admin.directory.{Directory, DirectoryScopes}
+import com.google.api.services.directory.model.{Group, Member}
+import com.google.api.services.directory.{Directory, DirectoryScopes}
 import com.google.api.services.pubsub.model.{PublishRequest, PubsubMessage}
 import com.google.api.services.pubsub.{Pubsub, PubsubScopes}
 import com.google.api.services.storage.model.Bucket
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/model/UserInfo.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/model/UserInfo.scala
@@ -27,13 +27,11 @@ trait WithAccessToken { val accessToken : OAuth2BearerToken }
   * @param userEmail the user's email address. Resolved to the owner if the request is from a pet.
   * @param accessToken the user's access token. Either a B2C JWT or a Google opaque token.
   * @param accessTokenExpiresIn number of seconds until the access token expires.
-  * @param userSubjectId the user id. Either a Google id (numeric) or a B2C id (uuid).
+  * @param id the user id. Either a Google id (numeric) or a B2C id (uuid).
   * @param googleAccessTokenThroughB2C if this is a Google login through B2C, contains the opaque
   *                                    Google access token. Empty otherwise.
   */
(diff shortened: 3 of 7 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -15,8 +15,8 @@ import com.google.api.client.googleapis.json.GoogleJsonResponseException
 import com.google.api.client.googleapis.services.AbstractGoogleClientRequest
 import com.google.api.client.http.HttpResponseException
 import com.google.api.client.json.jackson2.JacksonFactory
-import com.google.api.services.admin.directory.model.{Group, Member}
-import com.google.api.services.admin.directory.{Directory, DirectoryScopes}
+import com.google.api.services.directory.model.{Group, Member}
```
MatthewBemis, line 18: minor fallout from having to update the workbench-google2 library

src/main/scala/org/broadinstitute/dsde/firecloud/model/UserInfo.scala
```diff
@@ -16,10 +19,22 @@ import akka.http.scaladsl.model.headers.OAuth2BearerToken
   *  This is so that we can use AccessToken for cookies we get from the browser, which do not come with the fields in
   *  UserInfo.
  */
+
 trait WithAccessToken { val accessToken : OAuth2BearerToken }
 
-case class UserInfo(userEmail: String, accessToken: OAuth2BearerToken, accessTokenExpiresIn: Long, id: String) extends WithAccessToken {
-  def getUniqueId = id
```
MatthewBemis, line 22: removed `getUniqueId` because it was unused

project/Merging.scala
```diff
@@ -12,6 +12,11 @@ object Merging {
     case x if x.contains("javax/activation") => MergeStrategy.first
     case x if x.contains("javax/annotation") => MergeStrategy.first
 
+    case x if x.endsWith("kotlin-stdlib.kotlin_module") => MergeStrategy.first
+    case x if x.contains("bouncycastle") => MergeStrategy.first
+    case x if x.endsWith("kotlin-stdlib-common.kotlin_module") => MergeStrategy.first
+    case x if x.endsWith("arrow-git.properties") => MergeStrategy.concat
```
MatthewBemis, line 18: seems to be fallout from the upgraded google2 package

src/main/scala/org/broadinstitute/dsde/firecloud/service/RegisterService.scala
```diff
@@ -53,10 +53,17 @@ class RegisterService(val rawlsDao: RawlsDAO, val samDao: SamDAO, val thurloeDao
     }
   }
 
+  def generateWelcomeEmail(userInfo: UserInfo): Notification = {
+    userInfo match {
+      case UserInfo(_, _, _, _, None) if userInfo.isB2C => AzurePreviewActivationNotification(WorkbenchUserId(userInfo.id))
+      case _ => ActivationNotification(WorkbenchUserId(userInfo.id))
+    }
+  }
```
MatthewBemis: I'll add code comments to this block because I'm realizing it's vague.

Essentially, we need to distinguish between the Azure-user case and the non-Azure-user case. The way to do that is to test if the user is using B2C, and then to look for a google access token (which is what the `None` is matching). If it's B2C and there's no Google access token, it's an Azure user. All other cases just use the standard activation notification.

src/main/scala/org/broadinstitute/dsde/firecloud/service/RegisterService.scala
```diff
@@ -53,10 +53,17 @@ class RegisterService(val rawlsDao: RawlsDAO, val samDao: SamDAO, val thurloeDao
     }
   }
 
+  def generateWelcomeEmail(userInfo: UserInfo): Notification = {
+    userInfo match {
```
jordansafer: Could we match on userInfo.googleAccessTokenThroughB2C instead? It will make the pattern matching more explicit and readable

jordansafer (approved):
Looks good, see my one minor comment on the pattern match

src/main/scala/org/broadinstitute/dsde/firecloud/service/RegisterService.scala
MatthewBemis: Definitely- I'll make that change

MatthewBemis: Done in 2729bf3c7a0d0071a7e8440fda60f590cad3e07f

src/main/scala/org/broadinstitute/dsde/firecloud/utils/StandardUserInfoDirectives.scala
```diff
@@ -14,10 +14,11 @@ trait StandardUserInfoDirectives extends UserInfoDirectives {
...
       headerValueByName("OIDC_CLAIM_expires_in") &
       headerValueByName("OIDC_CLAIM_email") &
-      optionalHeaderValueByName("OAUTH2_CLAIM_google_id")
+      optionalHeaderValueByName("OAUTH2_CLAIM_google_id") &
+      optionalHeaderValueByName("OAUTH2_CLAIM_idp_access_token")
     ) tmap {
-      case (token, userId, expiresIn, email, googleIdOpt) => {
-        UserInfo(email, OAuth2BearerToken(token), expiresIn.toLong, googleIdOpt.getOrElse(userId))
+      case (token, userId, expiresIn, email, googleIdOpt, googleTokenOpt) => {
+        UserInfo(email, OAuth2BearerToken(token), expiresIn.toLong, googleIdOpt.getOrElse(userId), googleTokenOpt.map(OAuth2BearerToken))
```
rtitle, line 21: Just thinking something through. We're setting the `UserInfo.userId` here as `googleIdOpt.getOrElse(userId)`. So it'll always be the google (numeric) ID for google users. Does this mean we can just test `UserInfo.isB2C` to test whether it's an azure user? So we don't really need to read the `OAUTH2_CLAIM_idp_access_token` header.

Note this logic to set the userId is kind of unique to Orch (Rawls [doesn't do it](https://github.com/broadinstitute/rawls/blob/develop/core/src/main/scala/org/broadinstitute/dsde/rawls/openam/StandardUserInfoDirectives.scala#L37)) -- the reason is essentially to make Thurloe backwards compatible with existing Google users through B2C.

MatthewBemis: Oh interesting, that's a good point would simplify things a little bit. On the other hand, reading the `OAUTH2_CLAIM_idp_access_token` header is more in line with behavior of the other services and doesn't rely on a backwards compatibility quirk that might just be a temporary(?) solution.

I'm somewhat ambivalent here. Do you have any strong preference that I change it?

rtitle: Circling back to his, I don't have a strong opinion. I think the current version works, was just thinking if could be simplified. I can buy the argument that this way is more understandable and future-proof.

rtitle (approved):

PR #1161, opened 2023-06-12
AJ-1076 check for strings that look like number literals

calypsomatic:
[AJ-1076](https://broadworkbench.atlassian.net/browse/AJ-1076)
If a string attribute consists of a number ending in 'd' or 'f', then scala will read that as indicating 'double' or 'float', coerce it to a number, and drop the 'd' or 'f'.  This PR assumes that such attributes will always be intended to be strings and specifically checks for these cases.

[AJ-1076]: https://broadworkbench.atlassian.net/browse/AJ-1076?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupport.scala
@@ -155,7 +155,8 @@ trait TSVFileSupport {
         // to AttributeString.
         case Success(doubleValue) if !Double.NegativeInfinity.equals(doubleValue)
           && !Double.PositiveInfinity.equals(doubleValue)
-          && !Double.NaN.equals(doubleValue) =>
+          && !Double.NaN.equals(doubleValue)
+        && !matchesLiteral(value) =>
           AttributeNumber(doubleValue)
         case _ => Try(BooleanUtils.toBoolean(value.toLowerCase, "true", "false")) match {
           case Success(booleanValue) => AttributeBoolean(booleanValue)
@@ -168,6 +169,10 @@ trait TSVFileSupport {
         }
       }
     }
+  }
+
+  def matchesLiteral(value: String): Boolean = {
+    value.toLowerCase().endsWith("d") || value.toLowerCase().endsWith("f")
   }
 
   /**
--- a/src/test/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupportSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupportSpec.scala
@@ -154,6 +154,12 @@ class TSVFileSupportSpec extends AnyFreeSpec with TSVFileSupport {
     }
   }
 
+  List("1234d", "-5678D", "9012f", "3456F") foreach { str =>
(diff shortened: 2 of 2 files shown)


src/test/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupportSpec.scala
```diff
@@ -154,6 +154,12 @@ class TSVFileSupportSpec extends AnyFreeSpec with TSVFileSupport {
     }
   }
 
+  List("1234d", "-5678D", "9012f", "3456F") foreach { str =>
```
yuliadub, line 157: have you tried 123i? curious if it would treat i as "integer" or not. part of me wonders if just doing regex with number + letter should default to string, no matter what the letter is?

calypsomatic: I tried out 'd', 'f',' i' and 'l' to see what would happen, and 'd' and 'f' were the only ones that were coerced to numbers.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupport.scala
```diff
@@ -155,7 +155,8 @@ trait TSVFileSupport {
         // to AttributeString.
         case Success(doubleValue) if !Double.NegativeInfinity.equals(doubleValue)
           && !Double.PositiveInfinity.equals(doubleValue)
-          && !Double.NaN.equals(doubleValue) =>
+          && !Double.NaN.equals(doubleValue)
+        && !matchesLiteral(value) =>
           AttributeNumber(doubleValue)
```
davidangb, line 160: I wondered if you could replace the entire `matchesLiteral` method with a library call. We already have apache-commons lang3, so this could be:
```suggestion
          && !Double.NaN.equals(doubleValue)
          && NumberUtils.isParsable(value) =>
            AttributeNumber(doubleValue)
```

buuuuuut this causes one other unit test to fail, on input of "42.": https://github.com/broadinstitute/firecloud-orchestration/blob/ad08ecd794b5a0824e7be3937b9eae55d9921804/src/test/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupportSpec.scala#L92 … which I think might be a bad test anyway? Do we really want "42." to be interpreted as a number?

davidangb (commented):
If you're willing to track down what "42." should do (see inline), I think using `NumberUtils.isParsable(value)` is more robust; but I'm ok with this PR as-is if that causes a lot of scope creep.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupport.scala
calypsomatic: `NumberUtils.isParsable` also fails on the min and max double values (e.g. 4.9E-324) so it's not looking promising

davidangb (approved):

PR #1172, opened 2023-07-05
AJ-1112 update rawls model

calypsomatic:
https://broadworkbench.atlassian.net/browse/AJ-1112
This PR is a follow on to https://github.com/broadinstitute/rawls/pull/2415, which changed the Rawls model, so here I am updating the model version.
Since the `WorkspaceDetails` model has also been updated since the last time the Rawls model was changed, I had to update all instances of it in the tests.

In all cases:


calypsomatic:
jenkins retest

davidangb (approved):

PR #1197, opened 2023-08-14
AJ-1128 check for json in tsv attribute upload

calypsomatic:
[AJ-1128](https://broadworkbench.atlassian.net/browse/AJ-1128)
Workspace attributes are all automatically typed as strings.  This led to a problem with attributes that were large/long json objects - too large to be strings.  If we check for json compatibility and assign these attributes to json if possible, the problem is solved.

In all cases:

[AJ-1128]: https://broadworkbench.atlassian.net/browse/AJ-1128?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupport.scala
@@ -126,7 +126,9 @@ trait TSVFileSupport {
       if (value.equals("__DELETE__"))
         RemoveAttribute(AttributeName.fromDelimitedName(name))
       else {
-        AddUpdateAttribute(AttributeName.fromDelimitedName(name), AttributeString(StringContext.processEscapes(value)))
+        //stringToTypedAttribute
+        AddUpdateAttribute(AttributeName.fromDelimitedName(name), stringToTypedAttribute(StringContext.processEscapes(value)))
+//        AddUpdateAttribute(AttributeName.fromDelimitedName(name), AttributeString(StringContext.processEscapes(value)))
       }
     }
   }
@@ -161,8 +163,12 @@ trait TSVFileSupport {
         case _ => Try(BooleanUtils.toBoolean(value.toLowerCase, "true", "false")) match {
           case Success(booleanValue) => AttributeBoolean(booleanValue)
           case Failure(_) =>
-            Try(value.parseJson.convertTo[AttributeEntityReference]) match {
-              case Success(ref) => ref
+            Try(value.parseJson) match {
+              case Success(jsVal) =>
+                  Try(jsVal.convertTo[AttributeEntityReference]) match {
+                    case Success(ref) => ref
+                    case Failure(_) => AttributeValueRawJson(value)
+                  }
               case Failure(_) => AttributeString(value)
             }
 
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupport.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupport.scala
@@ -126,9 +126,7 @@ trait TSVFileSupport {
       if (value.equals("__DELETE__"))
         RemoveAttribute(AttributeName.fromDelimitedName(name))
       else {
-        //stringToTypedAttribute
-        AddUpdateAttribute(AttributeName.fromDelimitedName(name), stringToTypedAttribute(StringContext.processEscapes(value)))
-//        AddUpdateAttribute(AttributeName.fromDelimitedName(name), AttributeString(StringContext.processEscapes(value)))
+        AddUpdateAttribute(AttributeName.fromDelimitedName(name), checkForJson(StringContext.processEscapes(value)))
       }
     }
   }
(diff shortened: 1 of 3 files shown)


src/test/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupportSpec.scala
```diff
@@ -139,6 +144,14 @@ class TSVFileSupportSpec extends AnyFreeSpec with TSVFileSupport {
       }
     }
 
+//    "should detect json values when applicable" in {
+//      jsonTestCases foreach {
+//        case (input, expected) => withClue(s"should handle potential reference: $input") {
+//          stringToTypedAttribute(input) shouldBe expected
```
calypsomatic: Ignore for now; this is from an earlier draft where I used `stringToTypedAttribute` instead of writing a new `checkForJson` method.  That required modifying `stringToTypedAttribute` which changed expected behavior for some types.  Could potentially return to this depending on how I end up dealing with parsing numeric attributes in `checkForJson`.

src/test/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupportSpec.scala
```diff
@@ -106,6 +108,9 @@ class TSVFileSupportSpec extends AnyFreeSpec with TSVFileSupport {
     val referenceTestCases = Map(
       """{"entityType":"targetType","entityName":"targetName"}""" -> AttributeEntityReference("targetType", "targetName")
     )
+    val jsonTestCases = Map(
+      """{"tables":{"sample":{"save":["bai",true,"bam",true,"participant",false,"sample",true]}}}""" -> AttributeValueRawJson("""{"tables":{"sample":{"save":["bai",true,"bam",true,"participant",false,"sample",true]}}}""")
```
calypsomatic: Ignore for now; see comment below.

src/test/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupportSpec.scala
```diff
@@ -23,8 +23,10 @@ class TSVFileSupportSpec extends AnyFreeSpec with TSVFileSupport {
       val attributes = getWorkspaceAttributeCalls(MockTSVLoadFiles.validWorkspaceAttributes)
       assertResult(attributes) {
         List(AddUpdateAttribute(AttributeName("default", "a1"), AttributeString("v1")),
-          AddUpdateAttribute(AttributeName("default", "a2"), AttributeString("2")),
-          AddUpdateAttribute(AttributeName("default", "a3"), AttributeString("[1,2,3]")))
+          AddUpdateAttribute(AttributeName("default", "a2"), AttributeValueRawJson("2")),
```
calypsomatic: Note these are changed from `AttributeString` to `AttributeValueRawJson`; see question in the PR description.

src/main/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupport.scala
```diff
@@ -165,12 +165,18 @@ trait TSVFileSupport {
...
         }
       }
     }
   }
 
+  def checkForJson(value: String): Attribute = {
+    Try(value.parseJson) match {
+        case Success(jsVal) => AttributeValueRawJson(value)
+        case Failure(_) => AttributeString(value)
+      }
```
davidangb, line 177: if you wanted to change as little as possible in this PR - that is, treat json objects as json but preserve existing behavior of everything else being treated as a string - you could check if the parsed `JsValue` is truly a json object:
```scala
    Try(value.parseJson) match {
      case Success(jso: JsObject) => AttributeValueRawJson(jso)
      case _ => AttributeString(value)
    }
```

calypsomatic: Ah thank you!  Saves me the trouble of digging into documentation to figure out what my options are for doing just that!

src/test/scala/org/broadinstitute/dsde/firecloud/service/TSVFileSupportSpec.scala
```diff
@@ -24,7 +24,9 @@ class TSVFileSupportSpec extends AnyFreeSpec with TSVFileSupport {
       assertResult(attributes) {
         List(AddUpdateAttribute(AttributeName("default", "a1"), AttributeString("v1")),
           AddUpdateAttribute(AttributeName("default", "a2"), AttributeString("2")),
-          AddUpdateAttribute(AttributeName("default", "a3"), AttributeString("[1,2,3]")))
+          AddUpdateAttribute(AttributeName("default", "a3"), AttributeString("[1,2,3]")),
+        AddUpdateAttribute(AttributeName("default", "a4"), AttributeValueRawJson("""{"tables":{"sample":{"save":["participant",false,"sample",true]}}}""")
+        ))
```
calypsomatic, line 29: Note: did not add a test specifically for the new method `checkForJson` since it is only used in `getWorkspaceAttributeCalls` and its behavior is checked in this test.

jladieu:
I also don't have a great answer for "how much does it matter what types are?", or, to be honest, the specifics of this PR, but in general whenever I see type loss that could be avoided, I try to avoid it because it's irreversible (and by this, I mean, we can only _infer_ types if we want to reverse the transformation, but never really be sure).  If there's a fix to the issue you're trying to solve that preserves type information through transformation without causing a bunch of unintended consequences, I'd bias toward holding onto the types for as long as possible.

nawatts (approved):

src/test/scala/org/broadinstitute/dsde/firecloud/mock/MockTSV.scala
```diff
@@ -3,6 +3,7 @@ package org.broadinstitute.dsde.firecloud.mock
 import akka.http.scaladsl.model.Multipart
 import akka.http.scaladsl.model.Multipart.FormData.BodyPart
 import org.broadinstitute.dsde.firecloud.utils.TSVLoadFile
+import org.broadinstitute.dsde.rawls.model.AttributeValueRawJson
```
davidangb: nit: unused import

davidangb (approved):
:+1: for a small, targeted fix. I also agree with @jladieu that in the long run it would be good to align this with `stringToTypedAttribute` and have the workspace attribute TSV and data table TSV use the same code - that's out of scope for this ticket though

PR #1220, opened 2023-09-27
AJ-1131: cache-control and pragma headers by default

davidangb:
https://broadworkbench.atlassian.net/browse/AJ-1131

Orchestration now adds `Cache-control: no-store` and `Pragma: no-cache` headers to any response that doesn't already have those headers, to prevent caching of any sensitive data (e.g. PHI) in users' browsers.

The implementation in this PR allows underlying services like Rawls or Sam - or individual APIs inside Orch - to override this behavior simply by specifying their own values for those headers. If the response already has a cache-control or pragma header, Orch will not overwrite it.

Note specifically that since Orch contains many passthroughs, if the service Orch is passing through to sets one of these headers, Orch will respect that header. When looking at Orch's Swagger UI, you can usually tell if an API is a passthrough by looking for `x-passthrough: true` under "Extensions" at the bottom of each API.

_Before this PR:_
caching-before

_After this PR:_
caching-after

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudApiService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudApiService.scala
@@ -4,6 +4,8 @@ import akka.actor.{ActorRefFactory, ActorSystem}
 import akka.event.Logging.LogLevel
 import akka.event.{Logging, LoggingAdapter}
 import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
+import akka.http.scaladsl.model.headers.CacheDirectives.{`no-cache`, `no-store`}
+import akka.http.scaladsl.model.headers.{RawHeader, `Cache-Control`}
 import akka.http.scaladsl.model.{HttpEntity, HttpRequest, StatusCodes}
 import akka.http.scaladsl.server
 import akka.http.scaladsl.server.Directives._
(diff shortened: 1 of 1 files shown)


nawatts (approved):

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudApiService.scala
```diff
@@ -159,6 +161,16 @@ trait FireCloudApiService extends CookieAuthedApiService
...
 
+  // Return "cache-control: no-store" and "pragma: no-cache" headers,
+  // if those headers don't already exist on the response. This is included
+  // in `routeWrappers` below, so it affects all responses from Orch.
+  // Note that many Orch APIs are passthroughs, and if the underlying
+  // service (Rawls, Sam, etc) already returns these headers, Orch
+  // will not overwrite them.
+  private val noCacheNoStore: Directive0 = respondWithDefaultHeaders(
+    `Cache-Control`(`no-store`),
+    RawHeader("Pragma", `no-cache`.value))
```
jladieu, line 172: Why use `RawHeader` for `Pragma`, but not for `Cache-Control`?

jladieu (approved):

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudApiService.scala
davidangb: I couldn't find a pre-defined header definition for `Pragma` in `akka.http.scaladsl.model.headers.*`, so I had to resort to the `RawHeader` for it

PR #1212, opened 2023-09-11
Update transport from 5.6.16 to 7.17.13

broadbot:
## About this PR
📦 Updates org.elasticsearch.client:transport from `5.6.16` to `7.17.13` ⚠

## Usage
✅ **Please merge!**

I'll automatically update this PR to resolve conflicts as long as you don't change it yourself.

If you'd like to skip this version, you can just close this PR. If you have any feedback, just mention me in the comments below.

Configure Scala Steward for your repository with a [`.scala-steward.conf`](https://github.com/scala-steward-org/scala-steward/blob/b83aae55d9dd000548c3b3c9b63d79636e7b3c8b/docs/repo-specific-configuration.md) file.

_Have a fantastic day writing Scala!_

<details>
<summary>⚙ Adjust future updates</summary>

Add this to your `.scala-steward.conf` file to ignore future updates of this dependency:
```
updates.ignore = [ { groupId = "org.elasticsearch.client", artifactId = "transport" } ]
```
Or, add this to slow down future updates of this dependency:
```
dependencyOverrides = [{
  pullRequests = { frequency = "30 days" },
  dependency = { groupId = "org.elasticsearch.client", artifactId = "transport" }
}]
```
</details>

<sup>
labels: library-update, early-semver-major, semver-spec-major, commit-count:1
</sup>

PR #1221, opened 2023-09-27
AJ-1330: fix .scala-steward.conf

davidangb:
our `.scala-steward.conf` was incorrect w/r/t ignored updates.

It set `updates.ignore` twice; the second invocation overwrote the first, and that's why Scala Steward generated #1212 even though we didn't want it to.

I've also added `mockserver-netty` to the ignore list, and tagged each ignore with its associated Jira ticket.


jladieu (approved):
I have no idea what scala-steward is but I am pretty sure you didn't do anything terrible here 😆

davidangb:
> I have no idea what scala-steward is

@jladieu https://github.com/scala-steward-org/scala-steward … it's a Scala version of Dependabot! (DSP runs an internal Scala Steward server; we don't use the central one)

nawatts (approved):

yuliadub (approved):

davidangb:
503 from Dockerhub; Jenkins retest

PR #1246, opened 2023-11-20
ID-925 Fix register-user integration test flakiness.

Ghost-in-a-Jar:
https://broadworkbench.atlassian.net/browse/ID-925

The register-user ui integration test is being flakey and I think that the registration flow in orchestration that we changes recently is the cause. In a for yield, futures run concurrently and not necessarily in order unless they depend on results from eachother. So I think what is happening is that sometimes the register request to sam finishes before the rest but when it doesnt we get an error from thurloe and the test fails. By awaiting the sam registration call we make sure it finishes first. 

I think this is probably affecting users in production that are trying to register, hard to say how many are actually hitting this.

It is generally good to avoid awaiting futures, but in this case i think it makes sense, register user isnt a high volume endpoint so I am not worried about the marginal performance loss. An alternative is to do future.onSuccess or something but I like the simplicity and straightforward nature of the await here. 

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/RegisterService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/RegisterService.scala
@@ -12,9 +12,7 @@ import akka.http.scaladsl.model.StatusCodes
 import org.broadinstitute.dsde.firecloud.FireCloudConfig.Sam
 import org.broadinstitute.dsde.workbench.model.WorkbenchUserId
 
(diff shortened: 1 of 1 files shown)


sjkobori (approved):

tlangs:
Hmmm. I really don't want to `Await`. I know its probably not a perf concern, but its really not good practice. 

Instead, can we use `Future.onComplete`, or something to force the completion of the Sam call before calling Thurloe?

tlangs (approved):

PR #1249, opened 2023-11-30
AJ-1496: remove /api/duos/consent/orsp/{orspId} API

davidangb:
The Consent service has removed the ability to look up an ORSP id and return the data use restrictions for that ORSP id. The API that Orch used to call for this now returns a 404 all the time.

Thus, this PR removes the call from Orch to Consent for this lookup, and removes Orch's own `/api/duos/consent/orsp/{orspId}` API. Technically, removing this API would be a breaking change. However, since Consent will _always_ return a 404, if Orch kept its API alive it would also always return a 404. Instead, let's remove the API entirely … which will also return a 404.

By removing the call from Orch to Consent for this lookup, Orch can subsequently delete a whole bunch of code and unit tests. In fact, Orch can even remove Consent as a subsystem from its `/status` API.

Once this merges and is stable, I will follow up sometime after with a PR to remove the consent url from Orch's config: https://github.com/broadinstitute/terra-helmfile/blob/0935fb0a1fd3494fd2d7bfc09f4ada55c981fa27/charts/firecloudorch/templates/config/_conf.tpl#L344

---

In all cases:


src/main/scala/org/broadinstitute/dsde/firecloud/service/LibraryServiceSupport.scala
```diff
@@ -36,51 +34,32 @@ trait LibraryServiceSupport extends DataUseRestrictionSupport with LazyLogging {
...
-            val newAttrs = replaceDataUseAttributes(ws.attributes.getOrElse(Map.empty), orspAttrs)
-            ws.copy(attributes = Option(newAttrs))
-          case _ =>
-            // this workspace does not have an ORSP id; leave it untouched
-            ws
-        }
-      }
-      annotatedWorkspaces map {w => indexableDocument(w, parentMap, ontologyDAO)}
     }
+    Future.successful(annotatedWorkspaces map { w => indexableDocument(w, parentMap, ontologyDAO) })
```
davidangb, line 62: changes in this class are the only real functional changes. Previously, we called `consentDAO.getRestriction(orspId)` if an ORSP id was present, and the remainder of the code was dedicated to handling the responses. Now, since that call will never succeed, the code is much simplified.

davidangb:
martha didn't start in automation tests. Jenkins retest.

davidangb:
martha didn't start again. Is this a persistent problem? Jenkins retest.

davidangb:
Jenkins retest.

davidangb:
Jenkins retest

jladieu:
Delightful red diff.  B-B-B-Bonus multiplier 💰 💰 💰 💰  for deleting Scala.

_(sorry Scala)_

jladieu (approved):
Sorry I meant to approve this first time around and just left an cheerleading comment instead!  Rectified.

calypsomatic (approved):

PR #1250, opened 2023-12-02
AJ-1490, AJ-423: update workbench-libs and mockserver

davidangb:
This PR covers:
* https://broadworkbench.atlassian.net/browse/AJ-423: Upgrade Orch's mockserver-netty to current
* https://broadworkbench.atlassian.net/browse/AJ-1490: Upgrade workbenchLibsHash for all workbench-libs libraries in orch

This PR intentionally does not cover other updates, such as updating workbench-libs in the automation/ subdirectory, or updating workbench-util (AJ-424), akka (AJ-421) or sentry libs (AJ-422). To keep scope down, those should be separate PRs.

Update Orch's three workbench-libs dependencies to latest versions.

This required a corresponding update to mockserver-netty. This update carries some syntax changes in test code. All of the code changes in this PR are purely for compatibility with the updates.

---

In all cases:


project/Merging.scala
```diff
@@ -15,6 +15,7 @@ object Merging {
     case x if x.endsWith("kotlin-stdlib.kotlin_module") => MergeStrategy.first
     case x if x.contains("bouncycastle") => MergeStrategy.first
     case x if x.endsWith("kotlin-stdlib-common.kotlin_module") => MergeStrategy.first
+    case x if x.endsWith("okio.kotlin_module") => MergeStrategy.first
```
davidangb, line 18: this fixes an issue in `sbt assembly`, which we run as part of deployment and tests, but developers typically do not run locally. `sbt assembly` compiles the [fat jar](https://www.baeldung.com/scala/sbt-fat-jar), and hit an error trying to de-duplicate `okio-jvm META-INF/okio.kotlin_module` and `okio META-INF/okio.kotlin_module`. This fix takes the same approach as the merge strategy for `kotlin-stdlib` and `otlin-stdlib-common` just above in this file.

src/test/scala/org/broadinstitute/dsde/firecloud/mock/MockWorkspaceServer.scala
```diff
@@ -276,7 +277,7 @@ object MockWorkspaceServer {
         request()
           .withMethod("GET")
           .withPath(s"${workspaceBasePath}/%s/%s/submissions/%s/workflows/%s"
-            .format(mockSpacedWorkspace.namespace, mockSpacedWorkspace.name, mockValidId, mockValidId)))
+            .format(mockSpacedWorkspace.namespace, UrlEscapers.urlPathSegmentEscaper().escape(mockSpacedWorkspace.name), mockValidId, mockValidId)))
```
davidangb, line 280: these are syntax/behavior differences in the updated mockserver library, as are almost all the other changes in this PR.

project/Dependencies.scala
```diff
@@ -95,7 +95,7 @@ object Dependencies {
     "com.github.pathikrit"          %% "better-files"        % "3.9.2",
 
     "org.scalatest"                 %% "scalatest"           % "3.2.17"   % "test",
-    "org.mock-server"                % "mockserver-netty"    % "3.11"  % "test", // TODO: upgrading higher causes failures, need to investigate
+    "org.mock-server"                % "mockserver-netty"    % "5.15.0"  % "test",
```
davidangb, line 98: the workbench-libs updates hit a conflict with Orch's older version of mockserver, so I needed to update mockserver too.

project/Dependencies.scala
```diff
@@ -6,7 +6,7 @@ object Dependencies {
   val jacksonV = "2.13.4"
   val jacksonHotfixV = "2.13.4.2" // for when only some of the Jackson libs have hotfix releases
   val nettyV = "4.1.101.Final"
-  val workbenchLibsHash = "084d25b"
+  val workbenchLibsHash = "8ccaa6d"
```
jladieu: Without reading anything else from this repo (eg: its readme), I am guessing this hash is just a github sha from the workbenchLibsRepo.

Might be worth an inline comment with a link to where that list of hashes is, for quicker lookup, assuming this is something that needs to be discovered & changed whenever we need to bump our dep here.

src/test/scala/org/broadinstitute/dsde/firecloud/mock/MockWorkspaceServer.scala
jladieu: I see you only had to do this URL escaping once in this PR...but it also looks pretty ugly -- if you had to do this 20 times, once for each test case, what would your solution look like?

I'm asking not because I want you to implement that solution (since this is just a one-off) but just curious what it'd look like in Scala :)

jladieu (approved):

calypsomatic (approved):

src/test/scala/org/broadinstitute/dsde/firecloud/mock/MockWorkspaceServer.scala
davidangb: I would take one of two routes:

wrap the escaping in something terser, like [`StreamingPassthrough.escapePathSegment`](https://github.com/broadinstitute/firecloud-orchestration/blob/91ad97a82008ca8742e5a63b23d2bc1984f10188/src/main/scala/org/broadinstitute/dsde/firecloud/utils/StreamingPassthrough.scala#L32-L34), so you end up with:
```scala
          .withPath(s"${workspaceBasePath}/%s/%s/submissions/%s/workflows/%s"
            .format(mockSpacedWorkspace.namespace, escapePathSegment(mockSpacedWorkspace.name), mockValidId, mockValidId)))
```
this isn't a big win, but it's nicer.

Alternately, I would make an implicit class that provides methods for strings; then, I could call my custom method directly on the thing I wanted to escape:
```scala
    implicit class StringEscaper(s: String) {
      def escape: String = UrlEscapers.urlPathSegmentEscaper().escape(s)
    }

…
          .withPath(s"${workspaceBasePath}/%s/%s/submissions/%s/workflows/%s"
            .format(mockSpacedWorkspace.namespace, mockSpacedWorkspace.name.escape, mockValidId, mockValidId)))


```

project/Dependencies.scala
davidangb: I added a comment (and yes, it is in the workbench-libs readme)

PR #1257, opened 2023-12-08
AJ-1485: de-flake unit test by removing random collisions

davidangb:
The `SubmissionApiServiceSpec` unit test has been flaky. See, for instance: https://github.com/broadinstitute/firecloud-orchestration/actions/runs/7145799461/job/19462223538?pr=1256

This PR hopefully fixes the flakiness. See my comments inline for why.

In all cases:

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/firecloud/mock/MockWorkspaceServer.scala
+++ b/src/test/scala/org/broadinstitute/dsde/firecloud/mock/MockWorkspaceServer.scala
@@ -69,9 +69,9 @@ object MockWorkspaceServer {
     WorkspaceState.Ready
   )
 
-  val mockValidId = randomPositiveInt()
-  val mockInvalidId = randomPositiveInt()
-  val alternativeMockValidId = randomPositiveInt()
+  val mockValidId = 3
+  val mockInvalidId = 6
+  val alternativeMockValidId = 9
 
   val mockValidSubmission = OrchSubmissionRequest(
     methodConfigurationNamespace = Option(randomAlpha()),


src/test/scala/org/broadinstitute/dsde/firecloud/mock/MockWorkspaceServer.scala
```diff
@@ -69,9 +69,9 @@ object MockWorkspaceServer {
     WorkspaceState.Ready
   )
 
-  val mockValidId = randomPositiveInt()
-  val mockInvalidId = randomPositiveInt()
-  val alternativeMockValidId = randomPositiveInt()
+  val mockValidId = 3
+  val mockInvalidId = 6
+  val alternativeMockValidId = 9
```
davidangb, line 74: These three IDs are used to trigger different responses from the mock workspace server. See, for instance, the expectations defined at https://github.com/broadinstitute/firecloud-orchestration/blob/db26b3c77ecaa0fe8df6c064096937ec4c0de6a7/src/test/scala/org/broadinstitute/dsde/firecloud/mock/MockWorkspaceServer.scala#L108-L112, where these three IDs should return `OK`, `BadRequest`, and `NotFound`.

In test runs where `randomPositiveInt()` returned different values for each id, all tests passed and everything was fine.

However, if `randomPositiveInt()` returned the same value for any of these ids, tests would mysteriously fail.

Why did we use randomness to generate these ids? Has this been a problem forever???

calypsomatic (approved):

ashanhol (approved):

PR #1308, opened 2024-03-13
AJ-1602: list-jobs API should aggregate Import Service and cWDS

davidangb:
AJ-1602

High-level changes:
* add the cWDS client library
* create a new `CwdsDAO` trait, with `Http*` and `Mock*` implementations
* when listing jobs:
    * call Import Service to list its jobs
    * if `cwds.enabled` (see https://github.com/broadinstitute/terra-helmfile/pull/5261):
        * call Rawls to translate the workspace namespace/name into a workspace id
        * call cWDS, using the workspace id, to list its jobs
    * else:
        *  skip Rawls and cWDS and return an empty list of jobs
    * concat the Import Service and cWDS results
* unit tests

open questions:
- [ ] do we need more error trapping/retries?
- [ ] do we need cWDS in swatomation?

for a future PR:
- [ ] similar Import Service/cWDS support for the create-job APIs

---

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
@@ -381,7 +381,11 @@ class EntityService(rawlsDAO: RawlsDAO, importServiceDAO: ImportServiceDAO, cwds
       // get jobs from Import Service
       importServiceJobs <- importServiceDAO.listJobs(workspaceNamespace, workspaceName, runningOnly)(userInfo)
       // get jobs from cWDS
-      cwdsJobs = cwdsDAO.listJobsV1(workspace.workspace.workspaceId, runningOnly)(userInfo)
+      cwdsJobs = if (cwdsDAO.isEnabled) {
+        cwdsDAO.listJobsV1(workspace.workspace.workspaceId, runningOnly)(userInfo)
+      } else {
+        List.empty
+      }
     } yield {
       // merge Import Service and cWDS results
       importServiceJobs.concat(cwdsJobs)
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
@@ -376,15 +376,15 @@ class EntityService(rawlsDAO: RawlsDAO, importServiceDAO: ImportServiceDAO, cwds
   def listJobs(workspaceNamespace: String, workspaceName: String, runningOnly: Boolean, userInfo: UserInfo): Future[List[ImportServiceListResponse]] = {
 
     for {
-      // translate workspace namespace/name to workspaceid
-      workspace <- rawlsDAO.getWorkspace(workspaceNamespace, workspaceName)(userInfo)
       // get jobs from Import Service
       importServiceJobs <- importServiceDAO.listJobs(workspaceNamespace, workspaceName, runningOnly)(userInfo)
       // get jobs from cWDS
-      cwdsJobs = if (cwdsDAO.isEnabled) {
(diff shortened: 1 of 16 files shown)


src/test/scala/org/broadinstitute/dsde/firecloud/EntityServiceSpec.scala
```diff
@@ -214,10 +217,72 @@ class EntityServiceSpec extends BaseServiceSpec with BeforeAndAfterEach {
     }
   }
 
+  "EntityService.listJobs" - {
+    "should concatenate results from cWDS and Import Service" in {
+      val importServiceResponse = List(
+        ImportServiceListResponse("jobId1", "status1", "filetype1", None),
+        ImportServiceListResponse("jobId2", "status2", "filetype2", None)
+      )
```
calypsomatic, line 225: It weirds me out that these two list have identical jobIds.  Not clear how/if that will impact the test

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpCwdsDAO.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  private def toImportServiceListResponse(cwdsJob: GenericJob) = {
+    ImportServiceListResponse(jobId = cwdsJob.getJobId.toString,
+      status = toImportServiceStatus(cwdsJob.getStatus),
+      filetype = cwdsJob.getJobType.getValue,
+      message = Option(cwdsJob.getErrorMessage))
+  }
+
+  private def toImportServiceStatus(cwdsStatus: GenericJob.StatusEnum): String = {
+    STATUS_TRANSLATION.getOrElse(cwdsStatus, "blah")
```
jladieu: Maybe this could throw an error if there's no translation?

src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
```diff
@@ -374,11 +374,18 @@ class EntityService(rawlsDAO: RawlsDAO, importServiceDAO: ImportServiceDAO, goog
...
+
+    for {
+      // translate workspace namespace/name to workspaceid
+      workspace <- rawlsDAO.getWorkspace(workspaceNamespace, workspaceName)(userInfo)
+      // get jobs from Import Service
+      importServiceJobs <- importServiceDAO.listJobs(workspaceNamespace, workspaceName, runningOnly)(userInfo)
+      // get jobs from cWDS
+      cwdsJobs = cwdsDAO.listJobsV1(workspace.workspace.workspaceId, runningOnly)(userInfo)
+    } yield {
+      // merge Import Service and cWDS results
```
jladieu, line 390: not for this PR, but maybe for the metrics theme: this might be a good spot to add some metric instrumentation to understand migration status:

- measure latency of each dispatch for comparison
- measure the # of jobs returned from each for comparison

jladieu (approved):
Looks good to me, all comments are optional/future.

src/test/scala/org/broadinstitute/dsde/firecloud/EntityServiceSpec.scala
davidangb: copy/paste error, fixed!

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpCwdsDAO.scala
davidangb: I added a code comment; I would prefer to not fail, but now this returns "Unknown" instead of the draft-for-early-review "blah"

jladieu (approved):

calypsomatic (approved):

PR #1310, opened 2024-03-14
AJ-1602: create imports in cWDS

davidangb:
requires https://github.com/broadinstitute/terra-helmfile/pull/5270

When creating a new data import:
* check to see if cWDS is enabled in this environment
* check to see which import formats (e.g. pfb, tdrexport) cWDS supports in this environment
* if cWDS is enabled and supports the format, send the data import to cWDS. Else, send it to Import Service.
* unit tests

---

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
@@ -122,8 +122,9 @@ object FireCloudConfig {
 
   object Cwds {
     private val cwds = config.getConfig("cwds")
-    val baseUrl = cwds.getString("baseUrl")
-    val enabled = cwds.getBoolean("enabled")
+    val baseUrl: String = cwds.getString("baseUrl")
+    val enabled: Boolean = cwds.getBoolean("enabled")
+    val supportedFormats: List[String] = cwds.getStringList("supportedFormats").asScala.toList
   }
 
   object FireCloud {
@@ -212,6 +213,7 @@ object FireCloudConfig {
   object ImportService {
     lazy val server: String = config.getString("importService.server")
     lazy val bucket: String = config.getString("importService.bucketName")
+    lazy val enabled: Boolean = config.optionalBoolean("importService.enabled").getOrElse(true)
   }
 
   implicit class RichConfig(val config: Config) {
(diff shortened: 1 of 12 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
```diff
@@ -212,6 +213,7 @@ object FireCloudConfig {
   object ImportService {
     lazy val server: String = config.getString("importService.server")
     lazy val bucket: String = config.getString("importService.bucketName")
+    lazy val enabled: Boolean = config.optionalBoolean("importService.enabled").getOrElse(true)
```
davidangb, line 216: not used in this PR. I'm envisioning at a future point we'll want to turn off Import Service in code but leave it running for easy rollback.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpCwdsDAO.scala
```diff
@@ -57,6 +60,35 @@ class HttpCwdsDAO(enabled: Boolean) extends CwdsDAO {
...
+      }
+    }
+
+    importRequest
+  }
+
+  protected[dataaccess] def toCwdsImportType(input: String): ImportRequest.TypeEnum = {
+    TYPE_TRANSLATION.getOrElse(input,
+      throw new FireCloudException("Import type unknown; possible values are: " + TYPE_TRANSLATION.keys.mkString))
+  }
```
davidangb, line 90: this throws an exception if it can't translate, as compared to `toImportServiceStatus` which returns a default value. The difference here is that cWDS only supports a known set of `TypeEnum` values and if we send a default, it'll cause an error downstream; the status value from `toImportServiceStatus` is mostly informational.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpCwdsDAO.scala
```diff
@@ -69,5 +101,14 @@ class HttpCwdsDAO(enabled: Boolean) extends CwdsDAO {
     STATUS_TRANSLATION.getOrElse(cwdsStatus, "Unknown")
   }
 
+  private def getApiClient(accessToken: String): ApiClient = {
+    // prepare the cWDS client
+    val apiClient: ApiClient = new ApiClient()
+    apiClient.setHttpClient(commonHttpClient)
+    apiClient.setBasePath(FireCloudConfig.Cwds.baseUrl)
+    apiClient.setAccessToken(accessToken)
```
jladieu: Just wanna double check...is this the right token?

jladieu (approved):
One comment inline about access token..but otherwise LGTM (scala: confirmed)

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpCwdsDAO.scala
davidangb: how many times can I make the same mistake?!??!?! ARRGARGAF

calypsomatic (approved):

PR #1305, opened 2024-03-11
AJ-1602: list-jobs API no longer a passthrough to Import Service

davidangb:
AJ-1602

The eventual goal is to query both cWDS and Import Service, ask each about their running jobs, and aggregate the results into a single response.

This PR is prep for that. Previously, the call to Import Service was a pure passthrough; Orch was blind to Import Service's results and just passed on the response. Now, in this PR, Orch can deserialize the Import Service results and is fully aware of them. We still pass on the results unchanged, but this is a necessary step for when we'll later need to aggregate results.

----

Have you read [CONTRIBUTING.md](../CONTRIBUTING.md) lately? If not, do that first.

I, the developer opening this PR, do solemnly pinky swear that:

- [x] I've followed [the instructions](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/CONTRIBUTING.md#api-changes) if I've made any changes to the API, _especially_ if they're breaking changes
- [x] I've updated the RC_XXX release ticket with any manual steps required to release this change
- [x] I've updated the [FISMA documentation](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/CONTRIBUTING.md#fisma-documentation-changes) if I've made any security-related changes, including auth, encryption, or auditing

In all cases:

- [ ] Get two thumbsworth of review and PO signoff if necessary
- [ ] Verify all tests go green
- [ ] Squash and merge; you can delete your branch after this **unless it's for a hotfix**. In that case, don't delete it!
- [ ] Test this change deployed correctly and works on dev environment after deployment

PR #1314, opened 2024-03-18
AJ-1602: getJob checks both cWDS and Import Service

davidangb:
AJ-1602

When asking for the status of a given import job:
* if cWDS is enabled, ask cWDS for that job
* if found in cWDS, return the job
* if cWDS failed to return the job, or was disabled, ask Import Service for that job and return whatever Import Service returns, even if it's an error/not found

At the implementation level, this required changing the get-job API from a pure passthrough to a full implementation. This is similar to what we did in #1305.

---

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
@@ -9,7 +9,7 @@ import akka.http.scaladsl.model.{HttpResponse, StatusCodes}
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.firecloud.EntityService._
 import org.broadinstitute.dsde.firecloud.FireCloudConfig.Rawls
-import org.broadinstitute.dsde.firecloud.dataaccess.ImportServiceFiletypes.{FILETYPE_PFB, FILETYPE_RAWLS}
+import org.broadinstitute.dsde.firecloud.dataaccess.ImportServiceFiletypes.FILETYPE_RAWLS
 import org.broadinstitute.dsde.firecloud.dataaccess.{CwdsDAO, GoogleServicesDAO, ImportServiceDAO, RawlsDAO}
 import org.broadinstitute.dsde.firecloud.model.ModelJsonProtocol._
 import org.broadinstitute.dsde.firecloud.model.{ModelSchema, _}
@@ -28,7 +28,6 @@ import scala.jdk.CollectionConverters._
 import scala.concurrent.{ExecutionContext, Future}
 import scala.io.Source
 import scala.language.postfixOps
-import scala.sys.process._
 import scala.util.{Failure, Success, Try}
 
 object EntityService {
@@ -406,6 +405,31 @@ class EntityService(rawlsDAO: RawlsDAO, importServiceDAO: ImportServiceDAO, cwds
     } yield {
       // merge Import Service and cWDS results
       importServiceJobs.concat(cwdsJobs)
+    }
+  }
+
+  def getJob(workspaceNamespace: String, workspaceName: String, jobId: String, userInfo: UserInfo): Future[ImportServiceListResponse] = {
+    // there is no way to tell just from a jobId if the job exists in cWDS or Import Service. So, we have to try both.
+
+    // if cWDS is enabled, query cWDS for job
+    val cwdsFuture: Future[ImportServiceListResponse] = if (cwdsDAO.isEnabled) {
+      rawlsDAO.getWorkspace(workspaceNamespace, workspaceName)(userInfo) map { workspace =>
+        cwdsDAO.getJobV1(workspace.workspace.workspaceId, jobId)(userInfo)
+      }
+    } else {
+      Future.failed(new FireCloudExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, s"Import $jobId not found")))
+    }
+
+    // if cWDS found the job, return it; else, try Import Service
+    cwdsFuture map { cwdsResponse =>
+      logger.info(s"Found job $jobId in cWDS")
+      cwdsResponse
+    } recoverWith {
+      case _ =>
+        importServiceDAO.getJob(workspaceNamespace, workspaceName, jobId)(userInfo) map { importServiceResponse =>
+          logger.info(s"Found job $jobId in Import Service")
+          importServiceResponse
+      }
     }
   }
 
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
@@ -426,7 +426,7 @@ class EntityService(rawlsDAO: RawlsDAO, importServiceDAO: ImportServiceDAO, cwds
       logger.info(s"Found job $jobId in Import Service")
       importServiceResponse
     } recover { importServiceError =>
-      logger.info(s"Job $jobId not found in either cWDS or Import Service: " + importServiceError.getClass.getName)
+      logger.info(s"Job $jobId not returned successfully by either cWDS or Import Service")
       importServiceError match {
         case fex: FireCloudExceptionWithErrorReport => throw fex
         case   t => throw new FireCloudExceptionWithErrorReport(ErrorReport(StatusCodes.InternalServerError, t))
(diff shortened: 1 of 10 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
```diff
@@ -409,6 +408,37 @@ class EntityService(rawlsDAO: RawlsDAO, importServiceDAO: ImportServiceDAO, cwds
...
+    val importServiceFuture: () => Future[ImportServiceListResponse] = () => importServiceDAO.getJob(workspaceNamespace, workspaceName, jobId)(userInfo) map { importServiceResponse =>
+      logger.info(s"Found job $jobId in Import Service")
+      importServiceResponse
+    } recover { importServiceError =>
+      logger.info(s"Job $jobId not found in either cWDS or Import Service: " + importServiceError.getClass.getName)
+      importServiceError match {
+        case fex: FireCloudExceptionWithErrorReport => throw fex
+        case   t => throw new FireCloudExceptionWithErrorReport(ErrorReport(StatusCodes.InternalServerError, t))
+      }
+    }
```
davidangb, line 438: note that by making this a function `() => Future[...]`, we don't invoke anything until/unless we call it with `.apply()` (or just `()`)

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/CwdsDAO.scala
```diff
@@ -9,10 +10,17 @@ trait CwdsDAO {
...
   def listJobsV1(workspaceId: String,
                  runningOnly: Boolean
                 )(implicit userInfo: UserInfo): List[ImportServiceListResponse]
 
+  @throws(classOf[ApiException])
+  def getJobV1(workspaceId: String,
+               jobId: String
+              )(implicit userInfo: UserInfo): ImportServiceListResponse
+
+  @throws(classOf[ApiException])
```
davidangb, line 23: Scala is just fine without the `@throws` annotation, but Mockito needs them to throw checked exceptions out of mocks.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpImportServiceDAO.scala
```diff
@@ -39,6 +38,22 @@ class HttpImportServiceDAO(implicit val system: ActorSystem, implicit val materi
...
+    userAuthedRequest(Get(importServiceUrl))(userInfo) flatMap { importServiceResponse =>
+      importServiceResponse.status match {
+        case StatusCodes.OK => Unmarshal(importServiceResponse).to[ImportServiceListResponse]
+        case _ =>
+          extractResponseBody(importServiceResponse.entity) flatMap { respBody =>
+            Future.failed(new FireCloudExceptionWithErrorReport(ErrorReport(importServiceResponse.status, respBody)))
+          }
+      }
+    }
+  }
```
davidangb, line 55: the `case _` is necessary here because the call to Import Service via `userAuthedRequest` is successful even when Import Service returns a 404 or 500 or whatever … yes, we successfully got a response from Import Service; that response happens to be an error.

src/test/scala/org/broadinstitute/dsde/firecloud/EntityServiceSpec.scala
```diff
@@ -26,6 +29,9 @@ import scala.concurrent.ExecutionContext.Implicits._
 import scala.concurrent.Future
 
 class EntityServiceSpec extends BaseServiceSpec with BeforeAndAfterEach {
+
+  implicit val errorReportSource: ErrorReportSource = ErrorReportSource("EntityServiceSpec")
+
```
davidangb, line 34: necessary to serialize/work with `FireCloudExceptionWithErrorReport` exceptions

src/test/scala/org/broadinstitute/dsde/firecloud/EntityServiceSpec.scala
```diff
@@ -387,6 +393,196 @@ class EntityServiceSpec extends BaseServiceSpec with BeforeAndAfterEach {
...
+      // get job via entity service
+      val actual = entityService.getJob("workspaceNamespace", "workspaceName", jobId, dummyUserInfo("mytoken")).futureValue
+
+      actual shouldBe importServiceResponse
+      verify(cwdsDAO, never).getJobV1(any[String], any[String])(any[UserInfo])
+      verify(mockRawlsDAO, never).getWorkspace(any[String], any[String])(any[UserInfo])
+    }
+  }
+
```
davidangb, line 585: I tried extracting the implementation for these tests into a reusable function and thus avoiding all the copy/paste. There's enough variation in the tests that the reusable function became quite heavy with business logic and conditional branching; I reverted and left the tests fully written out like this.

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceApiServiceSpec.scala
```diff
@@ -1159,45 +1159,6 @@ class WorkspaceApiServiceSpec extends BaseServiceSpec with WorkspaceApiService w
...
-        s"Passthrough should not pass unrecognized HTTP verbs for $pathUnderTest" in {
-          (Delete(s"$pathUnderTest/dummyJobId")
-            ~> dummyUserIdHeaders(dummyUserId)
-            ~> sealRoute(workspaceRoutes)) ~> check {
-            status should equal(MethodNotAllowed)
-          }
-        }
-      }
-    }
-
```
davidangb, line 1200: these tests are no longer valid since the API is no longer a passthrough.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
```diff
@@ -409,6 +408,37 @@ class EntityService(rawlsDAO: RawlsDAO, importServiceDAO: ImportServiceDAO, cwds
...
+      }
+    } else {
+      Future.failed(new FireCloudExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, s"Import $jobId not found")))
+    }
+
+    val importServiceFuture: () => Future[ImportServiceListResponse] = () => importServiceDAO.getJob(workspaceNamespace, workspaceName, jobId)(userInfo) map { importServiceResponse =>
+      logger.info(s"Found job $jobId in Import Service")
+      importServiceResponse
+    } recover { importServiceError =>
+      logger.info(s"Job $jobId not found in either cWDS or Import Service: " + importServiceError.getClass.getName)
```
nawatts: Is this always accurate? If the error is something other than 404, should this log "not found"?

nawatts (approved):

src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
```diff
@@ -409,6 +408,37 @@ class EntityService(rawlsDAO: RawlsDAO, importServiceDAO: ImportServiceDAO, cwds
...
+
+    // if cWDS is enabled, query cWDS for job
+    val cwdsFuture: Future[ImportServiceListResponse] = if (cwdsDAO.isEnabled) {
+      rawlsDAO.getWorkspace(workspaceNamespace, workspaceName)(userInfo) map { workspace =>
+        val cwdsResponse = cwdsDAO.getJobV1(workspace.workspace.workspaceId, jobId)(userInfo)
+        logger.info(s"Found job $jobId in cWDS")
+        cwdsResponse
+      }
+    } else {
+      Future.failed(new FireCloudExceptionWithErrorReport(ErrorReport(StatusCodes.NotFound, s"Import $jobId not found")))
```
calypsomatic: I'm a bit confused about the exact logic going on here - it seems like we create a `FireCloudExceptionWithErrorReport` if cWDS is not enabled - does that also happen if getJobV1 throws a 404?

src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
davidangb: good catch, I have updated the log statement to be (hopefully) more generic.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
davidangb: I added a code comment to explain, and changed the thrown exception to be the same one that cWDS would have thrown. I don't think it makes a functional difference since the `recoverWith` (Scala version of `catch`) handles all exceptions, but I agree it's more accurate/understandable.

calypsomatic (approved):

PR #1321, opened 2024-04-03
Azure control plane

dvoet:
https://broadworkbench.atlassian.net/browse/TOAZ-345

When orch needs to run without various services, e.g. Agora, the associated config is replaced with `enabled=false` and the associated DAO, e.g. `AgoraDAO`, is replaced with a dummy implementation (a proxy) that throws an error for every method call (similar to rawls). See the ticket for list of services that can be disabled. Disabled services should not be checked by the health check.

Much of orch's config was loaded at startup time into `val`s. Any non-optional config for disableable services were switched to `lazy`.

In doing this I discovered that the error handling on startup was faulty, exceptions did not cause the JVM to terminate and were lost unless explicitly handled. Fixed.

In all cases:

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/firecloud/webservice/StatusApiServiceSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/firecloud/webservice/StatusApiServiceSpec.scala
@@ -5,9 +5,10 @@ import org.broadinstitute.dsde.firecloud.HealthChecks
 import org.broadinstitute.dsde.firecloud.service.{BaseServiceSpec, StatusService}
 import org.broadinstitute.dsde.workbench.util.health.StatusJsonSupport.StatusCheckResponseFormat
 import org.broadinstitute.dsde.workbench.util.health.Subsystems._
-import org.broadinstitute.dsde.workbench.util.health.{HealthMonitor, StatusCheckResponse}
+import org.broadinstitute.dsde.workbench.util.health.{HealthMonitor, StatusCheckResponse, Subsystems}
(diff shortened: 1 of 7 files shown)


src/test/scala/org/broadinstitute/dsde/firecloud/webservice/StatusApiServiceSpec.scala
```diff
@@ -61,8 +62,12 @@ class StatusApiServiceSpec extends BaseServiceSpec with StatusApiService with Sp
...
       Get(statusPath) ~> statusRoutes ~> check {
         val statusCheckResponse = responseAs[StatusCheckResponse]
-        val expectedSystems = Set(Agora, GoogleBuckets, LibraryIndex, OntologyIndex, Rawls, Sam, Thurloe)
-        assertResult(expectedSystems) { statusCheckResponse.systems.keySet }
+        val expectedSystems = Set(
+          Agora, Rawls, Sam, Thurloe,
+          Subsystems.withName(GoogleServicesDAO.serviceName),
+          Subsystems.withName(SearchDAO.serviceName),
+          Subsystems.withName(OntologyDAO.serviceName))
+          assertResult(expectedSystems) { statusCheckResponse.systems.keySet }
```
jladieu: nit: un-indent?

jladieu (approved):

nawatts (approved):

nawatts:
Might be worth a comment in FirecloudConfig explaining why some values are lazy? Without the PR description, there doesn't seem to be a clear pattern.

PR #1323, opened 2024-04-04
disable anonymized email when google is disabled

dvoet:
https://broadworkbench.atlassian.net/browse/TOAZ-345

missed this in the first PR, only generate an anonymous email group when google is enabled

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
@@ -129,13 +129,12 @@ object FireCloudConfig {
 
   object FireCloud {
     private val firecloud = config.getConfig("firecloud")
-    val baseUrl = firecloud.getString("baseUrl")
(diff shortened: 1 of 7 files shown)


nawatts (approved):

jladieu (approved):

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
```diff
@@ -129,13 +129,12 @@ object FireCloudConfig {
 
   object FireCloud {
     private val firecloud = config.getConfig("firecloud")
-    val baseUrl = firecloud.getString("baseUrl")
```
davidangb, line 132: Was this just extraneous?

davidangb (approved):
One question inline

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
dvoet: it wasn't in use so I guess so

PR #1342, opened 2024-04-26
AJ-1790 cwds async

calypsomatic:
https://broadworkbench.atlassian.net/browse/AJ-1790

Orch should now route async tsv uploads to cWDS, if cWDS is enabled and supports the `rawlsjson` file type.  
Adds an `isUpsert` flag to the import request options to distinguish `update` type uploads. Although these are not available from the UI, it is apparently possible through the API and we need to maintain parity with import service, which respects these differences.

Requires https://github.com/broadinstitute/terra-helmfile/pull/5467 to come into effect.

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
@@ -227,8 +227,7 @@ class EntityService(rawlsDAO: RawlsDAO, importServiceDAO: ImportServiceDAO, cwds
     val insertedObject = googleServicesDAO.writeObjectAsRawlsSA(bucketToWrite, fileToWrite, dataBytes)
     val gcsPath = s"gs://${insertedObject.bucketName.value}/${insertedObject.objectName.value}"
 
-    val importRequest = AsyncImportRequest(gcsPath, FILETYPE_RAWLS)
-//    val importRequest = AsyncImportRequest(gcsPath, FILETYPE_RAWLS, Some(ImportOptions(None, Some(isUpsert))))
+    val importRequest = AsyncImportRequest(gcsPath, FILETYPE_RAWLS, Some(ImportOptions(None, Some(isUpsert))))
 
     if (cwdsDAO.isEnabled && cwdsDAO.getSupportedFormats.contains(importRequest.filetype.toLowerCase)) {
       // translate the workspace namespace/name into an id
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
@@ -228,7 +228,22 @@ class EntityService(rawlsDAO: RawlsDAO, importServiceDAO: ImportServiceDAO, cwds
     val gcsPath = s"gs://${insertedObject.bucketName.value}/${insertedObject.objectName.value}"
 
     val importRequest = AsyncImportRequest(gcsPath, FILETYPE_RAWLS)
-    importServiceDAO.importJob(workspaceNamespace, workspaceName, importRequest, isUpsert)(userInfo)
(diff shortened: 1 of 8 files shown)


src/test/scala/org/broadinstitute/dsde/firecloud/EntityServiceSpec.scala
```diff
@@ -586,7 +622,7 @@ class EntityServiceSpec extends BaseServiceSpec with BeforeAndAfterEach {
   private def getEntityService(mockGoogleServicesDAO: MockGoogleServicesDAO = new MockGoogleServicesDAO,
                                mockImportServiceDAO: MockImportServiceDAO = new MockImportServiceDAO,
                                rawlsDAO: MockRawlsDAO = new MockRawlsDAO,
-                               cwdsDAO: MockCwdsDAO = new MockCwdsDAO) = {
+                               cwdsDAO: MockCwdsDAO = new MockCwdsDAO(false)) = {
```
calypsomatic, line 625: The tests that require `MockCwdsDAO` to be enabled specifically enable it, so by default it should be disabled.

src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
```diff
@@ -227,8 +227,22 @@ class EntityService(rawlsDAO: RawlsDAO, importServiceDAO: ImportServiceDAO, cwds
...
+      val cwdsJob = cwdsDAO.importV1(workspace.workspace.workspaceId, importRequest)(userInfo)
+        // massage the cWDS job into the response format Orch requires
+        val asyncImportResponse = AsyncImportResponse(url = importRequest.url,
+          jobId = cwdsJob.getJobId.toString,
+          workspace = WorkspaceName(workspaceNamespace, workspaceName))
+        RequestComplete(Accepted, asyncImportResponse)
+    }
+    } else {
+      importServiceDAO.importJob(workspaceNamespace, workspaceName, importRequest, isUpsert)(userInfo)
+    }
```
davidangb, line 245: nit: could use some formatting, the indentation is off here (IntelliJ does fine formatting Scala)

davidangb (approved):
this all looks correct to me, assuming it works as expected in your bee!

src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
```diff
@@ -227,8 +227,22 @@ class EntityService(rawlsDAO: RawlsDAO, importServiceDAO: ImportServiceDAO, cwds
     val insertedObject = googleServicesDAO.writeObjectAsRawlsSA(bucketToWrite, fileToWrite, dataBytes)
     val gcsPath = s"gs://${insertedObject.bucketName.value}/${insertedObject.objectName.value}"
 
-    val importRequest = AsyncImportRequest(gcsPath, FILETYPE_RAWLS)
-    importServiceDAO.importJob(workspaceNamespace, workspaceName, importRequest, isUpsert)(userInfo)
+    val importRequest = AsyncImportRequest(gcsPath, FILETYPE_RAWLS, Some(ImportOptions(None, Some(isUpsert))))
+
+    if (cwdsDAO.isEnabled && cwdsDAO.getSupportedFormats.contains(importRequest.filetype.toLowerCase)) {
```
nawatts, line 232: Is there a need to check here if Rawls JSON is supported? It's hard-coded to be.

```suggestion
    if (cwdsDAO.isEnabled) {
```

nawatts (approved):

src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
calypsomatic: Somewhere in there it's getting the info from orch's config, as I discovered by running it in my bee without updating the config.  If we leave this check in here, then I can merge this PR before updating the conf and before cWDS is actually able to take these types of requests without any problem.

nawatts: 🤦 I was looking at a test.

PR #1346, opened 2024-05-02
AJ-1790 Use cwds bucket

calypsomatic:
https://broadworkbench.atlassian.net/browse/AJ-1790
Orch should publish async imports to cWDS's bucket when importing with cWDS instead of import service's bucket.
Relies on https://github.com/broadinstitute/terra-helmfile/pull/5490

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
@@ -218,9 +218,11 @@ class EntityService(rawlsDAO: RawlsDAO, importServiceDAO: ImportServiceDAO, cwds
                           rawlsCalls: Seq[EntityUpdateDefinition], userInfo: UserInfo): Future[PerRequestMessage] = {
     import spray.json._
 
+    val useCWDS = cwdsDAO.isEnabled && cwdsDAO.getSupportedFormats.contains(FILETYPE_RAWLS)
+
     // generate unique name for the file-to-upload
     val fileToWrite = GcsObjectName(s"incoming/${java.util.UUID.randomUUID()}.json")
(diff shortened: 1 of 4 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
```diff
@@ -229,19 +231,24 @@ class EntityService(rawlsDAO: RawlsDAO, importServiceDAO: ImportServiceDAO, cwds
...
 
-    if (cwdsDAO.isEnabled && cwdsDAO.getSupportedFormats.contains(importRequest.filetype.toLowerCase)) {
+    if (useCWDS) {
+      importToCWDS(workspaceNamespace, workspaceName, userInfo, importRequest)
+    } else {
+      importServiceDAO.importJob(workspaceNamespace, workspaceName, importRequest, isUpsert)(userInfo)
+    }
+  }
+
+  private def importToCWDS(workspaceNamespace: String, workspaceName: String, userInfo: UserInfo, importRequest: AsyncImportRequest
```
calypsomatic, line 241: I added in this quick little refactor to avoid duplicated code (had meant to do it in a previous PR and forgot).  Please double-check that I haven't regressed anything here.

nawatts (approved):

davidangb (approved):
did not test it myself, but code looks good.

can you update the local-dev conf in this PR too, so running orch locally does the right thing? https://github.com/broadinstitute/firecloud-orchestration/blob/develop/local-dev/templates/firecloud-orchestration.conf.ctmpl

This relies on Orch's SA having permissions to write to cWDS's bucket. From looking in cloud console, I think this is ok but it will need testing.

jladieu (approved):
Nice!

PR #1347, opened 2024-05-03
AJ-1812 Improve bucket security for handoff to cwds

jladieu:
Include `workspaceId` in GCS path for cwds.  This will be used by cwds to guarantee that only `gs://` URIs pointing to the same `workspaceId` as requested can be used.  This prevents potential hijacking of other unauthorized workloads in the bucket.

- [ ] ~I've followed [the instructions](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/CONTRIBUTING.md#api-changes) if I've made any changes to the API, _especially_ if they're breaking changes~ (N/A)
- [ ] ~I've updated the RC_XXX release ticket with any manual steps required to release this change~ (N/A)
- [ ] ~I've updated the [FISMA documentation](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/CONTRIBUTING.md#fisma-documentation-changes) if I've made any security-related changes, including auth, encryption, or auditing~ (N/A)

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
@@ -227,14 +227,14 @@ class EntityService(rawlsDAO: RawlsDAO, importServiceDAO: ImportServiceDAO, cwds
         val bucketToWrite = GcsBucketName(FireCloudConfig.Cwds.bucket)
         val fileToWrite = GcsObjectName(s"to-cwds/${workspaceId}/${java.util.UUID.randomUUID()}.json")
         val gcsPath = writeDataToGcs(bucketToWrite, fileToWrite, dataBytes)
-        val importRequest = getImportRequest(gcsPath, isUpsert)
+        val importRequest = getRawlsJsonImportRequest(gcsPath, isUpsert)
         importToCWDS(workspaceNamespace, workspaceName, workspaceId, userInfo, importRequest)
       }
     } else {
       val bucketToWrite = GcsBucketName(FireCloudConfig.ImportService.bucket)
       val fileToWrite = GcsObjectName(s"incoming/${java.util.UUID.randomUUID()}.json")
       val gcsPath = writeDataToGcs(bucketToWrite, fileToWrite, dataBytes)
-      val importRequest = getImportRequest(gcsPath, isUpsert)
+      val importRequest = getRawlsJsonImportRequest(gcsPath, isUpsert)
       importServiceDAO.importJob(workspaceNamespace, workspaceName, importRequest, isUpsert)(userInfo)
     }
   }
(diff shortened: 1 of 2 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
```diff
@@ -220,36 +220,48 @@ class EntityService(rawlsDAO: RawlsDAO, importServiceDAO: ImportServiceDAO, cwds
...
+    // massage the cWDS job into the response format Orch requires
+    Future.successful(
+      RequestComplete(Accepted, AsyncImportResponse(url = importRequest.url,
         jobId = cwdsJob.getJobId.toString,
-        workspace = WorkspaceName(workspaceNamespace, workspaceName))
-      RequestComplete(Accepted, asyncImportResponse)
-    }
+        workspace = WorkspaceName(workspaceNamespace, workspaceName)))
+    )
   }
```
davidangb, line 264: now that `importToCWDS` no longer includes `getWorkspace`, there's no need for it to return a `Future`. It can be simplified by removing `Future[...]` from its return type and by removing `Future.successful(…)` from its implementation:
```scala
  private def importToCWDS(workspaceNamespace: String, workspaceName: String, workspaceId: String, userInfo: UserInfo, importRequest: AsyncImportRequest
                          ): PerRequestMessage = {
    // create the job in cWDS
    val cwdsJob = cwdsDAO.importV1(workspaceId, importRequest)(userInfo)
    // massage the cWDS job into the response format Orch requires
    RequestComplete(Accepted, AsyncImportResponse(url = importRequest.url,
      jobId = cwdsJob.getJobId.toString,
      workspace = WorkspaceName(workspaceNamespace, workspaceName)))
  }
```

then, in the two places it is called, you can change `getWorkspaceId … flatMap … importToCWDS` to `getWorkspaceId … map … importToCWDS`

davidangb (commented):
one comment inline about simplifying signatures, otherwise looks good!

davidangb (approved):

src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
jladieu: Nice!!!

src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
```diff
@@ -220,36 +220,47 @@ class EntityService(rawlsDAO: RawlsDAO, importServiceDAO: ImportServiceDAO, cwds
...
-        jobId = cwdsJob.getJobId.toString,
-        workspace = WorkspaceName(workspaceNamespace, workspaceName))
-      RequestComplete(Accepted, asyncImportResponse)
-    }
+  private def writeDataToGcs(bucket: GcsBucketName, objectName: GcsObjectName, data: Array[Byte]): String = {
+    val insertedObject = googleServicesDAO.writeObjectAsRawlsSA(bucket, objectName, data)
+    s"gs://${insertedObject.bucketName.value}/${insertedObject.objectName.value}"
+  }
+
+  private def getImportRequest(gcsPath: String, isUpsert: Boolean): AsyncImportRequest = {
```
nawatts: Since this class also deals with other import types, maybe this method could have a more specific name?

```suggestion
  private def getRawlsJsonImportRequest(gcsPath: String, isUpsert: Boolean): AsyncImportRequest = {
```

nawatts (approved):

src/main/scala/org/broadinstitute/dsde/firecloud/EntityService.scala
jladieu: Good call.  Pushed a commit to fix this.

jladieu:
Verified in bee, gonna merge.

jladieu:
Paired with https://github.com/DataBiosphere/terra-workspace-data-service/pull/775

PR #1355, opened 2024-05-10
AJ-1749 Remove import-service from firecloud-orchestration

jladieu:
Remove import service.  Once https://github.com/broadinstitute/terra-helmfile/pull/5556 is live, cWDS will effectively field 100% of import traffic and all import-service backend resources will be deactivated.

Process:

* Rename ImportServiceTypes -> LegacyServiceTypes (IDE refactor only)
* Move `LegacyFileTypes` to `CwdsDao` (Manual refactor)
* Rename ImportServiceDAO methods to match CwdsDao (IDE refactor only)
* Make `MockCwdsDao` like `MockImportServiceDao` (manual)
* Eliminate `importServiceDao.getJobV1` (manual)
* Eliminate `importServiceDao.listJobsV1` (manual)
* Remove `importV1` from ImportService: and then make all the tests
still pass for vaguely the same reasons as before, except using cWDS.
* Prune references to ImportService
* Make supportedFormats consistent

Related PRs:
* terra-helmfile: https://github.com/broadinstitute/terra-helmfile/pull/5538
* rawls: https://github.com/broadinstitute/rawls/pull/2866

I, the developer opening this PR, do solemnly pinky swear that:

- [x] ~I've followed [the instructions](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/CONTRIBUTING.md#api-changes) if I've made any changes to the API, _especially_ if they're breaking changes~ (N/A)
- [x] ~I've updated the RC_XXX release ticket with any manual steps required to release this change~ (N/A)
- [x] ~I've updated the [FISMA documentation](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/CONTRIBUTING.md#fisma-documentation-changes) if I've made any security-related changes, including auth, encryption, or auditing~ (N/A)

In all cases:

- [x] Get two thumbsworth of review and PO signoff if necessary
- [x] Verify all tests go green
- [x] Squash and merge; you can delete your branch after this **unless it's for a hotfix**. In that case, don't delete it!
- [x] Test this change deployed correctly and works on dev environment after deployment

PR #1361, opened 2024-05-16
AJ-1749: respect importService.enabled for listing jobs

davidangb:
Respect the `importServiceEnabled` [config setting](https://github.com/broadinstitute/terra-helmfile/blob/7a8fa345af1cba66e09b50117c37ec97aaab9f8d/values/app/firecloudorch/live/prod.yaml#L10) when listing jobs. This will allow us to disable import service and drain traffic from it, before committing fully to https://github.com/broadinstitute/firecloud-orchestration/pull/1355.

Tested in my BEE, and working as expected:
* This branch + Import Service pod turned off + `importServiceEnabled` conf flag `true`: errors on Data tab
* This branch + Import Service pod turned off + `importServiceEnabled` conf flag `false`: no errors

Requires https://github.com/broadinstitute/terra-helmfile/pull/5555 for full effect, but the two PRs can be merged independently.

---

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/Boot.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/Boot.scala
@@ -98,7 +98,7 @@ object Boot extends App with LazyLogging {
     // can be disabled
     val agoraDAO: AgoraDAO = whenEnabled[AgoraDAO](FireCloudConfig.Agora.enabled, new HttpAgoraDAO(FireCloudConfig.Agora))
     val googleServicesDAO: GoogleServicesDAO = whenEnabled[GoogleServicesDAO](FireCloudConfig.GoogleCloud.enabled, new HttpGoogleServicesDAO(FireCloudConfig.GoogleCloud.priceListUrl, GooglePriceList(GooglePrices(FireCloudConfig.GoogleCloud.defaultStoragePriceList, UsTieredPriceItem(FireCloudConfig.GoogleCloud.defaultEgressPriceList)), "v1", "1")))
-    val importServiceDAO: ImportServiceDAO = whenEnabled[ImportServiceDAO](FireCloudConfig.ImportService.enabled, new HttpImportServiceDAO)
+    val importServiceDAO: ImportServiceDAO = whenEnabled[ImportServiceDAO](FireCloudConfig.ImportService.enabled, new HttpImportServiceDAO(FireCloudConfig.ImportService.enabled))
     val shibbolethDAO: ShibbolethDAO = whenEnabled[ShibbolethDAO](FireCloudConfig.Shibboleth.enabled, new HttpShibbolethDAO)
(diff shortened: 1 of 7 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/Boot.scala
```diff
@@ -98,7 +98,7 @@ object Boot extends App with LazyLogging {
     // can be disabled
     val agoraDAO: AgoraDAO = whenEnabled[AgoraDAO](FireCloudConfig.Agora.enabled, new HttpAgoraDAO(FireCloudConfig.Agora))
     val googleServicesDAO: GoogleServicesDAO = whenEnabled[GoogleServicesDAO](FireCloudConfig.GoogleCloud.enabled, new HttpGoogleServicesDAO(FireCloudConfig.GoogleCloud.priceListUrl, GooglePriceList(GooglePrices(FireCloudConfig.GoogleCloud.defaultStoragePriceList, UsTieredPriceItem(FireCloudConfig.GoogleCloud.defaultEgressPriceList)), "v1", "1")))
-    val importServiceDAO: ImportServiceDAO = whenEnabled[ImportServiceDAO](FireCloudConfig.ImportService.enabled, new HttpImportServiceDAO)
+    val importServiceDAO: ImportServiceDAO = whenEnabled[ImportServiceDAO](FireCloudConfig.ImportService.enabled, new HttpImportServiceDAO(FireCloudConfig.ImportService.enabled))
```
nawatts, line 101: I don't think we can use `FireCloudConfig.ImportService.enabled` as an argument to `HttpImportServiceDAO`. If `FireCloudConfig.ImportService.enabled` is false, then `HttpImportServiceDAO` won't be used: `whenEnabled` will replace it with the result of `DisabledServiceFactory.newDisabledService`.

jladieu (approved):
Looks good to me, but Nick's point also makes sense...can you do this with another boolean?

(Pre-emptively approving)

src/main/scala/org/broadinstitute/dsde/firecloud/Boot.scala
davidangb: This is true, and it makes some of the code in this PR pointless. BUT, the end result is fine. `DisabledServiceFactory.newDisabledService` returns a proxy whose `isEnabled` method returns false and throws an error for other methods. All we care about is that `isEnabled` returns false, because that short-circuits everything else.

Unit tests pass, swat tests pass, my manual testing in a bee looks good

nawatts: Ok. I think I was getting this mixed up with Rawls, where ImportService is higher level and talks to both Import Service and CWDS, and thought this was going to end up disabling the entire import API.

nawatts (approved):

src/main/scala/org/broadinstitute/dsde/firecloud/Boot.scala
jladieu: ☝️

PR #1373, opened 2024-06-13
AJ-1854 add concurrency groups to GH workflows

calypsomatic:
https://broadworkbench.atlassian.net/browse/AJ-1854

Adds concurrency groups to 3 workflows that run for pull requests.  Runs them only on `pull_request`, not also on `push`.  The one that creates a bee already uses `always()` for bee-destroy so there shouldn't be any rogue bees left lying around - verified by canceling after the bee had been created and saw that bee-destroy was run.

In all cases:

### DIFF
--- a/.github/workflows/es-int-tests.yml
+++ b/.github/workflows/es-int-tests.yml
@@ -1,15 +1,12 @@
 name: Elasticsearch Integration Tests
 
 on:
-  push:
-    paths-ignore:
-      - 'README.md'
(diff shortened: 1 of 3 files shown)


davidangb (commented):
I think all three of these workflows should run on push to `develop`, which probably changes the invocation for the `group` value

nawatts (approved):

davidangb (approved):

PR #1422, opened 2024-09-03
[WOR-1808] Add workspace settings routes

trholdridge:
These changes update Orchestration to use the newly published rawls model which includes workspace settings, and add documentation for workspace settings to the Swagger. Also, there are changes made to support routing for `workspace/v2` endpoints in general.

---

In all cases:

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -5239,6 +5239,304 @@ paths:
           description: Internal Server Error
           content: {}
       x-passthrough: false
+  /api/workspaces/v2/{workspaceNamespace}/{workspaceName}:
+    delete:
+      tags:
+        - WorkspacesV2
+      summary: Delete workspace
+      description: Starts a job to delete a workspace and all of its contents.
+      operationId: deleteWorkspaceV2
+      parameters:
+        - $ref: '#/components/parameters/workspaceNamespaceParam'
+        - $ref: '#/components/parameters/workspaceNameParam'
+      responses:
+        202:
+          description: Delete request accepted. Workspace bucket will be deleted within
+            24 hours.
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/WorkspaceDeletionResult'
+        400:
+          description: Workspace is not in a deletable state.
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        403:
+          description: Insufficient permissions to delete workspace (must be owner)
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        409:
+          description: Workspace is already being deleted.
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        404:
+          description: Workspace not found
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/ErrorReport'
+        500:
+          $ref: '#/components/responses/RawlsInternalError'
+  /api/workspaces/v2/{workspaceNamespace}/{workspaceName}/clone:
+    post:
+      tags:
+        - WorkspacesV2
+      summary: Clone workspace
+      description: Clone a workspace, asynchronously if possible
+      operationId: cloneWorkspace
+      parameters:
+        - $ref: '#/components/parameters/workspaceNamespaceParam'
+        - $ref: '#/components/parameters/workspaceNameParam'
+      requestBody:
+        description: Request for new workspace
+        content:
+          'application/json':
+            schema:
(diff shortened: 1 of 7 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceV2ApiService.scala
```diff
@@ -0,0 +1,38 @@
...
+
+  val workspaceV2Routes: Route =
+    pathPrefix("api" / "workspaces" / "v2") {
+      pathPrefix(Segment / Segment) {
+        (workspaceNamespace, workspaceName) => {
+          val workspaceV2Path = encodeUri(rawlsWorkspacesV2Root + "/%s/%s".format(workspaceNamespace, workspaceName))
+          path("settings") {
+            get {
+              requireUserInfo() { _ =>
+                passthrough(workspaceV2Path + "/settings", HttpMethods.GET)
```
davidangb: consider using `streamingPassthrough` instead (it's generally more performant) - this would also allow you to just proxy the "settings" url for all methods instead of explicitly specifying GET and PUT.

aherbst-broad (approved):

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceV2ApiService.scala
```diff
@@ -0,0 +1,34 @@
...
+        (workspaceNamespace, workspaceName) => {
+          val workspaceV2Path = encodeUri(rawlsWorkspacesV2Root + "/%s/%s".format(workspaceNamespace, workspaceName))
+          path("settings") {
+            requireUserInfo() { _ =>
+              streamingPassthrough(workspaceV2Path + "/settings")
+            }
+          }
+        }
+      }
+    }
```
davidangb, line 16: I know you're still working on this PR, but what do you think about simplifying this to:
```scala
  val workspaceV2Routes: Route =
    pathPrefix("api" / "workspaces" / "v2") {
      streamingPassthrough(rawlsWorkspacesV2Root)
    }
```

in other words, any request to Orch starting with `/api/workspaces/v2` is proxied to the same request in Rawls.

This would expose a bunch (8?) v2 Rawls APIs which are not currently proxied through Orch - but are publicly accessible anyway. And, it is forward-compatible to support new v2 APIs as they are developed, and it keeps the proxying logic nice and simple.

If you don't want to expose those other APIs, I suggest at least removing the `requireUserInfo()` directive. Since Rawls enforces auth already, no need to enforce it here and then again at the Rawls layer.

src/test/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceV2ApiServiceSpec.scala
```diff
@@ -0,0 +1,88 @@
...
+      }
+
+      s"OK status is returned for HTTP PUT if Rawls returns it" in {
+        stubRawlsService(HttpMethods.PUT, settingsPath, OK)
+        Put(settingsPath) ~> dummyUserIdHeaders(dummyUserId) ~> sealRoute(workspaceV2Routes) ~> check {
+          status should equal(OK)
+        }
+      }
+    }
+  }
```
davidangb, line 21: I'm honestly not convinced these unit tests are worth it. These really seem to be testing two things:
* behavior of the `streamingPassthrough` directive, which is already tested in `StreamingPassthroughSpec`
* behavior of the third-party mockserver library and how well you configured mockserver in this unit test

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/WorkspaceV2ApiService.scala
trholdridge: I committed changes which gut the service and tests to match this setup. Just want to run it by you to make sure I got the idea right.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -5238,6 +5239,382 @@ paths:
...
+                $ref: '#/components/schemas/ErrorReport'
+        500:
+          $ref: '#/components/responses/RawlsInternalError'
+  /api/workspaces/v2/{workspaceNamespace}/{workspaceName}/clone:
+    post:
+      tags:
+        - WorkspacesV2
+      summary: Clone workspace
+      description: Clone a workspace, asynchronously if possible
+      operationId: cloneWorkspace
```
davidangb: nit: validation at https://editor.swagger.io/ shows that `cloneWorkspace` is a duplicate operationId, one of the two should be renamed

davidangb (approved):
looks great to me, thanks for dealing with all my requests! One nit inline, giving this a proactive :+1:.

I did not test it myself, assuming you did

marctalbott (approved):

PR #1441, opened 2024-09-30
AJ-59: save TSV to workspace bucket

davidangb:
https://broadworkbench.atlassian.net/browse/AJ-59

A new API which will save a TSV file to the current workspace's bucket.

The TSV file is (intentionally) the exact same format as the TSV file you would download via your browser. Inside the bucket, it saves to "/tsvexport/${entityType}/${entityType}-${timestamp}.tsv" (or .zip for set tables).

The API returns the path in the bucket to which it saved the TSV.

Here's a screenshot after calling the API twice for the "pedigree" table. The second invocation only downloaded a couple columns; that's why the sizes are so different:
Screenshot 30-09-2024 at 16 47

_see my comments inline for reviewing!_

---

In all cases:

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -3650,6 +3650,8 @@ paths:
         - Entities
       summary: |
         Write TSV file containing workspace entities of the specified type to the workspace bucket.
+      description: |
+        Requires write access to the workspace.
       operationId: bucketWriteEntitiesTSV
       parameters:
         - $ref: '#/components/parameters/workspaceNamespaceParam'
@@ -3680,7 +3682,6 @@ paths:
         414:
           description: |
             URI length exceeds the configured limit of 2048 characters.
-            Please use the POST endpoint when it's necessary to supply a large number of attribute names.
           content: {}
         500:
           description: Internal Server Error
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -3644,6 +3644,48 @@ paths:
           description: Internal Server Error
           content: {}
       x-passthrough: false
+  /api/workspaces/{workspaceNamespace}/{workspaceName}/entities/{entityType}/tsv/save:
+    post:
+      tags:
+        - Entities
+      summary: |
+        Write TSV file containing workspace entities of the specified type to the workspace bucket.
+      operationId: bucketWriteEntitiesTSV
+      parameters:
+        - $ref: '#/components/parameters/workspaceNamespaceParam'
+        - $ref: '#/components/parameters/workspaceNameParam'
+        - $ref: '#/components/parameters/entityTypeParam'
+        - name: attributeNames
+          in: query
+          description: comma separated list of ordered attribute names to be in downloaded
+            tsv
+          schema:
+            type: string
+        - name: model
+          in: query
+          description: firecloud (default) or flexible
+          schema:
+            type: string
+      responses:
+        200:
+          description: URL to saved file
+          content:
+            text/plain:
+              schema:
+                type: string
+                format: binary
+        404:
+          description: Workspace or entity type does not exist
+          content: {}
+        414:
+          description: |
+            URI length exceeds the configured limit of 2048 characters.
+            Please use the POST endpoint when it's necessary to supply a large number of attribute names.
+          content: {}
+        500:
+          description: Internal Server Error
+          content: {}
+      x-passthrough: false
   /api/workspaces/{workspaceNamespace}/{workspaceName}/entityQuery/{entityType}:
     get:
       tags:
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -3607,7 +3607,7 @@ paths:
       tags:
         - Entities
       summary: |
-        TSV file containing workspace entities of the specified type
+        Download a TSV file containing workspace entities
       operationId: downloadEntitiesTSV
       parameters:
         - $ref: '#/components/parameters/workspaceNamespaceParam'
@@ -3638,18 +3638,19 @@ paths:
         414:
           description: |
             URI length exceeds the configured limit of 2048 characters.
-            Please use the POST endpoint when it's necessary to supply a large number of attribute names.
+            When necessary to supply a large number of attribute names, please use one of:
+            * POST /api/workspaces/{workspaceNamespace}/{workspaceName}/entities/{entityType}/tsv
+            * POST /cookie-authed/workspaces/{workspaceNamespace}/{workspaceName}/entities/{entityType}/tsv
           content: {}
         500:
           description: Internal Server Error
           content: {}
       x-passthrough: false
-  /api/workspaces/{workspaceNamespace}/{workspaceName}/entities/{entityType}/tsv/save:
     post:
       tags:
         - Entities
       summary: |
-        Write TSV file containing workspace entities of the specified type to the workspace bucket.
+        Save TSV file containing workspace entities to the workspace bucket.
       description: |
         Requires write access to the workspace.
       operationId: bucketWriteEntitiesTSV
@@ -3657,17 +3658,21 @@ paths:
         - $ref: '#/components/parameters/workspaceNamespaceParam'
         - $ref: '#/components/parameters/workspaceNameParam'
         - $ref: '#/components/parameters/entityTypeParam'
-        - name: attributeNames
-          in: query
-          description: comma separated list of ordered attribute names to be in downloaded
-            tsv
-          schema:
-            type: string
-        - name: model
-          in: query
-          description: firecloud (default) or flexible
-          schema:
-            type: string
+      requestBody:
+        content:
+          application/x-www-form-urlencoded:
+            schema:
+              type: object
+              properties:
+                attributeNames:
+                  type: string
+                  description: comma separated list of ordered attribute names to
+                    be in downloaded tsv
+                model:
+                  type: string
+                  description: flexible or firecloud (firecloud model will be used
+                    by default)
+        required: true
       responses:
         200:
           description: URL to saved file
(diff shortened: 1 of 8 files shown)


src/main/resources/swagger/api-docs.yaml
```diff
@@ -3644,6 +3644,49 @@ paths:
...
+          description: Workspace or entity type does not exist
+          content: {}
+        414:
+          description: |
+            URI length exceeds the configured limit of 2048 characters.
+          content: {}
+        500:
+          description: Internal Server Error
+          content: {}
+      x-passthrough: false
```
davidangb, line 3689: This API signature is intentionally a copy of the existing download-TSV-to-browser API. I figured they should be as similar as possible.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/GoogleServicesDAO.scala
```diff
@@ -28,6 +29,7 @@ trait GoogleServicesDAO extends ReportsSubsystemStatus {
   val fetchPriceList: Future[GooglePriceList]
   
   def writeObjectAsRawlsSA(bucketName: GcsBucketName, objectKey: GcsObjectName, objectContents: Array[Byte]): GcsPath
+  def writeObjectAsRawlsSA(bucketName: GcsBucketName, objectKey: GcsObjectName, tempFile: File): GcsPath
```
davidangb, line 32: new variation of `writeObjectAsRawlsSA` which accepts a temp file pointer as input instead of a materialized byte array

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -147,25 +149,27 @@ class HttpGoogleServicesDAO(priceListUrl: String, defaultPriceList: GooglePriceL
...
-    implicit val logger: SelfAwareStructuredLogger[IO] = Slf4jLogger.getLogger[IO]
-
-    // create the storage service, using the Rawls SA credentials
-    // the Rawls SA json creds do not contain a project, so also specify the project explicitly
-    val storageResource = GoogleStorageService.resource(FireCloudConfig.Auth.rawlsSAJsonFile, Option.empty[Semaphore[IO]],
-       project = Some(GoogleProject(FireCloudConfig.FireCloud.serviceProject)))
+    // call the upload implementation
+    val dataStream: Stream[IO, Byte] = Stream.emits(objectContents).covary[IO]
+    streamUploadObject(getStorageResource, bucketName, objectKey, dataStream)
+  }
```
davidangb, line 155: I broke up the pre-existing `writeObjectAsRawlsSA` method to extract reusable parts, then invoked those reusable parts from the two `writeObjectAsRawlsSA` signatures.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -81,45 +76,84 @@ class ExportEntitiesByTypeActor(rawlsDAO: RawlsDAO,
...
+      } else {
+        (ContentType(MediaTypes.`text/tab-separated-values`, HttpCharsets.`UTF-8`), s"$entityType.tsv")
+      }
+      val disposition = `Content-Disposition`.apply(ContentDispositionTypes.attachment, Map("filename" -> fileName))
+      HttpResponse(entity = HttpEntity.fromFile(contentType, tempFile.toJava), headers = List(keepAlive, disposition))
+    }
+  }.recoverWith {
+    // Standard exceptions have to be handled as a completed request
+    case t: Throwable => handleStandardException(t)
+  }
```
davidangb, line 99: I extracted the reusable part - saving TSV to a temp file - of the old `streamEntities()` method into a new reusable `entitiesToTempFile()` method. Then, `streamEntities()` and the new `streamEntitiesToWorkspaceBucket()` both invoke the reusable part.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -81,45 +76,84 @@ class ExportEntitiesByTypeActor(rawlsDAO: RawlsDAO,
...
+    rawlsDAO.getWorkspace(workspaceNamespace, workspaceName)(userInfo) flatMap { workspaceResponse =>
+      val workspaceBucket = GcsBucketName(workspaceResponse.workspace.bucketName)
+
+      // verify write permissions
+      val isPermitted = workspaceResponse.accessLevel match {
+        case Some(accessLevel) => accessLevel >= WorkspaceAccessLevels.Write
+        case None => false
+      }
+      if (!isPermitted)
+        throw new FireCloudExceptionWithErrorReport(errorReport = ErrorReport(StatusCodes.Forbidden, s"You must have at least write access."))
```
davidangb, line 119: this permission implementation - checking what the workspace reports instead of querying Sam - is strange but it 1) saves an extra API call and 2) is consistent with how Orch checks permissions other places. I waffle on whether it is the right thing to do. But in the end, Orch isn't the permission enforcer; this check just exists to give a nicer error message in cases where the user only has read permissions.

src/test/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeServiceSpec.scala
```diff
@@ -1,24 +1,28 @@
...
+import org.broadinstitute.dsde.firecloud.model._
+import org.broadinstitute.dsde.firecloud.webservice.{CookieAuthedApiService, ExportEntitiesApiService}
+import org.broadinstitute.dsde.workbench.model.google.{GcsBucketName, GcsObjectName}
+import org.mockito.ArgumentMatchers.any
+import org.mockito.Mockito.{reset, spy, times, verify}
+import org.scalatest.BeforeAndAfterEach
 
-import scala.concurrent.{Await, ExecutionContext}
 import scala.concurrent.duration._
+import scala.concurrent.{Await, ExecutionContext}
```
davidangb, line 23: IntelliJ -> Optimize Imports had a lot to say!

src/main/resources/swagger/api-docs.yaml
```diff
@@ -3644,6 +3644,49 @@ paths:
           description: Internal Server Error
           content: {}
       x-passthrough: false
+  /api/workspaces/{workspaceNamespace}/{workspaceName}/entities/{entityType}/tsv/save:
+    post:
```
davidangb, line 3648: I don't like this URL signature … any suggestions for something better? The existing download-TSV API is `GET /api/workspaces/{workspaceNamespace}/{workspaceName}/entities/{entityType}/tsv`; I think the download-to-bucket API should be similar to that, but "…/tsv/save" feels hacky.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -3638,7 +3638,51 @@ paths:
         414:
           description: |
             URI length exceeds the configured limit of 2048 characters.
-            Please use the POST endpoint when it's necessary to supply a large number of attribute names.
+            When necessary to supply a large number of attribute names, please use one of:
+            * POST /api/workspaces/{workspaceNamespace}/{workspaceName}/entities/{entityType}/tsv
+            * POST /cookie-authed/workspaces/{workspaceNamespace}/{workspaceName}/entities/{entityType}/tsv
```
davidangb, line 3643: A few tweaks to the summary and description of the existing GET tsv endpoint. No functional changes; just clarifying the doc.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -3638,7 +3638,51 @@ paths:
...
+          content: {}
+        500:
+          description: Internal Server Error
+          content: {}
+      x-passthrough: false
+    post:
```
davidangb, line 3649: this `POST` endpoint is the new download-to-bucket endpoint. It lives at `/api/workspaces/{workspaceNamespace}/{workspaceName}/entities/{entityType}/tsv`. So, it's parallel to the existing download tsv API:
* GET: download to browser
* POST: save to bucket

src/main/resources/swagger/api-docs.yaml
```diff
@@ -3638,7 +3638,51 @@ paths:
...
+              properties:
+                attributeNames:
+                  type: string
+                  description: comma separated list of ordered attribute names to
+                    be in downloaded tsv
+                model:
+                  type: string
+                  description: flexible or firecloud (firecloud model will be used
+                    by default)
+        required: true
```
davidangb, line 3675: this request body mirrors the request body at `POST /cookie-authed/workspaces/{workspaceNamespace}/{workspaceName}/entities/{entityType}/tsv`

calypsomatic (approved):
Works for me!

src/main/resources/swagger/api-docs.yaml
kevinmarete: I like this simplified approach POST vs GET

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
kevinmarete: I agree, it's a simplified check, and since the workspace has already performed the necessary validations, there's no need for duplication.

src/main/resources/swagger/api-docs.yaml
```diff
@@ -3638,7 +3638,51 @@ paths:
...
+      responses:
+        200:
```
kevinmarete, line 3677: Should this be returning a 202 and returning a job ID since it was to be an async request?

kevinmarete (approved):
I Left a few comments and questions overall this looks good to me.

src/main/resources/swagger/api-docs.yaml
davidangb: ah. I admit I didn't follow the letter of the ticket, and left this as a synchronous API for simplicity. We could always add async functionality as a follow-on.

PR #1452, opened 2024-10-09
CORE-123: new file-pairing API for data uploader

davidangb:
Introduces a new API at
```
POST /api/workspaces/{workspaceNamespace}/{workspaceName}/entities/{entityType}/paired-tsv
```

This API will:
1. List the files in the workspace's bucket, filtered to a given bucket prefix
2. Attempt to pair those files together based on [Illumina paired-end file naming conventions](https://support.illumina.com/help/BaseSpace_Sequence_Hub_OLH_009008_2/Source/Informatics/BS/NamingConvention_FASTQ-files-swBS.htm) as well as other well-known naming conventions supplied by Product
3. Generate and download a TSV containing the results of those file pairings

The driver use case for this API is the "Data Uploader" in Terra UI, though we may find that scripters/notebook users also want to use the API.

I have tested this running locally against ~100,000 files in a bucket, and the file-matching portion of the algorithm executes in < 2 seconds. The end result is a 30MB TSV so the API is slow overall, but the size is unavoidable at that scale.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
@@ -369,4 +369,15 @@ class HttpGoogleServicesDAO(priceListUrl: String, defaultPriceList: GooglePriceL
     getScopedServiceAccountCredentials(firecloudAdminSACreds, authScopes)
       .refreshAccessToken().getTokenValue
   }
+
+  override def listBucket(bucketName: GcsBucketName, prefix: Option[String]): List[GcsObjectName] = {
+    val listAttempt = getStorageResource.use { storageService =>
+      storageService.listObjectsWithPrefix(GcsBucketName(""), prefix.getOrElse(""), maxPageSize = 5000).compile.toList
+    }
+
+    // TODO: recurse if there are more files in the bucket and we need to paginate
+
+    // execute the upload
+    listAttempt.unsafeRunSync()
+  }
 }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
@@ -372,10 +372,11 @@ class HttpGoogleServicesDAO(priceListUrl: String, defaultPriceList: GooglePriceL
 
   override def listBucket(bucketName: GcsBucketName, prefix: Option[String]): List[GcsObjectName] = {
     val listAttempt = getStorageResource.use { storageService =>
-      storageService.listObjectsWithPrefix(GcsBucketName(""), prefix.getOrElse(""), maxPageSize = 5000).compile.toList
+      storageService.listObjectsWithPrefix(bucketName, prefix.getOrElse(""), maxPageSize = 5000, isRecursive = true).compile.toList
     }
 
-    // TODO: recurse if there are more files in the bucket and we need to paginate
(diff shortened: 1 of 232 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
```diff
@@ -400,4 +400,22 @@ class HttpGoogleServicesDAO(priceListUrl: String, defaultPriceList: GooglePriceL
...
+
+  override def listBucket(bucketName: GcsBucketName,
+                          prefix: Option[String],
+                          recursive: Boolean = true
+  ): List[GcsObjectName] = {
+    // listObjectsWithPrefix handles paginating through results if there are more results than
+    // the `maxPageSize` setting.
+    val listAttempt = getStorageResource.use { storageService =>
+      storageService
+        .listObjectsWithPrefix(bucketName, prefix.getOrElse(""), maxPageSize = 2000, isRecursive = recursive)
```
kevinmarete, line 412: Could we make `maxPageSize` a parameter in the API to support a paging solution using `limit` and `offset`? This would help manage cases where there are a large number of files in a bucket.

kevinmarete (approved):
This looks good to me. I just added a comment on implementing a paging solution to handle a large number of files.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpGoogleServicesDAO.scala
davidangb: The `listObjectsWithPrefix` in workbench-libs handles pagination internally; it will automatically make multiple requests to ensure it lists the entire bucket. The `maxPageSize` here controls how large the individual requests are, but it will always list the entire bucket. Therefore, `maxPageSize` is just a performance knob. With that given, do you still think it's necessary to expose it as an API parameter?

kevinmarete: Thanks for clarifying! Since `maxPageSize` only affects performance and not the final result, keeping it within the implementation may be better than exposing it as an API parameter. This way, we can adjust it internally for efficiency without adding complexity to the API.

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
```diff
@@ -369,4 +384,81 @@ class ExportEntitiesByTypeActor(rawlsDAO: RawlsDAO,
...
+    // retrieve workspace so we can get its bucket
+    rawlsDAO.getWorkspace(workspaceNamespace, workspaceName)(userInfo) map { workspaceResponse =>
+      val workspaceBucket = GcsBucketName(workspaceResponse.workspace.bucketName)
+
+      // list all files in bucket which match matchingOptions.prefix
+      logger.info("listing bucket files ...")
+      val fileList: List[GcsObjectName] =
+        googleServicesDao.listBucket(workspaceBucket, Option(matchingOptions.prefix), recursive)
+
+      logger.info(s"found ${fileList.length} files")
```
dvoet, line 424: do we want to place a limit on this based on the number of files returned?

dvoet (approved):

src/main/scala/org/broadinstitute/dsde/firecloud/service/ExportEntitiesByTypeActor.scala
davidangb: added in 9505b6e09871ef060f002b74a893e463a1ae54ec

dvoet (approved):

PR #1475, opened 2024-10-31
[CORE-134] Fix daylight savings time test bug

trholdridge:
Ticket: [https://broadworkbench.atlassian.net/browse/CORE-134](https://broadworkbench.atlassian.net/browse/CORE-134)

Previously, if the local timezone had an offset change in the next 30 days, certain tests would fail. I didn't figure out the full root cause, but it probably has to do with parts of the actual implementation using seconds (`NihLink(eraCommonsUsername, iat + 60 * 60 * 24 * 30)`) and the tests using days. Changing just `minusDays` to `minusSeconds` made it work, but I converted the `plusDays` as well for good measure.

The tests always passed in GH actions, so to replicate the problem and verify the solution (only within 30 days of a time changeover!):

```
git checkout 1646e20daaeca31c11e40fcc59e49986aa8833b;
TZ="America/New_York" sbt "testOnly *NihServiceUnitSpec";
git checkout 1e297af8e4fdc5b85f4faecab4f57fa43c1bdd81;
TZ="America/New_York" sbt "testOnly *NihServiceUnitSpec";
```

[https://xkcd.com/2867/](https://xkcd.com/2867/)

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/firecloud/service/NihServiceUnitSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/firecloud/service/NihServiceUnitSpec.scala
@@ -47,7 +47,7 @@ import java.nio.charset.StandardCharsets
 import java.security.{KeyPairGenerator, PrivateKey}
 import java.time.Instant
 import java.util.{Base64, UUID}
-import scala.concurrent.duration.Duration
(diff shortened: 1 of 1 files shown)


calypsomatic (approved):
I *think* this makes sense!

davidangb (approved):
nice, thank you!

PR #1482, opened 2024-11-05
[AN-242] Update MethodQuery schema in Swagger docs

sam-schu:
Jira ticket: https://broadworkbench.atlassian.net/browse/AN-242

## Context
The MethodQuery schema in Swagger is used only by a single endpoint: POST `/api/methods`. The description of this endpoint specifies that `namespace`, `name`, `payload`, and `entityType` must be included in the request body, which has been verified, but the MethodQuery schema does not list these properties as required. Also, the schema provides default values for all of its properties, none of which are actually used by Orchestration in the case that they are not provided.

## Changes
- Updated the MethodQuery schema to make the `namespace`, `name`, `payload`, and `entityType` properties required and to change all default property values to example property values.

## Testing
- Tested in a BEE that the Swagger page is updated.

## Visual aids
Before:
MethodQuery Swagger schema before image

After:
MethodQuery Swagger schema after image

In all cases:

- [x] Squash and merge. Make sure your branch deletes; GitHub should do this for you.

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -7759,24 +7759,35 @@ components:
         namespace:
           type: string
           description: Namespace which contains AgoraEntity.
+          example: YOUR_NAMESPACE
         name:
           type: string
           description: Name of the AgoraEntity.
+          example: BWA
         synopsis:
           type: string
           description: Synopsis which contains AgoraEntity.
+          example: Quickly aligns short nucleotide sequences.
         snapshotComment:
           type: string
           description: Snapshot comment of AgoraEntity
+          example: Improved spline reticulation
         documentation:
           type: string
           description: Documentation of the AgoraEntity.
+          example: |
+            BWA is a software package for mapping low-divergent sequences
+            against a large reference genome, such as the human genome.
+            It consists of three algorithms: BWA-backtrack, BWA-SW and BWA-MEM.
         payload:
           type: string
           description: Payload of method -- must be in WDL format
+          example: |
+            task wc {File in_file command { cat ${in_file} | wc -l } output { Int count = read_int(stdout()) }}
         entityType:
           type: string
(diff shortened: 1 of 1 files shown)


LizBaldo (commented):
Would it make sense to move what used to be the default into an example section in the description fields instead? It does look like useful information, but I agree that they are not defaults

aednichols (commented):
The defaults are pretty much already used like examples, I do find them useful in [this context](https://agora.dsde-prod.broadinstitute.org/#/Methods/post_api_v1_methods__namespace___name___snapshotId_).

Screenshot 2024-11-05 at 11 28 20

sam-schu:
@aednichols I removed the defaults in line with the [specification](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.1.md#properties), which states that "[t]he default value represents what would be assumed by the consumer of the input as the value of the schema if one is not provided." Do we often enough use defaults contrary to this definition to represent examples that are not actually used as defaults by the backend that I should leave them as is?

aednichols:
If there's a way to adopt the spec and preserve the example payloads, I'm on board with that.

aednichols:
It looks like Leonardo does something reasonable with the `example:` key.

 [Swagger link](https://leonardo.dsde-prod.broadinstitute.org/#/admin/updateApps)
[code link](https://github.com/DataBiosphere/leonardo/blob/823027a72300d8b911509783826df36eb2e109cf/http/src/main/resources/swagger/api-docs.yaml#L2961-L2969)

LizBaldo (approved):

sam-schu:
@LizBaldo @aednichols Please take a look at the updated PR comment and let me know what you think. I added the example property values directly to the schema, and I confirmed that this does populate the "Example Value" for the request body in the endpoint that uses it.

aednichols (approved):
Nice 👍 

Most likely, this Swagger was written before `example` existed. Good to get things up to date.

PR #1484, opened 2024-11-05
consolidate and simplify passthroughs

davidangb:
for discussion.

* Consolidate multiple `passthrough` routes into `streamingPassthrough`
  * I'd like to mark `passthrough` as deprecated; perhaps wait until we remove some more usages of it
* Consolidate multiple `*ApiService` route classes into a single passthrough-only class
  * Makes it easier to see routes at a glance.
  * Eventually, I suggest that all passthroughs exist in this single class, and the only routes left in other classes are non-passthroughs
* Remove unit tests for passthrough-only routes. The `streamingPassthrough` directive has its own unit tests, so there's not much to test here.
  * this change feels a little squirrely. Should we add back in a unit test to confirm we are routing passthroughs to the correct underlying service?

---

In all cases:


marctalbott (commented):
I'm a fan of this approach. Easier to add new downstream routes and non-passthrough routes are more obvious

kevinmarete (approved):
Nice, this simplifies things!

calypsomatic (approved):

PR #1486, opened 2024-11-07
[CORE-53] Refactor storage cost estimate to use Rawls

trholdridge:
Rawls now supports getting workspace bucket location, so Orchestration doesn't have to directly request this information from Google, which was causing incorrect behavior around requester pays. This change modifies the storage cost estimate API so that it uses Rawls to get bucket location (for rate calculations). It also adds an optional `userProject` query parameter so that a reader may use the API to get a storage cost estimate on a requester pays workspace if they provide a project to bill.

For UI users, this will fix the error in getting estimated cost on requester pays workspaces for writers and owners. It won't look any different for readers, because the UI hides estimated cost from readers in the first place.

-----

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/GoogleServicesDAO.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/GoogleServicesDAO.scala
@@ -4,7 +4,7 @@ import akka.http.scaladsl.model.HttpResponse
 import better.files.File
 import com.google.api.services.storage.model.Bucket
 import org.broadinstitute.dsde.firecloud.model.WithAccessToken
-import org.broadinstitute.dsde.rawls.model.ErrorReportSource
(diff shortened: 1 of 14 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/GoogleServicesDAO.scala
```diff
@@ -39,7 +38,5 @@ trait GoogleServicesDAO extends ReportsSubsystemStatus {
 
   def publishMessages(fullyQualifiedTopic: String, messages: Seq[String]): Future[Unit]
 
-  def getBucket(bucketName: String, petKey: String): Option[Bucket]
```
trholdridge, line 42: I can just leave this + its implementations if that is better. It is not used anywhere else in orch which is why I deleted it.

davidangb (approved):
I didn't test it myself, but this looks great - thank you for all the cleanup!

marctalbott (approved):

PR #1502, opened 2024-11-26
CORE-191: list/get/create import jobs works for requester-pays workspace

davidangb:
Creating, getting, and listing async data table import jobs did not work in requester-pays workspaces.

As part of these operations, Orchestration queries Rawls. It sends a workspace namespace+name, gets a response, and extracts the workspaceId from the response. However, the request it sent was for the entire workspace object, including bucket info. This would cause the request to fail for RP workspaces.

This PR introduces a new, optimized `getWorkspaceId` method for Orch to query Rawls. This request does not ask anything about buckets and therefore hits no problems in RP workspaces.

--- 

In all cases:

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockRawlsDAO.scala
+++ b/src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockRawlsDAO.scala
@@ -200,7 +200,7 @@ class MockRawlsDAO extends RawlsDAO {
   private val rawlsWorkspaceWithAttributes = WorkspaceDetails(
     "attributes",
     "att",
-    "id",
+    "00000000-0000-0000-0000-000000000000",
     "", // bucketname
     Some("wf-collection"),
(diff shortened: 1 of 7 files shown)


src/test/scala/org/broadinstitute/dsde/firecloud/dataaccess/MockRawlsDAO.scala
```diff
@@ -200,7 +200,7 @@ class MockRawlsDAO extends RawlsDAO {
   private val rawlsWorkspaceWithAttributes = WorkspaceDetails(
     "attributes",
     "att",
-    "id",
+    "00000000-0000-0000-0000-000000000000",
```
davidangb, line 203: I needed to update this example test data; `UUID.fromString("id")` would cause problems. This is more realistic test data anyway.

calypsomatic (approved):

dvoet (approved):

PR #1518, opened 2024-12-17
CORE-207: TSV download optimizations in makeRow()

davidangb:
This PR is an optimization found while researching CORE-207. This should be performance only; it has no functional impact to end users.

The `makeRow()` function is called, as its name implies, for each row of a TSV during download. Thus, it feels worth optimizing.

1. **optimize use of collections within `TSVFormatter.makeRow`**. Previously, we looped over all attributes in the entity, performed a Seq.indexOf() call to determine the column number that attribute should reside in, made a map of the column number->value, and then created a Seq of values based on the map's keys. Now, we loop over the headers in order and extract each column from the entity, still in order. This is less work and makes the `makeRow` function ~1.5x faster.
2. **use List instead of IndexedSeq when iterating**. We were using `IndexedSeq` all over the place. `IndexedSeq` is optimized more for random access, while `List` (which extends `LinearSeq`) is optimized more for sequential access. Our access was sequential. Changing this added another ~1.4x speedup, for a total of ~2.2x.

Raw numbers from the JMH benchmark:
  | ops/sec | error +/- | vs.baseline | avg ops/sec | avg vs. baseline
-- | -- | -- | -- | -- | --
baseline | 244,195.39 | 19,087.54 |   | 247,734.34 |  
  | 240,787.36 | 21,066.81 | 98.60% |   |  
  | 258,220.27 | 14,516.10 | 105.74% |   |  
Map->Seq for makerow | 388,638.98 | 48,430.52 | 159.15% | 398,863.48 | 161.00%
  | 375,149.40 | 43,361.68 | 153.63% |   |  
  | 325,224.09 | 45,584.79 | 133.18% |   |  
  | 315,577.37 | 51,806.07 | 129.23% |   |  
  | 410,244.36 | 38,258.30 | 168.00% |   |  
IndexedSeq -> List | 578,346.66 | 8,294.20 | 236.84% | 555,535.68 | 224.25%
  | 559,476.43 | 72,216.37 | 229.11% |   |  
  | 546,860.46 | 24,416.72 | 223.94% |   |  
  | 537,459.14 | 25,614.79 | 220.09% |   |  

---

In all cases:


benchmarks/src/main/scala/org/broadinstitute/dsde/firecloud/utils/TsvFormatterBenchmark.scala
```diff
@@ -1,31 +1,68 @@
...
-
-  @Benchmark
-  def tsvSafeStringWithTab(blackHole: Blackhole, inputs: Inputs): String = {
-    val result = TSVFormatter.tsvSafeString(inputs.inputWithTab)
+  def makeEntityRows(blackHole: Blackhole, entityData: EntityData): List[List[String]] = {
+    val result =
+      TSVFormatter.makeEntityRows(entityData.entityType, entityData.entities, entityData.headers)(entityData.model)
     blackHole.consume(result)
     result
   }
```
davidangb, line 68: I deleted the two example benchmarks and replaced with a benchmark that is actually useful for the current case.

dvoet (approved):

marctalbott (approved):

PR #1550, opened 2025-01-28
[CORE-253] add dbgap members to allowlist

calypsomatic:
https://broadworkbench.atlassian.net/browse/CORE-253

Relies on https://github.com/broadinstitute/terra-helmfile/pull/6021 .  (Tests are failing until that PR is merged) Now all items in an nih whitelist are required to have 4 elements: `fileName` and `rawlsGroup` as before, but additionally `phsId` and `consentGroup`.  When syncing whitelist users, orch will now look up the group associated with `phsId` and `consentGroup` and add the results in along with any users from the allowlist file.  

In all cases:

### DIFF
--- a/src/test/scala/org/broadinstitute/dsde/firecloud/service/NihServiceUnitSpec.scala
+++ b/src/test/scala/org/broadinstitute/dsde/firecloud/service/NihServiceUnitSpec.scala
@@ -76,8 +76,8 @@ class NihServiceUnitSpec extends AnyFlatSpec with Matchers with BeforeAndAfterEa
   val userTcgaAndTarget = genSamUser();
   val userTcgaOnly = genSamUser();
   val userTargetOnly = genSamUser();
-  val userDbGapList = genSamUser();
-  val userDbGapGroup = genSamUser();
+  val userDbGap = genSamUser();
+  val dbGapGroupEmail = WorkbenchEmail(UUID.randomUUID().toString + "@email.com")
 
   // DateTimes must be modified in seconds instead of days to match implementation
   val secondsIn30Days = 30.days.toSeconds.toInt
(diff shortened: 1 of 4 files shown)


src/test/scala/org/broadinstitute/dsde/firecloud/service/NihServiceUnitSpec.scala
```diff
@@ -206,7 +229,7 @@ class NihServiceUnitSpec extends AnyFlatSpec with Matchers with BeforeAndAfterEa
   }
 
   private def verifyTargetGroupSynced(): Unit = {
-    val emailsToSync = Set(WorkbenchEmail(userTcgaAndTarget.email.value), WorkbenchEmail(userTargetOnly.email.value))
+    val emailsToSync = Set(userTcgaAndTarget.email, userTargetOnly.email)
```
calypsomatic, line 220: Was there a reason to pull out the value of the email and then turn it back into a `WorkbenchEmail`?

src/test/scala/org/broadinstitute/dsde/firecloud/service/NihServiceUnitSpec.scala
```diff
@@ -102,14 +118,18 @@ class NihServiceUnitSpec extends AnyFlatSpec with Matchers with BeforeAndAfterEa
...
   )
 
   val linkedAccountsByExternalId = Map(
     userNoAllowlistsLinkedAccount.linkedExternalId -> userNoAllowlistsLinkedAccount,
     userTcgaAndTargetLinkedAccount.linkedExternalId -> userTcgaAndTargetLinkedAccount,
     userTcgaOnlyLinkedAccount.linkedExternalId -> userTcgaOnlyLinkedAccount,
-    userTargetOnlyLinkedAccount.linkedExternalId -> userTargetOnlyLinkedAccount
+    userTargetOnlyLinkedAccount.linkedExternalId -> userTargetOnlyLinkedAccount,
+    userDbGapListLinkedAccount.linkedExternalId -> userDbGapListLinkedAccount,
+    userDbGapGroupLinkedAccount.linkedExternalId -> userDbGapGroupLinkedAccount
```
calypsomatic: Instead of changing existing tests, I left them without any results from the search by consentGroup + phsId, and instead created a new test where we get one user from that and another user from the allowlist file for a new allowlist.  Not sure what kind of situations we should expect to exist or to test for, so I'd love some suggestions as to what combinations I should set up here for testing.

src/test/scala/org/broadinstitute/dsde/firecloud/service/NihServiceUnitSpec.scala
dvoet: probably not, it it was probably an instance of `RawlsEmail` or something once upon a time

src/test/scala/org/broadinstitute/dsde/firecloud/service/NihServiceUnitSpec.scala
```diff
@@ -272,6 +295,33 @@ class NihServiceUnitSpec extends AnyFlatSpec with Matchers with BeforeAndAfterEa
...
+    val emailsToSync = Set(userDbGapList.email, userDbGapGroup.email)
+    val nihStatus = Await
+      .result(nihService.syncAllowlistAllUsers("RAS"), Duration.Inf)
+      .asInstanceOf[PerRequest.RequestComplete[StatusCode]]
+      .response
+
+    nihStatus should be(StatusCodes.NoContent)
+    verify(googleDao, never()).getBucketObjectAsInputStream(FireCloudConfig.Nih.whitelistBucket, "tcga-whitelist.txt")
+    verify(googleDao, times(1))
+      .getBucketObjectAsInputStream(FireCloudConfig.Nih.whitelistBucket, "dbgap_phs002409_c1_whitelist.txt")
```
dvoet: do we need to verify these since this is not trying to test these conditions

src/test/scala/org/broadinstitute/dsde/firecloud/service/NihServiceUnitSpec.scala
```diff
@@ -272,6 +295,33 @@ class NihServiceUnitSpec extends AnyFlatSpec with Matchers with BeforeAndAfterEa
     verifyTargetGroupSynced()
   }
 
+  it should "sync all users by including groups found with consentGroup + phsId" in {
+    when(ecmDao.getActiveLinkedEraAccounts(ArgumentMatchers.eq(UserInfo(adminAccessToken, ""))))
+      .thenReturn(Future.successful(Seq(userDbGapListLinkedAccount)))
+    when(thurloeDao.getAllUserValuesForKey(any[String])).thenReturn(Future.successful(Map.empty))
+
+    val emailsToSync = Set(userDbGapList.email, userDbGapGroup.email)
```
dvoet: I don't understand what these 2 emails are. They seem to be users but that is not what I was expecting the test to be looking for. The test should be regardless of users. Even when there are no linked users, we should verify that the configured dbgap group gets added

calypsomatic: Ah, it needs to be the email of the *group*?   Apparently that went over my head and I was think it was *user* emails.  Should be mostly the same, just need to update some names in the test.

kevinmarete (approved):

dvoet (approved):

PR #1554, opened 2025-02-07
[CORE-175] Add getStorageCostEstimateV2

calypsomatic:
https://broadworkbench.atlassian.net/browse/CORE-175
Adds a new `getStorageCostEstimateV2` API and deprecates the original.  `getStorageCostEstimateV2` uses the `getBucketUsageV2` API added in https://github.com/broadinstitute/rawls/pull/3179 and returns both the cost estimate and the usage and bytes. 

In all cases:

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -46,6 +46,62 @@ userprofile {
 }
 
 googlecloud {
+  storagePriceList = {
+    "STANDARD": 0.02,
+    "NEARLINE": 0.01,
+    "COLDLINE": 0.004,
+    "ARCHIVE": 0.0012,
+    "REGIONAL": 0.02,
+    "MULTI_REGIONAL": 0.026
+  }
+
+  priceListUrl = "https://cloudpricingcalculator.appspot.com/static/data/pricelist.json"
+  # egress key previouly was "CP-COMPUTEENGINE-INTERNET-EGRESS-NA-NA"
+  priceListEgressKey = "CP-INTERNET-EGRESS-STANDARD-TIER-FROM-NA"
+  priceListStorageKey = "CP-BIGSTORE-STANDARD"
+
+  defaultStoragePriceList = {
+    "us-central1": 0.02,
+    "us-east1": 0.02,
+    "us-east4": 0.023,
+    "us-west4": 0.023,
+    "us-west1": 0.02,
+    "us-west2": 0.023,
+    "us-west3": 0.023,
+    "europe-west1": 0.02,
+    "europe-west2": 0.023,
+    "europe-west3": 0.023,
+    "europe-central2": 0.023,
+    "europe-west4": 0.020,
+    "europe-west6": 0.025,
+    "europe-north1": 0.020,
+    "northamerica-northeast1": 0.023,
+    "northamerica-northeast2": 0.023,
+    "asia-east1": 0.02,
+    "asia-east2": 0.023,
+    "asia-northeast1": 0.023,
+    "asia-southeast2": 0.023,
+    "asia-northeast2": 0.023,
+    "asia-northeast3": 0.023,
+    "asia-southeast1": 0.02,
+    "australia-southeast1": 0.023,
+    "australia-southeast2": 0.023,
+    "southamerica-east1": 0.035,
+    "asia-south1": 0.023,
+    "asia-south2": 0.023,
+    "us": 0.026,
+    "europe": 0.026,
+    "asia": 0.026,
+    "asia1": 0.046,
+    "eur4": 0.036,
+    "nam4": 0.036
+  }
+
+  defaultEgressPriceList = {
+    "10240": 0.085,
+    "153600": 0.065,
+    "153601": 0.045
+  }
 }
 
 firecloud {
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -46,53 +46,6 @@ userprofile {
 }
 
 googlecloud {
-  priceListUrl = "https://cloudpricingcalculator.appspot.com/static/data/pricelist.json"
-  # egress key previouly was "CP-COMPUTEENGINE-INTERNET-EGRESS-NA-NA"
-  priceListEgressKey = "CP-INTERNET-EGRESS-STANDARD-TIER-FROM-NA"
-  priceListStorageKey = "CP-BIGSTORE-STANDARD"
-
-  defaultStoragePriceList = {
-                               "us-central1": 0.02,
-                               "us-east1": 0.02,
-                               "us-east4": 0.023,
-                               "us-west4": 0.023,
-                               "us-west1": 0.02,
-                               "us-west2": 0.023,
-                               "us-west3": 0.023,
-                               "europe-west1": 0.02,
-                               "europe-west2": 0.023,
-                               "europe-west3": 0.023,
-                               "europe-central2": 0.023,
-                               "europe-west4": 0.020,
-                               "europe-west6": 0.025,
-                               "europe-north1": 0.020,
-                               "northamerica-northeast1": 0.023,
-                               "northamerica-northeast2": 0.023,
-                               "asia-east1": 0.02,
-                               "asia-east2": 0.023,
-                               "asia-northeast1": 0.023,
-                               "asia-southeast2": 0.023,
-                               "asia-northeast2": 0.023,
-                               "asia-northeast3": 0.023,
-                               "asia-southeast1": 0.02,
-                               "australia-southeast1": 0.023,
-                               "australia-southeast2": 0.023,
-                               "southamerica-east1": 0.035,
-                               "asia-south1": 0.023,
-                               "asia-south2": 0.023,
-                               "us": 0.026,
-                               "europe": 0.026,
-                               "asia": 0.026,
-                               "asia1": 0.046,
-                               "eur4": 0.036,
-                               "nam4": 0.036
-                             }
-
-  defaultEgressPriceList = {
-                            "10240": 0.085,
-                            "153600": 0.065,
-                            "153601": 0.045
-                           }
 }
 
 firecloud {
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -4692,6 +4692,8 @@ paths:
         - Workspaces
       summary: Calculate an estimate of the monthly storage cost for the workspace
         bucket
+      deprecated: true
+      description: Deprecated in favor of /api/workspaces/v2/{workspaceNamespace}/{workspaceName}/storageCostEstimate
       operationId: getStorageCostEstimate
       parameters:
         - $ref: '#/components/parameters/workspaceNamespaceParam'
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -3077,6 +3077,8 @@ paths:
       tags:
         - Workspaces
       summary: Get bucket usage
+      deprecated: true
+      description: Deprecated in favor of /api/workspaces/v2/{workspaceNamespace}/{workspaceName}/storageCostEstimate, which returns usage as well as cost
       operationId: getBucketUsage
       parameters:
         - $ref: '#/components/parameters/workspaceNamespaceParam'
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -5710,6 +5710,30 @@ paths:
                 $ref: '#/components/schemas/ErrorReport'
         500:
           $ref: '#/components/responses/RawlsInternalError'
+  /api/workspaces/v2/{workspaceNamespace}/{workspaceName}/storageCostEstimate:
+    get:
+      tags:
+        - WorkspacesV2
+      summary: Get current bytes usage and calculate an estimate of the monthly storage cost for the workspace
+        bucket
+      operationId: getStorageCostEstimateV2
+      parameters:
+        - $ref: '#/components/parameters/workspaceNamespaceParam'
+        - $ref: '#/components/parameters/workspaceNameParam'
+      responses:
+        200:
+          description: Successful Request
+          content:
+            'application/json':
+              schema:
+                $ref: '#/components/schemas/BucketStorageUsageAndCost'
+        404:
+          description: Workspace not found
+          content: { }
+        500:
+          description: Internal Error
+          content: { }
+      x-passthrough: false
   /cookie-authed/workspaces/{workspaceNamespace}/{workspaceName}/entities/{entityType}/tsv:
     get:
       tags:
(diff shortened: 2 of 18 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -63,21 +55,17 @@ class WorkspaceService(protected val argUserToken: WithAccessToken,
 
   implicit val userToken: WithAccessToken = argUserToken
 
+  val priceList = Map("STANDARD" -> 0.02, "NEARLINE" -> 0.01, "COLDLINE" -> 0.004, "ARCHIVE" -> 0.0012, "REGIONAL" -> 0.02, "MULTI-REGIONAL" -> 0.02) //TODO: dma?
+
   def getStorageCostEstimate(workspaceNamespace: String,
                              workspaceName: String,
                              userProject: Option[GoogleProjectId]
```
calypsomatic, line 72: On `userProject`: The api description says "When specified, bill google storage requests to this project", although it looks like it's not doing anything except checking which location to use for the cost estimate.  So I think we don't need this parameter anymore, but taking it out would mean changing the API, which I was hoping would not need to be done!  Thoughts?

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -63,21 +55,17 @@ class WorkspaceService(protected val argUserToken: WithAccessToken,
 
   implicit val userToken: WithAccessToken = argUserToken
 
+  val priceList = Map("STANDARD" -> 0.02, "NEARLINE" -> 0.01, "COLDLINE" -> 0.004, "ARCHIVE" -> 0.0012, "REGIONAL" -> 0.02, "MULTI-REGIONAL" -> 0.02) //TODO: dma?
```
calypsomatic: Whoops, TODO ["dra"](https://cloud.google.com/storage/docs/storage-classes) not "dma".  Not sure what the price for that would be, or what exactly it would be coded as when returned from the google metrics API.  Seems like I should have a catch-all for any values that are not in this list, but unclear what the cost should be.  Ideas?

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -63,21 +64,25 @@ class WorkspaceService(protected val argUserToken: WithAccessToken,
 
   implicit val userToken: WithAccessToken = argUserToken
 
+  val priceList = Map("STANDARD" -> 0.02,
+                      "NEARLINE" -> 0.01,
+                      "COLDLINE" -> 0.004,
+                      "ARCHIVE" -> 0.0012,
+                      "REGIONAL" -> 0.02,
+                      "MULTI-REGIONAL" -> 0.02
+  ) // TODO: dra?
```
calypsomatic: https://cloud.google.com/storage/docs/storage-classes
 Not sure what the price for DRA would be, or what exactly it would be coded as when returned from the google metrics API. Seems like I should have a catch-all for any values that are not in this list, but unclear what the cost should be. Ideas?

src/main/resources/reference.conf
```diff
@@ -46,53 +46,6 @@ userprofile {
...
-                               "eur4": 0.036,
-                               "nam4": 0.036
-                             }
-
-  defaultEgressPriceList = {
-                            "10240": 0.085,
-                            "153600": 0.065,
-                            "153601": 0.045
-                           }
 }
```
davidangb, line 105: this leaves an empty `googlecloud {}` stanza, which looks strange; I worry someone will try to clean it up later. I realize `FireCloudConfig` tries to read that stanza. How about one of these?
```
googlecloud {
  # can be overriden by the main conf file
  # enabled = true
}
```
or just
```
googlecloud {
  enabled = true
}
```
since enabled=true is the default?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
```diff
@@ -72,6 +72,11 @@ class HttpRawlsDAO(implicit val system: ActorSystem,
   ): Future[BucketUsageResponse] =
     authedRequestToObject[BucketUsageResponse](Get(rawlsBucketUsageUrl(ns, name)))
```
davidangb, line 73: can this v1 `getBucketUsage` now be removed?

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/RawlsDAO.scala
```diff
@@ -51,6 +51,10 @@ trait RawlsDAO extends LazyLogging with ReportsSubsystemStatus {
     FireCloudConfig.Rawls.workspacesUrl + s"/$workspaceNamespace/$workspaceName/bucketUsage"
   )
```
davidangb, line 53: if `getBucketUsage()` is removed, this can be removed too

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -63,21 +64,25 @@ class WorkspaceService(protected val argUserToken: WithAccessToken,
 
   implicit val userToken: WithAccessToken = argUserToken
 
+  val priceList = Map("STANDARD" -> 0.02,
+                      "NEARLINE" -> 0.01,
+                      "COLDLINE" -> 0.004,
+                      "ARCHIVE" -> 0.0012,
+                      "REGIONAL" -> 0.02,
+                      "MULTI-REGIONAL" -> 0.02
```
davidangb: 1) I know you just deleted stuff from reference.conf … what do you think about putting these prices into a conf file and reading them on startup? If/when google changes their prices, it's easier to tweak a conf file than to make changes here in the code.

2) I'm not convinced this is the right structure for pricing. This has pricing for single-region standard/nearline/coldline/archive, and then a single price for multi-regional. Storage pricing is the matrix of (geography) * (storage class), so you could have single-region nearline or multi-region standard. It looks like Terra limits buckets to either a single region or the "US" multi-region, [here](https://github.com/broadinstitute/rawls/blob/11859052b9387d1809118c71b507c8c4a60fac6e/core/src/main/scala/org/broadinstitute/dsde/rawls/workspace/WorkspaceService.scala#L2088-L2103) so you can limit prices to those geographies. Terra UI limits regions to `us-central1` and `northamerica-northeast1` but Rawls looks much more open.

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
davidangb: https://cloud.google.com/storage/pricing has pricing for DRA … though since Terra controls these buckets and doesn't use DRA, I wonder if it's necessary

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
davidangb: `userProject ` exists for handling requester-pays buckets. I think the bucket usage API will throw an error on requester-pays buckets unless you specify some other google project for billing. We should not change this signature, or we might start throwing errors for RP buckets.

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -63,21 +64,25 @@ class WorkspaceService(protected val argUserToken: WithAccessToken,
...
-      bucketOptions.location.toLowerCase,
-      priceList.prices.cpBigstoreStorage("us")
-    )
     // Convert bytes to GB since rate is based on GB.
-    val estimate: BigDecimal = BigDecimal(bucketUsage.usageInBytes) / (1024 * 1024 * 1024) * rate
-    RequestComplete(WorkspaceStorageCostEstimate(f"$$$estimate%.2f", bucketUsage.lastUpdated))
+    val estimate: BigDecimal = bucketUsage.metrics
+      .map(metric => BigDecimal(metric.valueInBytes) / (1024 * 1024 * 1024) * priceList(metric.storageClass))
+      .sum
+    RequestComplete(WorkspaceStorageCostEstimate(f"$$$estimate%.2f", Some(DateTime.now())))
```
davidangb: see my previous comment on reference.conf; this probably needs to re-include the bucket location for an accurate price.

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
admull: userProject is for Requester Pays. Do we know if these metrics calls cost money? We should be billing the requester for such queries on requester pays buckets.

davidangb (commented):
See comments inline - I think our "price list" needs to include location information

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
admull: I think we can omit DRA. We should probably report N/A to the UI for anything that is unexpected, that way the user will know something is wrong. Ideally, we would want to know if there's anything non-standard out there, but seems very unlikely.

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
davidangb: Following up … if we're choosing not to care about the geography, then I don't understand the `REGIONAL` and `MULTI-REGIONAL` keys in this map … I would think the map would only contain storage classes as keys.

calypsomatic: Empirically, I've seen those returned from google as the storage class.

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
calypsomatic: It's not easy to tell for sure if the metrics calls cost money, but I think I've convinced myself they haven't.  The [non-chargeable metrics](https://cloud.google.com/stackdriver/pricing#metrics-non-chargeable) links to the [metrics page](https://cloud.google.com/monitoring/api/metrics_gcp) that includes the one I'm calling.

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpRawlsDAO.scala
calypsomatic: i think so!

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
admull: REGIONAL and MULTI-REGIONAL can exist if the bucket is not in Autoclass I think? Or it might be legacy? This shouldn't be the case since we migrated all Terra buckets to Autoclass, but theoretically possible?

[MULTI-REGIONAL pricing](https://cloud.google.com/storage/pricing#multi-regions) should be 0.026. 

If we want pricing based on storage class for multi-regional we can pull that too. However, we don't need to pull the full price sheets for all regions. We really only want to officially support us-central1 for now.

calypsomatic: Hm most or all of the examples I looked at had REGIONAL or MULTI-REGIONAL so not sure if that's something we need to look into?

admull: Is it a category/sub-category? Like REGIONAL -> NEARLINE? Can you share some examples on Slack?

kevinpalis (approved):
I don't think I have much to contribute to the conversation for now. Deferring to David on this one.

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -80,6 +83,34 @@ class WorkspaceService(protected val argUserToken: WithAccessToken,
...
+      bucketUsage <- rawlsDAO.getBucketUsageV2(workspaceNamespace, workspaceName)
+    } yield {
+      // Convert bytes to GB since rate is based on GB.
+      val (totalBytes, totalEstimate) = bucketUsage.metrics.foldLeft((BigDecimal(0), BigDecimal(0))) {
+        case ((sumBytes, sumEstimate), metric) =>
+          val bytes = BigDecimal(metric.valueInBytes)
+          val estimate = bytes / (1024 * 1024 * 1024) * storagePriceList(metric.storageClass)
+          (sumBytes + metric.valueInBytes, sumEstimate + estimate)
+      }
+      WorkspaceStorageUsageAndCostEstimate(f"$$$totalEstimate%.2f", totalBytes, Some(DateTime.now()))
```
calypsomatic: I was working on catching NoSuchElementException in the case of unexpected storage classes, which would show a "key not found" error since it's not in the `storagePriceList` map, and reformatting it to a different error message that explains the problem.  But as I struggled to do proper error handling in scala, I began to wonder if we really need to watch for this particular error, or if it should just be thrown as is since it's not clear the user will care?

src/main/resources/swagger/api-docs.yaml
```diff
@@ -4692,6 +4692,8 @@ paths:
         - Workspaces
       summary: Calculate an estimate of the monthly storage cost for the workspace
         bucket
+      deprecated: true
+      description: Deprecated in favor of /api/workspaces/v2/{workspaceNamespace}/{workspaceName}/storageCostEstimate
```
davidangb, line 4698: deprecate the v1 `bucketUsage` API too?

src/main/scala/org/broadinstitute/dsde/firecloud/model/Workspace.scala
```diff
@@ -99,5 +99,10 @@ case class RawlsGroupMemberList(userEmails: Option[Seq[String]] = None,
 
 case class WorkspaceStorageCostEstimate(estimate: String, lastUpdated: Option[DateTime])
 
+case class WorkspaceStorageUsageAndCostEstimate(estimate: String,
+                                                usageInBytes: BigDecimal,
+                                                lastUpdated: Option[DateTime]
+)
```
davidangb: 1. I'd rather see `estimate` be a `BigDecimal` or some other kind of number, then let the UI or other calls do currency formatting.
2. Any reason for `lastUpdated` to be wrapped in an `Option`? I don't see any cases where it can be `None`.

src/main/scala/org/broadinstitute/dsde/firecloud/service/WorkspaceService.scala
```diff
@@ -24,6 +24,7 @@ import org.broadinstitute.dsde.rawls.model.AttributeUpdateOperations.{
 }
 import org.broadinstitute.dsde.rawls.model.WorkspaceACLJsonSupport._
 import org.broadinstitute.dsde.rawls.model._
+import org.joda.time.DateTime
```
davidangb: joda-time is deprecated in favor of the JDK's `java.time`; see https://www.joda.org/joda-time/. I know we still have joda-time elsewhere in Orch. But if you don't need it, could you use something else, like `Instant`?

davidangb (commented):
only minor comments; see inline

src/main/resources/swagger/api-docs.yaml
calypsomatic: hm i guess we could deprecate it on rawls too if we're feeling plucky

src/main/scala/org/broadinstitute/dsde/firecloud/model/Workspace.scala
calypsomatic: Those are both holdovers from copying the v1 API.  Will fix.

davidangb (approved):

PR #1570, opened 2025-02-25
CORE-308 Use okhttp rite

dvoet:
https://broadworkbench.atlassian.net/browse/CORE-308

An observation from the logs is that for orch, the connect timeouts only occur using Sam's client library. Orch only uses sam client library for 2 calls,`getUserStatusInfo` and `bulkMembershipUpdateV2`. All other calls use a hand rolled client using akka http. Granted, `getUserStatusInfo` is called on every api call so the prevalence of the error might be biased towards that call.

The crux of this change is to use a shared instance of the okhttp client configured for large bandwidth to sam. As it was, each new `ApiClient` was creating a bunch of resources for single use that were meant to be reused. This will hopefully lead to fewer connection attempts and fewer timeouts.

An ancillary change is to move a call to sam into the SamDAO (I don't know why it wasn't there to start... probably plumbing issues).

In all cases:


src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpSamDAO.scala
```diff
@@ -214,8 +219,16 @@ class HttpSamDAO(implicit
...
+    val apiClient = newApiClient(user)
+    val sam = new UsersApi(apiClient)
+    val callback = new SamApiCallback[UserStatusInfo]("getUserStatusInfo")
+    sam.getUserStatusInfoAsync(callback)
+    callback.future
+  }
+
   private def newApiClient(user: WithAccessToken) = {
-    val apiClient = new ApiClient()
+    val apiClient = new ApiClient(httpClient)
```
dvoet, line 231: This is the important change. `httpClient` is reused across all sam calls using the client library

src/main/scala/org/broadinstitute/dsde/firecloud/dataaccess/HttpSamDAO.scala
```diff
@@ -57,6 +58,10 @@ class HttpSamDAO(implicit
     with SprayJsonSupport {
 
   val timeout: FiniteDuration = 1.minute
+  private val dispatcher = new Dispatcher()
+  dispatcher.setMaxRequests(1000)
+  dispatcher.setMaxRequestsPerHost(100)
```
dvoet, line 63: I am just guessing here. I will circle back to make these configurable if this ends up helping

src/main/scala/org/broadinstitute/dsde/firecloud/utils/EnabledUserDirectives.scala
```diff
@@ -25,6 +26,8 @@ trait EnabledUserDirectives extends LazyLogging with SprayJsonSupport with Statu
   implicit val executionContext: ExecutionContext
   implicit val materializer: Materializer
 
+  val samDao: SamDAO
```
dvoet, line 29: changes in this file are related to moving the `getUserStatusInfo` call to samDAO

src/test/scala/org/broadinstitute/dsde/firecloud/utils/EnabledUserDirectivesSpec.scala
```diff
@@ -2,144 +2,43 @@ package org.broadinstitute.dsde.firecloud.utils
...
     with Matchers
     with ScalatestRouteTest
-    with BeforeAndAfterAll
-    with SprayJsonSupport {
+    with SprayJsonSupport
+    with MockitoSugar {
 
   implicit override val executionContext: ExecutionContext = scala.concurrent.ExecutionContext.Implicits.global
 
+  val samDao: SamDAO = mock[SamDAO](RETURNS_SMART_NULLS)
```
dvoet, line 34: can use mockito now instead of a mock http server

marctalbott (approved):

kevinmarete (approved):

PR #1581, opened 2025-03-11
[CORE-315] Expand DataTable Autogeneration to Support ONT Sequencing Patterns

kevinmarete:
https://broadworkbench.atlassian.net/browse/CORE-315

This PR extends the Data Table Autogeneration feature to support ONT (Oxford Nanopore Technologies) sequencing patterns. New regex patterns have been added to detect and classify ONT files, allowing them to be processed alongside Illumina files in the same data table.

[Test-Output-files.zip](https://github.com/user-attachments/files/19199255/Output-files.zip)

In all cases:

- [x] Squash and merge. Make sure your branch deletes; GitHub should do this for you.

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/filematch/FileMatcher.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/filematch/FileMatcher.scala
@@ -2,7 +2,11 @@ package org.broadinstitute.dsde.firecloud.filematch
 
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.firecloud.filematch.result.{FailedMatchResult, FileMatchResult, SuccessfulMatchResult}
-import org.broadinstitute.dsde.firecloud.filematch.strategy.{FileRecognitionStrategy, IlluminaPairedEndStrategy}
+import org.broadinstitute.dsde.firecloud.filematch.strategy.{
+  FileRecognitionStrategy,
+  IlluminaPairedEndStrategy,
+  OntSingleReadStrategy
+}
 
 import java.nio.file.Path
 
@@ -16,7 +20,8 @@ import java.nio.file.Path
 class FileMatcher extends LazyLogging {
 
   // the list of recognition strategies to use
-  private val matchingStrategies: List[FileRecognitionStrategy] = List(new IlluminaPairedEndStrategy())
+  private val matchingStrategies: List[FileRecognitionStrategy] =
+    List(new IlluminaPairedEndStrategy(), new OntSingleReadStrategy())
 
   /**
     * Given a list of files, pair up those files according to our known recognition strategies.
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/filematch/FileMatcher.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/filematch/FileMatcher.scala
@@ -1,7 +1,12 @@
 package org.broadinstitute.dsde.firecloud.filematch
 
 import com.typesafe.scalalogging.LazyLogging
-import org.broadinstitute.dsde.firecloud.filematch.result.{FailedMatchResult, FileMatchResult, SuccessfulMatchResult}
+import org.broadinstitute.dsde.firecloud.filematch.result.{
+  FailedMatchResult,
+  FileMatchResult,
+  PartialMatchResult,
+  SuccessfulMatchResult
+}
 import org.broadinstitute.dsde.firecloud.filematch.strategy.{
   FileRecognitionStrategy,
   IlluminaPairedEndStrategy,
@@ -53,25 +58,32 @@ class FileMatcher extends LazyLogging {
     */
   private def performPairing(pathList: List[Path]): List[FileMatchResult] = {
     // find every path in the incoming pathList that is recognized by one of our known patterns
-    val desiredPairings: List[SuccessfulMatchResult] = findFirstFiles(pathList)
+    val (successfulMatches, partialMatches) = findFirstFiles(pathList).partition {
+      case _: SuccessfulMatchResult => true
+      case _: PartialMatchResult    => false
+    }
 
-    // remove the recognized firstFiles from the outstanding pathList
-    val remainingPaths: List[Path] = pathList diff desiredPairings.map(_.firstFile)
+    // remove the recognized firstFiles and partialMatches from the outstanding pathList
+    val remainingPaths: List[Path] = pathList.diff(
+      successfulMatches.collect { case s: SuccessfulMatchResult => s.firstFile } ++
+        partialMatches.collect { case p: PartialMatchResult => p.firstFile }
+    )
 
     // process the recognized "read 1" files, and look for their desired pairings in the outstanding pathList.
     // this will result in either SuccessfulMatchResult when the desired pairing is found, or PartialMatchResult
     // when the desired pairing is not found
-    val pairingResults: List[FileMatchResult] = findSecondFiles(remainingPaths, desiredPairings)
+    val pairingResults: List[FileMatchResult] =
+      findSecondFiles(remainingPaths, successfulMatches.collect { case s: SuccessfulMatchResult => s })
 
     // remove the recognized "read 2" files from the outstanding pathList
     val unrecognizedPaths: List[Path] = remainingPaths diff pairingResults.collect { case s: SuccessfulMatchResult =>
       s.secondFile
     }
     // translate the unrecognized paths into a FileMatchResult
-    val unrecognizedResults: List[FailedMatchResult] = unrecognizedPaths.map(path => FailedMatchResult(path))
+    val unrecognizedResults: List[FailedMatchResult] = unrecognizedPaths.map(FailedMatchResult(_))
 
     // return results, sorted by firstFile
-    (pairingResults ++ unrecognizedResults).sortBy(r => r.firstFile)
+    (pairingResults ++ partialMatches ++ unrecognizedResults).sortBy(_.firstFile)
   }
 
   /**
@@ -79,10 +91,11 @@ class FileMatcher extends LazyLogging {
     * @param pathList the list of files to inspect
     * @return pairing results
     */
-  private def findFirstFiles(pathList: List[Path]): List[SuccessfulMatchResult] =
+  private def findFirstFiles(pathList: List[Path]): List[FileMatchResult] =
     pathList.collect { path =>
       tryPairingStrategies(path) match {
         case success: SuccessfulMatchResult => success
+        case partial: PartialMatchResult    => partial
       }
     }
 
@@ -116,11 +129,12 @@ class FileMatcher extends LazyLogging {
     val strategyHit = matchingStrategies.collectFirst(strategy =>
       strategy.matchFirstFile(file) match {
         case success: SuccessfulMatchResult => success
+        case partial: PartialMatchResult    => partial
       }
     )
     strategyHit match {
       // The current file is recognized by one of our recognition strategies
-      case Some(desiredResult: SuccessfulMatchResult) => desiredResult
+      case Some(desiredResult: FileMatchResult) => desiredResult
       // the current file is not recognized
       case _ => FailedMatchResult(file)
     }
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/filematch/FileMatcher.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/filematch/FileMatcher.scala
@@ -69,8 +69,7 @@ class FileMatcher extends LazyLogging {
     // process the recognized "read 1" files, and look for their desired pairings in the outstanding pathList.
     // this will result in either SuccessfulMatchResult when the desired pairing is found, or PartialMatchResult
     // when the desired pairing is not found
-    val pairingResults: List[FileMatchResult] =
-      findSecondFiles(remainingPaths, successfulMatches.collect { case s: SuccessfulMatchResult => s })
+    val pairingResults: List[FileMatchResult] = findSecondFiles(remainingPaths, successfulMatches)
 
     // remove the recognized "read 2" files from the outstanding pathList
     val unrecognizedPaths: List[Path] = remainingPaths diff pairingResults.collect { case s: SuccessfulMatchResult =>
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/filematch/FileMatcher.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/filematch/FileMatcher.scala
@@ -58,16 +58,13 @@ class FileMatcher extends LazyLogging {
     */
   private def performPairing(pathList: List[Path]): List[FileMatchResult] = {
     // find every path in the incoming pathList that is recognized by one of our known patterns
-    val (successfulMatches, partialMatches) = findFirstFiles(pathList).partition {
-      case _: SuccessfulMatchResult => true
-      case _: PartialMatchResult    => false
-    }
+    val matches = findFirstFiles(pathList)
+    val successfulMatches: List[SuccessfulMatchResult] = matches.collect { case x: SuccessfulMatchResult => x }
+    val partialMatches: List[PartialMatchResult] = matches.collect { case x: PartialMatchResult => x }
 
     // remove the recognized firstFiles and partialMatches from the outstanding pathList
-    val remainingPaths: List[Path] = pathList.diff(
-      successfulMatches.collect { case s: SuccessfulMatchResult => s.firstFile } ++
-        partialMatches.collect { case p: PartialMatchResult => p.firstFile }
-    )
+    val remainingPaths: List[Path] =
+      pathList.diff(successfulMatches.map(_.firstFile) ++ partialMatches.map(_.firstFile))
 
     // process the recognized "read 1" files, and look for their desired pairings in the outstanding pathList.
     // this will result in either SuccessfulMatchResult when the desired pairing is found, or PartialMatchResult
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/filematch/result/SuccessfulMatchResult.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/filematch/result/SuccessfulMatchResult.scala
@@ -16,4 +16,7 @@ case class SuccessfulMatchResult(firstFile: Path, secondFile: Path, id: String) 
 object SuccessfulMatchResult {
   def fromStrings(firstFile: String, secondFile: String, id: String): SuccessfulMatchResult =
     SuccessfulMatchResult(new java.io.File(firstFile).toPath, new java.io.File(secondFile).toPath, id)
+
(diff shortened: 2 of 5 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/filematch/strategy/OntSingleReadStrategy.scala
```diff
@@ -0,0 +1,41 @@
...
+object OntSingleReadStrategy {
+  private val fastqGz = ".fastq.gz"
+  // if the first file starts with ${key} and ends with any of the values in the list, then this is a single read file
+  val FILE_STARTS_ENDINGS: Map[String, List[String]] = Map(
+    "Complete_barcode" -> List(fastqGz, ".clean.fastq"),
+    "Incomplete_barcode" -> List(fastqGz),
+    "Repeat_barcode" -> List(fastqGz),
+    "barcode" -> List(fastqGz)
+  )
+}
```
davidangb, line 21: This will ONLY match on files that start with "Complete_barcode", "Incomplete_barcode", "Repeat_barcode", or "barcode". My understanding from the requirements of the Jira ticket is that the filenames can start with anything, as long as it includes "barcode". Adam wrote:

> … naming convention of some kind of \*barcode#####.*\.fastq or \*barcode#####.\*.fastq.gz pattern.

I suggest switching the matching strategy to use a regex or other kind of check that allows for more flexibility. You could probably get away with a `.contains()` and `endsWith()`, though I think regex will be clearer and more precise in the end.

Edited to add: this also will only match specific file endings; `.clean.fastq` is ONLY allowed for files starting with "Complete_barcode"

davidangb (changes requested):
See inline - I don't think this satisfies the requirement in the ticket

src/test/scala/org/broadinstitute/dsde/firecloud/filematch/FileMatcherSpec.scala
```diff
@@ -60,6 +60,37 @@ class FileMatcherSpec extends AnyFreeSpec with Matchers {
...
+
+        val expected = List(
+          FailedMatchResult.fromString("Complete_barcode01.fasta.gz"),
+          FailedMatchResult.fromString("Incomplete_barcode03.bam"),
+          FailedMatchResult.fromString("unknown_file.txt")
+        )
+        val actual = new FileMatcher().pairFiles(input)
+
         actual shouldBe expected
       }
```
davidangb, line 95: Could you include an additional unit test that uses multiple match strategies? Something like:
```scala
      "should use multiple strategies in the same inputs" in {
        // input includes hits on both IlluminaPairedEndStrategy and OntSingleReadStrategy
        val input = List(
          "Sample1_01.fastq.gz",
          "Complete_barcode01.fastq.gz",
          "Sample1_02.fastq.gz",
          "Incomplete_barcode03.fastq.gz",
          "Sample2_01.fastq.gz",
          "Sample2_02.fastq.gz"
        )

        val expected = List(
          PartialMatchResult.fromStrings("Complete_barcode01.fastq.gz", "Complete_barcode01"),
          PartialMatchResult.fromStrings("Incomplete_barcode03.fastq.gz", "Incomplete_barcode03"),
          SuccessfulMatchResult.fromStrings("Sample1_01.fastq.gz", "Sample1_02.fastq.gz", "Sample1"),
          SuccessfulMatchResult.fromStrings("Sample2_01.fastq.gz", "Sample2_02.fastq.gz", "Sample2")
        )
        val actual = new FileMatcher().pairFiles(input)

        actual shouldBe expected
      }
```

kevinmarete: Added: https://github.com/broadinstitute/firecloud-orchestration/pull/1581/commits/08b26008cf56abd0147316734df2cf527c484802

src/main/scala/org/broadinstitute/dsde/firecloud/filematch/strategy/OntSingleReadStrategy.scala
kevinmarete: Fixed: https://github.com/broadinstitute/firecloud-orchestration/pull/1581/commits/9e4f4bd50a36798a2572f893ee499fea772253bc

src/main/scala/org/broadinstitute/dsde/firecloud/filematch/strategy/OntSingleReadStrategy.scala
```diff
@@ -0,0 +1,41 @@
...
+  }
+}
+
+object OntSingleReadStrategy {
+  // if the first file contains ${key} and ends with any of the values in the list, then this is a single read file
+  val FILE_CONTAIN_ENDINGS: Map[String, List[String]] = Map(
+    "Complete_barcode" -> List(".clean.fastq"),
+    "barcode" -> List(".fastq", ".fastq.gz")
+  )
+}
```
davidangb, line 21: I still don't think this is right. You may want to check with Adam to verify requirements - but the requirement as written in Jira is:

> … naming convention of some kind of \*barcode#####.*\.fastq or \*barcode#####.\*.fastq.gz pattern.

which means the only logic we should care about is:
* ends with `.fastq` or `.fastq.gz`
* contains `barcode` before the file extension (should this be case-sensitive or not?)

And we don't need any considerations for "Complete_" or ".clean". As written, this will not return the right results for input of "Repeat_barcode4.clean.fastq.gz", for instance.

kevinmarete: I have updated the matcher: https://github.com/broadinstitute/firecloud-orchestration/pull/1581/commits/315ac74b654e0eda3016592087cced3dd56dd72b and added more test cases to validate output: https://github.com/broadinstitute/firecloud-orchestration/pull/1581/commits/14012df22439dc3abcde2cb0bbf54b73d47300a7

src/main/scala/org/broadinstitute/dsde/firecloud/filematch/strategy/OntSingleReadStrategy.scala
```diff
@@ -0,0 +1,44 @@
...
+      // the file is not recognized
+      case None => FailedMatchResult(path)
+    }
+  }
+}
+
+object OntSingleReadStrategy {
+  // if the file contains ${key} and ends with any value in the list, it's an ONT single read file
+  val FILE_CONTAIN_ENDINGS: Map[String, List[String]] = Map(
+    "barcode" -> List(".clean.fastq", ".clean.fastq.gz", ".fastq", ".fastq.gz")
```
marctalbott: According to the ticket, the endings here should be `.*.fastq` or `.*.fastq.gz`. Is that still correct? If so, `clean` should be removed in favor of a wildcard

kevinmarete: Fixed: https://github.com/broadinstitute/firecloud-orchestration/pull/1581/commits/016d7477bccb04ee687baad65e436e3fc51aa842

src/main/scala/org/broadinstitute/dsde/firecloud/filematch/strategy/OntSingleReadStrategy.scala
```diff
@@ -0,0 +1,44 @@
...
+    }
+  }
+}
+
+object OntSingleReadStrategy {
+  // if the file contains ${key} and ends with any value in the list, it's an ONT single read file
+  val FILE_CONTAIN_ENDINGS: Map[String, List[String]] = Map(
+    "barcode" -> List(".fastq", ".fastq.gz")
+  )
+}
```
davidangb, line 21: since there is now only a single key in this map now, could you reduce complexity and just write out the regex verbatim? Something like this:
```scala
object OntSingleReadStrategy {
  // if the file contains barcode### and ends with .fastq or .fastq.gz, it's an ONT single read file
  val PATTERN: Regex = new Regex("(?i)(.*barcode\\d+).*\\.fastq(?:.gz)*$")
  // to be all Scala-like, use the .r shorthand:
  // val PATTERN: Regex = "(?i)(.*barcode\\d+).*\\.fastq(?:.gz)*$".r
}
```

Note that in this regex, I've added `()` around everything before the file extension. If you went with this approach, you could then use this capture group to get the row's id instead of doing the extra work of `.stripSuffix(value).replaceAll("\\..*$", "")`. The `matchFirstFile` implementation could be simplified a lot to something like:
```scala
    PATTERN.findFirstMatchIn(path.toString) match {
      case Some(regexMatch) => // matched successfully; retrieve the capture group out of the regexMatch variable
      case None             => // failed to match
    }
```

If you went with Adam's potential feature of not even caring about "barcode" or numbers, the regex becomes even yet more simple.

src/main/scala/org/broadinstitute/dsde/firecloud/filematch/strategy/OntSingleReadStrategy.scala
```diff
@@ -0,0 +1,44 @@
+package org.broadinstitute.dsde.firecloud.filematch.strategy
+
+import org.broadinstitute.dsde.firecloud.filematch.result.{FailedMatchResult, FileMatchResult, SuccessfulMatchResult}
+import OntSingleReadStrategy.FILE_CONTAIN_ENDINGS
+
+import java.nio.file.{Path, Paths}
+import scala.util.matching.Regex
+
+class OntSingleReadStrategy extends FileRecognitionStrategy {
```
marctalbott, line 9: Could we simplify all of this? The `Map[String, Regex]` and the object all feel a bit complicated considering we're just trying to find out if the regex matches the filename and then capturing part of that match. 

We're always matching on the same regex, can we use that to see if the filename matches and to capture the id?

kevinmarete: Done: https://github.com/broadinstitute/firecloud-orchestration/pull/1581/commits/e97d1d1ef61c78ebb10ad5b46e46a1c200c31c1a

src/main/scala/org/broadinstitute/dsde/firecloud/filematch/strategy/OntSingleReadStrategy.scala
kevinmarete: Done: https://github.com/broadinstitute/firecloud-orchestration/pull/1581/commits/e97d1d1ef61c78ebb10ad5b46e46a1c200c31c1a

src/main/scala/org/broadinstitute/dsde/firecloud/filematch/strategy/OntSingleReadStrategy.scala
```diff
@@ -0,0 +1,21 @@
...
+import OntSingleReadStrategy.PATTERN
+
+import java.nio.file.{Path, Paths}
+import scala.util.matching.Regex
+
+class OntSingleReadStrategy extends FileRecognitionStrategy {
+
+  override def matchFirstFile(path: Path): FileMatchResult =
+    PATTERN.findFirstMatchIn(path.toString) match {
+      case Some(regexMatch) => SuccessfulMatchResult(path, Paths.get(""), regexMatch.group(1))
```
marctalbott: Should this be a `PartialMatchResult` instead of passing in an empty path for the second file?

src/main/scala/org/broadinstitute/dsde/firecloud/filematch/strategy/OntSingleReadStrategy.scala
```diff
@@ -0,0 +1,21 @@
...
+      case None             => FailedMatchResult(path)
+    }
```
marctalbott: Optional, but I just learned you could pattern match like this so I figured I'd share. It might obfuscate the fact that a regex is being used too much though, I haven't made up my mind.
```suggestion
    path.toString match {
      case PATTERN(id) => PartialMatchResult(path, id)
      case _           => FailedMatchResult(path)
    }
```

src/main/scala/org/broadinstitute/dsde/firecloud/filematch/strategy/OntSingleReadStrategy.scala
```diff
@@ -0,0 +1,21 @@
...
+}
+
+object OntSingleReadStrategy {
+  // if the file contains barcode### and ends with .fastq or .fastq.gz, it's an ONT single read file
+  val PATTERN: Regex = "(?i)(.*barcode\\d+).*\\.fastq(?:.gz)*$".r
```
marctalbott: I believe the `*` at the end should be a `?` or `Complete_barcode01.fastq.gz.gz.gz.gz.gz.gz` will match. 

You can also use `"""regex"""` to avoid having to escape your back slashes

kevinmarete: Done: https://github.com/broadinstitute/firecloud-orchestration/pull/1581/commits/84e78009d6a1a332e8fb08e51c98414837c3a902

src/main/scala/org/broadinstitute/dsde/firecloud/filematch/strategy/OntSingleReadStrategy.scala
kevinmarete: Done: https://github.com/broadinstitute/firecloud-orchestration/pull/1581/commits/9ccf4944d496541fed79a497fdd1f31a24e50f05

src/main/scala/org/broadinstitute/dsde/firecloud/filematch/strategy/OntSingleReadStrategy.scala
kevinmarete: Using a `PartialMatchResult` causes the matcher to generate an overall `FailedMatchResult`. To address this, I overloaded `SuccessfulMatchResult` to optionally accept two parameters (filePath1, id), as shown here: https://github.com/broadinstitute/firecloud-orchestration/pull/1581/commits/6713d87751ef556da69ebbbd4b7c461f178cebd4

src/main/scala/org/broadinstitute/dsde/firecloud/filematch/result/SuccessfulMatchResult.scala
```diff
@@ -16,4 +16,7 @@ case class SuccessfulMatchResult(firstFile: Path, secondFile: Path, id: String)
 object SuccessfulMatchResult {
   def fromStrings(firstFile: String, secondFile: String, id: String): SuccessfulMatchResult =
     SuccessfulMatchResult(new java.io.File(firstFile).toPath, new java.io.File(secondFile).toPath, id)
+
+  def apply(firstFile: Path, id: String): SuccessfulMatchResult =
+    SuccessfulMatchResult(firstFile, null, id)
```
davidangb: hmmm … intentionally using a null is a big no-no in Scala land. I'd rather this return a `PartialMatchResult` as Marc suggested, then have `FileMatcher` handle those partials better. I don't think this would be too bad:
* FileMatcher.findFirstFiles changes to return List[FileMatchResult] and includes both SuccessfulMatchResult and PartialMatchResult
* FileMatcher.performPairing only calls findSecondFiles on the successes and passes partials untouched

If you don't like using PartialMatchResult in this way, you could also add another FileMatchResult subclass which indicates "I found the first file but please don't even look for a second file"

kevinmarete: Done: https://github.com/broadinstitute/firecloud-orchestration/pull/1581/commits/bf414fcf7acfdedd9a1b42d76449d4b9c10677b2

src/main/scala/org/broadinstitute/dsde/firecloud/filematch/FileMatcher.scala
```diff
@@ -48,36 +58,44 @@ class FileMatcher extends LazyLogging {
     */
   private def performPairing(pathList: List[Path]): List[FileMatchResult] = {
     // find every path in the incoming pathList that is recognized by one of our known patterns
-    val desiredPairings: List[SuccessfulMatchResult] = findFirstFiles(pathList)
+    val (successfulMatches, partialMatches) = findFirstFiles(pathList).partition {
+      case _: SuccessfulMatchResult => true
+      case _: PartialMatchResult    => false
+    }
```
davidangb: optional; your call: if you handled the casting up here, you could do away with multiple `collect`s later in the code:
```scala
    val matches = findFirstFiles(pathList)
    val successfulMatches: List[SuccessfulMatchResult] = matches.collect { case x: SuccessfulMatchResult => x }
    val partialMatches: List[PartialMatchResult] = matches.collect { case x: PartialMatchResult => x }
```
The single call to `partition` is more efficient than two calls to `collect` … but there are multiple calls to `collect` later in the code.

davidangb (approved):
one last Scala optimization comment inline, your call on accepting or not. Giving this a :thumb:. I know this PR has been through a lot, thank you for seeing it through!

src/main/scala/org/broadinstitute/dsde/firecloud/filematch/FileMatcher.scala
kevinmarete: Done: https://github.com/broadinstitute/firecloud-orchestration/pull/1581/commits/7c0ad600848207b2832ffcf34886b88b21518d7a

src/main/scala/org/broadinstitute/dsde/firecloud/filematch/FileMatcher.scala
```diff
@@ -48,36 +58,41 @@ class FileMatcher extends LazyLogging {
...
+    // remove the recognized firstFiles and partialMatches from the outstanding pathList
+    val remainingPaths: List[Path] =
+      pathList.diff(successfulMatches.map(_.firstFile) ++ partialMatches.map(_.firstFile))
 
     // process the recognized "read 1" files, and look for their desired pairings in the outstanding pathList.
     // this will result in either SuccessfulMatchResult when the desired pairing is found, or PartialMatchResult
     // when the desired pairing is not found
-    val pairingResults: List[FileMatchResult] = findSecondFiles(remainingPaths, desiredPairings)
+    val pairingResults: List[FileMatchResult] =
+      findSecondFiles(remainingPaths, successfulMatches.collect { case s: SuccessfulMatchResult => s })
```
davidangb: remove this `.collect` too, then all good!

davidangb (approved):
one more .collect to remove, another proactive 👍

marctalbott (approved):

src/main/scala/org/broadinstitute/dsde/firecloud/filematch/FileMatcher.scala
kevinmarete: done: https://github.com/broadinstitute/firecloud-orchestration/pull/1581/commits/b2e49bb33fea29750828cc767dbcafe7cb9664e2

PR #1596, opened 2025-03-24
Revert "CORE-69: update guava, workbench-google2; exclude from google2"

davidangb:
Reverts broadinstitute/firecloud-orchestration#1595

This change started properly in a bee, but is failing to start in dev. Reverting so I can figure it out.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -6,7 +6,7 @@ object Dependencies {
   val jacksonV = "2.18.3"
   val jacksonHotfixV = "2.18.3" // for when only some of the Jackson libs have hotfix releases
   val nettyV = "4.1.119.Final"
-  val workbenchLibsHash = "70c7d82" // see https://github.com/broadinstitute/workbench-libs readme for hash values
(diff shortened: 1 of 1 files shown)


dvoet (approved):

project/Dependencies.scala
```diff
@@ -38,7 +38,7 @@ object Dependencies {
     // TODO: can these move to sbt's dependencyOverrides?
     "io.netty"                       % "netty-handler"       % nettyV, // netty is needed by the Elasticsearch client at runtime
     "org.apache.lucene"              % "lucene-queryparser"  % "6.6.6", // pin to this version; it's the latest compatible with our elasticsearch client
-    "com.google.guava"               % "guava"               % "33.4.5-jre",
+    "com.google.guava"               % "guava"               % "33.4.0-jre",
```
pshapiro4broad, line 41: Does this project use `guava`? I don't see it imported anywhere and `sbt assembly` works with this line commented out. But maybe it's needed for some kind of runtime dependency, I didn't try running it.

davidangb: I will investigate that when reverting the revert!

pshapiro4broad (approved):

PR #1607, opened 2025-04-03
CORE-443 deny email patterns from access to auth domains driven by telemetry

dvoet:
https://broadworkbench.atlassian.net/browse/CORE-443

In all cases:


kevinmarete (approved):

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
```diff
@@ -274,6 +274,9 @@ class NihService(val samDao: SamDAO,
     } yield ()
   }
 
+  private def allowedNihMembers(members: Set[WorkbenchEmail]): Set[WorkbenchEmail] =
+    members.filterNot(email => FireCloudConfig.Nih.denyEmailPatterns.exists(_.matches(email.value)))
+
```
davidangb, line 288: recommendation, feel free to do in a follow-on PR: log when a user is denied because of this block list so we can follow up with some record of the blocklist behaving

davidangb (approved):

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
dvoet: good idea

PR #1622, opened 2025-04-08
CORE-382: delete previously-ignored data library swat tests

davidangb:
CORE-382

Fully delete the swat tests related to the FireCloud Data Library. These tests were previously ignored in #1584; we are now deleting them entirely.


automation/src/test/resources/AsyncImportSpec-pfb-expected-entities.json
davidangb, line 1: this one isn't actually related to Data Library, but I noticed it was an orphan. The code that used it was deleted way back in #1391.

marctalbott (approved):

kevinmarete (approved):

PR #1634, opened 2025-05-13
CORE-496 disable allowlists and tolerate missing dbgap groups

dvoet:
https://broadworkbench.atlassian.net/browse/CORE-496

In prep for https://broadworkbench.atlassian.net/browse/CORE-489, implement a disabled config switch for each allowlist in orch. When set to true, ignore them when a user links or unlinks their NIH account and treat them as empty when the explicit call to sync them is made from ncbiaccess. 

Disabling an allowlist means that users won't be added to the associated auth domain group based on telemetry files processed by the ncbiaccess cronjob. Instead they will gain access via RAS which triggers membership in a different group. This new group is a sub group of the auth domain group (there is also a bug fix in this PR to the process that makes the new group a sub group of the auth domain group, tolerating when the new group does not exist yet).

In all cases:

### DIFF
--- a/src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
+++ b/src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
@@ -48,7 +48,8 @@ case class NihStatus(linkedNihUsername: Option[String] = None,
 case class NihAllowlist(name: String,
                         groupToSync: WorkbenchGroupName,
                         fileName: String,
-                        dbGapPermission: DbGapPermission
+                        dbGapPermission: DbGapPermission,
+                        disabled: Boolean
 )
 
 case class NihDatasetPermission(name: String, authorized: Boolean)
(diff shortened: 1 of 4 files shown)


src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
```diff
@@ -274,6 +282,12 @@ class NihService(val samDao: SamDAO,
     } yield ()
   }
 
+  private def getSamGroupEmail(groupName: WorkbenchGroupName): Future[Option[WorkbenchEmail]] =
+    samDao.getGroupEmail(groupName)(getAdminAccessToken).map(Option.apply).recover {
+      case e: FireCloudExceptionWithErrorReport if e.errorReport.statusCode.contains(StatusCodes.NotFound) =>
+        None
```
dvoet, line 288: bug fix, handle the case where the group is not found

calypsomatic (approved):

davidangb (approved):

PR #1648, opened 2025-07-11
[CORE-561] Added NIH Resources Api

kevinmarete:
Ticket: https://broadworkbench.atlassian.net/browse/CORE-561

---

In all cases:

### DIFF
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -2175,6 +2175,20 @@ paths:
           description: Internal server error unlinking NIH account
           content: {}
       x-passthrough: false
+  /api/nih/resources:
+    get:
+      tags:
+        - NIH
+      summary: Retrieves info about a user's NIH Resources
+      operationId: NIHResources
+      responses:
+        200:
+          description: Successfully retrieved NIH Resources (without Linked NIH Username)
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/NihResources'
+      x-passthrough: false
   /api/notifications/general:
     get:
       tags:
@@ -7359,6 +7373,13 @@ components:
           type: integer
           format: int64
           default: 0
+    NihResources:
+      type: array
+      properties:
+        name:
+          type: string
+        authorized:
+          type: boolean
     NotificationType:
       required:
         - description
--- a/src/main/resources/swagger/api-docs.yaml
+++ b/src/main/resources/swagger/api-docs.yaml
@@ -2183,7 +2183,7 @@ paths:
       operationId: NIHResources
       responses:
         200:
-          description: Successfully retrieved NIH Resources (without Linked NIH Username)
+          description: Successfully retrieved NIH Resources
           content:
             application/json:
               schema:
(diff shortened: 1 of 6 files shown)


calypsomatic (approved):

src/main/resources/swagger/api-docs.yaml
```diff
@@ -2175,6 +2175,20 @@ paths:
...
       x-passthrough: false
+  /api/nih/resources:
+    get:
+      tags:
+        - NIH
+      summary: Retrieves info about a user's NIH Resources
+      operationId: NIHResources
+      responses:
+        200:
+          description: Successfully retrieved NIH Resources (without Linked NIH Username)
```
davidangb: nit: remove "(without Linked NIH Username)"; that only makes sense if you are comparing this API to /api/nih/status

src/main/resources/swagger/api-docs.yaml
```diff
@@ -7359,6 +7373,13 @@ components:
           type: integer
           format: int64
           default: 0
+    NihResources:
+      type: object
+      properties:
+        datasetPermissions:
+          type: array
+          items:
+            $ref: '#/components/schemas/DatasetPermission'
```
davidangb, line 7380: This comment is purely informational, I am not requesting changes to this PR:

Since you've now defined a `NihResources` schema component, you could reuse it inside `NihStatus` using `allOf`. See how OpenAPI handles inheritance and polymorphism here: https://swagger.io/docs/specification/v3_0/data-models/inheritance-and-polymorphism/

You'd end up with a `NihStatus` schema that looks like:
```yaml
    NihStatus:
      allOf:
        - $ref: "#/components/schemas/NihResources"
        - required:
            - linkedNihUsername
          type: object
          properties:
            linkedNihUsername:
              type: string
            linkExpireTime:
              type: integer
              format: int64
              default: 0
```
(I haven't tested this, syntax might be wrong)

Again - what you have in this PR is just fine; I'm adding this comment in case you haven't yet seen OpenAPI features like this

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
```diff
@@ -167,6 +170,11 @@ class NihService(val samDao: SamDAO,
         }
     }
 
+  def getNihResources(userInfo: UserInfo): Future[NihResources] =
+    getAllAllowlistGroupMemberships(userInfo).map { allowlistMembership =>
+      NihResources(allowlistMembership)
+    }
+
```
davidangb, line 177: Optional, your call: you don't actually need a new `NihResources` model object. You could have this return
```scala
NihStatus(datasetPermissions = allowlistMembership, linkedNihUsername = None, linkExpireTime = None)
```
which would result in the same API response payload to the end user.

Having a separate model object DOES make the intent clear in the code, so I am fine with either way, up to you

davidangb (commented):
one minor nit, two totally optional comments. Thank you!

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
kevinmarete: I will keep this as is!

davidangb (approved):

PR #1659, opened 2025-08-18
CORE-609: update last step of nightly automation test to reflect CRDC offboarding

davidangb:
Followup to #1658, which was only a partial fix.

Since we no longer sync TARGET and TCGA dataset permissions, we can't use those datasets in the nightly test. This updates the test to use other datasets.

As a bonus, this PR also includes a fix to allow kicking off test runs using an arbitrary branch of test code.

I kicked off a test run using this code and it passed; see https://github.com/broadinstitute/firecloud-orchestration/actions/runs/17043664395/job/48314904895.

### DIFF
--- a/.github/workflows/orch-nightly-swat-tests.yml
+++ b/.github/workflows/orch-nightly-swat-tests.yml
@@ -6,6 +6,12 @@ on:
   schedule:
     - cron: "30 9 * * *"
   workflow_dispatch:
+    inputs:
+      test-code-ref:
+        description: "The branch, tag or SHA of test automation code to execute."
+        default: ''
+        required: false
+        type: string
 
 env:
   RUN_NAME: '${{ github.event.repository.name }}-${{ github.run_id }}-${{ github.run_attempt }}'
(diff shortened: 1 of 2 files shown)


.github/workflows/orch-nightly-swat-tests.yml
```diff
@@ -89,7 +95,7 @@ jobs:
               "bee-name": "orch-${{ github.run_id }}-${{ matrix.terra-env }}",
               "ENV": "${{ matrix.testing-env }}",
               "test-context": "${{ needs.prepare-configs.outputs.test-context }}",
-              "ref": "${{ needs.prepare-configs.outputs.tag }}"
+              "ref": "${{ inputs.test-code-ref || needs.prepare-configs.outputs.tag }}"
```
davidangb, line 98: this allows specifying which branch of test code to run, instead of always using the latest tag.

kevinmarete (approved):

samanehsan (approved):

PR #1387, opened 2024-07-03
AJ-1784: update to Akka module versions 24.05

davidangb:
See https://akka.io/blog/news/2024/05/17/akka-24.05-released

Upgrades to this latest version of Akka libraries.

These are pretty big jumps, but I have read through the migration guides and do not see anything that needs attention.

---

Have you read [CONTRIBUTING.md](../CONTRIBUTING.md) lately? If not, do that first.

I, the developer opening this PR, do solemnly pinky swear that:

- [x] I've followed [the instructions](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/CONTRIBUTING.md#api-changes) if I've made any changes to the API, _especially_ if they're breaking changes
- [x] I've updated the RC_XXX release ticket with any manual steps required to release this change
- [x] I've updated the [FISMA documentation](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/CONTRIBUTING.md#fisma-documentation-changes) if I've made any security-related changes, including auth, encryption, or auditing

In all cases:

- [ ] Get two thumbsworth of review and PO signoff if necessary
- [x] Verify all tests go green
- [ ] Squash and merge; you can delete your branch after this **unless it's for a hotfix**. In that case, don't delete it!
- [ ] Test this change deployed correctly and works on dev environment after deployment

PR #1623, opened 2025-04-09
CORE-382: delete Data Library

davidangb:
CORE-382

* Delete Data Library APIs and their implementations:
    *  `/api/library` APIs
    * `/duos` APIs
    * `/api/workspaces/{namespace}/{name}/catalog` APIs
* Change some APIs to be passthroughs, now that their implementations don't need Data Library features:
    * clone-workspace
    * delete-workspace
    * workspace updateAttributes

With all this change, there are some Rawls, terra-helmfile, and probably Sam cleanup to do. I'll take care of that followup work as part of the same Jira ticket.

PR #1675, opened 2025-10-02
CORE-716: downgrade akka, akka-http

davidangb:
This _downgrades_ akka and akka-http to the most recent versions which:
* are available in Maven Central
* use the standard Apache license instead of akka's BSL

Akka and akka-http versions greater than these are only available from repo.akka.io, which requires registration and - as of recently - a unique repository token. Since July 2024, as of PR #1387, we had successfully been using repo.akka.io without a token but now they require one.

This trajectory for akka is disappointing, so I am downgrading instead of buying in.

FWIW, I did a search through both the `broadinstitute` and `DataBiosphere` repos, and Orchestration is the ONLY repo that contained the string "repo.akka.io".

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -78,11 +78,9 @@ pullRequests.grouping = [ { name = "minor_patch", title = "CORE-69: Minor and pa
 # Defaults to empty `[]` which mean Scala Steward will not ignore dependencies.
 # updates.ignore = [ { groupId = "org.acme", artifactId="foo", version = "1.0" } ]
 
-# Elasticsearch and its Lucene dependency should not be automatically updated;
-# these require nontrivial changes to calling code and business logic. See AJ-266.
 updates.ignore = [
-  { groupId = "org.elasticsearch.client", artifactId = "transport" },   # AJ-266
-  { groupId = "org.apache.lucene", artifactId = "lucene-queryparser" } # AJ-266
+    # Akka updates may use a different license (BSL); don't automatically upgrade.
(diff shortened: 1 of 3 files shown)


calypsomatic (commented):
Is there a setting to prevent scala steward from trying to update it?

davidangb:
> Is there a setting to prevent scala steward from trying to update it?

@calypsomatic I've added this to Scala Steward ignores in the most recent commit. Your comment also sparked a little more research, and I've downgraded even further for license reasons.

.scala-steward.conf
```diff
@@ -78,11 +78,9 @@ pullRequests.grouping = [ { name = "minor_patch", title = "CORE-69: Minor and pa
 # Defaults to empty `[]` which mean Scala Steward will not ignore dependencies.
 # updates.ignore = [ { groupId = "org.acme", artifactId="foo", version = "1.0" } ]
 
-# Elasticsearch and its Lucene dependency should not be automatically updated;
-# these require nontrivial changes to calling code and business logic. See AJ-266.
 updates.ignore = [
-  { groupId = "org.elasticsearch.client", artifactId = "transport" },   # AJ-266
```
calypsomatic, line 84: Just checking - this is safe to remove because we don't use elastic search anymore, right?

davidangb: correct, that is gone as of #1623

calypsomatic (approved):

samanehsan (approved):

PR #1704, opened 2026-05-21
[CTM-43] Remove shibboleth code

calypsomatic:
https://broadworkbench.atlassian.net/browse/CTM-43

Shibboleth has been shut down, so we no longer need the `ShibbolethDao` in orch.  It was only used in the `/nih/callback` api, which not used anywhere (the terra-ui code that calls it is dead code).  This PR removes that API along with the associated shibboleth and nih code that is no longer needed.

In all cases:


aednichols (approved):
Nice!

src/main/scala/org/broadinstitute/dsde/firecloud/model/Profile.scala
```diff
@@ -117,12 +117,6 @@ case class NihLink(linkedNihUsername: String, linkExpireTime: Long) extends mapp
   require(ProfileValidator.nonEmpty(linkedNihUsername), "linkedNihUsername must be non-empty")
 }
 
-// For parsing the decoded JWT received from Shibboleth
-case class ShibbolethToken(eraCommonsUsername: String, iat: Long) {
-  // Link should expire in 30 days
-  def toNihLink: NihLink = NihLink(eraCommonsUsername, iat + 60 * 60 * 24 * 30)
-}
-
```
davidangb, line 125: can the `NihLink` class also be removed? Is that actually used elsewhere?

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
```diff
@@ -29,7 +23,6 @@ import org.broadinstitute.dsde.workbench.client.sam.model.{BulkMembershipUpdateR
 import org.broadinstitute.dsde.workbench.model.{WorkbenchEmail, WorkbenchGroupName, WorkbenchUserId}
 import org.broadinstitute.dsde.workbench.util2.messaging.ReceivedMessage
 import org.slf4j.LoggerFactory
-import pdi.jwt.{Jwt, JwtAlgorithm}
```
davidangb, line 32: I think this allows `"com.github.jwt-scala" %% "jwt-core"` to be removed entirely as a library dependency - is that safe?

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/NihApiService.scala
```diff
@@ -34,17 +34,9 @@ trait NihApiService extends Directives with RequestBuilding with EnabledUserDire
...
         pathPrefix("nih") {
-          // api/nih/callback: accept JWT, update linkage + lastlogin
-          path("callback") {
-            post {
-              entity(as[JWTWrapper]) { jwtWrapper =>
-                complete(nihServiceConstructor().updateNihLinkAndSyncSelf(userInfo, jwtWrapper))
-              }
-            }
+          path("status") {
+            complete(nihServiceConstructor().getNihStatus(userInfo))
```
davidangb, line 38: is this status API still used?

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
```diff
@@ -164,13 +164,6 @@ object FireCloudConfig {
     lazy val maxFileMatchingFileCount = firecloud.getInt("max-filematching-bucket-files")
   }
 
-  object Shibboleth {
-    // lazy - only required when shibboleth is enabled
-    private lazy val shibboleth = config.getConfig("shibboleth")
-    lazy val publicKeyUrl = shibboleth.getString("publicKeyUrl")
-    val enabled = shibboleth.optionalBoolean("enabled").getOrElse(true)
-  }
-
```
davidangb, line 173: Can you also remove the `shibboleth {...}` block from conf? I assume that'll be over in terra-helmfile

davidangb (approved):
:+1:, with questions about additional deletions. Fine with me if you want to handle the additional deletions in this PR or in a follow-on PR.

src/main/scala/org/broadinstitute/dsde/firecloud/service/NihService.scala
calypsomatic: done

src/main/scala/org/broadinstitute/dsde/firecloud/FireCloudConfig.scala
calypsomatic: done

src/main/scala/org/broadinstitute/dsde/firecloud/webservice/NihApiService.scala
calypsomatic: I do see calls to it in the logs, so I'll leave it in for now

PR #1715, opened 2026-07-09
CTM-581 NCBIAccess decommissioning 

aednichols:
Narrow removal to let us delete the `whitelists` and `whitelistBucket` keys from the config in terra-helmfile repository. There is definitely dead code left, but I have been struggling to disentangle what is and is not removable.

I tried to work forwards from the principle that I turned off the bucket update cron job in https://github.com/broadinstitute/terra-helmfile/pull/6461, so it's OK to assume the bucket contents are empty or irrelevant going forward.

### DIFF
--- a/local-dev/templates/firecloud-orchestration.conf
+++ b/local-dev/templates/firecloud-orchestration.conf
@@ -118,8 +118,6 @@ firecloud {
 }
 
 nih {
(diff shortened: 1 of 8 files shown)


jgainerdewar (approved):
Once CI passes, LGTM!

